PR #4, opened 2018-12-05
Returns GPS info in Photo::get response

hermzz:


anthosz:
Hello,

I just try and I have no gps datas in the informations of the photo:
image

I just try with & without exiftool: same issue

Best regards,

PR #25, opened 2019-01-07
Fixes #22

### LINKED ISSUE #22: Settings show error due to missing 'user.css' after clean install
After a clean install of Lychee-Laravel, the app shows an error in the UI (API-Error) because it tries to load the user.css, but it is not available.

bennettscience:
Added a script in composer after the install dump to make a user.css
file in the `dist` directory to prevent console or UI errors.

I haven't worked with composer scripts at all, which is why I'm doing this as a PR rather than a straight commit. It worked fine on my local machine when testing.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,11 @@
         ],
         "post-autoload-dump": [
             "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
-            "@php artisan package:discover"
+            "@php artisan package:discover",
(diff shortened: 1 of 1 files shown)


ildyria:
@bennettscience when you are doing a merge, I would prefer if we do a "Squash and merge" instead of just "Merge".
Both will create a merge commit but the Squash make the history cleaner: it will remove all the WIP commits etc...

PR #29, opened 2019-01-11
Installs Homestead

hermzz:
It seems we're all using vagrant/homestead, might as well make it easier
for newcomers to setup their dev machine.

I noticed the `Homestead.yaml` file is in the `.gitignore`. Maybe it makes sense to create a basic `Homestead.yaml.dist`?


aliases
```diff
@@ -0,0 +1,264 @@
...
+    pids=$(pidof /usr/bin/Xvfb)
+
+    if [ ! -n "$pids" ]; then
+        Xvfb :0 -screen 0 1280x960x24 &
+    fi
+
+    php artisan dusk "$@"
+}
+
+function php56() {
```
ildyria: this is not required as we do not support it.

aliases
```diff
@@ -0,0 +1,264 @@
...
+    sudo update-alternatives --set php /usr/bin/php5.6
+    sudo update-alternatives --set php-config /usr/bin/php-config5.6
+    sudo update-alternatives --set phpize /usr/bin/phpize5.6
+}
+
+function php70() {
```
ildyria: this is not required as we do not support it.

ildyria (commented):
I do not use Vagrant/Homestead but I agree with this PR.
Can you check if the two points I mentioned are necessary, if so, please advise and this will be merged. :+1:

ildyria:
@bennettscience you are Using Vagrant I think, can you check this?

aliases
bennettscience: Agreed. Laravel doesn't run on PHP < 7.1, so we can remove this.

aliases
bennettscience: Agreed.

bennettscience:
I don't see any problem with it, but I'm not sure about the "we're all using Vagrant" comment. I tend to shy away from dependency bloat, especially if you're using a low-bandwidth connection (like myself). But, that's mainly personal preference rather than long-term project usability. 

Another, more general concern, is that this would overwrite environments already configured on the next pull.

ildyria:
> Another, more general concern, is that this would overwrite environments already configured on the next pull.

What do you mean ?

ildyria (approved):

bennettscience:
Nevermind...it's been a long day. I was concerned that adding a tracked `Vagrantfile` to the repo would overwrite mine (or someone else's), which is already configured. But if it's untracked, it should be okay.

d7415:
> I tend to shy away from dependency bloat

Late to the party but I strongly agree with this. I'm not convinced that pulling the dev dependencies should automatically install homestead if only one dev is using it...

ildyria:
I broke the history tree (my bad). Just fixed it...
You might want to do a 
```sh
# just to be sure create a save branch
git checkout -b save
# switch back to master
git checkout master

# reset 5 commits
git reset --hard HEAD~5

# pull the new history
git pull origin master

# compare the changes with the save branch and cherry pick the commits you are missing on the save branch
git cherry-pick <commit id>
...
git push

```

hermzz:
Regarding bloat: the two new composer packages (`laravel/homestead` and `symfony/yaml`) come at under 1MB, given that the current `vendor/` folder sits at ~60MB I don't think this a huge deal. I think making this more dev friendly is worth the extra MB.

d7415:
> given that the current `vendor/` folder sits at ~60MB

That's its own problem...

> making this more dev friendly

Agreed in principle, but I think that's a stretch if you're the only developer using it.

It's merged now anyway.

Aleshus:
Can someone show his Homestead.yml here? I am getting 404 - No input file specified. on /php/index.php (Session::init) . I think that the project cant find SessionController.php but I dont know why for now.

Thanks

ildyria:
Sadly aside from Hermzz none of use use homestead.

I know that there is this:
https://github.com/LycheeOrg/Lychee/blob/master/Vagrantfile

and I also know that @trigun539 seems to have managed to get this working https://github.com/LycheeOrg/Lychee/issues/654#issuecomment-656277876

If someone would provide us with a Homestead.yml we would happily merge it.

trigun539:
Hello all, here is the Homestead.yaml that I used.

```yaml
---
ip: "192.168.10.10"
memory: 2048
cpus: 2
provider: virtualbox

authorize: ~/.ssh/id_rsa.pub

keys:
    - ~/.ssh/id_rsa

folders:
    - map: ~/opensource/Lychee
      to: /var/www/html/Lychee
    - map: ~/opensource/LycheeOrg.github.io
      to: /var/www/html/LycheeOrg.github.io

sites:
    - map: lychee.test
      to: /var/www/html/Lychee/public
    - map: lycheeorg.test
      to: /var/www/html/LycheeOrg.github.io

databases:
    - homestead

features:
    - mariadb: false
    - ohmyzsh: true
    - webdriver: false
```

The one extra item to do is update your `/etc/hosts` file to the following, add the following line:

```bash
192.168.10.10 lychee.test lycheeorg.test
``` 

After this, you should be able to run both the lychee org site and the lychee application. Hope this helps.

Aleshus:
Ok, I found the issue with  404 - No input file specified. There is a discusion about it here : https://github.com/LycheeOrg/Lychee/issues/52  and when clean new vagrant instance is started, nginx configuration is broken. Second issue is that image is not able to update packages. It looks like it is very old image :) 

You all who dont use vagrant, you are on docker? I looked fastly on this https://github.com/LycheeOrg/Lychee-Docker/blob/master/Dockerfile and I think that for develop puposes it needs some tweeks - install ssh, xdebug etc......

ildyria:
To develop I use Virtualbox, here is my setup : https://lycheeorg.github.io/docs/distributions.html#virtualbox-with-visual-studio-code

PR #32, opened 2019-01-19
Update/readme

ildyria:


ildyria:
If we are going to attrack more people here, it would be nice to at least have a decent landing page (even if incomplete).

PR #36, opened 2019-01-27
Integrate Lychee-Frame

ildyria:
Heavily inspired from https://github.com/chrobs/Lychee_frame from @chrobs .
Aim to solve part of https://github.com/LycheeOrg/Lychee/issues/162

Lychee-frame displays the Starred pictures as in a frame. They rotate every 30 seconds (maybe be a future parameter ?).

To activate it, you need to set the `Mod_Frame` to 1 in advanced settings.
You can then access it at http://example.com/frame

Code is simpler (FrameController.php) :
```php
/**
 * @return false|string
 */
function init()
{
  // check if Module is enabled
  // (it is not the case by default, user has to go in adv. settings to enable it)
  if (Configs::get_value('Mod_Frame') == '1') {
    $photo = Photo::where('star', '=', 1)->inRandomOrder()->first();

    // check if there are star picture
    if ($photo != null) {
      $thumb = Config::get('defines.urls.LYCHEE_URL_UPLOADS_THUMB').$photo->thumbUrl;
      if ($photo->medium == '1') {
        $url = Config::get('defines.urls.LYCHEE_URL_UPLOADS_MEDIUM').$photo->url;
      }
      else {
        $url = Config::get('defines.urls.LYCHEE_URL_UPLOADS_BIG').$photo->url;
      }

      return view('frame', [
        'url'   => $url,
        'thumb' => $thumb
      ]);
    }
    else {     // else redirect home
      return redirect()->route('home');
    }
  }
  else { // else redirect home
    return redirect()->route('home');
  }
}
```

And the html (blade template) side:
```blade
<img id="background" src="{{ $thumb }}" />
<canvas id="background_canvas"></canvas>
<div id="noise"></div>
<div class="image_container">
    <img id="picture" src="{{ $url }}" />
</div>
```

Additionally it solves a bug in Lychee-Frame: when pictures are in landscape mode, the css did not work (picture is not centered vertically). This has been solved.

Also this is a very basic implementation which reloads the page every 30 seconds.
It would be nice if we used AJAX to query the server for the next picture.


chrobs:
Salut!

I'm sorry that I didn't find the time to do it myself, but this looks very good :)

To improve it further, I thought about extending the Tags Feature such that albums may have tags as well. Additionally, add a parameter or selector for this frame page such that not only the starred photos may be shown but the user can choose to show either starred ones or based on the photos'/albums' tags.

Motivation is, that the selected photos to be shown in the digital frame may vary, depending on mood or think of multiple frames for different family members accessing the same lychee instance.

cheers,
chrobs

ildyria:
I am merging this as it contains some change on the template part.

The frame functionality is disabled by default. Be aware that, it does not provide photo protections: every pictures with a star will be displayed.

the url is: https://example.com/lychee/frame

ildyria:
> To improve it further, I thought about extending the Tags Feature such that albums may have tags as well. Additionally, add a parameter or selector for this frame page such that not only the starred photos may be shown but the user can choose to show either starred ones or based on the photos'/albums' tags.
>
> Motivation is, that the selected photos to be shown in the digital frame may vary, depending on mood or think of multiple frames for different family members accessing the same lychee instance.

Please open an issue for that. :)

chrobs:
Nothing easier than [that](https://github.com/LycheeOrg/Lychee-Laravel/issues/42) ;)

Thanks for your work!

PR #37, opened 2019-01-29
Converts the PhotoFunctions class into a service

hermzz:
Using Dependency Injection is better than statically calling class
methods.


ildyria:
I did not know that.

Thanks a lot for your insight. :+1:

PR #43, opened 2019-01-30
Converts the AlbumFunctions class into a service

hermzz:


hermzz:
I had actually converted a bunch of other albumIDs before I noticed that sometimes they can be strings, but reverted them back and I think we're safe here.

ildyria:
Yep. I quickly checked that before merging. I was just pointing it out. :+1:

PR #55, opened 2019-02-02
Moves and renames `PhotoFunctions::exists` to `Photo::isDuplicate`.

hermzz:


app/Photo.php
```diff
@@ -35,6 +35,21 @@ public function owner()
...
 
 
+    /**
+     * @param string $checksum
+     * @param $photoID
+     * @return array|false Returns a subset of a photo when same photo exists or returns false on failure.
+     */
+    public function isDuplicate(string $checksum, $photoID = null)
+    {
+        $sql = $this->where('checksum', '=', $checksum);
```
ildyria: I am not sure about the `$this->where` I think it should be `Photo::where()` no ?

hermzz: Neither am I, but I tested it and it worked? Let me do some more digging.

hermzz: I found this series of [best practices](https://github.com/alexeymezenin/laravel-best-practices#fat-models-skinny-controllers) that has a relevant example. The rest of the tips are reasonable as well.

ildyria: Hooo nice :+1: 
I basically did bad in every examples. :x

PR #59, opened 2019-02-06
Creates handlers for the different image processing libraries 

hermzz:


app/Image/GdHandler.php
```diff
@@ -0,0 +1,207 @@
...
+				$image = imagerotate($image, 90, 0);
+				imageflip($image, IMG_FLIP_HORIZONTAL);
+				break;
+
+			case 8:
+				$image = imagerotate($image, 90, 0);
+				break;
+		}
+
+		imagejpeg($image, $path, self::COMPRESSION_QUALITY);
```
ildyria: This is wrong.
With this function we are also modifying the "originals".
It should be 100.
Cf here:
https://github.com/LycheeOrg/Lychee-Laravel/pull/59/files#diff-2407954faff3ae8fbd036080db62e09fL210

hermzz: Good catch! I'll get that fixed.

Issue #69, opened 2019-02-10
Call to unknown function ImagickHandler::get_quality()


Chostakovitch:
PR #67 removed the function `get_quality()` but it is still referenced [here](https://github.com/hermzz/Lychee-Laravel/blob/be314914a77d56edf5ffca2d5631c5e750ff5a58/app/Image/GdHandler.php#L49) and [here](https://github.com/hermzz/Lychee-Laravel/blob/be314914a77d56edf5ffca2d5631c5e750ff5a58/app/Image/GdHandler.php#L86). 

Same with `GdHandler`.

Chostakovitch:
See PR #70

PR #70, opened 2019-02-10
Replace removed accessor for compression quality

Chostakovitch:
Fix for issue #69


ildyria:
Next time you can put as comment:
Fixes #69 
This way, when we merges this request, it will automatically close the issue. :)

PR #96, opened 2019-02-28
Checks album permissions to hide them correctly (fixes #74)

### LINKED ISSUE #74: Hidden Sub Albums are still visible
Create an album `test`, set it to `public` and `visible`.
Create an album `test2` in `test` set it to `public` and `hidden`.
Log out, go into `test`. `test2` will be visible.

hermzz:
Won't return albums that are hidden that do not belong to the user
requesting them.

### DIFF
--- a/app/Album.php
+++ b/app/Album.php
@@ -102,19 +102,27 @@ class Album extends Model
 	}
 
 
-
-	public function get_albums()
+	/**
+	 * Recursively returns the tree structure of albums. Private user albums are returned
+	 * only if `$userId` is set.
+	 *
+	 * @param  int   $userId
+	 * @return array
+	 */
+	public function get_albums(int $userId = null) : array
 	{
-		$return = array();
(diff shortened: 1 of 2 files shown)


app/Album.php
```diff
@@ -102,19 +102,27 @@ public function gen_thumbs($return)
...
-			$album_t = $album->gen_thumbs($album_t);
-			$return[$i] = $album_t;
-			$i++;
+		$subAlbums = [];
+		foreach ($this->children as $subAlbum) {
+			$album = $subAlbum->prepareData();
+			$album['albums'] = $subAlbum->get_albums($userId);
+			$album = $subAlbum->gen_thumbs($album);
+
+			if ($subAlbum['visible'] === '1' || ($userId === $subAlbum->owner->id)) {
```
ildyria: You are missing the admin case: `$userId == 0`

Issue #93, opened 2019-02-27
Feature proposal: better support for movies


kamil4:
I can't believe I'm suggesting this since I generally hate taking movies, but...

The current support is minimal at best. Only the square thumbnails are generated; why not at least the small variant as well? No useful metadata is extracted (at least things like width/height/fps/duration would be helpful).

I'll be happy to give it a shot next, once HiDPI is in.

For metadata, should I extend the `photos` table or should I simply overload some of the existing fields to avoid a need for database migration?

ildyria:
> For metadata, should I extend the `photos` table or should I simply overload some of the existing fields to avoid a need for database migration?

I think you can overload. The `type` is the part making the difference.

kamil4:
Thanks for the reminder! This one's finished; closing.

PR #110, opened 2019-03-04
Support for small thumbs and metadata for videos

kamil4:
Addresses #93. The bulk of the work is on the server side (here); I'll submit a separate PR for the front end.

`PhotoFunctions::createVideoThumb()` is renamed to `createVideoImages()`. It now also generates thumb@2x, small, and small@2x variants. It also extracts metadata (width, height, frame rate, creation time, location, duration). For metadata that goes beyond static images (frame rate, duration), I overloaded `focal` and `aperture`, respectively. The patch also fixes a strange behavior of video thumbs having a different basename than the video file. Another fixed bug is the deletion of the extracted frame image when it's no longer needed. The function includes extensive diagnostics when invoked from the console.

Speaking of the console, a new `movies` command is added that will extract any missing thumbnails and metadata.

For videos, we can't build the name of the small/small@2x files by appending `$photo.url` since that one is a video file, not a jpeg file, so we need to check and use `thumbUrl` instead.


app/ModelFunctions/PhotoFunctions.php
```diff
@@ -261,7 +487,7 @@ public function add(array $file, $albumID_in = 0)
 		// should not be needed
 		// Verify video
 		$mimeType = $file['type'];
-		if (!in_array($mimeType, $this->validVideoTypes, true)) {
```
ildyria: `$this->` was fine.

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -388,7 +614,7 @@ public function add(array $file, $albumID_in = 0)
 			}
 
 			// Create Thumb
-			if (!in_array($photo->type, $this->validVideoTypes, true)) {
```
ildyria: `$this->` was fine

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -543,9 +771,9 @@ public function getValidImageTypes(): array
 	 * @param  string $type
 	 * @return boolean
 	 */
-	public function isValidVideoType(string $type): bool
+	public static function isValidVideoType(string $type): bool
```
ildyria: no need of `static` here.

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -543,9 +771,9 @@ public function getValidImageTypes(): array
...
 	{
-		return in_array($type, $this->validVideoTypes, true);
```
ildyria: `$this->` was fine.

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -557,7 +785,7 @@ public function isValidVideoType(string $type): bool
 	 */
 	public function getValidVideoTypes(): array
 	{
-		return $this->validVideoTypes;
```
ildyria: `$this->` was fine.

app/Photo.php
```diff
@@ -130,9 +130,16 @@ public function prepareData()
 			$photo['license'] = $this->license;
 		}
 
+		if (PhotoFunctions::isValidVideoType($this->type)) {
```
ildyria: @hermzz there should be a way to avoid this static call right?

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -39,7 +40,7 @@ class PhotoFunctions
 	/**
 	 * @var array
 	 */
-	public $validVideoTypes = array(
+	public static $validVideoTypes = array(
```
ildyria: You should not need the `static` here.

kamil4:
Yeah, this patch turned out rather ugly, no arguments there. :frowning_face:

I was actually trying to *avoid* code duplication but that lead to the rather complex and lengthy `createVideoImages()`. This function is invoked in two contexts: from the web server when new videos are added, and from the console through the new artisan `videos` command, to add the missing data to previously added movies.

It was my desire to support that second context that resulted in most of the ugliness. I wanted the `videos` command to behave like the recently patched `medium` and friends, meaning that we would count towards the limit only the videos for which we changed something (so that a subsequent invocation of the same command could make progress). But that meant that `createVideoImages()`, if invoked in that context, needs to track (in `$save`) if it actually made any changes to `$photo`, and provide that information to the caller. The code to do that is relatively simple but verbose and kind-of ugly.

`Command` is passed to `createVideoImages()` specifically to indicate that second context. I could've passed a true/false flag instead, of course, but the way it's done makes it simple to output diagnostics messages from within `createVideoImages()` using `line()` (I haven't actually investigated: would ordinary PHP print/echo also work - what's the reason behind that `line()`?).

In the first context (new video being uploaded) there's no need to check the "old" values of `$photo->height`, etc, since, for a freshly allocated Photo object, they will all be 0/null/''. It's only in the second context that we test, to report any changes to the caller and also to avoid unnecessary work.

As to the `static` changes, I needed to invoke `isValidVideoType()` in various contexts (especially from App\Photo) and this was the way I was able to make it work. I'm happy to change it back if there's an easier way, of course.

One way I can think of to make this code much cleaner would be to completely remove the support for the second (Console) context from `createVideoImages()`. If only the upload of new videos was supported, the code would be *much* tighter. But a large section of `createVideoImages()` would then need to be duplicated within `Commands/videos.php`, though again, since only one context would need to be supported there, the code would be somewhat tighter. Let me know if you would like me to do it this way or if we should keep the current approach and just fix some of the worst ugliness.

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -75,33 +76,258 @@ public function __construct(Extractor $metadataExtractor, ImageHandlerInterface
 	 * @param string $path
 	 * @return string Path of the thumbnail
 	 */
-	public function createVideoThumb(Photo $photo, string $path) : string
-{
+	public function createVideoImages(Photo $photo, string $path, Command $videos = null) : string
```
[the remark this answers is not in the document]
hermzz: Agreed, either log all this stuff with `Logs` or you can print out some of this info back in the Command after this method returns. 

On a related note, this method is now insanely long. It probably make sense to refactor this and a) move it out to another class, and b) split this method out into smaller methods.

I'm also not really sure what the purpose of the `$save` variable is for?

app/Photo.php
hermzz: Yeah, why not just `strpos($this->type, 'video') === 0`? If it's been uploaded we already know it's valid, we just need to check if it's a video or image.

ildyria:
> One way I can think of to make this code much cleaner would be to completely remove the support for the second (Console) context from createVideoImages(). If only the upload of new videos was supported, the code would be much tighter. But a large section of createVideoImages() would then need to be duplicated within Commands/videos.php, though again, since only one context would need to be supported there, the code would be somewhat tighter. Let me know if you would like me to do it this way or if we should keep the current approach and just fix some of the worst ugliness.

I would say just use the Log and return a true if the creation was successful. If it was not, you can say in the command, that they need to check the logs. :)

kamil4:
Thank you both for your feedback! I'll redo this patch, preferably on top of @hermzz's refactored createMedium (#115); I'll see if I can reduce code duplication by reusing some of that new code. I'm also thinking of moving the metadata extraction into, well, the Metadata\Extractor, where it should fit much better. It may take me a day or two to get it done; in the meantime, the merged front end changes are safe with the unmodified server.

hermzz:
Probably a good idea. I made a small attempt at improving the metadata extractor here: https://github.com/hermzz/Lychee-Laravel/tree/metadata-extract. Feel free to grab anything from there if you like.

kamil4:
@hermzz, what's your plan for merging that `metadata-extract` work? I like how it's half the size of the current version but, considering that it seems to be a complete rewrite, I'm reluctant to put any of my code in there until yours is merged...

hermzz:
I'm not planning to do anything with that anytime soon. I thought it was going to be a quick-ish matter, but it was turning out to be quite involved and I don't have the time for such a large feature. I'll add some comments to commit 386474d outlining what I think needed to be done to get it finished.

kamil4:
OK, I pretty much reimplemented this patch from scratch. I hope it will be more acceptable now.

Extraction of video metadata is now integrated into the regular metadata extractor.

`createVideoThumb()` was renamed to `extractVideoFrame()` as that's the only thing it does now. The regular `createThumb()` and `createSmallerImages()` are now used for videos as well (the extracted video frame is passed as an extra argument).

A new `video_data` console command extracts the metadata and generates thumbs.

The patch addresses two related old bugs:
- Extracted video frames were saved to `/tmp` but never deleted.
- Thumbs for videos had a different basename than video files.

I also fixed two unrelated bugs:
- The timestamp of original (big) files was being set to January 1st, 1970 on upload.
- small2x/medium2x thumb generation was recently broken by commit 82ff5ce: the 2x variants would overwrite the regular ones.

app/Console/Commands/video_data.php
```diff
@@ -0,0 +1,150 @@
...
+						$this->line('Renamed thumb to match the video file');
+					}
+				}
+			}
+
+			if (file_exists($url)) {
+				$info = $this->metadataExtractor->extract($url, $photo->type);
+
+				if ($info['width']) {
+					$this->line('Extracted metadata');
```
ildyria: this line should probably be above the if statement.

app/Console/Commands/video_data.php
```diff
@@ -0,0 +1,150 @@
...
+					}
+					if ($frame_tmp !== '') {
+						$this->line('Extracted video frame for thumbnails');
+						if ($photo->thumbUrl === '' || $photo->thumb2x === 0) {
+							if (!$this->photoFunctions->createThumb($photo, $frame_tmp)) {
+								Logs::error(__METHOD__, __LINE__, 'Could not create thumbnail for video');
+							}
+						}
+						if ($photo->small === '' || $photo->small2x === '') {
+							$this->photoFunctions->createSmallerImages($photo, $frame_tmp);
```
ildyria: doesn't this also creates mediums ?

app/Http/Controllers/LogController.php
```diff
@@ -55,6 +55,8 @@ static public function clearNoise()
 		orWhere('function', '=', 'App\ModelFunctions\PhotoFunctions::add')->
 		orWhere('function', '=', 'App\ModelFunctions\PhotoFunctions::createMedium')->
 		orWhere('function', '=', 'App\ModelFunctions\PhotoFunctions::createThumb')->
+		orWhere('function', '=', 'App\ModelFunctions\PhotoFunctions::createVideoThumb')->
+		orWhere('function', '=', 'App\ModelFunctions\PhotoFunctions::extractVideoFrame')->
```
ildyria: thanks :+1:

app/Photo.php
```diff
@@ -141,9 +148,9 @@ public function prepareData()
 		}
 
 		if ($this->medium2x != '') {
-			$url2x = explode('.', $this->url);
-			$url2x = $url2x[0].'@2x.'.$url2x[1];
-			$photo['medium2x'] = Config::get('defines.urls.LYCHEE_URL_UPLOADS_MEDIUM').$url2x;
+			$photoName2x = explode('.', $photoName);
+			$photoName2x = $photoName2x[0].'@2x.'.$photoName2x[1];
```
ildyria: this can be done around line 139. This avoids to do it twice: once for medium2x and once for small2x

app/Console/Commands/video_data.php
kamil4: Actually, the metadata extractor doesn't return a success indicator so we don't know if it succeeded to extract anything until we check the contents of the `info` array. So that's why I wrote it that way. It's still not perfect because I suppose it could be that `width` could not be extracted but the `duration` could... Maybe I'm overthinking it. We could also just drop that message altogether? Let me know...

app/Console/Commands/video_data.php
kamil4: No, I modified `createSmallerImages()` not to generate the mediums if the frame argument is non-empty.

app/Photo.php
kamil4: Right. I wrote it that way because all the 2x stuff is optional (and off by default) so I thought we shouldn't do it unless it's needed. But I agree that readability suffers as a result. I'll fix it.

app/Console/Commands/video_data.php
ildyria: Then shouldn't you test `if ($info['width'] != '')` instead ? :)

kamil4: Good point :smiley: Actually, that field is an int not a string, but it's still not right. I'll fix it.

PR #115, opened 2019-03-05
Refactors PhotoFunctions::createMedium

hermzz:
Re-jigs stuff around a bit to tidy it up, renames `createMedium` to
`resizePhoto`, and makes it depend on convention a bit more to make the
code less verbose and more generic.

Creates a new generic command `generate_thumbs` that replaces the other
really specific `small`, `small2x`, `medium`, and `medium2x`
commands that only differed in some small details.


app/Console/Commands/generate_thumbs.php
```diff
@@ -0,0 +1,105 @@
...
+			$this->line('No picture requires '.$type.'.');
+			return 0;
+		}
+
+		foreach ($photos as $photo) {
+			if ($this->photoFunctions->resizePhoto(
+				$photo,
+				$type,
+				intval(Configs::get_value('medium_max_width')),
+				intval(Configs::get_value('medium_max_height')))
```
ildyria: I think here you may want to take `small_max_width` also and `small_max_height` depending if it is `medium` or `small`.

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -409,48 +356,86 @@ public function add(array $file, $albumID_in = 0)
...
+			return false;
+		}
+
+		$resWidth = $resHeight = 0;
+		if (!$this->imageHandler->scale($url, $uploadFolder.$filename, $maxWidth, $maxHeight, $resWidth, $resHeight)) {
+			Logs::error(__METHOD__, __LINE__, 'Failed to '.$type.' resize image');
+			return false;
+		}
+
+		$photo->{$type} = $resWidth . 'x' . $resHeight;
```
ildyria: clever ! :+1:

app/Console/Commands/generate_thumbs.php
```diff
@@ -0,0 +1,105 @@
...
+		set_time_limit($this->argument('timeout'));
+
+		$basicType = $type;
+		if (($split = strpos($basicType, '2')) !== false) {
+			$basicType = substr($basicType, 0, $split);
+		}
+
+		$this->line(
+			sprintf(
+				'Will attempt to generate up to %s medium (%dx%d) images with a timeout of %d seconds...',
```
kamil4: Here I guess you should replace `medium` with `$type`?

hermzz:
Good catch, on second review the command was all kinds of broken.

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -409,48 +356,86 @@ public function add(array $file, $albumID_in = 0)
...
+			$this->resizePhoto($photo, 'medium2x', $mediumMaxWidth * 2, $mediumMaxHeight * 2);
 		}
 
-		return $this->save($photo, $albumID);
+		$smallMaxWidth = intval(Configs::get_value('small_max_width'));
+		$smallMaxHeight = intval(Configs::get_value('small_max_height'));
+		$this->resizePhoto($photo, 'small', $smallMaxWidth, $smallMaxHeight);
+
+		if (Configs::get_value('small_2x') === '1') {
+			$this->resizePhoto($photo, 'small2x', $mediumMaxWidth * 2, $mediumMaxHeight * 2);
```
kamil4: I think these should be `$smallMaxWidth * 2` and `$smallMaxHeight *2`?

ildyria: Good catch !

PR #116, opened 2019-03-05
Reapply changes lost in 7e14093

kamil4:
I noticed some strange misbehavior with HiDPI images. On investigation I found a minor bug in the front end (I will submit a separate commit for that) but also a major issue on the server side: changes to `App/Album.php` from commit 0a43348 got lost. @ildyria, this appears to have happened during your commit 7e14093.
This PR reapplies the lost changes.

### DIFF
--- a/app/Album.php
+++ b/app/Album.php
@@ -79,12 +79,13 @@ class Album extends Model
 	{
 
 		$return['thumbs'] = array();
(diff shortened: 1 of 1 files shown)


ildyria:
Thanks, nicely spotted :+1:

PR #129, opened 2019-03-08
Refactoring

ildyria:
Lots of rewriting... Started writing comments.

### DIFF
--- a/app/Configs.php
+++ b/app/Configs.php
@@ -16,6 +16,8 @@ class Configs extends Model
 
 	private static $cache = null;
 
+
+
 	public static function arrayify($query)
 	{
 		$configs = $query->get();
@@ -100,6 +102,10 @@ class Configs extends Model
 			Logs::error(__METHOD__, __LINE__, $config->getErrors());
 			return false;
 		}
+
+		// invalidate cache.
+		self::$cache = null;
+
 		return true;
 	}
 
--- a/app/Configs.php
+++ b/app/Configs.php
@@ -18,19 +18,19 @@ class Configs extends Model
 
 
 
-	public static function arrayify($query)
-	{
-		$configs = $query->get();
-
-		$return = array();
-
-		// Add each to return
-		foreach ($configs as $config) {
-			$return[$config->key] = $config->value;
-		}
-
-		return $return;
-	}
+//	public static function arrayify($query)
+//	{
+//		$configs = $query->get();
+//
+//		$return = array();
+//
+//		// Add each to return
+//		foreach ($configs as $config) {
+//			$return[$config->key] = $config->value;
+//		}
+//
+//		return $return;
+//	}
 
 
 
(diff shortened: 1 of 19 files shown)


app/Configs.php
```diff
@@ -8,95 +8,80 @@
...
-	 * @param string $key
-	 * @param bool $public
-	 * @return bool
-	 */
-	static protected function inExceptArray(string $key, bool $public)
+	private static $cache = null;
+
+
+
+	public static function arrayify($query)
```
hermzz: Eloquent does something for this: https://laravel.com/docs/5.8/collections#method-pluck

ildyria: Niiiiice :astonished:  Thanks !

app/Http/Controllers/IndexController.php
```diff
@@ -28,16 +28,17 @@ public function __construct(ConfigFunctions $configFunctions)
...
 		{
-			$lang = Lang::get_lang(Configs::where('key', '=', 'lang')->first());
+			$lang = Lang::get_lang(Configs::get_value('lang', 'en'));
 
 			$infos = $this->configFunctions->get_pages_infos();
 
 			$menus = Page::menu()->get();
 
-			$title = Configs::get_value('site_title');
+			$title = Configs::get_value('site_title', 'Lychee v4');
```
hermzz: Given that `Lychee v4` is repeated a bunch of times, wouldn't it make sense to define it in one of the `config/` files?

ildyria: good point.

app/Http/Controllers/PageController.php
```diff
@@ -28,15 +28,17 @@ public function __construct(ConfigFunctions $configFunctions)
 
 	function page(Request $request, $page)
 	{
+		Configs::get();
```
hermzz: This is a bit awkward, is there a way to make `Configs::get_value` check if the cache is not set and call `Configs::get` from within?

ildyria: it is supposed to be the case but for some reason it is not working:
https://github.com/LycheeOrg/Lychee-Laravel/blob/refactoring/app/Configs.php#L66

app/ModelFunctions/SessionFunctions.php
```diff
@@ -0,0 +1,75 @@
...
+	 */
+	public function noLogin()
+	{
+
+		$configs = Configs::get();
+
+		// Check if login credentials exist and login if they don't
+		if (isset($configs['username']) && $configs['username'] === '' &&
+			isset($configs['password']) && $configs['password'] === '') {
+			Session::put('login', true);
```
d7415: I don't like the indentation here, but I can't think of a better way either.

database/migrations/2019_03_08_135817_private_config_fields.php
```diff
@@ -0,0 +1,91 @@
...
+	 * @return void
+	 */
+	public function up()
+	{
+		if (Schema::hasTable('configs')) {
+			Schema::table('configs', function ($table) {
+				$table->tinyInteger('confidentiality')->after('cat')->default(0);
+			});
+
+			// set confidentiality to 3 for those (never returned)
```
d7415: s/3/4/

Issue #139, opened 2019-03-12
Subalbums and password protection


kamil4:
Album-level password protection doesn't seem to work in combination with subalbums.

I first tried a password-protected top-level album with a public sub-album. In public mode (not logged in), providing the password opened the top-level album as expected, but clicking on the sub-album wasn't working. JS console had a `TypeError: albums.getByID(...) is undefined` error logged.

I tried the opposite combination: an unprotected top-level album with a password-protected subalbum. That didn't work either: clicking on the subalbum results in an empty album view and, eventually, same JS error. More worrying, the subalbum was not marked as password protected and the thumbs of the three pictures inside were displayed, so it loks like the server was only too happy to serve them -- without a password being provided.

Looking at the source code, it looks like at least `albums.getByID()` has no support for subalbums (I guess it would need to descend recursively in search for an album ID?).

My basic question though would be: is this even supposed to work? I can imagine that with Lychee-Laravel being multi-user, perhaps password protection at the album level is seen as an anachronism? A similar effect could, I'm guessing, be achieved by sharing an album/subalbums with a password-protected, locked, read-only "guest" account. So let me know if you even want to be bothered with such reports...

ildyria:
Fixed.

joszz:
This seems to still not work properly?
I also wanted to password protect an album with subalbums.
After logging in with the password no subalbums are displayed

d7415:
> After logging in with the password no subalbums are displayed

Have you set the subalbums to public? (with or without password)

joszz:
Well yes, if I set the subalbums to public it will show.
It would be handy though that I could inherit the password protection (or all other visibility properties for that mather) from the top album.
Now if I want to properly protect this album and it's subalbums, a user would have to provide the password constantly when navigating through album and then subalbum. Bit of a bad user experience. Although it might make sense to do this, having different passwords set for different sub albums for some use cases.
Maybe not really a bug, but also not the best solution for everybody. 

If I set the album to hidden it will be invisible from the top album (which makes sense I guess, it does what it says it does, only accessible through direct link).

p.s.
I'm on latest Lychee 4.0.4

ildyria:
> It would be handy though that I could inherit the password protection (or all other visibility properties for that mather) from the top album.

This is somewhat already the case. In order to access the sub album directly you still need to provide the password for the parent.

[...]

PR #142, opened 2019-03-14
Don't return thumbs of password-protected subalbums

kamil4:
Server-side fix for #139

This basically duplicates the condition from `AlbumFunctions::prepare_albums()`, only this time for subalbums, and in a less round-about way :smirk:

### DIFF
--- a/app/Album.php
+++ b/app/Album.php
@@ -5,6 +5,7 @@ namespace App;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Facades\Config;
 use Illuminate\Support\Facades\Hash;
+use Illuminate\Support\Facades\Session;
 
 class Album extends Model
 {
@@ -121,7 +122,9 @@ class Album extends Model
 
 				$album = $subAlbum->prepareData();
 				$album['albums'] = $subAlbum->get_albums($userId);
-				$album = $subAlbum->gen_thumbs($album);
+				if ($subAlbum->password === null || Session::get('login')) {
+					$album = $subAlbum->gen_thumbs($album);
+				}
 
 				$subAlbums[] = $album;
 			}


ildyria:
It is slightly more complex, you also need to consider whether the user is an admin or a regular user and if the sub album is shared with that regular user...

kamil4:
I played with it a bit and you are correct. In multi-user mode, it does not work correctly (the password protection is, basically, ignored). I thought the condition may be too simple, which is exactly why I spelled out where I copied it from. :smiley:

Making the multi-user mode work robustly will be a lot more work, though. There seem to be a lot of cases that are not handled correctly at the moment. E.g., the front end seems to think that a regular user can edit any public album, irrespective of who owns it (the server seems to enforce user separation, but the front end gets rather confused when uploads or deletes fail)... I've also seen weird behavior when moving photos between shared albums.

What I'm trying to say is: can we consider the fixing of passwords in multi-user mode be a separate problem, to be addressed after (or while) general multi-user support is made more robust? We could keep the issue #139 open to remind us that there are other cases that need fixing but, in the mean time, I believe the patches I submitted do correctly fix the publicMode case (hopefully without introducing any new problems).

kamil4:
The front end side https://github.com/LycheeOrg/Lychee-front/pull/51 needs to be merged as well to complete the fix... While github gives me permission to do the merge now (and @d7415 has OK'ed the change), I wasn't sure how you want to handle such merges policy-wise so I chose to wait. If it's OK for me to merge approved PRs, should I be choosing "Squash and merge"?

FYI, I am interested in working on a comprehensive set of fixes to close the majority of such corner cases and unexpected interactions between the server and the front end, once I complete a few smaller projects. I think the first step should be documenting the current behavior to ensure that everybody is on the same page with respect to how things work and also how they are _supposed_ to work, before doing any major hacking. Where should such discussions go? The outcomes should probably end up as a set of issues but I'm not sure if starting the discussion within an issue is the right approach?

d7415:
I was waiting for @ildyria's review on -front. I only gave a cursory look and checked that it didn't break v3.

In general terms, yes, so long as everyone's happy feel free to merge. And yes, squash and merge is the preferred way in Lychee so that we can keep a "cleaner" tree.

The behaviour could be sketched out in an issue, or you can use the organisation.

Issue #94, opened 2019-02-27
Feature proposal: fullscreen button


kamil4:
I know I can hit F12 to switch my desktop browser to full screen mode but I think some people would appreciate a "full screen" button on the screen that does that. Many galleries seem to have it. What would you say about adding one to the frontend?

ildyria:
Isn't it F11 ? :smile: 

If you want to give it a shot sure. I don't really see it a priority but I don't have anything against.
However if you do it, please add a keybind for it to `F` (Lightroom keybinding :wink: )

kamil4:
LOL, clearly I've been hitting that F12 too much lately, it got etched on my brain :wink:

Sure, I absolutely agree that it's low priority. Clearly I'm reaching the bottom of my personal TODO list if I mention such things :smiley:

ildyria:
Well, there are some bugs left that are higher in the list, such as : 
https://github.com/LycheeOrg/Lychee-Laravel/issues/91
https://github.com/LycheeOrg/Lychee-Laravel/issues/88
https://github.com/LycheeOrg/Lychee-Laravel/issues/58
https://github.com/LycheeOrg/Lychee-Laravel/issues/48 < This one would be very nice. :)

ildyria:
done.

PR #146, opened 2019-03-17
Add fullscreen support to album and photo views

kamil4:
Finally takes care of my pet project #94

Front end part is at https://github.com/LycheeOrg/Lychee-front/pull/52


kamil4:
My latest commit should address this. @d7415 do I need to do the same with https://github.com/LycheeOrg/Lychee/pull/228? I added the locale strings there but then I realized that index.html isn't localized...

d7415:
> My latest commit should address this. @d7415 do I need to do the same with [LycheeOrg/Lychee#228](https://github.com/LycheeOrg/Lychee/pull/228)? I added the locale strings there but then I realized that index.html isn't localized...

I think it's fine. I don't think Lychee does any localisation itself, only the front-end.

Lychee-front does update some of index.html (the star one in particular I've noticed when playing) but that's out of scope (and I *think* falls back to the values in `lychee_locale.js` anyway)

PR #153, opened 2019-03-21
Fixes #151 by adding a fail back to GD

### LINKED ISSUE #151: Big images crashes Imagick
```
    2019-03-19 16:29:08 -- error -- App\ModelFunctions\PhotoFunctions::resizePhoto -- 435 -- Failed to small resize image
    2019-03-19 16:29:08 -- error -- App\Image\ImagickHandler::scale -- 57 -- width or height exceeds limit `/var/www/Lychee/public/uploads/big/ff5589de6c95442137971ca29d616ec8.jpg' @ error/cache.c/OpenPixelCache/3912
    2019-03-19 16:29:08 -- error -- App\ModelFunctions\PhotoFunctions::resizePhoto -- 435 -- Failed to medium resize image
    2019-03-19 16:29:08 -- error -- App\Image\ImagickHandler::scale -- 57 -- width or height exceeds limit `/var/www/Lychee/public/uploads/big/ff5589de6c95442137971ca29d616ec8.jpg' @ error/cache.c/OpenPixelCache/3912
    2019-03-19 16:29:08 -- notice -- App\ModelFunctions\PhotoFunctions::add -- 343 -- ff5589de6c95442137971ca29d616ec8.jpeg
    2019-03-19 16:29:08 -- error -- App\Image\ImagickHandler::crop -- 92 -- width or height exceeds limit `/var/www/Lychee/public/uploads/big/ff5589de6c95442137971ca29d616ec8.jpg' @ error/cache.c/OpenPixelCache/3912
    2019-03-19 16:29:08 -- error -- App\Image\ImagickHandler::crop -- 92 -- width or height exceeds limit `/var/www/Lychee/public/uploads/big/ff5589de6c95442137971ca29d616ec8.jpg' @ error/cache.c/OpenPixelCache/3912
    2019-03-19 16:29:07 -- notice -- App\ModelFunctions\PhotoFunctions::createThumb -- 100 -- Photo URL is ff5589de6c95442137971ca29d616ec8.jpg
```
The said picture is 59MB, 20434x5282 pixel

ildyria:
Fixes #151 by adding a failback to GD, I have not encountered a case where GD fails yet.
Another way would be to actually increases the values in `/etc/imagick/policy.xml` But some users may not have access to those values.

### DIFF
--- a/app/Image/ImageHandler.php
+++ b/app/Image/ImageHandler.php
@@ -43,7 +43,7 @@ class ImageHandler implements ImageHandlerInterface
 	public function scale(string $source, string $destination, int $newWidth, int $newHeight, int &$resWidth, int &$resHeight): bool
 	{
 		$i = 0;
-		while(!$this->engines[$i]->scale($source,$destination,$newWidth,$newHeight,$resWidth,$resHeight))
+		while($i < count($this->engines) && !$this->engines[$i]->scale($source,$destination,$newWidth,$newHeight,$resWidth,$resHeight))
 		{
 			$i++;
 		}
@@ -62,7 +62,7 @@ class ImageHandler implements ImageHandlerInterface
 	public function crop(string $source, string $destination, int $newWidth, int $newHeight): bool
 	{
 		$i = 0;
-		while(!$this->engines[$i]->crop($source,$destination,$newWidth,$newHeight))
+		while($i < count($this->engines) && !$this->engines[$i]->crop($source,$destination,$newWidth,$newHeight))
 		{
 			$i++;
 		}


hermzz:
I'm not a fan of this commit as it introduces an extra dependency where it doesn't belong. It shouldn't be the responsibility of the image class to determine what to do if there's an error. This would make more sense if we introduced an overarching image processing manager of sorts that loaded all the image processing libraries and ran through them in order if there were errors.

I think the better solution to this bug is to add a config option for the `Imagick::setResourceLimit` values and giving the user feedback upon failure explaining that the limit was exceeded and how to increase it.

ildyria:
> I think the better solution to this bug is to add a config option for the Imagick::setResourceLimit values and giving the user feedback upon failure explaining that the limit was exceeded and how to increase it.

I tried the Imagick::setResourceLimit solution. I was not able to make it work. :(

kamil4:
> I tried the Imagick::setResourceLimit solution. I was not able to make it work. :(

Yeah, from my reading, you can't exceed the limits set in `policy.xml`.

> This would make more sense if we introduced an overarching image processing manager of sorts that loaded all the image processing libraries and ran through them in order if there were errors.

The abstraction is already there, isn't it? If `ImageHandlerInterface` became a full-fledged class, this new code would fit in there perfectly...

ttys3:
this will cause GD as an 'Must Have Enabled' php extension.

kamil4:
I thought GD was already required? Can the server run without it? I'm pretty sure I used it unconditionally in a database migration...

d7415:
> I thought GD was already required?

[Yup](https://github.com/LycheeOrg/Lychee-Laravel/wiki/Install#0-install-dependencies)

ildyria:
I need to rework on this PR...

ildyria:
Waiting for feedback from @hermzz  :)

app/Image/ImageHandler.php
```diff
@@ -0,0 +1,86 @@
...
+	 * Rotates and flips a photo based on its EXIF orientation.
+	 *
+	 * @param string $path
+	 * @param array $info
+	 * @return array
+	 */
+	public function autoRotate(string $path, array $info): array
+	{
+
+		return $this->engines[0]->autoRotate($path,$info);
```
hermzz: Why is this hard coded to on;y use the first engine? Wouldn't it make sense to do the same thing we do with `crop` and `scale`?

ildyria: I didn't see any code to detect if autoRotate fails. :s

hermzz: Good point :-)

PR #155, opened 2019-03-22
Ensure correct padding of id on both ends

### LINKED ISSUE #123: Album sort by creation date is not by creation date but by id
I added all "correct" created_at and updated_at manually (as they got not migrated when I did the migration). 

image

image

It looks like it is sorting by id though.

kamil4:
Should fix #123.

I encountered the same problem as @SerenaButler but without migration from v3. Basically, if the fourth digit after the decimal point is `0` (one in ten chance, obviously), `id` ends up being a digit short so sorting by it is broken.

I replaced the custom code we had with a single call to good old C-style `sprintf()` which gets us pretty much everything we want here: guaranteed 4 decimal digits and 14 digits total, 0-padded on the left if need (which it never will be, unless we move back to 2001). All I kept was the code that removes the dot.

### DIFF
--- a/app/ModelFunctions/Helpers.php
+++ b/app/ModelFunctions/Helpers.php
@@ -13,13 +13,12 @@ class Helpers
 	{
 
 		// Generate id based on the current microtime
-		$id = str_replace(' ', '', microtime(true));
+		// Ensure 4 digits after the decimal point, 15 characters
+		// total (including the decimal point), 0-padded on the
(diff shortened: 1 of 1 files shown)


d7415:
Now I need to go find that conversation...

d7415:
Bonus: 90% of the time, ID matches v3. I might look into patching this in v3....

PR #162, opened 2019-03-29
skip csrf verification if an api key is set and present

alex-phillips:
Check the `Authorization` header to make sure an API key is present and matches what is in the config table (assuming what is a valid string)

### DIFF
--- a/app/Http/Middleware/VerifyCsrfToken.php
+++ b/app/Http/Middleware/VerifyCsrfToken.php
@@ -2,32 +2,50 @@
 
 namespace App\Http\Middleware;
 
+use App\Configs;
 use Closure;
 use Illuminate\Foundation\Http\Middleware\VerifyCsrfToken as Middleware;
-use Illuminate\Http\Request;
-use App\Configs;
 
 class VerifyCsrfToken extends Middleware
 {
-    /**
-     * The URIs that should be excluded from CSRF verification.
-     *
-     * @var array
-     */
-    protected $except = [
-        // entry point...
-        '/php/index.php'
-    ];
+	/**
+	 * The URIs that should be excluded from CSRF verification.
+	 *
+	 * @var array
+	 */
+	protected $except = [
+		// entry point...
+		'/php/index.php'
+	];
 
-    public function handle(Request $request, Closure $next)
-    {
-        if ($request->is('api/*')) {
-            $apiKey = Configs::get_value('api_key');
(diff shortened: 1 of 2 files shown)


ildyria:
Hi @alex-phillips, thank you a lot for your PR. :+1: 
On your installation, assuming you did `composer --install` (without the `--no-dev`) can you try `vendor/bin/phpunit` in your root (well where you would do `php artisan ...`.

It seems to break on some tests ( https://github.com/LycheeOrg/Lychee-Laravel/blob/master/tests/Feature/HtmlTest.php ).

app/Http/Middleware/VerifyCsrfToken.php
```diff
@@ -15,4 +18,16 @@ class VerifyCsrfToken extends Middleware
         // entry point...
         '/php/index.php'
     ];
+
+    public function handle(Request $request, Closure $next)
+    {
+        if ($request->is('api/*')) {
+            $apiKey = Configs::get_value('api_key');
+            if ($apiKey && $request->header('Authorization') === $apiKey) {
```
ildyria: Do you want it as an part of the header or would it be more interesting to have it as a `POST` field ? :)

ildyria: I'm not asking it to be changed, I am just genuinely curious. :smiley:

alex-phillips: I'm OK with either. I know the `Authorization` header is common, but that's also usually for JWTs or OAuth tokens, so having it as a header could be confusing to some people? I'm OK with whichever you prefer as it's an easy change.

alex-phillips:
@ildyria thanks for the feedback!  I get the following output when I run phpunit:

```
PHPUnit 7.5.6 by Sebastian Bergmann and contributors.

..S                                                                 3 / 3 (100%)

Time: 343 ms, Memory: 20.00 MB

OK, but incomplete, skipped, or risky tests!
Tests: 3, Assertions: 26, Skipped: 1.
```

d7415:
> It seems to break on some tests ( https://github.com/LycheeOrg/Lychee-Laravel/blob/master/tests/Feature/HtmlTest.php ).

Really interesting that it only fails with the dev tools... Looks like a 500 error when it's expecting something else? A blank, maybe?

ildyria:
I fixed the error 500. :)

PR #179, opened 2019-04-10
use php_script_no_limit instead of php_script_limit, do not introduce db log write operation when there 's no DB configed 

ttys3:
I think maybe `php_script_no_limit` is better
when `no limit` is true, we know that it'  unlimited

set `php_script_limit` to 1 then it is unlimited, I think is ambiguous  and confused.

-------------------------------------------------------------

app/Configs.php is a Base helper class

`Logs::notice(__METHOD__, __LINE__, "hasImagick : false");`

this call will cause composer install failure,

because when there is no imagick extension loaded, 

`artisan package:discover` will fail, because no DB configed now

### DIFF
--- a/app/Configs.php
+++ b/app/Configs.php
@@ -122,7 +122,8 @@ class Configs extends Model
 		if ((bool) (extension_loaded('imagick') && self::get_value('imagick', '1') == '1')) {
 			return true;
 		}
-		Logs::notice(__METHOD__, __LINE__, "hasImagick : false");
+		//this call will cause composer install failure (artisan package:discover will fail, because no DB config now)
+		//Logs::notice(__METHOD__, __LINE__, "hasImagick : false");
 		return false;
 	}
 
--- a/app/Configs.php
+++ b/app/Configs.php
@@ -100,6 +100,10 @@ class Configs extends Model
 	{
 
 		$config = Configs::where('key', '=', $key)->first();
+		//first() may return null, fixup 'Creating default object from empty value' error
+		if ($config == null) {
+			return true;
+		}
 		$config->value = $value;
 		if (!$config->save()) {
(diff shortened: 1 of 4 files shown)


ildyria:
Hi @ttys3 thank you for your PR.

Commit 3c5744d is not useful, this has already been fixed: https://github.com/LycheeOrg/Lychee-Laravel/commit/5e13f3e1a31585974e5f0244d2e5600065043348 :)

app/Configs.php
```diff
@@ -122,7 +122,8 @@ public static function hasImagick()
 		if ((bool) (extension_loaded('imagick') && self::get_value('imagick', '1') == '1')) {
 			return true;
 		}
-		Logs::notice(__METHOD__, __LINE__, "hasImagick : false");
```
ildyria: cf comment.

database/migrations/2019_04_03_141031_php_script_limit.php
```diff
@@ -18,7 +18,7 @@ public function up()
```
ildyria: This file should not be modified.
While it will work for any new user, it will not pass to the users who already have the latest step installed.
The correct approach is to create a new migration:
`php artisan make:migration rename_php_script_limit`
and add in it something like:
```php
	    Configs::where('key','php_script_limit')->update([
			    'key' => 'php_script_no_limit'
	    ]);
```
Take example on https://github.com/LycheeOrg/Lychee-Laravel/blob/master/database/migrations/2019_02_19_193500_rename_justified_layout.php

ildyria: done.

ttys3:
thanks for the `migration file`

PR #185, opened 2019-04-10
Fill created_at by the takestamp of v3

### LINKED ISSUE #89: Migration Lychee 3 -> 4 // Album Creations dates not migrated but overwritten
All Album Dates now February 2019 ... :-(

image

ildyria:
Fixes #89 

Can someone check that PR before merge ?

### DIFF
--- a/database/migrations/2018_08_15_102039_move_albums.php
+++ b/database/migrations/2018_08_15_102039_move_albums.php
@@ -25,9 +25,7 @@ class MoveAlbums extends Migration
 					$album->public = $result->public;
 					$album->visible_hidden = $result->visible;
 					$album->license = $result->license;
-					if ($result->takestamp != '') {
-						$album->created_at = date('Y-m-d H:i:s', substr($result->takestamp, 0, -4));
-					}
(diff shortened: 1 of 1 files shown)


database/migrations/2018_08_15_102039_move_albums.php
```diff
@@ -1,58 +1,61 @@
...
+				foreach ($results as $result) {
+					$album = new Album();
+					$album->id = $result->id;
+					$album->title = $result->title;
+					$album->description = $result->description;
+					$album->public = $result->public;
+					$album->visible_hidden = $result->visible;
+					$album->license = $result->license;
+					if ($result->takestamp != '') {
+						$album->created_at = date('Y-m-d H:i:s', substr($result->takestamp, 0, -4));
```
kamil4: I believe this should be `sysstamp`, not `takestamp`, and you should pass it directly, without `substr`, since it's in full seconds, not like `id` (see https://github.com/LycheeOrg/Lychee-Laravel/issues/89#issuecomment-467019143).

ildyria: you might be right. :D

PR #201, opened 2019-04-15
Add new route for update directly from web page.

ildyria:
This is an experimental feature which would allow an admin to directly do a `git pull` from the web.


app/Http/Controllers/UpdateController.php
```diff
@@ -0,0 +1,71 @@
...
+		}
+		catch (Exception $e)
+		{
+			return Response::error($e->getMessage());
+		}
+
+		if ($up_to_date)
+		{
+			$output = [];
+			chdir('../');
```
kamil4: Don't you need to return to the original working directory when done?
If so, wouldn't it be easier for this chdir be part of the `command` (`cd .. && git/php`)?

routes/web.php
```diff
@@ -94,6 +94,8 @@
 Route::post('/api/Diagnostics',                 'DiagnosticsController@show');
 Route::get('/Diagnostics',                      'DiagnosticsController@show');
 
+Route::get('/api/Update',                       'UpdateController@do')->middleware('admin');
```
kamil4: So how will this be invoked? An admin-only menu option in the front end? If you have that code ready, can you push it so that I can actually try using this? :smile:

ildyria: Given that it is a get request you could also directly access it via https://example.com/api/Update :)
But as you suggested, I added a button in the front-end.

app/Http/Controllers/UpdateController.php
ildyria: We don't need to return to the original working directory in this specific case.
I prefer to use the `chdir` of php to allow a failure before actually trying the command. :)
The less I do in `exec` the happier I am.

ildyria:
Also note that the update via web feature is disabled by default (and will require an `artisan migrate`)

ildyria:
I merged it because this can mostly be tested ... in the branch master. I trust that the code is fine and should not break too much. :)

d7415:
> The less I do in `exec` the happier I am.

> I made it pretty defensive.

:+1:

d7415:
> A bit late now, but a skim read looks fine. It's a shame we can't trigger the migration from within Laravel rather than shelling out.

Have you tried
```php
Artisan::call('migrate');
```

?

ildyria:
I did not, but I don't know how this interact with the production mode. :/

PR #202, opened 2019-04-16
Use portable detection of duplicate keys

kamil4:
This will hopefully be a more portable way of detecting duplicate ids. `23505` is for PgSql, `23000` should be good for the rest.


ildyria:
I will need to add tests to CI for this.

PR #211, opened 2019-04-20
Check parent_id when creating albums

kamil4:
When checking upload permissions, check parent_id used with Album::add requests.

### DIFF
--- a/app/Http/Middleware/UploadCheck.php
+++ b/app/Http/Middleware/UploadCheck.php
@@ -67,7 +67,7 @@ class UploadCheck
     public function album_check(Request $request, int $id)
     {
         if ($request->has('albumID') || $request->has('parent_id')) {
-            $albumID = $request[$request->has('albumID') ? 'albumID' : 'parent_id'];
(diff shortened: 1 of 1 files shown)


app/Http/Middleware/UploadCheck.php
```diff
@@ -66,8 +66,8 @@ public function handle(Request $request, Closure $next)
      */
     public function album_check(Request $request, int $id)
     {
-        if ($request->has('albumID')) {
-            $albumID = $request['albumID'];
+        if ($request->has('albumID') || $request->has('parent_id')) {
+            $albumID = $request[$request->has('albumID') ? 'albumID' : 'parent_id'];
```
ildyria: While I see how this works and the beauty for this ternary, I think it is slightly too complex to read. I think this would be easier to read:
```php
$albumID = $request->has('albumID') ? $request[ 'albumID'] : $request['parent_id'];
```

kamil4: Sure, no problem. I'm not emotionally attached to that ternary :wink:

PR #212, opened 2019-04-21
Don't set file timestamp if takestamp is null

kamil4:
If an imported photo has no EXIF, `$info['takestamp']` is actually `null`.

I'm not sure if the existing two conditions will ever trigger but since they shouldn't hurt I just added one more...


ildyria:
This looks fine to me. :)

PR #218, opened 2019-04-24
MoreWording: Selected instead All

SerenaButler:
Missed an "All" ..

### DIFF
--- a/app/Locale/French.php
+++ b/app/Locale/French.php
@@ -45,7 +45,7 @@ final class French implements LangInterface
 			'RENAME'			=> 'Renommer',
 			'RENAME_ALL'		=> 'Renommer la sélection',
 			'MERGE'				=> 'Fusionner',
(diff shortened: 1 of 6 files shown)


app/Locale/French.php
```diff
@@ -45,7 +45,7 @@ static public function get_locale()
 			'RENAME'			=> 'Renommer',
 			'RENAME_ALL'		=> 'Renommer la sélection',
 			'MERGE'				=> 'Fusionner',
-			'MERGE_ALL'			=> 'Tout fusionner',
+			'MERGE_ALL'			=> 'Fusioner la sélection',
```
ildyria: Missing a `n` at `Fusioner` => `Fusionner` :smile: 
You have the correct form LIKE THE LINE JUST ABOVE.

SerenaButler: Hmpf ......

PR #225, opened 2019-04-30
Add Italian Translation

edoaxyz:
@ildyria here you have it


d7415:
Thanks for this. Some keys are still missing. I'm not sure what you used as your source, but I *think* [English.php](https://github.com/LycheeOrg/Lychee-Laravel/blob/master/app/Locale/English.php) is the most complete.Standing out are `'SEARCH'` and `'MORE'`, but there are probably more. Once they're all present, the Travis tests should pass.

d7415:
> I'm not sure what you used as your source,

Ah, looking at history, maybe the v3 code...

edoaxyz:
I've used [this](https://github.com/LycheeOrg/Lychee/blob/master/php/Locale/English.php) one as source

ildyria:
@edoaxyz if you installed with `composer install` instead of `composer install --no-dev` you have access to  `phpunit` which you can launch via `vendor/bin/phpunit`.
It is pretty usefull to see where things are missing. :)

PR #232, opened 2019-05-05
Always return albums using a tree structure

kamil4:
Note that this needs corresponding changes to the front end.

Also adds a missing sort for shared albums.


kamil4:
This is actually a great time for this discussion as far as I'm concerned as the next few items on my todo list have to do with user authentication and authorization (specifically, adjusting owner ids when admin add/moves objects around other users' albums, and restricting other users' access to non-public items).

So sure, let's do it right! The problem though is that I don't really understand the object model of Lychee-Laravel and what's clean and what isn't. So I would appreciate more in-depth explanation of where things should or shouldn't go. For instance, when you say _adding the Session dependency to the model_, what specifically do you mean? Any references to Session in `app/Album.php` or `app/Photo.php`? Does the same restriction apply to any of the files in subdirectories (I expect to need to add some to `ModelFunctions/` and `Http/Controllers/` but there's plenty there already so I guess those are fine?). Where should that `AlbumModelFunctions` class that you mention go -- under `ModelFunctions/`? There's already `AlbumFunctions.php` there -- is that no good?

ildyria:
Currently we have:
- Controllers
- Models
- Dependency Injection (https://designpatternsphp.readthedocs.io/en/latest/) which relies on Singletons.

A while ago @hermzz posted this: https://www.codementor.io/pardeepkumar905/working-with-thin-controller-and-fat-model-concept-in-laravel-nwl7ljsst

My view is:
- simple actions on the model stays on the model file e.g. `Photo.php`,
- if an action require more complex logic which depends on the model and additional information (e.g. the user id) then I would rather see it in `ModelFunctions/PhotoFunctions.php`
- avoid complex logic in the Controllers `Http/Controllers/...` and use them mostly to do checks or implement the logic at a higher level and rely on `ModelFunctions` for complex stuff.

Sometimes we have functions which does not depends on a Controller or a Model or are required in a very specific case, e.g.:
- `Lang` which manages all the localization
- `Image` for the case between GD and Imagick.
- `Extractor` to get the metadata or the info from Github

You will notice that most `ModelFunctions` are defined as singleton in `Providers/AppServiceProvider.php`

As I see it Models should be somewhat "stand alone".

The question left is, do we want to have `Session::get('UserID')` in ModelFunctions? For that one I will let @hermzz and @d7415 decide, I don't really have strong feelings.

d7415:
> For that one I will let @hermzz and @d7415 decide,

I have no idea.

ildyria:
If it were just on me, I would avoid also Session calls into the Model function and let them in the controller. But that is just me.

kamil4:
This was very helpful -- thanks!

kamil4:
I moved `get_albums()` to `AlbumFunctions`. Let met know if this is what you had in mind.

PR #233, opened 2019-05-05
Only delete user's photos from Unsorted

kamil4:
Ordinary users could delete everybody else's photos from Unsorted (even though they couldn't see them).

### DIFF
--- a/app/Http/Controllers/AlbumController.php
+++ b/app/Http/Controllers/AlbumController.php
@@ -336,7 +336,7 @@ class AlbumController extends Controller
 
 		$no_error = true;
 		if ($request['albumIDs'] == '0') {
-			$photos = Photo::Unsorted()->get();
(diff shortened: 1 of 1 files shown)


kamil4:
I'll have more generic questions for you about what should go where in the discussion in #232 but specifically regarding your middleware comment: I don't understand how that could do anything in this case?

`Album::delete` already goes through the `UploadCheck` middleware, which OKs the request, and rightly so, because every user with upload privileges is supposed to be able to invoke Delete on the Unsorted album. Isn't that all that the middleware can do: run a binary check on whether a request should be passed to a controller or not?

Here the solution was not to add another binary check, but to fix the implementation by adding a filter so that only the photos belonging to the user are purged from the album. Is that something that could be done in middleware?

ildyria:
Indeed you are right, I did not realize any user could invoke delete on unsorted. This would pass the middle ware obviously. My mistake.

PR #236, opened 2019-05-06
Cleaning up sessions

ildyria:
More refactoring...

### DIFF
--- a/app/ControllerFunctions/ReadAccessFunctions.php
+++ b/app/ControllerFunctions/ReadAccessFunctions.php
@@ -41,7 +41,7 @@ class ReadAccessFunctions
 	public function albums($albumID)
 	{
 		// if we are logged in then we have access
-		// we do not use this function to check if there is proper acess in case of sharing... yet?
+		// we do not use this function to check if there is proper access in case of sharing... yet?
 		if ($this->sessionFunctions->is_logged_in()) {
 			return 1; // access granted
 		}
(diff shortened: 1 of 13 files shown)


app/Http/Middleware/UploadCheck.php
```diff
@@ -17,134 +19,169 @@
...
+			return response('false');
+		}
+
+		$ret = $this->photo_check($request, $user_id);
+		if ($ret === false) {
+			return response('false');
+		}
+
+		// why do we want this ?
+		// I don't remember >_<
```
kamil4: It will only be used for the `Sharing::Delete` API call. It seems to check if the sharing(s) to be deleted concern the album(s) owned by the current user.

ildyria: Yep gotcha. Thanks for bracktracking that one.

app/ControllerFunctions/ReadAccessFunctions.php
```diff
@@ -0,0 +1,86 @@
...
+	 * if 2 : album is private
+	 * if 3 : album is password protected and require user input
+	 *
+	 * @param $albumID
+	 * @return int
+	 */
+	public function albums($albumID)
+	{
+		// if we are logged in then we have access
+		// we do not use this function to check if there is proper acess in case of sharing... yet?
```
d7415: s/acess/access/

d7415: And yes, some proper access control would be good, but complicated and out of scope here.

kamil4: _proper access control_ is something I'm highly interested in working on but if somebody beats me to it, I won't complain of course :smiley:. This clean-up should obviously go in first anyway. I'm hoping to try it out tonight.

PR #237, opened 2019-05-06
WIP Import from cli

ildyria:
I found this lovely work by @Spotlight0xff  :smiley:

### DIFF
--- a/app/Console/Commands/import.php
+++ b/app/Console/Commands/import.php
@@ -18,7 +18,7 @@ class import extends Command
 	 *
 	 * @var string
 	 */
-	protected $signature = 'lychee:import {dir : directory to import} {--flatten} {--album_id : Album ID to import to}
+	protected $signature = 'lychee:import {dir : directory to import} {--flatten} {--album_id= : Album ID to import to}
 							{--owner_id=0 : User}';
 
 	/**
--- a/app/Console/Commands/import.php
+++ b/app/Console/Commands/import.php
@@ -118,7 +118,7 @@ class import extends Command
 		}
 
 
-		$dir_iterator = new \RecursiveDirectoryIterator("/home/spotlight/photos", \FilesystemIterator::SKIP_DOTS);
+		$dir_iterator = new \RecursiveDirectoryIterator($directory, \FilesystemIterator::SKIP_DOTS);
 		$iterator = new \RecursiveIteratorIterator($dir_iterator, \RecursiveIteratorIterator::SELF_FIRST);
 		foreach ($iterator as $file) {
 			if ($file->isDir()) {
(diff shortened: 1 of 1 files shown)


ildyria:
I did not test this. @d7415 can you have a look ?

kamil4:
Yeah, with the work that I'm planning, there should be no need for messing around with `owner_id`. Imported photos will simply inherit the ownership from the album.

app/Console/Commands/import.php
```diff
@@ -0,0 +1,173 @@
...
+				'id',
+				'title',
+				'parent_id'
+			]);
+			$this->print_album($albums, 0 /* parent */, 0 /* indent */);
+			do {
+				if ($album_title != null) {
+					$this->line('No album '.$album_title.' found. Try again.');
+				}
+				$album_title = $this->ask('album to insert into (title)?');
```
d7415: If we're providing a list of IDs with the titles, wouldn't it be easier to ask for the ID? Also potentially less typing, less chance of special characters...

app/Console/Commands/import.php
```diff
@@ -0,0 +1,173 @@
...
+
+	/**
+	 * Execute the console command.
+	 *
+	 * @return mixed
+	 */
+	public function handle()
+	{
+		$directory = $this->argument('dir');
+		$flatten = $this->option('flatten');
```
d7415: Good idea, but unused.

ildyria:
I am also thinking if we could not re-use the following : https://github.com/LycheeOrg/Lychee-Laravel/blob/master/app/Http/Controllers/ImportController.php#L168 or so we don't have code duplication.

app/Console/Commands/import.php
kamil4: Also, I don't think that titles need to be unique...

Spotlight0xff:
Hey guys!

Yeah this was written in order to allow me to sync my google photos to the lychee instance.
I wanted a way to just sync all the photos (which are downloaded using [timeliner](https://github.com/mholt/timeliner), with subfolders corresponding to the years/months) to a single album.
I don't have experience writing PHP-Laravel code, so the code is definitely not ideal.

Very cool that you found it and consider it to include it somehow though!

app/Console/Commands/import.php
Spotlight0xff: Definitely! It was more for testing, not meant to be production-ready :)

alex-phillips:
FWIW, I'm working on a standalone CLI for lychee (both API versions supported) that allows uploading, tagging, searching, filtering, etc. of all photos in lychee. I will update the README with more info, but you can find the repo here: https://github.com/alex-phillips/lychee-cli

ildyria:
Niiiice :open_mouth:

ildyria:
This has been fixed with https://github.com/LycheeOrg/Lychee-Laravel/pull/427 I believe.

PR #238, opened 2019-05-07
video thumb2x: fixup 'Undefined offset: 1' error

ttys3:
for some reason,  a video may has no thumbnail, 

add this PR to prevent PHP report error.

### DIFF
--- a/app/Photo.php
+++ b/app/Photo.php
@@ -291,7 +291,7 @@ class Photo extends Model
 		// Parse paths
 		$photo['thumbUrl'] = Config::get('defines.urls.LYCHEE_URL_UPLOADS_THUMB').$this->thumbUrl;
 
-		if ($this->thumb2x == '1') {
+		if ($this->thumb2x == '1' && $this->thumbUrl !== '') {
 			$thumbUrl2x = explode(".", $this->thumbUrl);
 			$thumbUrl2x = $thumbUrl2x[0].'@2x.'.$thumbUrl2x[1];
 			$photo['thumb2x'] = Config::get('defines.urls.LYCHEE_URL_UPLOADS_THUMB').$thumbUrl2x;


kamil4:
This combination (`thumb2x == 1` while `thumbUrl` is empty) is in principle not supposed to be possible. However, there was a bug, which I recently fixed in aa2769664e9d3013ddbb3696f800255fb47e8c36, that could cause it.

Still, I would call this an instance of a corrupted database. @ildyria, should we be working this case around in the code or should we instead add a database migration that will fix it? If in the code, then the fix in the PR is incomplete -- there are at least two more places that would need to be fixed.

ildyria:
@kamil4 the less code we need to handle mistakes cases of the past the better.
 I would rather see a migration to fix the corrupt databases.

kamil4:
I thought about this some more. The reason why this happens is what in retrospect I consider to be a poor choice of the default value of the `thumb2x` column in the `photos` table: it defaults to `1`. I coded it that way because I thought it would simplify the database migration: most photos have a 2x thumbnail so I felt that defaulting to `1` and resetting to `0` for the few rows that don't have a 2x thumbnail would be more efficient.

An unintended side effect, however, is that if something goes unexpectedly wrong while adding a new photo and a thumb is not generated but the operation goes through, we may end up with a row with no thumb but with `thumb2x` set to `1`, as reported by @ttys3. As I wrote, I fixed one such bug recently, but in the end this just seems needlessly fragile.

So I'm wondering if it wouldn't be a good idea to change the default to `0`, thus preventing such problems from recurring in the future?

d7415:
> So I'm wondering if it wouldn't be a good idea to change the default to `0`, thus preventing such problems from recurring in the future?

Sounds sensible to me.

kamil4:
@ttys3 I just submitted PR #243, which is meant to be a safer/more complete alternative to what you proposed. Would you mind trying it out? You will need to run `php artisan migrate` after checking it out -- it should locate and fix any problematic entries in the `photos` table.

ttys3:
@kamil4  thanks for the PR.  it's better than my monky patch.

PR #239, opened 2019-05-07
Feature: import from server or url add video support

ttys3:
in normal photo upload mode,
we can upload video.

but when import from server , or  import from URL,
this code just ignored the video, 
this PR will  keep the feature consistent


kamil4:
I never used that functionality before but I just tested it with a sample video file and it worked fine.

PR #243, opened 2019-05-08
Use a safe default for thumb2x

kamil4:
Also fix any corrupt entries in the photos table.

This is meant to be a better alternative to #238.

### DIFF
--- a/database/migrations/2019_05_07_202107_fix_thumb2x_default.php
+++ b/database/migrations/2019_05_07_202107_fix_thumb2x_default.php
@@ -1,0 +1,47 @@
+<?php
+
+use App\Photo;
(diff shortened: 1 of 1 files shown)


ildyria:
Well it should be speeding up: you do 1 SQL request instead of pulling all the photos e.g. 1000 and then executing 50 SQL request to fix 50 pictures.

d7415:
> Well it should be speeding up: you do 1 SQL request instead of pulling all the photos e.g. 1000 and then executing 50 SQL request to fix 50 pictures.

No, I meant *that specific line*. The `WHERE thumb2x=1` part. But like I say, that part's negligible in whichever direction. The batch job in general should definitely be an improvement.

PR #244, opened 2019-05-08
Ensure that ownership follows parent album

kamil4:
This is a first attempt to fix an issue where admin could create/move photos and albums around resulting in mixed-ownership album content, etc.

This patch ensure that all add, move, and merge operations performed by admin change the ownership of moved objects to that of the destination album, if any.

I tried to take advantage of the new session infrastructure but I wasn't always successful, e.g., in AlbumFunctions (is it possible to invoke SessionFunction from them?).

### DIFF
--- a/app/Http/Controllers/AlbumController.php
+++ b/app/Http/Controllers/AlbumController.php
@@ -7,6 +7,7 @@ use App\Album;
 use App\Logs;
 use App\ModelFunctions\AlbumFunctions;
 use App\ModelFunctions\Helpers;
+use App\ModelFunctions\SessionFunctions;
 use App\Photo;
 use App\Response;
 use Illuminate\Http\Request;
@@ -22,14 +23,21 @@ class AlbumController extends Controller
 	 */
 	private $albumFunctions;
 
+	/**
+	 * @var SessionFunctions
+	 */
+	private $sessionFunctions;
+
 
 
 	/**
 	 * @param AlbumFunctions $albumFunctions
+	 * @param SessionFunctions $sessionFunctions
 	 */
-	public function __construct(AlbumFunctions $albumFunctions)
+	public function __construct(AlbumFunctions $albumFunctions, SessionFunctions $sessionFunctions)
 	{
 		$this->albumFunctions = $albumFunctions;
+		$this->sessionFunctions = $sessionFunctions;
 	}
 
 
@@ -384,14 +392,25 @@ class AlbumController extends Controller
 		$no_error = true;
 		foreach ($photos as $photo) {
 			$photo->album_id = $albumID;
+			if ($this->sessionFunctions->is_admin()) {
+				// Admin can merge albums between users.  Make sure that the
+				// ownership changes in the process.
+				$photo->owner_id = $album->owner_id;
+			}
 			$no_error &= $photo->save();
 		}
 
 		$albums = Album::whereIn('parent_id', $albumIDs)->get();
 		$no_error = true;
-		foreach ($albums as $album) {
-			$album->parent_id = $albumID;
-			$no_error &= $album->save();
+		foreach ($albums as $album_t) {
+			$album_t->parent_id = $albumID;
+			if ($this->sessionFunctions->is_admin()) {
+				// Admin can merge albums between users.  Make sure that the
+				// ownership changes in the process.
+				$album_t->owner_id = $album->owner_id;
+				$no_error &= $this->albumFunctions->setContentsOwner($album_t->id, $album->owner_id);
+			}
+			$no_error &= $album_t->save();
 		}
 
 		$albums = Album::whereIn('id', $albumIDs)->get();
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/AlbumController.php
```diff
@@ -384,14 +392,25 @@ function merge(Request $request)
 		$no_error = true;
 		foreach ($photos as $photo) {
 			$photo->album_id = $albumID;
+			if ($this->sessionFunctions->is_admin()) {
```
ildyria: actually this check is unnecessary. If you are a normal user:
`$photo->owner_id` = `$album->owner_id`
So the result will be the same.

app/Http/Controllers/AlbumController.php
```diff
@@ -384,14 +392,25 @@ function merge(Request $request)
...
 		}
 
 		$albums = Album::whereIn('parent_id', $albumIDs)->get();
 		$no_error = true;
-		foreach ($albums as $album) {
-			$album->parent_id = $albumID;
-			$no_error &= $album->save();
+		foreach ($albums as $album_t) {
+			$album_t->parent_id = $albumID;
+			if ($this->sessionFunctions->is_admin()) {
```
ildyria: idem as above. :)
The only gain you have is a slight code speed up for non-admin users.

app/Http/Controllers/AlbumController.php
```diff
@@ -417,18 +436,29 @@ function move(Request $request)
...
 
+		$ownerId = null;
 		if ($albumID != 0) {
 			$album = Album::find($albumID);
 			if ($album === null) {
 				Logs::error(__METHOD__, __LINE__, 'Could not find specified albums');
 				return 'false';
 			}
+
+			if ($this->sessionFunctions->is_admin()) {
```
ildyria: idem.

app/Http/Controllers/AlbumController.php
```diff
@@ -417,18 +436,29 @@ function move(Request $request)
...
+				// Admin can move albums between users.  Make sure that the
+				// ownership changes in the process.
+				$ownerId = $album->owner_id;
+			}
 		}
 
 		$albums = Album::whereIn('id', $albumIDs)->get();
 		$no_error = true;
 		foreach ($albums as $album) {
 			$album->parent_id = ($albumID == 0 ? null : $albumID);
```
ildyria: I will redo this part.

app/Http/Controllers/PhotoController.php
```diff
@@ -282,9 +282,23 @@ function setAlbum(Request $request)
 
 		$albumID = $request['albumID'];
 
+		$album = null;
+		if ($this->sessionFunctions->is_admin() && $albumID !== '0') {
```
ildyria: you don't need to check if it is an admin or not.

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -41,15 +41,26 @@ public function is_smart_album($albumID)
 	 */
 	public function create(string $title, int $parent_id): Album
 	{
-		$num = Album::where('id', '=', $parent_id)->count();
-		// id cannot be 0, so by definition if $parent_id is 0 then...
```
ildyria: True but it is safer: What if you create an album with a wrong `$parent_id`?

d7415: I'm amused that the line you picked is your own comment, but I see your point. Though this may go back to the other question of how much checking do we want to do for bad data... I guess if one user/browser/session deleted an album between loading the page and creating the subalbum this could be required..

ildyria:
> I tried to take advantage of the new session infrastructure but I wasn't always successful, e.g., in AlbumFunctions (is it possible to invoke SessionFunction from them?).

If you have it in the constructor then it should be fine.

app/Http/Controllers/AlbumController.php
kamil4: The reason I did it like that was that in some cases the Album object was not readily available and needed to be looked up. So I didn't want to slow down the common case (a regular user moving their own objects around) to handle a corner case. But I take your point that it was making the code look more complicated.

ildyria: :)

app/ModelFunctions/AlbumFunctions.php
kamil4: I see that you rewrote this code in the meantime, but here was my thinking last night:

1. If `$parent_id` is wrong then we'll never get here because the `UploadCheck` middleware will flag it. Although now that I think about it, it will let it right through for the admin.

2. The test seemed inconsistent with the rest of the code. It seems to me that in most other places if an album doesn't exist, we fail hard. Here we had this weird fallback where we instead create the album at the top level. This is a case that's not supposed to happen so why not fail and return an error to the user?

ildyria: I'm fine with failing hard too. :)

kamil4:
OK, I'll test it again tonight with @ildyria's changes and if it continues working fine, I'll merge it.

Note that it would help with the testing if #232 (and its front end partner https://github.com/LycheeOrg/Lychee-front/pull/95) were merged first. I encountered a few situations last night where the front end was failing in the corner cases that are inconsistent with the current albums representation.

PR #255, opened 2019-05-17
Fix user access

ildyria:
Work in Progress.

Just making a draft of the pull request so we can see the complete changes.


kamil4:
Thanks! This is definitely ready for testing at this point, and I was going to submit a PR later today summarizing the bugs fixed and asking for input on additional issues I identified that should be dealt with in the future (you probably noticed the FIXMEs and TODOs in the comments).

kamil4:
OK, I believe this is ready for merging now, but I thought I would provide some additional explanation:

I modified `Album::gen_thumbs` to take the list of [sub]albums to consider. This makes it possible to avoid leaking private or password protected pictures.

`Album::get_all_sub_albums` is now a private method used only by `update_min_max_takestamp`. All other previous callers use the new `get_sub_albums` from `AlbumFunctions`, which considers access permissions and so avoids leaking private pictures. BTW, there was a subtle bug in `get_all_sub_albums` that prevented it from returning any subalbums more than one level deep, so some min/max takestamps in existing deeply nested galleries are probably wrong, and some thumbs that should've been displayed were not, so don't be surprised if some existing albums suddenly display different thumbs than before.

Subalbums of password-protected albums are no longer being leaked to the front end via `albums.json.albums[].albums`. That field is not available anymore until a password is provided, just like album thumbs.

I changed the behavior of password-protected albums, taking advantage of the fact that in v4 the server caches their status in a Session variable. The new behavior is that once a password has been provided, the album is effectively unlocked for the duration of the session and it acts like any other album, including the displaying of thumbs in the albums view. This requires a (trivial) patch to the front end, which I will also be submitting. A much larger patch to the front end to prevent it from attempting to modify other user's albums is on my personal TODO list, BTW. It's not urgent as the server already correctly enforces upload restrictions.

Questions:

What is the `DemoController` supposed to do? I've never used it. I tried to test it by appending `/demo` to the URL of a gallery but I got some semi-debug output dump.

What is that `sysstamp` field we send to the front end in `AlbumFunctions::prepare_albums` and `SearchController::search` supposed to be used for? I can't find any references to it in the JS code in the front end. :confused:

`update_min_max_takestamp` and `reset_takestamp` strike me as criminally inefficient :wink:. I mean, I understand that an SQL query will be much faster than PHP code, but if that query is not needed 99% of the time, perhaps a few extra lines of PHP code should be in order, especially as we hear about people with tens of thousands of pictures in the gallery? Also, `update_min_max_takestamp` is clearly broken right now: add a picture to a subalbum and see what happens to the takestamps of the parent album -- nothing.

ildyria:
May I suggest to add a migration which would update all those missing min max takestamp ? :)

d7415:
> What is the `DemoController` supposed to do? I've never used it. I tried to test it by appending `/demo` to the URL of a gallery but I got some semi-debug output dump.

That makes me wonder if it's just a demo provided by Laravel... @ildyria ?

> What is that `sysstamp` field we send to the front end in `AlbumFunctions::prepare_albums` and `SearchController::search` supposed to be used for? I can't find any references to it in the JS code in the front end. 

I wonder if that's meant to be `sysdate`?

ildyria:
> > What is the `DemoController` supposed to do? I've never used it. I tried to test it by appending `/demo` to the URL of a gallery but I got some semi-debug output dump.
> 
> That makes me wonder if it's just a demo provided by Laravel... @ildyria ?

It generates the api answers that you can then use to simulate the api calls. I made it for https://lycheeorg.github.io/demo/

app/ModelFunctions/SessionFunctions.php
```diff
@@ -132,11 +146,34 @@ public function has_visible_album($albumID)
 
 		$visible_albums = Session::get('visible_albums');
 		$visible_albums = explode('|', $visible_albums);
-		$found = false;
-		foreach ($visible_albums as $visible_album) {
-			$found |= ($visible_album == $albumID);
+
+		return in_array($albumID, $visible_albums);
```
d7415: :+1:

d7415:
> It generates the api answers that you can then use to simulate the api calls. I made it for https://lycheeorg.github.io/demo/

Ahhh, of course!

kamil4:
@ildyria This patch only fixes the takestamps partially as they weren't my focus here. I agree with your request to add a database migration, but can we delay it to a future PR that actually fully fixes the takestamps? For now, a workaround would be to trigger `reset_takestamp`, e.g., by deleting any image (possibly a just added one). That should rebuild all takestamps in all albums (you see why I call this stuff criminally inefficient? :smiley:), hopefully correctly this time (I haven't actually checked).

@d7415 `sysdate` is generated as well, and in a different format, so it's not a case of a typo. I checked with v3 and it doesn't pass `sysstamp` to the front end either. I think `sysstamp` must be a mistake, or an idea that was abandoned halfway through. I'll remove it in a future PR after verifying that it doesn't brake anything.

So are we good merging this PR at this point?

d7415:
> For now, a workaround would be to trigger `reset_takestamp`

Without looking, could we call that in a migration?

> @d7415 `sysdate` is generated as well, and in a different format, so it's not a case of a typo. I checked with v3 and it doesn't pass `sysstamp` to the front end either. I think `sysstamp` must be a mistake, or an idea that was abandoned halfway through. I'll remove it in a future PR after verifying that it doesn't brake anything.

So it is... It looks like v3 uses `sysstamp` in the DB and `sysdate` for the front end (and converts accordingly).

> So are we good merging this PR at this point?

No objections here. I'll leave it to you and @ildyria.

app/ControllerFunctions/ReadAccessFunctions.php
```diff
@@ -45,7 +41,12 @@ public function __construct(SessionFunctions $sessionFunctions)
 	 */
 	public function album($albumID, bool $obeyHidden = false)
 	{
-		if (in_array($albumID, array('f', 's', 'r', '0'))) {
+		if (in_array($albumID, array(
+			'f',
+			's',
+			'r',
+			'0'
+		))) {
```
d7415: Is there a reason for laying the array out like this?

ildyria: nah it's just the formatting done by my Phpstorm when I do CTRL+L ^^"

kamil4:
I tested the branch with @ildyria's changes as well as with my change to remove the two instances of `sysstamp`. I believe this is ready for merging now.

kamil4:
> Without looking, could we call that in a migration?

Yes, absolutely. There's a chance though that I'll move that whole function to the migration file, as right now I don't expect that we'll need to keep such a... sledgehammer approach in the back end.

Issue #49, opened 2018-09-17
How to allow public to use tag search function?


ildyria:
From : https://github.com/electerious/Lychee/issues/738

> I have tagged various photos with descriptive labels. I'd like public users to be able to search on tags and / or photo titles to find relevant images as other gallery systems allow. Unless the user is logged in the search box is invisible.

scratcher28:
I suggest implementing it with teamtnt/tntsearch.

scratcher28:
No deps, no need to run full reindex, check docs.
I definitely will use it in my fork.

copperschnack:
OK, let's see. I just created my first pull request above.

ildyria:
Hi, thank you for your PR on Lychee-front and Lychee.
Can you also port it for the current repository (aka this one) ? :)
If you need help feel free to drop by on the gitter (https://gitter.im/LycheeOrg/Lobby) and ping either me or @d7415 or @kamil4  :+1:

ildyria:
Fixed.

PR #256, opened 2019-05-20
public search functionality w/ setting, see #49

copperschnack:
Hi @ildyria,

we tried to integrate our backend PR for https://github.com/LycheeOrg/Lychee-Laravel/issues/49 into the current Lychee-Laravel repository.

malte.

PS: I do not like query builders too much ;-)

### DIFF
--- a/resources/views/gallery.blade.php
+++ b/resources/views/gallery.blade.php
@@ -98,7 +98,6 @@
         <a class="button button--eye" id="button_share" title="{{ $locale['SHARE_PHOTO'] }}">
             <svg class="iconic"><use xlink:href="#eye"></use></svg>
         </a>
-        <a class="header__divider"></a>
         <a class="button button--info" id="button_info" title="{{ $locale['ABOUT_PHOTO'] }}">
             <svg class="iconic"><use xlink:href="#info"></use></svg>
         </a>


app/Http/Controllers/SearchController.php
```diff
@@ -85,7 +86,13 @@ public function search(Request $request)
...
 		 */
 		// for now we only look in OUR pictures
-		$query = Photo::where('owner_id', '=', $id);
+		$query = Photo::where('id', '>', 0);
+		$query->where(function (Builder $query) use ($id) {
+			$query->where('owner_id', '=', $id);
+			if (Configs::get_value('public_search', '0') == '1') {
+				$query = $query->orWhere('public', '>=', 0);
+			}		
+		});
```
kamil4: This is not the right way to do it -- but I understand where you're coming from, since the original code was taking shortcuts so it wasn't correct either (hence the `for now we only look in OUR pictures` comment). I can think of at least two problems here:

1. Checking the `public` flag of a `Photo` is insufficient. For photos in albums, it's actually the `public` flag of the _album_ that determines if an image is public or not.

2. This will ignore shared albums (albums that another user is sharing privately with the current user). But this problem was there before your changes as well.

I'll say more about how I think this _should_ be done below, when talking about Albums.

app/Http/Controllers/SearchController.php
```diff
@@ -85,7 +86,13 @@ public function search(Request $request)
 		 * Photos
 		 */
 		// for now we only look in OUR pictures
-		$query = Photo::where('owner_id', '=', $id);
+		$query = Photo::where('id', '>', 0);
```
kamil4: What is the purpose of this check? Aren't ids always greater than 0 ?

app/Http/Controllers/SearchController.php
```diff
@@ -100,15 +107,23 @@ function (Builder $query) use ($id, $escaped_term) {
 		if ($photos != null) {
 			$i = 0;
 			foreach ($photos as $photo) {
-				$return['photos'][$i] = $photo->prepareData();
-				++$i;
+				if (Session::has('UserID') || $photo->get_public()) {
```
kamil4: This should probably use `ReadAccessFunctions::photo` instead...

app/Http/Controllers/SearchController.php
```diff
@@ -100,15 +107,23 @@ function (Builder $query) use ($id, $escaped_term) {
...
+					++$i;
+				}
 			}
 		}
 
 		/**
 		 * Albums
 		 */
-		$query = Album::where('owner_id', '=', $id);
+		$query = Album::where('id', '>', 0);
```
kamil4: Some question as about Photo query above.

app/Http/Controllers/SearchController.php
```diff
@@ -100,15 +107,23 @@ function (Builder $query) use ($id, $escaped_term) {
...
+		$query->where(function (Builder $query) use ($id) {
+			$query->where('owner_id', '=', $id);
+			if (Configs::get_value('public_search', '0') == '1') {
+				$query = $query->orWhere('public', '=', 1);
+			}		
+		});
```
kamil4: Again, this is not correct:

1. It will leak hidden albums (and their contents).

2. It will leak contents from password-protected albums.

3. It will not return shared albums (that issue predates your PR).

I've been thinking as well about how to make Search safe in public mode, though I never got around to doing anything about it. But here's how I think this needs to be done:

1. The code needs to recursively traverse the albums tree from the root down, checking the read access permissions at every level. Shared albums should be included.

2. In the process, an array of `albumIDs` of accessible albums needs to be constructed. This should include all subalbums. Hidden albums or their subalbums should not be included (even if public and not hidden). The contents of password-protected albums should not be included (again, even if public and not password-protected), unless the password has already been provided.

3. Subsequent queries, both on the Albums and Photos (see above), need to use that array as a filter.

We already had such a filter: by `owner_id`, but it was not sufficient and obviously won't work in the public mode.

Actually, we already have the code that does all of the above in other places, specifically:

- `App\Http\Controllers\AlbumsController::get` extracts all the accessible top-level albums
- `App\ModelFunctions\AlbumFunctions::prepare_albums` extracts the accessible sub-albums (we only need the access check and `get_albums` from this one)

I think the right way to do it would be to refactor `App\Http\Controllers\AlbumsController::get`, moving the bulk of its code to a new method that could also be used from search.

kamil4: One more thing:
- `AlbumFunctions::get_sub_albums` can probably be useful here as, unlike `get_albums`, it returns results in the format we need for SQL queries...

app/Http/Controllers/SearchController.php
copperschnack: As I said:

> PS: I do not like query builders too much ;-)

I added the first dummy condition to get the brackets around the first non-dummy WHERE-condition with the OR-components, ie. getting from "here" ...

`select * from `photos` where `id` > ? and `owner_id` is null or `public` >= ? and (`title` like ? or `description` like ? or `tags` like ?) [9,0,"%privat%","%privat%","%privat%"] `

... to "there":

`select * from `photos` where `id` > ? and (`owner_id` is null or `public` >= ?) and (`title` like ? or `description` like ? or `tags` like ?) [0,0,"%privat%","%privat%","p%rivat%"] `

app/Http/Controllers/SearchController.php
copperschnack: > 1. Checking the `public` flag of a `Photo` is insufficient. For photos in albums, it's actually the `public` flag of the _album_ that determines if an image is public or not.

And if it was sufficient, it would even beimplemented wrong with `public >= 0`. This was just another trick for the query generator to first fetch all photos. Non-public images are not getting into the response object, see your next comment: https://github.com/LycheeOrg/Lychee-Laravel/pull/256/files#r285669443

app/Http/Controllers/SearchController.php
copperschnack: > PS: I do not like query builders too much ;-)

Sorry... the WHERE-brackets again, see above.

app/Http/Controllers/SearchController.php
copperschnack: OK, thanks. Here it gets pretty obvious that I neither know Laravel generally nor Lychee particularly in depth. I tried to implement the public search with minimal invasion into the existing code - which clearly introduces some security risks. I will try to comprehend your comment and see if I can do some adjustments. 

By now, I did not realize that there now is an album hirarchy or a possibility to share between registered users. I apologize for the shortcuts.

ildyria: hahaha no worries. Indeed the v4 has multi user support (locked or not, with upload right or not) and sharing between users also available. And we also have some other plans in mind e.g. rss feed...

ildyria: I will try to improve your Laravel code. :)

kamil4:
FYI, I'm working on getting this code in shape; it'll probably be another day before it's ready...

copperschnack:
> FYI, I'm working on getting this code in shape; it'll probably be another day before it's ready...

Thanks - I already found another problem with the implementation: If the public search finds a public photo in a private album, an API error is thrown when the photo's detail page gets invoked :-|

kamil4:
> If the public search finds a public photo in a private album, an API error is thrown when the photo's detail page gets invoked

I'm not sure if I fully understand the scenario you're describing (do you mean just clicking on a photo in the search results gives an error?) but the approach I'm pursing is that search will **not** return public photos in private albums. Such photos are effectively hidden -- you normally need to know the exact URL to view them. Only photos that you can get to through the browsing of the albums tree from the root level will be considered. Otherwise we could leak contents that isn't meant to be seen by everybody.

kamil4:
Sorry, I don't know what just happened. I tried to push my changes to copperschnack:master from my local copperschnack-master but instead git pushed from my master.

kamil4:
Sigh, I think I understand what I did wrong. I did:
```
git push --force copperschnack master
```
But I think I should've done:
```
git push --force copperschnack copperschnack-master:master
```
Anyway, it looks like github won't let me fix it (it auto-closed the PR so it won't let me push to @copperschnack's repo anymore). Do I need to open a new PR for this one as well?

ildyria:
>  Do I need to open a new PR for this one as well?

Yep.

PR #258, opened 2019-05-23
Fix footer

ildyria:


ildyria:
Seems to be still a bit buggy. DO NOT MERGE.

PR #260, opened 2019-05-26
Add username to Session::init response

kamil4:
This is a prerequisite to larger changes on the front end side to prevent it from attempting to edit shared albums.


kamil4:
Given its triviality and the fact that https://github.com/LycheeOrg/Lychee-front/pull/106 depends on it, I'm merging without an explicit review.

PR #264, opened 2019-05-28
added concat_space option to PhpCs config

scratcher28:
Just want to make the code more readable ;)


ildyria:
Well this solve our troubles... Next time create a new branch so we don't need to push on your master. This will save us time.

PR #265, opened 2019-05-28
slightly better readme

ildyria:
Just slightly better looking header


ildyria:
No need to check Travis, only readme.md is modified.

PR #267, opened 2019-05-29
Public search

kamil4:
New PR for public search, after I messed up #256...

This one's a little messed up as well, as it includes bogus deletions in public/dist (I hate rebasing...), but hopefully we have a way to deal with that...


app/ModelFunctions/AlbumFunctions.php
```diff
@@ -340,4 +348,55 @@ public function get_sub_albums($parentAlbum, $return = array())
...
+			'albums' => null,
+			'shared_albums' => null,
+		);
+
+		if ($this->sessionFunctions->is_logged_in()) {
+			$id = $this->sessionFunctions->id();
+			$user = User::find($id);
+
+			if ($id == 0) {
+				$return['albums'] = Album::where('owner_id', '=', 0)
```
ildyria: aren't we missing a `with(['children'])` here?

d7415:
> as it includes bogus deletions in public/dist (I hate rebasing...), but hopefully we have a way to deal with that...

We should just be able to regenerate the files once the front end is merged and add them in another commit before we squash.

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -340,4 +348,55 @@ public function get_sub_albums($parentAlbum, $return = array())
...
+					->orderBy('owner_id', 'ASC')
+					->orderBy(Configs::get_value('sortingAlbums_col'), Configs::get_value('sortingAlbums_order'))
+					->get();
+			} else {
+				if ($user == null) {
+					Logs::error(__METHOD__, __LINE__, 'Could not find specified user ('.Session::get('UserID').')');
+
+					return null;
+				} else {
+					$return['albums'] = Album::where('owner_id', '=', $user->id)
```
ildyria: aren't we missing a `with(['children'])` here?

app/Http/Controllers/SearchController.php
```diff
@@ -72,51 +91,94 @@ public function search(Request $request)
...
-		/**
-		 * Albums.
+		/*
+		 * Photos.
+		 *
+		 * Begin by eliminating albums we can see but can't access (i.e.,
+		 * password-protected ones).
 		 */
-		$query = Album::where('owner_id', '=', $id);
+		for ($i = 0; $i < count($albumIDs);) {
```
ildyria: isn't this for loop useless ?
I mean it removes albumIDs from which we do not have access to the album.
However, unless I'm wrong **by construction**, they have this property. See line 103 104 111 and 112.

app/ModelFunctions/AlbumFunctions.php
kamil4: Pardon my ignorance, but I never actually understood what those `with` clauses do so I honestly can't answer your question :-(. All I can say is that I just moved this code from the `AlbumsController`, so that's the way it was there and it seemed to work fine :smile:.

app/Http/Controllers/SearchController.php
kamil4: This is a subtle point. Lines 103-104 and 111-112 ensure that we do not descend *into* an inaccessible album. However, the id of that inaccessible album _is_ included in `albumIDs` (see lines 102 and 110). That way we can match on the name of that inaccessible album (so, say, if every password-protected album has "with password" in its name, searching for "password" will return their complete list).

When searching for photos, however, which are inside the albums, we need to remove those inaccessible albums so that we don't leak the contents of password-protected albums.

Believe me, these code paths were tested and in fact this loop was added after the rest of the code, when I noticed during testing that the code was leaking photos.

app/ModelFunctions/AlbumFunctions.php
ildyria: Have a look at that: https://laravel.com/docs/5.8/eloquent-relationships#eager-loading :)

kamil4: Ah, now I understand why I couldn't understand them -- they have no semantic meaning, they are optimization hints :smile:. Right, so I guess `children` is likely to come in handy given that the callers will travel down the subalbum tree. Should we indicate `owner` as well, like we do for shared albums? I'm guessing it may come in handy when checking permissions using `readAccessFunctions->album()`?

kamil4: OK, I tried adding `with(['children'])` and also `with(['owner', 'children'])` but I failed to notice any significant difference in the total number of SQL queries performed. Is it still worth adding then?

PR #270, opened 2019-06-04
Add disk usage to the Diagnostic pages

### LINKED ISSUE #269: [Enhancement] Display disk space usage
Hi,

I've installed this project at home on a Raspberry Pi to share photos with my family. It's a really nice tool so far, great job for that :+1:

I have a limited amount of disk space, so I would like to see how much disk space is used by media on my instance of Lychee, so that I can expect when I need to extend the available space.

Is it possible to have this information on the current version ?

ildyria:
Fixes #269


app/Metadata/DiskUsage.php
```diff
@@ -0,0 +1,165 @@
...
+	 * @return string
+	 */
+	public function getSymbolByQuantity($bytes)
+	{
+		$symbols = array(
+			'B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB',
+		);
+		$exp = intval(floor(log($bytes) / log(1024)));
+
+		return sprintf('%.2f %s', ($bytes / pow(1024, floor($exp))), $symbols[$exp]);
```
kamil4: I think that `floor` is not needed here since you're already doing `intval(floor())` above.

app/Metadata/DiskUsage.php
```diff
@@ -0,0 +1,165 @@
...
+	 */
+	public function get_total_space()
+	{
+		if (!$this->is_win()) {
+			$ds = disk_total_space('/');
+
+			return $this->getSymbolByQuantity($ds);
+		}
+
+		return 'NAN';
```
kamil4: Shouldn't we be querying `C:\` on Windows, or something?
I don't really care, I thought I would just throw in a question :smiley:.

app/Metadata/DiskUsage.php
```diff
@@ -0,0 +1,165 @@
...
+	public function get_lychee_space()
+	{
+		// we currently are in public. we want Lychee
+		$ds = $this->getTotalSize('../');
+
+		return $this->getSymbolByQuantity($ds);
+	}
+
+	/**
+	 * Return the space takem by the upload folder (Big, Medium, Small, Thumbs).
```
kamil4: Typo. `takem` -> `taken`.

app/Metadata/DiskUsage.php
ildyria: gotcha.

app/Metadata/DiskUsage.php
ildyria: We could check `C:\` I don't really care either.

kamil4:
I like this approach and I think it's a good idea to put it in Diagnostics. There may be value in calculating usage per-album on demand in addition to this, but that's a task for another day...

ildyria:
> There may be value in calculating usage per-album on demand in addition to this, but that's a task for another day...

That would fit a statistic page :)

PR #273, opened 2019-06-06
Fix takestamps

kamil4:
This should correct all instances of incorrect album min/max takestamps and as a bonus be computationally much less expensive.

### DIFF
--- a/app/Album.php
+++ b/app/Album.php
@@ -4,6 +4,7 @@
 
 namespace App;
 
+use App\Logs;
 use Eloquent;
 use Exception;
 use Illuminate\Database\Eloquent\Builder;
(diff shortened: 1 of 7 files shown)


app/Album.php
```diff
@@ -246,9 +245,94 @@ public function update_min_max_takestamp()
...
-	 * the albums that were modified, plus recursively their parents).
+	 * Update album's min_takestamp and max_takestamp based on changes made
+	 * to the album content.  If needed, recursively updates parent album(s).
+	 *
+	 * @param array $takestamps: an array with the takestamps of changed
+	 *                           elements; for albums needs to include both min and max takestamps
+	 *                           (including null elements in the array is safe)
+	 * @param bool  $adding:     true if adding new content, false if removing
+	 */
+	public function update_takestamps(array $takestamps, bool $adding)
```
ildyria: This function should return true of false depending on failure.

kamil4: Should be fixed with the latest commit...

PR #277, opened 2019-06-12
more tests

ildyria:
This should be fine. Just added more tests.


ildyria:
PR crashes on Travis side but this is not due to the actual changes.

PR #278, opened 2019-06-13
Swap width and height for rotated videos

kamil4:
Video preview was being displayed incorrectly because `width` and `height` were not swapped to match the video rotation tag.


ildyria:
agreed.

Issue #100, opened 2019-03-01
Logged in Session last only a few hours


SerenaButler:
Lychee3 Logged in Sessions last forever, Laravel just for some hours.

kamil4:
Worse, you get a cryptic "Server API error" or something message. Since I tend to mess with server APIs when coding, I'm never initially sure if it's the timeout or if I introduced a new bug...

ildyria:
It is "user settable"

in `.env.example` : https://github.com/LycheeOrg/Lychee-Laravel/blob/master/.env.example#L24
```php
SESSION_LIFETIME=120
```

in `config/session.php` : https://github.com/LycheeOrg/Lychee-Laravel/blob/master/config/session.php
```php
    /*
    |--------------------------------------------------------------------------
    | Session Lifetime
    |--------------------------------------------------------------------------
    |
    | Here you may specify the number of minutes that you wish the session
    | to be allowed to remain idle before it expires. If you want them
    | to immediately expire on the browser closing, set that option.
    |
    */
    'lifetime' => env('SESSION_LIFETIME', 120),
    'expire_on_close' => false,
```

May require documentation, if someone want to add this to the wiki.

kamil4:
I think the cryptic error message the user gets after the session times out _is_ an issue though. Do you think it would be hard to catch it and display a neat "Session timed out" kind of information to the user?

SerenaButler:
I never got a cryptic message though. I were just logged out.

SerenaButler:
There are situations "loosing" the session while in an non-public area. Reloading the page can then result in an empty page or /indeed) in a cryptic message. If session is gone reloading a non public page should refer to the next higher public page.

ildyria:
that should be handled by Lychee-front.

SerenaButler:
Ok :-)

kamil4:
I had a look at it last night. After the session times out, the next request from the front end results in the following error in laravel.log:
```
local.ERROR: The payload is invalid. {"exception":"[object] (Illuminate\\Contracts\\Encryption\\DecryptException(code: 0): The payload is invalid. at /home/kamil/public_html/tmp/Lychee-Laravel/vendor/laravel/framework/src/Illuminate/Encryption/Encrypter.php:195)
```
Followed by a stacktrace 51 frames deep...

[...]

PR #279, opened 2019-06-13
Catch DecryptException and notify the front end

kamil4:
Server side of the fix for #100. Front end side will follow.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -3,6 +3,7 @@
 namespace App\Exceptions;
 
 use Exception;
+use Illuminate\Contracts\Encryption\DecryptException;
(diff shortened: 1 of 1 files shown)


app/Exceptions/Handler.php
```diff
@@ -47,6 +49,10 @@ public function report(Exception $exception)
 	 */
 	public function render($request, Exception $exception)
 	{
+		if ($exception instanceof DecryptException && $exception->getMessage() === 'The payload is invalid.') {
+			return response()->json(['error' => 'Session timed out'], 400);
```
ildyria: I'm starting to think that we should consider uniforming the output format of Lychee and Lychee-Laravel to a Json array in all cases instead of just returning an id or true or false or an Json array...
But this should obviously not be done in that PR. :)

PR #287, opened 2019-06-17
fixes lack of documentation

### LINKED ISSUE #40: Add code documentation
Even if the functions are pretty self explanatory. It would be nice to actually have a small documentation above each one.

If that could be done before the release of the V4 that would be top though.

ildyria:
fixes #40

### DIFF
--- a/app/Http/Controllers/PhotoController.php
+++ b/app/Http/Controllers/PhotoController.php
@@ -47,6 +47,8 @@ class PhotoController extends Controller
 	}
 
 	/**
+	 * Given a photoID and albumID returns the data of the photo.
+	 *
 	 * @param Request $request
 	 *
 	 * @return array|string
@@ -55,7 +57,6 @@ class PhotoController extends Controller
 	{
 		$request->validate([
 			'albumID' => 'string|required',
-
 			'photoID' => 'string|required',
 		]);
 
@@ -78,6 +79,9 @@ class PhotoController extends Controller
 	}
 
 	/**
+	 * Return a random public photo (starred)
+	 * This is used in the Frame Controller.
+	 *
 	 * @return string
 	 */
 	public function getRandom()
@@ -100,6 +104,8 @@ class PhotoController extends Controller
 	}
 
 	/**
+	 * Add a function given an AlbumID.
+	 *
 	 * @param Request $request
 	 *
 	 * @return false|string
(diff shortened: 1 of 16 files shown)


app/Http/Controllers/PhotoController.php
```diff
@@ -201,6 +213,10 @@ public function setDescription(Request $request)
 	}
 
 	/**
+	 * Define if a photo is public.
+	 * We do not advise the use of this and would rather see people use albums visibility
+	 * This would highly simplify the code if we remove this. Do we really want to keep it ?
+	 *
```
kamil4: One case where you can't use album visibility is for Unsorted. But the whole Unsorted always struck me as a bit of a strange feature. If we want to support pictures being displayed directly at the top level (without putting them in albums) then that's fine, but that's not what Unsorted does. When you upload a picture at the top level (albums view) it ends up in Unsorted. When you make it public, it also shows up in the Public smart album, but it's still not visible at the top level. What's the use case here? No idea...

app/Http/Controllers/PhotoController.php
```diff
@@ -440,6 +467,8 @@ public function duplicate(Request $request)
 	}
 
 	/**
+	 * Return a photo as an archive just to annoy @SerenaButler.
```
kamil4: LOL :smiley:. I actually I have it on my TODO list to fix it (I will comment in https://github.com/LycheeOrg/Lychee-front/issues/119).

app/Http/Controllers/SettingsController.php
```diff
@@ -140,6 +215,14 @@ public function setPublicSearch(Request $request)
 		return (Configs::set('public_search', '0')) ? 'true' : 'false';
 	}
 
+	/**
+	 * Show image overlay by default or not
+	 * (white text in the bottom right corner).
```
kamil4: Bottom _left_, I believe?

app/Http/Controllers/PhotoController.php
ildyria: use case is being able to share a single picture in a private album.
You can then access the picture with the view.php (in Lychee v3) and a similar system in the v4.

app/Http/Controllers/SettingsController.php
ildyria: yes, the other right.

app/Http/Controllers/PhotoController.php
kamil4: Yeah, I know. I guess what I was trying to say is that I would find the existing use case a lot more compelling if it applied to top-level images as well -- but there's currently no such thing as top-level images...

Issue #47, opened 2018-11-17
Make Starred Album Public?


seonwoolee:
How can I make the Starred album public? I can make all my starred photos public, but the starred album still isn't public

ildyria:
@kamil4 while you are at it... :laughing:

kamil4:
Heh... When I was implementing public Recent I asked myself this question: should any others be public as well? Starred could be but I figured that nobody would want to since in publicMode we don't even display the stars (though stars do influence the album thumbs).

I guess somebody does want to after all. Oh well, shouldn't be too much work to add. The front end should actually be good as I made that part fully generic but the server will need changes, of course.

kamil4:
Closed via #293.

ckizer:
I'm completely confused. I imported 200 images. How can I make them all public? How can I set an album to be public? The only thing I see is i have to click on each image 1 by 1 add it to a new public album. This will take about 6 weeks.

ckizer:
How do I make an album public? I can't figure out any settings for this? 
<img width="409" alt="Screen Shot 2019-09-05 at 5 57 49 PM" src="https://user-images.githubusercontent.com/191738/64393525-b9f91200-d006-11e9-8a5f-e25a56c246fb.png">

kamil4:
I don't believe that you are using Lychee-Laravel (v4), but rather the current git HEAD of Lychee v3. Which, I can confirm, appears to be broken; I'll fix that. In the meantime, however, I would strongly recommend that you actually do try Lychee-Laravel. We just released version 4 beta 1, which you can find [here](https://github.com/LycheeOrg/Lychee-Laravel/releases/tag/v4.0.0-beta.1).

In the non-broken version, clicking on the leftmost eye icon brings up a dialog window where you can make an album public by clicking on the large slider.

Also, in the future, please open a new issue rather than adding unrelated comments to issues that are already closed.

PR #293, opened 2019-06-20
Public recent

### LINKED ISSUE #176: Option to make "Recents" folder public
For now, other (unregistered) users have no tools to track albums. No RSS, no other streams and if there are sub-albums, even with "sort-by-date" - need to go into each and check if there is anything new on top.
As admin user we have "smart-album" called "Recent", it holds latest uploaded photos automatically. But it's not public and no option to make it public.
Maybe add option (may be available in "Settings -> More" option, to keep settings GUI unmodified) to make it public and "days for recent" configurable?
This would allow people to "follow" albums.

kamil4:
fixes #176.

Adds support for public Recent and configurable recent age.

Also fixes a related Search issue where password-protected subalbums would not be included in results. Commentary was improved to better explain what is happening there.

I did not create a new SmartAlbums category because I didn't think it was worth it for just two entries. Maybe make it a part of a future comprehensive cleanup in that area?


kamil4:
For the sake of completeness: fixes #176. Front end side is in https://github.com/LycheeOrg/Lychee-front/pull/123.

kamil4:
Public Starred support added, as per #47.

Note that this branch touches some of the same code as https://github.com/LycheeOrg/Lychee-Laravel/tree/image_protection. Manual rebasing will be needed on whichever goes second...

kamil4:
With the refactored infrastructure it was really easy to fix the leaks in `getRandom` so I pushed that as well. This was hopefully the last new feature on this branch :smiley:.

ildyria:
Rebased on master. Advised:
```
git checkout master
git branch -D public-recent
git fetch
git pull
git checkout public-recent
```

d7415:
Skim read. Looks good!

PR #294, opened 2019-06-20
Fix tags delete

kamil4:
I never use tags but while testing something I noticed that the attempt to remove the last tag of an image results in a server error.


ildyria:
Can you try to add a few tests in https://github.com/LycheeOrg/Lychee-Laravel/blob/master/tests/Feature/PhotosTest.php
e.g. near line
https://github.com/LycheeOrg/Lychee-Laravel/blob/master/tests/Feature/PhotosTest.php#L174
(test the tag removal).

kamil4:
Test done. An existing test extended. They even still pass :-).

PR #295, opened 2019-06-25
Image protection

ildyria:
Add symbolic linking in order to hide the direct addresses of pictures. 

The idea is to create symbolic links that will be removed after a set period of time. This avoid a too important slowdown of the server (by e.g. checking if the user visiting has access right etc).

- This functionality is disabled by default.
- This functionality is also disabled for admins (we may want to also remove this for users, not external visitors)
- Default symbolic link life time is 1 day.

Link address is composed as follow:
```php
$file_name = hash('sha256', $salt . '|' . $original) . $extension;
```
where `$original` is the `url` or `thumbUrl` field of a photo:
```php
$photo_name = md5(microtime()) . $extension
```
and `$salt` is:
```php
$now = now();
stringval($now)
```

The goal of this approach is to avoid someone finding
`upload/small/1234567890abcdef.jpg` or `upload/thumb/1234567890abcdef.jpeg` and deduce the url of the full sized picture: `upload/big/1234567890abcdef.jpg`

In such way we protect the access to full sized pictures.


kamil4:
As I suspected, after merging `public-recent` this now conflicts. Could you rebase?

app/ModelFunctions/SymLinkFunctions.php
```diff
@@ -1,31 +1,53 @@
...
+		if ($this->sessionFunctions->is_admin() && Configs::get_value('SL_for_admin', '0') === '0') {
+			return null;
 		}
 
-		$sym = SymLink::where('photo_id', $photo->id)->orderBy('created_at', 'DESC')->first();
+		$sym = null;
+
+		$sym = SymLink::where('photo_id', $photo->id)
+			->orderBy('created_at', 'DESC')
+			->first();
```
kamil4: I was meaning to ask: how can we have more than one SymLink row per photo? Is that due to race conditions?

app/Http/Controllers/IndexController.php
```diff
@@ -71,6 +79,7 @@ public function phpinfo()
 	 */
 	public function gallery()
 	{
+		$this->symLinkFunctions->remove_outdated();
```
kamil4: So will this run every time a new user connects to the gallery? Perhaps there should be some throttling? Then again, it's a single DB query...

app/ModelFunctions/SymLinkFunctions.php
```diff
@@ -64,4 +88,21 @@ public function clearSymLink()
...
 	}
+
+	/**
+	 * Remove outdated SymLinks.
+	 *
+	 * @return bool
+	 */
+	public function remove_outdated()
+	{
+		$symlinks = SymLink::where('created_at', '<', now()->subDays(intval(Configs::get_value('SL_life_time_days', '7')))->toDateTimeString())->get();
```
kamil4: Should the fallback here be `7` or `1`?

ildyria: I don't know how long we want our symlink to stay live by default. 1 day ? 3 days ? 7 days ?

app/Http/Controllers/IndexController.php
ildyria: well it is a bit more than a single DB query as it's a query for every picture symlinked...

app/ModelFunctions/SymLinkFunctions.php
ildyria: Race condition was not in my mind but more the fact that we may want to extend the work to 1 symlink per visitor... I don't know yet...

ildyria:
> As I suspected, after merging `public-recent` this now conflicts. Could you rebase?

done. :)

app/ModelFunctions/SymLinkFunctions.php
d7415: At one per visitor per photo it *must* be slower than an auth check...

app/ModelFunctions/SymLinkFunctions.php
kamil4: Sorry, I should've been more clear what I meant. Right now the fallback (`7`) is inconsistent with the default value set in the database migration (`1`). I believe they should be the same. This is what I meant with my question.

routes/web.php
```diff
@@ -22,6 +22,8 @@
 Route::get('/phpinfo', 'IndexController@phpinfo')->middleware('admin');
 Route::get('/gallery', 'IndexController@gallery')->name('gallery');
 
+Route::get('/images/{img}', 'ImageController@view');
```
kamil4: Sorry, I may have asked this before, but what is this for? It doesn't seem to be defined anywhere...

kamil4:
@ildyria I pushed a commit to your branch that synchronizes `getSmartAlbumsThumbs` with the changes you made to `gen_thumbs` (in particular, the use of `SymLinkFunctions::find`)

app/ModelFunctions/SymLinkFunctions.php
ildyria: Yeah I know. But I'm still undecided on the default value. ^^

app/ModelFunctions/SymLinkFunctions.php
ildyria: @d7415 actually no even at one per visitor per photo it is not slower.
The reason is you only generate the symlink once. After that it is only a `get` on the symlink.
When the visitor access the picture we don't have check, it is directly served with the content of the symlink.
A authorization based system would require to serve the `image/size/id` url, go through Laravel for **each** image call before actually serving the image.

routes/web.php
ildyria: Yeah that was the beginning of the auth based system before I considered symlinks. I will remove it.

app/ModelFunctions/SymLinkFunctions.php
ildyria: let's go for 3.

app/Http/Controllers/PhotoController.php
```diff
@@ -84,6 +90,20 @@ public function get(Request $request)
 
 		$return = $photo->prepareData();
 		$this->symLinkFunctions->getUrl($photo, $return);
+		if (!$this->sessionFunctions->is_logged_in()) {
```
kamil4: I find it suspicious that this functionality is guarded by `is_logged_in`. Haven't we recently fixed many instances of that? Being logged in shouldn't be enough to disable this -- the logged in user should be either the admin or the owner of the album...

app/Photo.php
```diff
@@ -309,6 +309,20 @@ public function prepareData()
 		return $photo;
 	}
 
+	/**
+	 * Downgrade the quality of the pictures.
+	 *
+	 * @param array $return
+	 */
+	public function downgrade(array &$return)
```
kamil4: Wouldn't it be better to return an empty string as `url` and make sure that the front end handles that correctly? With this change, I think the front end will be unable to tell the dimensions of the image pointed to by `url` (until now, we could use Photo's `width` and `height`). We may not be needing that at the moment, but I think if we implement image zoom, etc, that users have requested, we may need that info...

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -180,6 +181,9 @@ public function photos(Builder $photos_sql)
 			// Turn data from the database into a front-end friendly format
 			$photo = $photo_model->prepareData();
 			$this->symLinkFunctions->getUrl($photo_model, $photo);
+			if (!$this->sessionFunctions->is_logged_in() && !$full_photo) {
```
kamil4: Same comment as before about the use of `is_logged_in`.

kamil4:
One more thought on full_photo. We probably need to exempt video files from any of this -- we keep their location in `url` and we can't just replace that with `medium` or `medium2x` -- unlike for images, we don't have downscaled versions of video files.

app/Http/Controllers/PhotoController.php
ildyria: Hooo right nicely spotted.
I guess I need to use `is current user` then. :)

app/Photo.php
ildyria: Nicely spotted. Also medium may not even exists...

ildyria:
I always forget that there might be videos...

Issue #216, opened 2019-04-23
New feature: Download selected images as Zip Archive


SerenaButler:
![image](https://user-images.githubusercontent.com/35893306/56576719-6fcc4280-65c9-11e9-9fd5-dbc423baa979.png)

ildyria:
I like it. :)

PR #296, opened 2019-07-03
Improve image download

### LINKED ISSUE #189: Downloading single images zipped is uneccessary ;-)
Downloading **single** images zipped (image view!) is (usually) uneccessary ;-) Comments?

kamil4:
Fixes #189 and, with front end changes, #216.

Note that since this modifies the server protocol, the corresponding front end changes (https://github.com/LycheeOrg/Lychee-front/pull/127) need to be merged at the same time.

* Add support for downloading multiple photos or albums in one go

* Add support for downloading albums with subalbums

* Don't zip single photos

* Disable caching

Rationale for particular changes:

I removed ZipStream options because what we had there had no effect. See the source code of the version we use (0.5.2): itdoesn't match any of the strings we were passing. Also, sending headers is a bad idea as it enables caching (see below).

I disabled caching because it was causing big problems with album downloads. A zip file would get cached by the web browser and then even if the album was modified, another download attempt would be served from that cache rather than by the server. It's less of an issue for photo downloads but I disabled it there as well for consistency but also because I think it's harmful to thrash the browser cache with download data: users are unlikely to need it again.


app/Http/Controllers/AlbumController.php
```diff
@@ -541,123 +550,183 @@ public function getArchive(Request $request)
...
-			// Check if album empty
-			if ($photos_sql->count() == 0) {
-				Logs::error(__METHOD__, __LINE__, 'Could not create ZipStream without images');
-
-				return 'false';
-			}
+		$response = new StreamedResponse(function () use ($albumIDs, $badChars) {
+			$zip = new ZipStream(null);
+
+			$UserId = $this->sessionFunctions->id();
```
ildyria: what happens here if the user is not logged in ?

ildyria:
```php
	/**
	 * Return the current ID of the user
	 * what happens when UserID is not set? :p.
	 *
	 * @return int
	 */
	public function id()	{
		return Session::get('UserID');
	}
```

app/Http/Controllers/AlbumController.php
```diff
@@ -541,123 +550,183 @@ public function getArchive(Request $request)
...
+						$dir = $album->title;
+						$photos_sql = Photo::set_order(Photo::where('album_id', '=', $albumID));
+						break;
+				} // switch (albumID)
 
-			$photos = $photos_sql->get();
-			foreach ($photos as $photo) {
-				$title = str_replace($badChars, '', $photo->title);
-				$url = Config::get('defines.urls.LYCHEE_URL_UPLOADS_BIG') . $photo->url;
+				$compress_album = function ($photos_sql, $dir, &$dirs, $parent_dir, $album) use (&$zip, $badChars, &$compress_album) {
```
ildyria: Is the use of a functor necessary ? Why not use a private method ?

app/Http/Controllers/AlbumController.php
```diff
@@ -541,123 +550,183 @@ public function getArchive(Request $request)
...
-				if (!empty($files)) {
-					$i = 1;
-					while (in_array($zipFileName, $files)) {
-						// Set new title for photo
-						$zipFileName = $zipTitle . '/' . $title . '-' . $i . $extension;
-						$i++;
+					$files = [];
+					$photos = $photos_sql->get();
+					foreach ($photos as $photo) {
+						$url = Config::get('defines.urls.LYCHEE_URL_UPLOADS_BIG') . $photo->url;
```
ildyria: this will need to use the Storage facade (after merging the image protection one).

app/Http/Controllers/AlbumController.php
kamil4: I believe we don't read the value unless somebody is logged on. This code was there before as well so clearly it works :smiley:.

ildyria: :laughing:

app/Http/Controllers/AlbumController.php
kamil4: Just laziness I guess :wink:. The code was in the function body so rather than moving it completely out, I created a nested function. I did something similar in the `PhotoController.php` as well -- would you like me to turn them both into private methods?

ildyria: I would be interested to see if those can be delegated to a dependency related to all the archiving stuff like we did for albumfunctions etc... :)

kamil4:
@ildyria Answering your question on top: yes, there is clearly similarity between the two `getArchive` methods but my gut feeling is that they are just sufficiently different that trying to extract the common parts will increase the code complexity rather than decrease it.

ildyria:
Do what you think is the best. :)

kamil4:
This should be ready for another pass... I rebased and switched to the Storage facade. I did not refactor the archiving stuff across albums and photos as I didn't think there was enough code duplication.

Late in the game I realized that I needed to add checks for `albumIDs` and `photoIDs` to the `read` middleware since until now it only needed to handle `albumID` and `photoID`. In the process I noticed that the checks in the `upload` middleware were kind-of lax. E.g., if a request had an `albumID` that was fine, we would never check if it also had an `albumIDs`. We also weren't checking if all items in `albumIDs` and such were actually valid, only if any were. I also simplified an ownership test in `photo_check`, which was optionally also testing the album; this is not needed anymore as we now have code that ensures no mixed album/content ownership.

ildyria:
We may want to only allow download archive for the not `s3` default storage option.

ildyria:
> Late in the game I realized that I needed to add checks for `albumIDs` and `photoIDs` to the `read` middleware since until now it only needed to handle `albumID` and `photoID`. In the process I noticed that the checks in the `upload` middleware were kind-of lax. E.g., if a request had an `albumID` that was fine, we would never check if it also had an `albumIDs`. We also weren't checking if all items in `albumIDs` and such were actually valid, only if any were. I also simplified an ownership test in `photo_check`, which was optionally also testing the album; this is not needed anymore as we now have code that ensures no mixed album/content ownership.

I will see if I can add tests to actually double check those.

kamil4:
> We may want to only allow download archive for the **not** `s3` default storage option.

Is that because `ZipStream` and `BinaryFileResponse` require source files to be locally accessible?

kamil4:
I just realized that `getArchive` doesn't check at all whether `downloadable` is set -- it's all client-side only at this point... I guess it should also check `full_photo`? If the client is not allowed to view an image, they shouldn't be allowed to download it either...

ildyria:
You can check `full_photo` for albums, but you should only trust it if the user is not logged in.

I would be in favor that if if `download` is set but not `full_photo` then provide the photo at the `url` of `Photo::downgrade()`.

> Is that because `ZipStream` and `BinaryFileResponse` require source files to be locally accessible?

You may want to have a look at : https://github.com/maennchen/ZipStream-PHP/wiki/Symfony-example :)

kamil4:
OK, the latest changes on this branch add a new global `downloadable` config that behaves just like `full_photo`: it's used for smart albums and for non-public albums.

`downloadable` is now consistently set for all albums returned by `Album::get` (it wasn't before for smart albums) and is also set for `Photo::get` to make front end's job easier.

`getArchive` methods now test whether the requested photos/albums are in fact downloadable.

Also, I added bailout code to `getArchive` in case of the S3 storage.

Finally, the branch includes changes needed to fix the standalone `view` mode (I have corresponding changes to v3 as well).

PR #307, opened 2019-07-16
Config upgrade

ildyria:
Fixes https://github.com/LycheeOrg/Lychee-Laravel/issues/206
Add sanity check to Diagnostics.


database/migrations/2019_07_16_093502_config_upgrade.php
```diff
@@ -0,0 +1,124 @@
...
+			Configs::where('key', '=', 'version')->update(['type_range' => INT]);
+			Configs::where('key', '=', 'username')->update(['type_range' => STRING_REQ]);
+			Configs::where('key', '=', 'password')->update(['type_range' => STRING_REQ]);
+			Configs::where('key', '=', 'check_for_updates')->update(['type_range' => BOOL]);
+			Configs::where('key', '=', 'sorting_Photos_col')->update(['type_range' => 'id|takestamp|title|description|public|star|type']);
+			Configs::where('key', '=', 'sorting_Photos_order')->update(['type_range' => 'ASC|DESC']);
+			Configs::where('key', '=', 'sorting_Albums_col')->update(['type_range' => 'id|title|description|public|max_takestamp|min_takestamp']);
+			Configs::where('key', '=', 'sorting_Albums_order')->update(['type_range' => 'ASC|DESC']);
+			Configs::where('key', '=', 'imagick')->update(['type_range' => BOOL]);
+			Configs::where('key', '=', 'dropboxKey')->update(['type_range' => STRING]);
```
kamil4: If we are standardizing on snake case, shouldn't this be `dropbox_key` then (or maybe `Dropbox_key`)?

routes/web.php
```diff
@@ -79,6 +79,7 @@
 Route::post('/api/Settings::getAll', 'SettingsController@getAll')->middleware('admin');
 Route::post('/api/Settings::saveAll', 'SettingsController@saveAll')->middleware('admin');
 Route::post('/api/Settings::setOverlayType', 'SettingsController@setImageOverlayType')->middleware('admin');
+Route::post('/api/Settings::setDropboxKey', 'SettingsController@setDropboxKey')->middleware('admin');
```
kamil4: Does that mean that setting the Dropbox key didn't work at all in v4 until now? :confused:

ildyria: yup :laughing:

Issue #321, opened 2019-08-18
Implement natural sorting


garster:
Album and photo title sort uses the database order by clause. This means a title that contain a sequence of numbers will not sort naturally. For example 1, 2, 10 will comes out at 1, 10, 2 due to how MySql/MariaDB orders string values.

It would be super cool if the [sorting was natural](https://en.wikipedia.org/wiki/Natural_sort_order). There are some natural sorting tools out there like PHP's natsort or Laravel sortBy with a callback. Or it could happen in the JS frontend. 

It would add some overhead so maybe it would be an option to turn on, or only enabled on string values. 

I could see this being of value for people that use the albums for things like addresses (real estate) or numbered landmarks, really anything where a number is used in the name.

![image](https://user-images.githubusercontent.com/4443025/63218927-8ae43480-c11b-11e9-9b12-08efc3365b47.png)

![image](https://user-images.githubusercontent.com/4443025/63218928-93d50600-c11b-11e9-853e-aa13d22e1cf1.png)

garster:
I found a way to implement it for subalbums. I will need to clean up, do it for top level albums and maybe I can do a PR.

Before
![image](https://user-images.githubusercontent.com/4443025/63220069-b9bad480-c134-11e9-81d6-4d6c4c58be9e.png)





After
![image](https://user-images.githubusercontent.com/4443025/63220061-97c15200-c134-11e9-89b4-49ee5bf897ec.png)

kamil4:
Thanks for taking the initiative on this! While I can see the potential value of this feature, to me personally it is not a priority, though other developers might feel different.

I wonder if we should replace the sorting we have with the natural sort or offer it as an option? Making it optional is of course safer; however, extending the GUI can be an order-of-magnitude more work than just working on the feature on the server in the first place... @ildyria, @d7415, what do you think? My feeling is that we should see performance of the two sorts on an album with >1,000 pictures and decide then.. You'll also first need to verify that Laravel's natural sort works correctly  on all datatypes of interest (some of the columns we can sort by contain booleans, numbers, or even dates).

[...]

PR #322, opened 2019-08-19
Introduce natural sorting

garster:
No doubt that DB ordering is way faster than sorting in code, but to me this is big deal. It would be great to get some speed numbers, but also big-picture, speed improvements could be done in other places too.

I don't know what Laravel uses under the hood for sorting (I can't find the source code), but PHP says it uses Quicksort which works well on a list that is already well sorted.

So my suggested sledgehammer here is to add a sortBy onto just about every orderBy. The DB will sort well then the sortBy will have another pass.

So far it is worked will on a album with 150 pics, I get the API response back in about 200ms. I have not done a comparison.

This PR is related to #321


app/ModelFunctions/AlbumFunctions.php
```diff
@@ -543,7 +545,9 @@ public function getToplevelAlbums()
 			if ($id == 0) {
 				$return['albums'] = Album::where('owner_id', '=', 0)
 					->where('parent_id', '=', null)
-					->orderBy(Configs::get_value('sorting_Albums_col'), Configs::get_value('sorting_Albums_order'))->get();
+					->orderBy(Configs::get_value('sorting_Albums_col'), Configs::get_value('sorting_Albums_order'))
+					->get()
+					->sortBy(Configs::get_value('sorting_Albums_col'), SORT_NATURAL, (Configs::get_value('sorting_Albums_order') === 'DESC'));
```
kamil4: So, given your follow-up timings from #321, I wonder if we should be doing orderBy here and in other such places at all... I think we'll need to try three versions:

* plain `orderBy` (current master)
* `orderBy` plus `sortBy` (your PR)
* just `sortBy`

I'll see if I can work on that tonight...

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -136,7 +136,9 @@ public function gen_thumbs($return, $album_list)
 		$photos = Photo::whereIn('album_id', $album_list)
 			->orderBy('star', 'DESC')
 			->orderBy(Configs::get_value('sorting_Photos_col'), Configs::get_value('sorting_Photos_order'))
-			->limit(3)->get();
+			->limit(3)
+			->get()
+			->sortBy(Configs::get_value('sorting_Photos_col'), SORT_NATURAL, (Configs::get_value('sorting_Photos_order') === 'DESC'));
```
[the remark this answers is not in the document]
garster: I did not test photos, and the sortBy must be after the ->get() so it will not work well in this situation. I think the sortBy should be removed here. I  will modify the code.

garster:
> I'm also surprised by the lack of changes to `set_order` in `Photo.php`. Did you verify that your branch sorts photos naturally in addition to albums?

I missed this, the trick is the sortBy has to come after the get() call. I traced the `set_order` function thru some others and found two places it is called. I made the change. I also saw it in DemoController but I did not make the change there.

app/ModelFunctions/AlbumFunctions.php
garster: Removing the OrderBys would simplify the code, I just did not want to get too crazy as an outside contributor.

kamil4:
Yes, I made a similar change in my tree but I didn't get to test it yet as I spent most of the last evening creating an album with >10,000 pictures to have something to test the performance of sorting with.

kamil4:
Sorry for the lack of updates. The tests I ran revealed some missing corner cases that were difficult to fix without large performance losses but I think I solved the major ones. I just need to clean up my code before pushing a commit on top of what's here.

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -517,11 +548,11 @@ public function setContentsOwner($albumID, int $ownerId)
...
 	 * their content).
 	 *
-	 * @param $parentAlbum
 	 * @param array $return
-	 * @param bool  $includePassProtected
+	 * @param $parentAlbum
+	 * @param bool $includePassProtected
 	 */
-	public function get_sub_albums($parentAlbum, array &$return, $includePassProtected = false)
+	public function get_sub_albums(array &$return, $parentAlbum, $includePassProtected = false)
```
ildyria: what is the type of `$parentAlbum` ?

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -556,52 +587,60 @@ public function getToplevelAlbums()
 			'shared_albums' => null,
 		);
 
+		$sortingCol = Configs::get_value('sorting_Albums_col');
+		$sortingOrder = Configs::get_value('sorting_Albums_order');
+
+		$customSort = function ($query) use ($sortingCol, $sortingOrder) {
```
ildyria: you really like functors. ^^

PR #332, opened 2019-09-11
Laravel6

ildyria:
Migrate from to Laravel 5.8 to Laravel 6.0


ildyria:
> Do you expect that we will be able to make use of any new functionality from Laravel 6 in Lychee?

https://laravel.com/docs/6.0/collections#lazy-collections

That one looks pretty interesting. :)

PR #334, opened 2019-09-14
add basic raw support.

### LINKED ISSUE #284: Feature request: Allow uploading raw files
Given that prepare backend for #204 #216 is soon to be applied I wonder whether it could be an idea to extend the feature to be able to upload and download raw files as well?

ildyria:
fixes #284 

We add a new config field: `raw_formats` it takes as input a string containing the extensions wanted. eg: `.tex|.cr2|.cr3`

**We do not do anything on raw files.** We just upload them to the new `uploads/raw` folder.

In order to avoid broken image logo we use `img/placeholder.png` instead of the the file.
The content of the file, e.g. for a `.tex` file is still visible via direct link.

Symlinking, Album Download, Photo Download should also work.

### DIFF
--- a/app/Http/Controllers/AlbumController.php
+++ b/app/Http/Controllers/AlbumController.php
@@ -738,7 +738,8 @@ class AlbumController extends Controller
 							continue;
 						}
 
-						$url = Storage::path('big/' . $photo->url);
+						$prefix_url = $photo->type == 'raw' ? 'raw/' : 'big/';
+						$url = Storage::path($prefix_url . $photo->url);
 						// Check if readable
 						if (!@is_readable($url)) {
 							Logs::error(__METHOD__, __LINE__, 'Original photo missing: ' . $url);
@@ -753,7 +754,11 @@ class AlbumController extends Controller
 						// Get extension of image
 						$extension = Helpers::getExtension($url, false);
 						// Set title for photo
-						$file = $title . $extension;
+						if ($prefix_url != 'raw/') {
+							$file = $title . $extension;
+						} else {
+							$file = $title;
+						}
 						// Check for duplicates
 						if (!empty($files)) {
 							$i = 1;
--- a/app/Http/Controllers/AlbumController.php
+++ b/app/Http/Controllers/AlbumController.php
@@ -738,7 +738,9 @@ class AlbumController extends Controller
 							continue;
 						}
 
-						$prefix_url = $photo->type == 'raw' ? 'raw/' : 'big/';
+						$is_raw = ($photo->type == 'raw');
+
+						$prefix_url = $is_raw ? 'raw/' : 'big/';
 						$url = Storage::path($prefix_url . $photo->url);
 						// Check if readable
 						if (!@is_readable($url)) {
@@ -746,26 +748,25 @@ class AlbumController extends Controller
 							continue;
 						}
 
+						// Get extension of image
+						$extension = Helpers::getExtension($url, false);
+
+						// Set title for photo
 						$title = str_replace($badChars, '', $photo->title);
 						if (!isset($title) || $title === '') {
 							$title = 'Untitled';
 						}
 
-						// Get extension of image
-						$extension = Helpers::getExtension($url, false);
-						// Set title for photo
-						if ($prefix_url != 'raw/') {
-							$file = $title . $extension;
-						} else {
-							$file = $title;
-						}
+						$file = $title . ($is_raw ? '' : $extension);
+
 						// Check for duplicates
 						if (!empty($files)) {
 							$i = 1;
 							$tmp_file = $file;
+							$pos = strrpos($tmp_file, '.');
 							while (in_array($tmp_file, $files)) {
 								// Set new title for photo
-								$tmp_file = $title . '-' . $i . $extension;
+								$tmp_file = substr_replace($file, '-' . $i, $pos, 0);
 								$i++;
 							}
 							$file = $tmp_file;
--- a/app/Http/Controllers/PhotoController.php
+++ b/app/Http/Controllers/PhotoController.php
@@ -152,9 +152,6 @@ class PhotoController extends Controller
 		$request->validate([
 			'albumID' => 'string|required',
 			'0' => 'required',
-			//            '0.*' => 'image|mimes:jpeg,png,jpg,gif',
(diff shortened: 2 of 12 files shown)


app/Http/Controllers/AlbumController.php
```diff
@@ -753,7 +754,11 @@ public function getArchive(Request $request)
 						// Get extension of image
 						$extension = Helpers::getExtension($url, false);
 						// Set title for photo
-						$file = $title . $extension;
+						if ($prefix_url != 'raw/') {
```
d7415: As it's the same block, I feel this should be the same test as L741, `$photo->type == 'raw'`. Or we could set a boolean up there e.g. `$raw = $photo->type == 'raw';`?

I'll try to get around to the rest, but I've just seen how long this diff is!

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -140,6 +140,54 @@ public function createThumb(Photo $photo, string $frame_tmp = '')
...
+	 * @param $file
+	 * @param $extension
+	 *
+	 * @return string
+	 */
+	private function kind($file, string $extension)
+	{
+		// check raw files
+		if (in_array(strtolower($extension), explode('|', Configs::get_value('raw_formats', '')), true)) {
+			return 'RAW';
```
kamil4: I don't really like the use of uppercase `'RAW'` here but lowercase `'raw'` everywhere else. Any reason for this?

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -200,31 +243,13 @@ public function add(array $file, $albumID_in = 0)
 				break;
 		}
 
-		// Verify extension
+//		// Verify extension
```
kamil4: Unnecessary change.

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -286,10 +315,19 @@ public function add(array $file, $albumID_in = 0)
...
 
-		$info = $this->metadataExtractor->extract($path, $mimeType);
+		if ($kind == 'RAW') {
+			$info = $this->metadataExtractor->bare();
+			$this->metadataExtractor->size($info, $path);
+			$this->metadataExtractor->validate($info);
+			$info['type'] = 'raw';
+		} else {
+			$info = $this->metadataExtractor->extract($path, $mimeType);
+		}
```
kamil4: So PHP built-in exif support only works with JPEGs, not RAWs? Bummer... I believe `exiftool` can handle both. We'll need to figure out what to do when we add `exiftool` support...

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -286,10 +315,19 @@ public function add(array $file, $albumID_in = 0)
...
 
 		// Use title of file if IPTC title missing
-		if ($info['title'] === '') {
+		if ($kind == 'RAW') {
+			$info['title'] = substr(basename($file['name']), 0, 30);
```
kamil4: I expected as much based on the earlier code but I still don't know why for RAWs you choose to keep the extension in the title?

app/ModelFunctions/PhotoFunctions.php
ildyria: I actually did not try the exif tool on raw files.

It seems that the native one should support it.

app/ModelFunctions/PhotoFunctions.php
ildyria: Because I would rather like to know the kind of file I'm dealing with directly by looking at its title rather than looking at the mimetype.

app/ModelFunctions/PhotoFunctions.php
ildyria: not really. Just that I prefer constants that are used for decision to be in uppercase. I don't really care to be honest.

ildyria:
> So we can store RAWs in Lychee now but we can't display them. Is that useful? Or is this PR merely the first step?

It can be seen as a first step yes. :)

app/ModelFunctions/PhotoFunctions.php
d7415: I think consistency is important, especially in the same function's return values

app/ModelFunctions/PhotoFunctions.php
d7415: I'm not sure about this either... I can see advantages, but...

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -234,7 +259,9 @@ public function add(array $file, $albumID_in = 0)
 		// Set paths
 		$tmp_name = $file['tmp_name'];
 		$photo_name = md5(microtime()) . $extension;
-		$path = Storage::path('big/' . $photo_name);
+
+		$path_prefix = $kind != 'RAW' ? 'big/' : 'raw/';
```
d7415: I don't know what I'd do instead, but it feels strange hard coding this in several places.

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -140,6 +140,54 @@ public function createThumb(Photo $photo, string $frame_tmp = '')
...
+	 * Returns 'video' if it is a video
+	 * Returns 'raw' if it is an accepted file (we only check extensions)
+	 * Returns 'error message' if it is something else.
+	 *
+	 * @param $file
+	 * @param $extension
+	 *
+	 * @return string
+	 */
+	private function kind($file, string $extension)
```
d7415: Is `kind` a good function name here when we're using `kind` for big/medium/small/etc in similar code paths?

app/Http/Controllers/PhotoController.php
```diff
@@ -152,9 +152,6 @@ public function add(Request $request)
 		$request->validate([
 			'albumID' => 'string|required',
 			'0' => 'required',
-			//            '0.*' => 'image|mimes:jpeg,png,jpg,gif',
-			'0.*' => 'image|mimes:jpeg,png,jpg,gif,mov,webm,mp4,ogv',
```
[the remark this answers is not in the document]
ildyria: I did not see it working in either of the case.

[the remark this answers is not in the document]
d7415: Ok.

PR #336, opened 2019-09-17
Import

### LINKED ISSUE #173: Import from Server dialog keeps showing after import is done
The "Import from Server" dialog keeps showing with "Importing" blinking. I can know when an import is over by tracking the logs table. That happened when importing around 300 images.

```
Diagnostics
    -----------
    Warning: Dropbox import not working. No property for dropboxKey.

    System Information
    ------------------
    Lychee Version:  3.2.13
    DB Version:      040000
    System:          Linux
    PHP Version:     7.2
    MySQL Version:   5.7.25-0ubuntu0.18.04.2
    Imagick:         1
    Imagick Active:  1
    Imagick Version: 1687
    GD Version:      2.2.5

    Config Information
    ------------------
    version:                 040000
    checkForUpdates:         1
    sortingPhotos_col:       takestamp
    sortingPhotos_order:     ASC
    sortingAlbums_col:       description
    sortingAlbums_order:     DESC
    imagick:                 1
    skipDuplicates:          1
    small_max_width:         0
    small_max_height:        360
    medium_max_width:        1920
    medium_max_height:       1080
    lang:                    en
    layout:                  0
    image_overlay:           1
    image_overlay_type:      exif
    default_license:         none
    compression_quality:     90
    full_photo:              1
    deleteImported:          1
    Mod_Frame:               0
    Mod_Frame_refresh:       30000
    landing_page_enable:     0
    landing_owner:           
    landing_title:           
    landing_subtitle:        
    landing_facebook:        
    landing_flickr:          
    landing_twitter:         
    landing_instagram:       
    landing_youtube:         
    landing_background:      
    thumb_2x:                1
    small_2x:                0
    medium_2x:               0
    site_title:              Lychee v4
    site_copyright_enable:   1
    site_copyright_begin:    2019
    site_copyright_end:      2019
```

### LINKED ISSUE #312: No feedback when time limit is reached during import
I've used the import from server feature to import a big folder. This triggers the default PHP time limit (according to the PHP error log), without showing a proper error in the browser.

In the browser, I still see:

image

This is using the current git master version (2270caf25b9f7729badaaa6f82adfae88a43de0f).

I realize that this can be solved by disabling or increasing the time limit (though perhaps having a smaller limit per image file, by calling `set_time_limit()` after each photograph, might be nicer than no limit at all), but regardless this should really provide some feedback to the user.

### LINKED ISSUE #313: [Enhancement] Show "delete" checkbox on server import dialog
When you import files from the server, these are deleted by default. This can be toggled in the (advanced) settings, but it would be nicer if you can (also) toggle this on each import separately, i.e. by adding a checkbox here:

image

On a related note, the above dialog currently warns that files will be deleted even when the config entry for that is disabled, which is a bit misleading (but that would also be fixed if a checkbox was added).

kamil4:
Fixes #173,
Fixes #312,
Fixes #313.

* Adds per-import choice of whether to delete original files
* Changes the default for `delete_imported` to no
* Adds status updates during import; this prevents http response timeouts for long imports
* Better problem reporting
* Monitors the server for low-memory condition
* Eliminates `php_no_limit_config` in favor of per-iteration reset

I was able to import over 13,000 pictures in one go without problems so I consider it stable at this point. The low-memory condition triggers only when `APP_DEBUG` is `true`.

Here are some screenshots for your viewing pleasure :wink:

Requires corresponding front-end changes; I'll submit a PR there in a minute.


app/Http/Controllers/ImportController.php
```diff
@@ -151,111 +159,184 @@ public function server(Request $request)
...
+
+		$response = new StreamedResponse();
+		$response->setCallback(function () use ($request, $delete_imported) {
+			// Surround the response in '"' characters to make it a valid
+			// JSON string.
+			echo '"';
+			$this->server_exec($request['path'], $request['albumID'], $delete_imported);
+			echo '"';
+		});
+		// nginx-specific voodoo, as per https://symfony.com/doc/current/components/http_foundation.html#streaming-a-response
```
ildyria: :laughing:

PR #339, opened 2019-09-20
First version of displaying coordinates on map

tmp-hallenser:
Backend functionality to display location data and map on the frontend.


config/secure-headers.php
```diff
@@ -377,6 +377,12 @@
         ],
         'img-src' => [
             'self' => true,
+            'allow' => [
+              'http://a.tile.osm.org/',
```
ildyria: add to which server these addresses points to (in this case openstreetmap).

kamil4:
Regarding the png images, given their tiny sizes, I wonder if it's even worth the bother of regular vs 2x? Under the circumstances I think I would just always display the 2x variants; the browser will downscale them as appropriate...

tmp-hallenser:
> Regarding the png images, given their tiny sizes, I wonder if it's even worth the bother of regular vs 2x? Under the circumstances I think I would just always display the 2x variants; the browser will downscale them as appropriate...

I think, this is how the leaflet plugin works. Of course, one could adjust it - but I don't think it's work the effort.

tmp-hallenser:
I hope I captured now all string constants.
Furthermore, I added an additional icon to this PR to visualize the direction of the image (needed by the PR of the front)

PR #344, opened 2019-09-20
Added missing german translations for string constants

tmp-hallenser:
Some string constants in the german localization were still in english.


d7415:
Thanks!

PR #346, opened 2019-09-25
Added Spanish translation

joserick:
On request of @ildyria in https://github.com/LycheeOrg/Lychee/pull/294#issuecomment-534539318

I hope it is useful :)


joserick:
Checking error...  :)

PR #352, opened 2019-09-28
Fix config categories and cleanup the config migrations.

ildyria:
- Compact the configs migrations into one file (what ever your update you are, this will normalize you).
- Compact some of the photos migrations.
- Solve the category miss match from PR #350


ildyria:
> but I guess we could, by renaming the migration...

Yep :laughing:

database/migrations/2019_09_28_171753_config_fix.php
```diff
@@ -0,0 +1,577 @@
...
+			[
+				'key' => 'check_for_updates',
+				'value' => '0',
+				'cat' => 'Admin',
+				'type_range' => BOOL,
+				'confidentiality' => '0',
+			],
+			[
+				'key' => 'sorting_Photos_col',
+				'value' => 'title',
```
kamil4: I believe this used to default to `takestamp`; is this an intentional change?

Considering that it's configurable, it's not a big deal, but I do think that we should strive for the defaults to be as unsurprising as possible. In this particular case, I'm not sure whether `title` or `takestamp` is better, but I thought I would bring it up...

database/migrations/2019_09_28_171753_config_fix.php
```diff
@@ -0,0 +1,577 @@
...
+			[
+				'key' => 'sorting_Photos_order',
+				'value' => 'ASC',
+				'cat' => 'Gallery',
+				'type_range' => 'ASC|DESC',
+				'confidentiality' => '2',
+			],
+			[
+				'key' => 'sorting_Albums_col',
+				'value' => 'max_takestamp',
```
kamil4: By the look of it, this one used to default to `description`, which in retrospect I agree probably wasn't the best choice... :wink:

database/migrations/2019_09_28_171753_config_fix.php
```diff
@@ -0,0 +1,577 @@
...
+			[
+				'key' => 'small_max_width',
+				'value' => '0',
+				'cat' => 'Image Processing',
+				'type_range' => INT,
+				'confidentiality' => '2',
+			],
+			[
+				'key' => 'small_max_height',
+				'value' => '360',
```
kamil4: I meant to ask this since forever: why does this default to `360`? justified-layout's `targetRowHeight` defaults to `320`, though as far as I can tell, the height is allowed to vary from half to twice as much, i.e., from `160` to `640`. Given that, `360` seems so... arbitrary. To ensure high quality, it should probably default to `640`...

database/migrations/2019_09_28_171753_config_fix.php
```diff
@@ -0,0 +1,577 @@
...
+			[
+				'key' => 'full_photo',
+				'value' => '1',
+				'cat' => 'Gallery',
+				'type_range' => BOOL,
+				'confidentiality' => '0',
+			],
+			[
+				'key' => 'delete_imported',
+				'value' => '1',
```
kamil4: I recently changed this to `0` since I felt that defaulting to a destructive behavior is not user-friendly. Though, given that it's configurable, it's not a big deal.

database/migrations/2019_09_28_171753_config_fix.php
```diff
@@ -0,0 +1,577 @@
...
+			[
+				'key' => 'Mod_Frame',
+				'value' => '1',
+				'cat' => 'Mod Frame',
+				'type_range' => BOOL,
+				'confidentiality' => '0',
+			],
+			[
+				'key' => 'Mod_Frame_refresh',
+				'value' => '30000',
```
kamil4: Another one I meant to mention for a while. I think the unit for this one is completely bonkers. It is milliseconds -- why?! Obviously, it should be seconds.

database/migrations/2019_09_28_171753_config_fix.php
```diff
@@ -0,0 +1,577 @@
...
+			[
+				'key' => 'landing_title',
+				'value' => 'John Smith',
+				'cat' => 'Mod Welcome',
+				'type_range' => STRING,
+				'confidentiality' => '2',
+			],
+			[
+				'key' => 'landing_subtitle',
+				'value' => 'Cat, Dogs & Humans Photography',
```
kamil4: Shouldn't that be _Cat**s**, Dogs & Humans Photography_? :smile:

database/migrations/2019_09_28_171753_config_fix.php
```diff
@@ -0,0 +1,577 @@
...
+			[
+				'key' => 'thumb_2x',
+				'value' => '1',
+				'cat' => 'Image Processing',
+				'type_range' => BOOL,
+				'confidentiality' => '2',
+			],
+			[
+				'key' => 'small_2x',
+				'value' => '1',
```
kamil4: This defaulted to `0` until now, I believe. Personally I always set it to `1` in my installations but I want to be sure that this is an intentional change. It does, after all, come with additional processing overheads and increased space and bandwidth requirements...

database/migrations/2019_09_28_171753_config_fix.php
```diff
@@ -0,0 +1,577 @@
...
+				'cat' => 'Image Processing',
+				'type_range' => BOOL,
+				'confidentiality' => '2',
+			],
+			[
+				'key' => 'medium_2x',
+				'value' => '1',
```
kamil4: Ditto...

database/migrations/2019_09_28_171753_config_fix.php
```diff
@@ -0,0 +1,577 @@
...
+			[
+				'key' => 'force_migration_in_production',
+				'value' => '0',
+				'cat' => 'Admin',
+				'type_range' => BOOL,
+				'confidentiality' => '3',
+			],
+			[
+				'key' => 'additional_footer_text',
+				'value' => 'config',
```
kamil4: Erm, don't you mean `'value' => ''` here?

database/migrations/2019_09_28_171753_config_fix.php
```diff
@@ -0,0 +1,577 @@
...
+			[
+				'key' => 'public_search',
+				'value' => '0',
+				'cat' => 'config',
+				'type_range' => BOOL,
+				'confidentiality' => '0',
+			],
+			[
+				'key' => 'gen_demo_js',
+				'value' => '1',
```
kamil4: What's the value of having it enabled for regular users?

database/migrations/2019_09_28_171753_config_fix.php
d7415: Good point. I'm not sure either...

database/migrations/2019_09_28_171753_config_fix.php
d7415: My guess would be that it's half of 720 (as lots of photos and screens are build around HD-type resolutions now)

database/migrations/2019_09_28_171753_config_fix.php
d7415: Agreed. Especially as there's a checkbox, I'd default to non-destructive.

database/migrations/2019_09_28_171753_config_fix.php
d7415: @ildyria only has one cat :wink:

database/migrations/2019_09_28_171753_config_fix.php
d7415: My gut says default to on (for both of these), but worth raising.

database/migrations/2019_09_28_171753_config_fix.php
d7415: Good point.

database/migrations/2019_09_28_171753_config_fix.php
ildyria: `takestamp` it is.

database/migrations/2019_09_28_171753_config_fix.php
ildyria: I am not changing the value if there is one already set.

database/migrations/2019_09_28_171753_config_fix.php
ildyria: because it goes out of the bat for the frame Javascript code (`timeout`).

database/migrations/2019_09_28_171753_config_fix.php
ildyria: The magic of copy paste. :'D

database/migrations/2019_09_28_171753_config_fix.php
ildyria: I think it was disabled, it's just that I copied the table from my test DB... :'D

database/migrations/2019_09_28_171753_config_fix.php
d7415: No, but @kamil4 changed the default in #336 (93448cc426794a3f465725b172aa3f04cf28d31b) and this would revert it.

database/migrations/2019_09_28_171753_config_fix.php
```diff
@@ -175,7 +176,7 @@ public function up()
 				'key' => 'sorting_Albums_col',
 				'value' => 'max_takestamp',
 				'cat' => 'Gallery',
-				'type_range' => 'id|title|description|public|max_takestamp|min_takestamp',
+				'type_range' => 'id|title|description|public|max_takestamp|min_takestamp|created_at',
```
d7415: I'm still not convinced that this is a good idea. It's still a big edge case.

ildyria: Well, we do not propose it visually. ;)

database/migrations/2019_09_28_171753_config_fix.php
kamil4: Oh, I'm aware of JS requirements. I was just pointing out that a user-level config should use a unit that a regular user will expect, and `ms` certainly isn't it. But that's an issue orthogonal to this PR so let's address it separately.

PR #355, opened 2019-09-30
Fix Import from Link

kamil4:
As reported by @d7415, _Import from Link_ stopped respecting the current album. Looks like I'm the one who broke it while working on _Import from Server_. Sorry!

So this trivial patch restores the previous behavior. Here's a question though: shouldn't we _always_ delete the imported file in this particular case? We are the ones who created the file by downloading the picture from the URL before importing it into Lychee. What's the point of keeping it around?

### DIFF
--- a/app/Http/Controllers/ImportController.php
+++ b/app/Http/Controllers/ImportController.php
@@ -133,7 +133,7 @@ class ImportController extends Controller
 				continue;
 			}
 			// Import photo
-			if (!$this->photo($tmp_name, Configs::get_value('delete_imported', '0'), $request['albumID'])) {
+			if (!$this->photo($tmp_name, true, $request['albumID'])) {
 				$error = true;
 				Logs::error(__METHOD__, __LINE__, 'Could not import file (' . $tmp_name . ')');
 				continue;
--- a/app/Http/Controllers/ImportController.php
+++ b/app/Http/Controllers/ImportController.php
@@ -133,7 +133,7 @@ class ImportController extends Controller
 				continue;
 			}
 			// Import photo
-			if (!$this->photo($tmp_name, $request['albumID'])) {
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/ImportController.php
```diff
@@ -133,7 +133,7 @@ public function url(Request $request)
 				continue;
 			}
 			// Import photo
-			if (!$this->photo($tmp_name, $request['albumID'])) {
+			if (!$this->photo($tmp_name, Configs::get_value('delete_imported', '0'), $request['albumID'])) {
```
ildyria: actually...
The parameters of `$this->photo` are :
```
	 * @param $path
	 * @param bool $delete_imported
	 * @param int  $albumID
```
I don't think that `Configs::get_value('delete_imported', '0')` returns a boolean...

ildyria:
> Shouldn't we always delete the imported file in this particular case? We are the ones who created the file by downloading the picture from the URL before importing it into Lychee. What's the point of keeping it around?

Agreed.

app/Http/Controllers/ImportController.php
kamil4: Duh! You would think I would know that, considering that I added that parameter, and pretty recently at that...

kamil4:
OK, I added to more commits:
* always delete the downloaded source in this case (including in the previously missing case of a duplicate),
* while testing it I noticed that we don't delete small@2x; it must've been a copy-and-paste error.

ildyria:
```
1) Tests\Feature\UpdateTest::test_do_logged
Response status code [500] does not match expected 200 status code.
Failed asserting that false is true.
```
Humm....

d7415:
I was expecting this to be longer... Thanks @kamil4!

PR #359, opened 2019-10-02
Adding functionality to display photos of an album on a map

tmp-hallenser:
This PR adds the functionality to display photos of an album (and its sub albums) on a map. 

* Adds map related string constants 
* Adds map button to header (only album and photo header)
* Creates new header for map view
* Adds map tag to container

Changes von front are covered by separate PR.


app/Http/Controllers/AlbumController.php
```diff
@@ -165,6 +165,108 @@ public function get(Request $request)
...
+	/**
+	 * Provided an albumID, returns the album with only map related data.
+	 *
+	 * @param Request $request
+	 *
+	 * @return array|string
+	 */
+	public function getPositionData(Request $request)
+	{
+		$request->validate(['albumID' => 'string|required']);
```
kamil4: Shouldn't `includeSubAlbums` also be here?

app/Http/Controllers/AlbumController.php
```diff
@@ -165,6 +165,108 @@ public function get(Request $request)
...
+				if ($this->sessionFunctions->is_logged_in()) {
+					$username = $album->owner->username;
+				}
+
+				$full_photo = $album->full_photo_visible();
+
+				$album_list = array();
+				if ($request['includeSubAlbums']) {
+					// Get all subalbums of the current album
+					$album_list = $this->albumFunctions->getAlbumIDsfromAlbumTree($this->albumFunctions->get_albums($album, $username));
```
kamil4: Shouldn't you be using `AlbumFunctions::get_sub_albums()` here? What you are doing seems more complicated than it needs to be, and more costly.

app/Http/Controllers/AlbumsController.php
```diff
@@ -57,4 +59,90 @@ public function get()
...
+
+		// All normal albums
+		$toplevel_albums = $this->albumFunctions->prepare_albums($toplevel['albums']);
+		// Add albums shared with users
+		$toplevel_albums = array_merge($toplevel_albums, $this->albumFunctions->prepare_albums($toplevel['shared_albums']));
+
+		$toplevel_album_ids = array();
+		$album_list = array();
+
+		foreach ($toplevel_albums as $album_iterator) {
```
kamil4: What is it that you are trying to do here? Generate an array of all accessible albums? `AlbumFunctions::getPublicAlbums()` already does that, and much more efficiently than the code here.

app/Http/Controllers/AlbumsController.php
```diff
@@ -57,4 +59,90 @@ public function get()
...
+
+			// Get all photos of subalbums -> you only call this function
+			// if you want all photos of subalbums
+			$album_list = array_merge($album_list, $this->albumFunctions->getAlbumIDsfromAlbumTree($this->albumFunctions->get_albums($album_tmp, $username)));
+		}
+
+		$album_list = array_merge($album_list, $toplevel_album_ids);
+		$photos_sql = Photo::whereIn('album_id', $album_list);
+
+		// Unsorted photos only visible for admin (album_id = null)
```
kamil4: Erm, no? Every user with upload rights can have unsorted photos.

See https://github.com/LycheeOrg/Lychee-Laravel/blob/master/app/Http/Controllers/SearchController.php#L193-L210 for an example of how to do this.

app/Http/Controllers/AlbumsController.php
```diff
@@ -57,4 +59,90 @@ public function get()
...
+		if ($this->sessionFunctions->is_admin()) {
+			$photos_sql = $photos_sql->orWhereNull('album_id');
+		}
+
+		$return['photos'] = $this->albumFunctions->photosLocationData($photos_sql, $full_photo);
+		$return['id'] = '';
+
+		// Remove all unnecessary data
```
kamil4: Why do you need to unset these? Where are they set?

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -676,4 +711,27 @@ function (Builder $query) {
...
 	}
+
+	/**
+	 * Provided an album tree structure, returns all album IDs.
+	 *
+	 * @param array $album_tree_structure
+	 *
+	 * @return array
+	 */
+	public function getAlbumIDsfromAlbumTree(array $album_tree_structure)
```
kamil4: If you use the existing methods I listed earlier, this should not be needed.

app/Locale/English.php
```diff
@@ -319,7 +325,15 @@ public static function get_locale()
 			'OVERLAY_EXIF' => 'Photo EXIF data',
 			'OVERLAY_DESCRIPTION' => 'Photo description',
 			'OVERLAY_DATE' => 'Photo date taken',
-			'MAP_DISPLAY_TEXT' => 'Display coordinates on map (OpenStreetMap):',
+			'MAP_DISPLAY_TEXT' => 'Enable maps (provided by OpenStreetMap):',
+			'MAP_DISPLAY_PUBLIC_TEXT' => 'Enable maps for public albums (provided by OpenStreetMap):',
+			'MAP_PROVIDER' => 'Provider of OpenStreetMap tiles:',
+			'MAP_PROVIDER_WIKIMEDIA' => 'Wikimedia',
+			'MAP_PROVIDER_OSM_ORG' => 'OpenStreetMap.org (no retina)',
```
kamil4: While not a major concern of course, I would prefer if we used industry-wide term _HiDPI_ rather than Apple-specific _retina_.

app/Http/Controllers/AlbumsController.php
```diff
@@ -70,78 +72,29 @@ public function getPositionData()
...
-		foreach ($toplevel_albums as $album_iterator) {
-			$toplevel_album_ids[] = $album_iterator['id'];
-
-			$album_tmp = Album::with('children')->find($album_iterator['id']);
-			$full_photo = $album_tmp->full_photo_visible();
+		$albumIDs = $this->albumFunctions->getPublicAlbums();
+
+		$query = Photo::with('album')->where(
+				function (Builder $query) use ($albumIDs) {
+					$query->whereIn('album_id', $albumIDs);
```
ildyria: This line does nothing.

app/Http/Controllers/AlbumsController.php
```diff
@@ -70,78 +72,29 @@ public function getPositionData()
...
-
-		foreach ($toplevel_albums as $album_iterator) {
-			$toplevel_album_ids[] = $album_iterator['id'];
-
-			$album_tmp = Album::with('children')->find($album_iterator['id']);
-			$full_photo = $album_tmp->full_photo_visible();
+		$albumIDs = $this->albumFunctions->getPublicAlbums();
+
+		$query = Photo::with('album')->where(
+				function (Builder $query) use ($albumIDs) {
```
ildyria: Hummm I am not happy with this function. There should be a simpler way to achieve that.

app/Http/Controllers/AlbumsController.php
kamil4: Erm, sure it does: it adds all the accessible albums to the query.

Let's recap what we are trying to do here so that there are no misunderstandings: we want a query that is going to go over all the photos accessible to the current user by simple browsing from the top level (i.e., hidden and password-protected albums excluded). `AlbumFunctions::getPublicAlbums()` already did the heavy lifting, and its result is in `$albumIDs`. We "just" need to add the unsorted photos. Which is tricky, since they don't belong to an album, and in multi-user case, are mixed together so also need to be filtered by `owner id`. That's what the code below this line does, and I don't think that the desired effect can be achieved with a static query, though I'll be happy to be proven wrong. At most, the line above can probably be moved outside the function, but will that change anything?

So what is your concern here exactly?

ildyria: nope, it should be `$query = $query->whereIn('album_id', ...)` or I am wrong. :)

kamil4: I totally missed that! Now I'm beginning to wonder how things work at all in the `SearchController` because there's the same bug there...

kamil4: No, wait... Look at https://github.com/LycheeOrg/Lychee-Laravel/blob/master/app/Http/Controllers/SearchController.php#L213-L221. It's the same "erroneous" syntax yet that one definitely has an effect?!

app/Http/Controllers/AlbumController.php
```diff
@@ -165,6 +165,109 @@ public function get(Request $request)
...
+				$photos_sql = Photo::select_unsorted(Photo::OwnedBy($UserId));
+				break;
+			default:
+				$album = Album::with('children')->find($request['albumID']);
+				if ($album === null) {
+					Logs::error(__METHOD__, __LINE__, 'Could not find specified album');
+
+					return 'false';
+				}
+				$return = $album->prepareData();
```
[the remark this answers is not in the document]
tmp-hallenser: you're right

app/Http/Controllers/AlbumController.php
```diff
@@ -165,6 +165,109 @@ public function get(Request $request)
...
+
+				break;
+		}
+
+		$return['photos'] = $this->albumFunctions->photosLocationData($photos_sql, $full_photo);
+
+		$return['id'] = $request['albumID'];
+
+		// Remove all unnecessary data
+		unset($return['albums']);
```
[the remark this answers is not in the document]
tmp-hallenser: correct

kamil4:
I have some cleanup to the front end that I'll submit shortly but I believe this is now ready for merge.

PR #361, opened 2019-10-03
build a "small" installer from scratch

### LINKED ISSUE #360: Add php-based install script for users without shell access
Primarily for the artisan steps of install
```sh
php artisan migrate
php artisan key:generate
```

ildyria:
Fixes #360 

to access it open `install.php` in your web browser.
It is not in the `public` folder to prevent unauthorized users to access it.

~Still need to add more safety in it (like disabling it completely if some file is found).~


ildyria:
> Still need to add more safety in it (like disabling it completely if some file is found).

Fixed.

installer/Helpers/RequirementsChecker.php
```diff
@@ -0,0 +1,144 @@
...
+			'supported' => $supported,
+		];
+
+		return $phpStatus;
+	}
+
+
+
+	/**
+	 * Check if exec is enabled. This will allow use to execute the migration
```
kamil4: Typo: allow _us_ to execute the migration

installer/Helpers/RequirementsChecker.php
```diff
@@ -0,0 +1,144 @@
...
+
+	/**
+	 * Get current Php version information.
+	 *
+	 * @return array
+	 */
+	private static function getPhpVersionInfo()
+	{
+		$currentVersionFull = PHP_VERSION;
+		preg_match("#^\d+(\.\d+)*#", $currentVersionFull, $filtered);
```
kamil4: I'm trying to understand what this does. Will it effectively strip any `-rc` strings and such, keeping just the numbers and dots?

installer/Templates/Permissions.php
```diff
@@ -0,0 +1,37 @@
...
+use Template;
+
+class Permissions implements Template
+{
+
+	public function print(array $input = [])
+	{
+		echo "\t".'<ul class="list">'."\n";
+		foreach ($input['permissions'] as $permission) {
+			echo "\t"."\t".'<li class="list__item list__item--permissions '
```
kamil4: Why isn't this simply `"\t\t"`?

installer/Templates/Requirements.php
```diff
@@ -0,0 +1,66 @@
...
+				echo "\t"."\t".'</small>'."\n";
+				echo "\t"."\t".'</strong>'."\n";
+				echo "\t"."\t".'<span class="float-right" >'."\n";
+				echo "\t"."\t".'<strong>'."\n";
+				echo $input['phpSupportInfo']['current'];
+				echo "\t"."\t".'</strong>'."\n";
+				echo "\t"."\t".'<i class="fa fa-fw fa-';
+				echo $input['phpSupportInfo']['supported'] ? 'check-circle-o'
+					: 'exclamation-circle';
+				echo ' row-icon" aria-hidden= "true"></i></span>'."\n";
```
kamil4: Unintended space after `aria-hidden=`?

installer/Templates/Requirements.php
```diff
@@ -0,0 +1,66 @@
...
+	public function print(array $input = [])
+	{
+		echo "<br>";
+		foreach (
+			$input['requirements'] as $type => $requirement
+		) {
+			echo "<br>";
+			echo "\t".'<ul class="list">';
+			echo "\t"."\t".'<li class="list__item list__title ';
+			echo $input['phpSupportInfo']['supported'] ? 'success' : 'error';
```
kamil4: So this loop will run for `php` and `apache`? Why should `phpSupportInfo` be used for anything in case of the `apache` iteration?

installer/Templates/Permissions.php
```diff
@@ -0,0 +1,37 @@
...
+				.($permission['isSet'] ? 'success' : 'error').'">'."\n";
+			echo "\t"."\t".$permission['folder']."\n";
```
kamil4: I'm having difficulty finding it in the source code (it must be somewhere in the CSS, I guess) but I don't like the fact that the pathnames displayed in the web browser start with an uppercase...

installer/Templates/Head.php
```diff
@@ -0,0 +1,103 @@
...
+			.'">';
+		echo '<i class="step__icon fa fa-server" aria-hidden="true"></i>';
+		echo '
+                    </li>';
+		echo '
+                    <li class="step__divider"></li>
+                    <li class="step__item '.($input['step'] == 'Env' ? 'active' : '')
+			.'">';
+		if ($this->get_index($input['step']) >= 3) {
+			echo '<a href="?step=env">
```
kamil4: Can you add a `title` or some such to these buttons so that there's an explanation when one hovers over? I don't find the icons to be particularly obvious :smile:.

installer/Templates/Env.php
```diff
@@ -0,0 +1,32 @@
...
+
+use Template;
+
+class Env implements Template
+{
+
+	public function print(array $input = [])
+	{
+		echo '<form method="post" action="?step=env">
+        <textarea class="textarea" name="envConfig">'.$input['env'].'</textarea>
```
kamil4: It's not immediately obvious that this is an editable area. Perhaps there should be a one-line header telling the user to edit the file as appropriate?

install.php
```diff
@@ -0,0 +1,41 @@
...
+
+// routes & dispatch
+if ($check == false)
+{
+	$controller = $routes->dispatch();
+	$view->apply($controller->view(), $controller->do());
+}
+else
+{
+	$view->apply('Migrate', $check);
```
kamil4: I'm not sure what this is supposed to do? The installer does generate `installed.log` for me but the next time I start it, it starts from the beginning as if nothing happened. Is that how it's supposed to work?

installer/Config.php
```diff
@@ -0,0 +1,113 @@
...
+			/*
+			|--------------------------------------------------------------------------
+			| Folders Permissions
+			|--------------------------------------------------------------------------
+			|
+			| This is the default Laravel folders permissions, if your application
+			| requires more permissions just add them to the array list bellow.
+			|
+			*/
+			'permissions'  => [
```
kamil4: I don't think that enforcing particular values of the permission bits is the right approach. For instance, what if I simply change the ownership of the whole Lychee tree to `www-data`? I don't need to set any group permissions then. Or, in my particular case, the directories are normally owned by `kamil:users` and changing them to `775` won't give the http daemon the permissions it needs but the installer will be happy. In `app/Http/Controllers/DiagnosticsController.php` we use `is_writeable()` which strikes me as much more robust (I am naively assuming that it's implemented sanely behind the scenes :smiley:).

ildyria:
> So why is `install.php` not in `public`? At least in my production Lychee installation, the top-level directory cannot be accessed via a URL and making it accessible strikes me as a dangerous precedent (what would prevent somebody from getting to `.env` then, which holds privileged information)?

@kamil4 thanks for your Review. I'll fix things. :)

installer/Helpers/RequirementsChecker.php
ildyria: yep

installer/Templates/Requirements.php
ildyria: It does not matter in the case of Apache as it does not have a text at the right (that is the thing colored). The coloring of `<strong>` is what matters.

install.php
ildyria: > but the next time I start it

That is not the case on my machine. Very weird.

installer/Templates/Env.php
ildyria: fixed.

installer/Templates/Head.php
ildyria: fixed.

installer/Templates/Permissions.php
ildyria: I did not look into it.

installer/Templates/Head.php
```diff
@@ -0,0 +1,103 @@
...
+		} else {
+			echo '<i class="step__icon fa fa-key" aria-hidden="true"></i>';
+		};
+		echo '
+                    </li>
+                    <li class="step__divider"></li>
+                    <li class="step__item '.($input['step'] == 'Requirements' ? 'active'
+				: '').'" title="Checking Requirements">';
+
+		if ($this->get_index($input['step']) == 1) {
```
kamil4: Shouldn't this be `>=` ?

installer/Templates/Head.php
```diff
@@ -0,0 +1,103 @@
...
+                    </li>
+                    <li class="step__divider"></li>
+                </ul>
+                <div class="main">';
+
+		if (isset($input['errors']))
+		{
+			echo '
+		            <p class="alert alert-danger text-center">
+		                <strong>Please fix the errors before going to the next step.</strong>
```
kamil4: I think that in case of errors being detected, there should be button the user can click to rerun the test that failed (you know, the user fixes the permission or whatnot in the background and then wants to continue). Right now I can click on the buttons at the top, but they don't actually rerun the tests; they just display the same thing. I need to reload the whole page in the browser, which is not exactly user-friendly.

installer/Controller/EnvController.php
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * @return array
+	 */
+	public function do()
+	{
+
+		if (isset($_POST['envConfig']))
+		{
+			file_put_contents('.env',$_POST['envConfig']);
```
kamil4: This results in ugly DOS-style CRLF line terminators, apparently per HTML specs. I guess you should filter out `\r` characters before writing?

installer/Controller/MigrationController.php
```diff
@@ -0,0 +1,46 @@
...
+
+class MigrationController implements Controller
+{
+
+	/**
+	 * @return array
+	 */
+	public function do()
+	{
+		$output = [];
```
kamil4: So even though hovering over the button for this step says _Creating the Database_, this doesn't actually create it. Is that impossible because we now support multiple database engines? If that's the case, then I guess an additional test step should be added after the environment is set up, to check whether the database exists?

installer/Templates/Permissions.php
```diff
@@ -8,29 +8,43 @@
...
+		$succ = '<i class="fa fa-fw fa-check-circle-o success"></i>';
+		foreach ($array_permission as $perm)
+		{
+			$ret[] = '<span class="perm float-right">'.($areSet & 1 ? $err : $succ).$perm.'</span>';
+			$areSet >>= 1;
+		}
+		$ret =  join(' ', $ret);
+		$ret = str_replace('file_','',$ret);
+		$ret = str_replace('!','not',$ret);
+		$ret = str_replace('is_',' ',$ret);
```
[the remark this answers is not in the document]
ildyria: hahaha :D But the condition I want to be available is more like "not readable" and "not writable".

installer/Middleware/InstallCheck.php
```diff
@@ -12,7 +12,9 @@ public function check()
 		$logfile = 'installed.log';
 		if (file_exists($logfile))
 		{
-			return ['lines' => [file_get_contents($logfile)]];
+			// we directly redirect to gallery
+			header("Location: /gallery");
```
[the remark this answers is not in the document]
ildyria: hummm noted.

installer/Config.php
```diff
@@ -39,19 +39,20 @@ class Config
...
-				'public/uploads/big/'    => '775',
-				'public/uploads/import/' => '775',
-				'public/uploads/medium/' => '775',
-				'public/uploads/raw/'    => '775',
-				'public/uploads/small/'  => '775',
-				'public/uploads/thumb/'  => '775',
+				'.'                      => 'file_exists|is_readable|is_writable',
+				'storage/framework/'     => 'file_exists|is_readable|is_writable',
+				'storage/logs/'          => 'file_exists|is_readable|is_writable',
+				'bootstrap/cache/'       => 'file_exists|is_readable|is_writable',
```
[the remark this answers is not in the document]
ildyria: They have to be writable. Because otherwise Laravel cannot create the log, sessions and cache files.

install.php
ildyria: fixed.

installer/Controller/MigrationController.php
ildyria: good point.

ildyria: but how...

installer/Templates/Head.php
ildyria: will see what I can do.

installer/Config.php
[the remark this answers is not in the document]
kamil4: I know about `storage/framework/` and `storage/logs/` and I agree, but I'm asking specifically about `bootstrap/cache/`. For me, nothing gets written there. Is it actually being used in your case?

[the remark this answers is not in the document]
ildyria: https://laravel.com/docs/5.7/structure#the-bootstrap-directory

> The `bootstrap` directory contains the `app.php` file which bootstraps the framework. This directory also houses a `cache` directory which contains framework generated files for performance optimization such as the route and services cache files.

This directory is being used if you do `artisan config:cache` `artisan route:cache`  and so one.

ildyria:
> * The so-called database creation step doesn't actually create a database?

I made it more clear to the user.

> * Client-side caching of the gallery redirect in case `installed.php` already exists. That's actually a major PITA... I was forced to use a separate browser profile with history disabled to be able to do any meaningful testing...

Should be fixed with a `http_response_code(307);`.

tmp-hallenser:
Looks good to me (with a nice UI :))

installer/Controller/MigrationController.php
```diff
@@ -0,0 +1,84 @@
...
+		return ['lines' => $output, 'errors' => $error];
+	}
+
+	/**
+	 * @return bool|null
+	 */
+	private function migrate(array &$output)
+	{
+		$res = 0;
+		exec('php artisan migrate', $output, $res);
```
d7415: Can we not use `Artisan::call('migrate', ['--force' => true]);` for the initial migration?

installer/Controller/MigrationController.php
```diff
@@ -0,0 +1,84 @@
...
+	{
+		return 'Migrate';
+	}
+
+	/**
+	 * @param array $output
+	 */
+	public function installed(array &$output)
+	{
+		$dateStamp = date('Y/m/d h:i:sa');
```
d7415: As above re ISO8601.

installer/Helpers/RequirementsChecker.php
```diff
@@ -0,0 +1,131 @@
...
+						$results['requirements'][$type][$requirement] = true;
+						if (!extension_loaded($requirement)) {
+							$results['requirements'][$type][$requirement]
+								= false;
+							$results['errors'] = true;
+						}
+					}
+
+					if ($this->checkExec()) {
+						$results['requirements'][$type]['Php exec() available'] = true;
```
d7415: Is it worth tracking the negative here?

database/migrations/2064_12_25_0000_generate_installed_log.php
```diff
@@ -0,0 +1,29 @@
...
+{
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		$dateStamp = date('Y/m/d h:i:sa');
+		$message = 'Lychee INSTALLED on ' . $dateStamp;
```
d7415: If we're only using this to block the installer, maybe we should change "on" to "before" here?

installer/Templates/Env.php
```diff
@@ -0,0 +1,35 @@
...
+{
+	public function print(array $input = [])
+	{
+		echo '
+        <ul class="list">
+        <li class="list__item list__item--env list__title error"><span><i class="fa fa-ban" aria-hidden="true"></i> <strong>Lychee does not create the database.</strong></span></li>
+        <li class="list__item list__item--env">1 - Manually create your database and then enter the sql details bellow.</li>
+        <li class="list__item list__item--env">2 - Edit the form below to reflect your desired configuration.</li>
+        </ul>
+		<strong>For more details of how those values are used, look into "config" folder.</strong><br/>
```
d7415: s/into/in the/

installer/Templates/Head.php
```diff
@@ -0,0 +1,100 @@
...
+		return array_search($step, $this->steps);
+	}
+
+	public function print(array $input = [])
+	{
+		echo '<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="utf-8">
+        <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
d7415: I think if a user needs this when *installing* Lychee then all hope is already lost. This should only be an issue pre IE8.

installer/Templates/Head.php
```diff
@@ -0,0 +1,100 @@
...
+                    <h1 class="header__title">' . $input['title'] . '</h1>
+                </div>
+                <ul class="step">
+                    <li class="step__divider"></li>
+                    <li class="step__item ' . ($input['step'] == 'Migrate' ? 'active' : '')
+			. '"  title="Apply the database migration">';
+		echo '<i class="step__icon fa fa-server" aria-hidden="true"></i>';
+		echo '
+                    </li>';
+		echo '
```
d7415: Why are these split out?

installer/Controller/MigrationController.php
ildyria: No, because `Artisan::call()` is not available yet. :)
This is bootstrapped, we do not load anything of Laravel yet.

database/migrations/2064_12_25_0000_generate_installed_log.php
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	/**
+	 * Reverse the migrations.
+	 *
+	 * @return void
+	 */
+	public function down()
+	{
+		// THIS WILL MAKE YOUR
```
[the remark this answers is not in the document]
ildyria: fixed.

database/migrations/2064_12_25_0000_generate_installed_log.php
```diff
@@ -0,0 +1,29 @@
...
+class GenerateInstalledLog extends Migration
+{
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		$dateStamp = date('Y/m/d h:i:sa');
```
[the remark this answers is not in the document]
ildyria: fixed.

installer/Config.php
```diff
@@ -0,0 +1,108 @@
...
+					'gd',
+				],
+				'apache' => ['mod_rewrite'],
+			],
+			/*
+			|--------------------------------------------------------------------------
+			| Folders Permissions
+			|--------------------------------------------------------------------------
+			|
+			| This is the default Laravel folders permissions, if your application
```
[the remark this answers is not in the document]
ildyria: fixed.

installer/Config.php
```diff
@@ -0,0 +1,108 @@
...
+class Config
+{
+	private $config
+		= [
+			/*
+			|--------------------------------------------------------------------------
+			| Server Requirements
+			|--------------------------------------------------------------------------
+			|
+			| This is the default Laravel server requirements, you can add as many
```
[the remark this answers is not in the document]
ildyria: fixed.

installer/Config.php
```diff
@@ -0,0 +1,108 @@
...
+			////						'mail_port'           => 'required|string|max:50',
+			////						'mail_username'       => 'required|string|max:50',
+			////						'mail_password'       => 'required|string|max:50',
+			////						'mail_encryption'     => 'required|string|max:50',
+			////						'pusher_app_id'       => 'max:50',
+			////						'pusher_app_key'      => 'max:50',
+			////						'pusher_app_secret'   => 'max:50',
+			//					],
+			//				],
+			//			],
```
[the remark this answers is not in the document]
ildyria: fixed.

installer/Controller/EnvController.php
```diff
@@ -0,0 +1,35 @@
...
+class EnvController implements Controller
+{
+	/**
+	 * @return array
+	 */
+	public function do()
+	{
+		if (isset($_POST['envConfig'])) {
+			file_put_contents('.env', str_replace("\r", '', $_POST['envConfig']));
+		}
```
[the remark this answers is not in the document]
ildyria: better ?

installer/Controller/MigrationController.php
ildyria: fixed.

database/migrations/2064_12_25_0000_generate_installed_log.php
ildyria: Well yes and no.
The file is created by the migration file, but we don't know if the migration has been run:

1. automatically by hand or on update `git pull; php artisan migrate` or via online update.
2. at install `git clone ...; composer install; php artisan migrate`
3. by the installer online https://www.example.com/install.php.

For **1.** then *before* would make sense. For **2.** and  **3.** then *on* makes more sense. :unamused:

installer/Templates/Head.php
ildyria: So what do I do there ? :sweat_smile:

installer/Templates/Head.php
ildyria: fixed.

installer/Templates/Head.php
d7415: I'd remove the `<meta http-equiv...>` line, but it's not going to hurt really. I may have done this review shortly after reading a summary of the mess that is user agents...

installer/Helpers/RequirementsChecker.php
ildyria: I would say yes. Because it tells the user why he cannot install it: we need the exec to apply the migration. ;)

installer/Controller/EnvController.php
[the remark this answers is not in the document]
d7415: Yup. I'm not sure the initialisation is necessary as they're set on the `if` and `else`, but it can't hurt.

installer/Helpers/RequirementsChecker.php
d7415: I think I meant displaying the result if `exec` is enabled. In hindsight I think not.

database/migrations/2064_12_25_0000_generate_installed_log.php
d7415: Fair point :+1:

installer/Helpers/RequirementsChecker.php
ildyria: Ah, I see what you mean.
Well the way I see it is that we display the actual dependencies, and show that they are satisfied.
I would rather see what dependencies I need and are checked. Rather than just a message "All dependencies are satisfied, you are good to go, Sir." :smiley:

PR #368, opened 2019-10-08
Skip duplicate albums with skip_duplicates

### LINKED ISSUE #366: skip_duplicates for Imported Folders?
I'm really enjoying lychee-laravel, but I notice that "skip_duplicates" doesn't apply to imported folders. I'm using google drive as a repository, then using drive to sync those folders/images to my server.

But when I choose "Import from Server" it imports all of the folder structure as Albums again. In my case I have C1/C2/C3/C4 as folders, with subfolders underneath. So even if i do a merge, my subfolders are duplicated.

Is there a simple code change (php etc) that I can do to avoid this? My album structure won't change, so even if I can disable "folder imports" and just import files into the appropriate structure?

thanks!

kamil4:
Fixes #366

### DIFF
--- a/app/Http/Controllers/ImportController.php
+++ b/app/Http/Controllers/ImportController.php
@@ -263,9 +263,6 @@ class ImportController extends Controller
 			return;
 		}
 
-		$contains['photos'] = false;
-		$contains['albums'] = false;
-
 		// We process breadth-first: first all the files in a directory,
 		// then the subdirectories.  This way, if the process fails along the
 		// way, it's much easier for the user to figure out what was imported
@@ -311,7 +308,6 @@ class ImportController extends Controller
 			$extension = Helpers::getExtension($file, true);
 			if (@exif_imagetype($file) !== false || in_array(strtolower($extension), $this->photoFunctions->validExtensions, true)) {
 				// Photo or Video
-				$contains['photos'] = true;
 				if ($this->photo($file, $delete_imported, $albumID) === false) {
 					$this->status_update('Problem: ' . $file . ': Could not import file');
 					Logs::error(__METHOD__, __LINE__, 'Could not import file (' . $file . ')');
@@ -330,7 +326,7 @@ class ImportController extends Controller
 			// Folder
 			$album = null;
 			if (Configs::get_value('skip_duplicates', '0') === '1') {
-				$album = Album::where('parent_id', '=', $albumID)
+				$album = Album::where('parent_id', '=', $albumID == 0 ? null : $albumID)
 					->where('title', '=', basename($dir))
 					->get()
 					->first();
@@ -345,7 +341,6 @@ class ImportController extends Controller
 				}
 			}
 			$newAlbumID = $album->id;
-			$contains['albums'] = true;
 			$this->server_exec($dir . '/', $newAlbumID, $delete_imported);
 		}
 	}
--- a/app/Http/Controllers/ImportController.php
+++ b/app/Http/Controllers/ImportController.php
@@ -6,6 +6,7 @@
 
 namespace App\Http\Controllers;
 
+use App\Album;
 use App\Configs;
 use App\Logs;
 use App\ModelFunctions\AlbumFunctions;
@@ -327,12 +328,21 @@ class ImportController extends Controller
 		// Album creation
 		foreach ($dirs as $dir) {
 			// Folder
-			$album = $this->albumFunctions->create(basename($dir), $albumID, $this->sessionFunctions->id());
-			// this actually should not fail.
-			if ($album === false) {
-				$this->status_update('Problem: ' . $basename($dir) . ': Could not create album');
-				Logs::error(__METHOD__, __LINE__, 'Could not create album in Lychee (' . basename($dir) . ')');
-				continue;
+			$album = null;
+			if (Configs::get_value('skip_duplicates', '0') === '1') {
+				$album = Album::where('parent_id', '=', $albumID)
+					->where('title', '=', basename($dir))
+					->get()
+					->first();
+			}
+			if ($album === null) {
+				$album = $this->albumFunctions->create(basename($dir), $albumID, $this->sessionFunctions->id());
+				// this actually should not fail.
+				if ($album === false) {
+					$this->status_update('Problem: ' . $basename($dir) . ': Could not create album');
+					Logs::error(__METHOD__, __LINE__, 'Could not create album in Lychee (' . basename($dir) . ')');
+					continue;
+				}
 			}
 			$newAlbumID = $album->id;
 			$contains['albums'] = true;
--- a/app/ModelFunctions/PhotoFunctions.php
+++ b/app/ModelFunctions/PhotoFunctions.php
@@ -312,7 +312,7 @@ class PhotoFunctions
 				@unlink($tmp_name);
 			}
 			// Check if the user wants to skip duplicates
-			if (Configs::get()['skip_duplicates'] === '1') {
+			if (Configs::get_value('skip_duplicates', '0') === '1') {
 				Logs::notice(__METHOD__, __LINE__, 'Skipped upload of existing photo because skipDuplicates is activated');
 
 				return Response::warning('This photo has been skipped because it\'s already in your library.');


avieros:
issue not solved on my testing - results in duplicate folders again...

avieros:
is it because my parent folder is "password protected" & "public" , and all subfolders are "public" 
is it ignoring those set flags, and assuming the folders are different?

d7415:
@avieros Are your photos in the duplicate albums? And is `skip_duplicates` set to 1? (separate questions for separate reasons)

d7415:
Pending @avieros's issues, it looks good to me :) As an aside, we should probably handle read errors better - we're currently counting them as "skipped" in the import dialog (the logs are correct)

kamil4:
@d7415 I find that the existing `skip_duplicates` feature has a rather counterintuitive semantics for duplicate pictures. It looks for duplicates exclusively based on the checksum of the file content. If it finds a duplicate, it doesn't add another copy to the current album at all. However, it doesn't mean that the duplicate is in the same album or has the same name. Also, the duplicate condition is currently not reported via the Import from Server dialog, so it can be rather confusing why some pictures aren't showing up, unless one checks the logs. Now that I think about it, the existing duplicate may not even belong to the same user, so this functionality can be used as a side-channel attack to determine if some other user of the same gallery has a given picture in their album already. Do we care?

kamil4:
> As an aside, we should probably handle read errors better - we're currently counting them as "skipped" in the import dialog (the logs are correct)

I think I may have coded it that way on purpose, reserving error reports (in red) for conditions that abort the whole import process. I believe my thinking was that skipping some files would be fairly common... Now the question is: should skipped duplicates be reported via the GUI?

kamil4:
@avieros I'm surprised it doesn't work for you; it seems to be pretty reliable over here. Can you do some more experimenting with simpler directory structures (maybe just a single subdirectory to begin)? Also, you mentioned in #366 using Google Drive -- are you somehow importing directly from it or are the files on a regular file system on your server by the time you import them?

It would help if you could include the output of `ls -lR` from your import directory and the log entries from `storage/logs/laravel.log` generated from the first and the second import (make sure that `APP_DEBUG=true` in top-level `.env` for this to be of value). Screenshots of the imported albums wouldn't hurt either :smile:.

d7415:
> It looks for duplicates exclusively based on the checksum of the file content. If it finds a duplicate, it doesn't add another copy to the current album at all. 

Yeah, probably a carry-over from v3? Having just updated the duplication code, would that fit here? We'd still have to check for location though...

> this functionality can be used as a side-channel attack to determine if some other user of the same gallery has a given picture in their album already. Do we care?

Not really. The attacker needs 1) the photo in question 2) an account on the server 3) no access to the album that photo is in. And it only shows existence, not location or metadata.

I think we should fix it, but to handle duplicates better, not as a security issue.

> I think I may have coded it that way on purpose, reserving error reports (in red) for conditions that abort the whole import process. I believe my thinking was that skipping some files would be fairly common... Now the question is: should skipped duplicates be reported via the GUI?

Best case would probably be a summary `Skipped 123 files and 5 folders as duplicates` to avoid noise. My next best would be a `Some files were skipped as duplicates. More information can be found in the logs`. If we were feeling really clever, make "`logs`" a link, and make that sentence admin-only :P I'd be inclined to separate out access errors though (even if, again, it's only a summary).

kamil4:
> > It looks for duplicates exclusively based on the checksum of the file content. If it finds a duplicate, it doesn't add another copy to the current album at all.
> 
> Yeah, probably a carry-over from v3? Having just updated the duplication code, would that fit here? We'd still have to check for location though...

The central location that checks for `skip_duplicates` is `PhotoFunctions::add`, so both _Import_ and the regular _Upload_ are affected. I'm not sure what the intended use case of `skip_duplicates` is, but the current behavior strikes me as as a bit of a misfeature. _Upload_ is a very explicit user action and I don't think that `skip_duplicates` should be able to affect it, yet right now it does. I think `skip_duplicates` should be applicable exclusively to _Import from Server_, which is designed for bulk imports. Adding duplicate detection there based on filenames rather than checksums, alongside what this PR is already doing for folders, would be trivial.

> I think we should fix it, but to handle duplicates better, not as a security issue.

Agreed...

> > should skipped duplicates be reported via the GUI?
> 
> Best case would probably be a summary `Skipped 123 files and 5 folders as duplicates` to avoid noise.

I had the same concerns that you did regarding the noise in case of the incremental bulk imports (which is the most likely use case for `skip_duplicates`, right?) so I like your idea! I'll see if I can do the summary server-side...

> If we were feeling really clever, make "`logs`" a link, and make that sentence admin-only :P

The whole _Import from Server_ is admin-only already... In general though, I don't like pointing users at the logs for "regular" errors; that should be reserved for debugging...

> I'd be inclined to separate out access errors though (even if, again, it's only a summary).

Yes, duplicates should be separate from access errors. So are you OK with how access errors are currently reported or is something missing there? Because it sounded like you found its current behavior confusing when you tried to use it...

> This shouldn't affect this PR though, unless you want to do more at once?

Well, since this PR is already dealing with `skip_duplicates`, we might just as well _really_ fix it :smile:.

avieros:
> @avieros I'm surprised it doesn't work for you; it seems to be pretty reliable over here. Can you do some more experimenting with simpler directory structures (maybe just a single subdirectory to begin)? Also, you mentioned in #366 using Google Drive -- are you somehow importing directly from it or are the files on a regular file system on your server by the time you import them?
> 
> It would help if you could include the output of `ls -lR` from your import directory and the log entries from `storage/logs/laravel.log` generated from the first and the second import (make sure that `APP_DEBUG=true` in top-level `.env` for this to be of value). Screenshots of the imported albums wouldn't hurt either 😄.

ok:

so, when i talk about google drive, i mean like this:

Google Drive folder:
Folder
    |--- Folder 
             |---- subfolder
                          |---- subfolder
             |---- subfolder

Etc. This folder structure and contents is sync by **rclone** copy google_drive:"/Folder' /gd_pics/ --copy-empty-src-folders .. At present i'm just testing with a few dummy images in different subfolders etc. It doesn't matter if they have images or not, i still get the dupe folders.. 

The structure is like so:
Top Level -> under that is 9 folders/albums -> each album has 6 subfolders (for different days) -> each of those 6 subfolders has 2 different subfolders (for different type of image)
The only folders that have images are the 2 deepest subfolders for each of the 6 subfolders, for each of the 9 top folders. Hope that makes sense.

So will Lychee recreate albums if no image? 

ls -lR (snippet)
`./Tours Photography/W4 - Galleries & Gardens/Wednesday - 20th November:
total 8
drwxr-xr-x 2 root root 4096 Oct  8 16:47 Candid Photos
drwxr-xr-x 2 root root 4096 Oct  8 16:47 Tour Group Photos`

i imported from server
then i imported from server again
logs here: [http://pastebin.zone/ghwMI8lN](url)

image here: [https://i.postimg.cc/3rLzxDvY/dupes.png](url)

kamil4:
Thanks, the log did the trick! Turns out that my PR does not work correctly when importing at the top level. However, if you import into an existing (possibly empty) album instead, it should work the way you want (please check). I will fix it so that it works at the top level as well, of course :smile:.

avieros:
> Thanks, the log did the trick! Turns out that my PR does not work correctly when importing at the top level. However, if you import into an existing (possibly empty) album instead, it should work the way you want (please check). I will fix it so that it works at the top level as well, of course 😄.

Excellent! Thanks so much... bingo! create a test album, import from server into that... import from server again, 0 duplication!! Awesome!

ildyria:
@kamil4 

> I will fix it so that it works at the top level as well, of course :smile: .

So I guess this is still WIP ?

kamil4:
> So I guess this is still WIP ?

Yes. But I'm planning to work on it this weekend.

kamil4:
In the end I committed just a minimal fix for the top-level duplicate album detection.

The rest of the changes we proposed above (changing the behavior of duplicate photo detection, reporting the number of skipped duplicate entries to the user, etc) is left for a future PR...

This should be merged in its current form as it does fix a real issue: duplicate detection of subalbums.

PR #369, opened 2019-10-12
Fix 356

### LINKED ISSUE #356: Diagnostic is too slow

Opening the Diagnostic page should not be that long. This is due to the commit check (request to Github, waiting etc).

### Solutions

- Cache the update check response every *x* days ?
- Remove the update check ?
- Add a link to the front-end to apply update (even if none is available).
- Add button to check for updates ?

ildyria:
Fixes #356 

To merge with https://github.com/LycheeOrg/Lychee-front/pull/156

Use caching to decrease the number of time we need to make a request to github.
I separated the Check for update from the Diagnostic. Need to find a way to have the check for update run in background (for another PR).
Added a button to check if an update is available, this button is only visible if git is functional.


app/Metadata/GitHubFunctions.php
```diff
@@ -212,18 +227,52 @@ public function get_behind_text()
...
+
+		/** @var Carbon|null $last_update */
+		$last_update = $this->get_update_age();
+		if (!$last_update) {
+			$last = 'unknown';
+			$end = '';
+		} else {
+			$last = now()->diffInDays($last_update);
+			$end = $last > 0 ? ' days' : '';
+			$last = ($last == 0 && $end = ' hours') ? now()->diffInHours($last_update) : $last;
```
kamil4: The single `=` in `$end = ' hours'` is not a typo here or in the next two lines, is it? That's at pretty peculiar way of doing this. Can't you use the comma operator after the `?` ? Or is that not available in PHP?

ildyria: Indeed it is not a typo.

I tried to think of simpler ways to do this, but multiple if statements made the code a bit long and not nice to read.
Here I am abusing the short-circuit evaluation of PHP. 
https://stackoverflow.com/questions/4216754/php-lazy-boolean-evaluation

if `$last` is `0` then this means that is was not the previous value (` days`) so we know that the `$end` will be ` hours`. I am using the fact that the `=` returns `1` when being executed.

Yes, I know it is smart code... But I could not find a prettier/more elegant way to do it.

kamil4: In C I would've written it like that _(untested code, may need some brackets etc, but you get the gist)_:

```
$last = ($last == 0) ? $end = ' hours', now()->diffInHours($last_update) : $last;
```

No idea if that works in PHP though?

app/Http/Controllers/DiagnosticsController.php
```diff
@@ -182,27 +206,36 @@ public function get_errors()
...
 		} elseif ($settings['dropbox_key'] == '') {
-			$errors[] = 'Warning: Dropbox import not working. dropbox_key is empty.';
+			$errors[]
+				= 'Warning: Dropbox import not working. dropbox_key is empty.';
 		}
 
 		// Check php.ini Settings
-		if (ini_get('max_execution_time') < 200 && ini_set('upload_max_filesize', '20M') === false) {
-			$errors[] = 'Warning: You may experience problems when uploading a large amount of photos. Take a look in the FAQ for details.';
+		if (ini_get('max_execution_time') < 200
```
kamil4: While not directly related to this PR, should we eliminate this check, now that we reset the timeout for each iteration of import or download?

ildyria: sure. but I guess in another PR. :)

PR #376, opened 2019-10-22
Use the source format for small/medium with GD

### LINKED ISSUE #375: Why images 'medium' are format jpg but with extencion .png?
Hi,

Why, when I upload a image in format PNG (Photo.png), Lychee create the 'medium' in format jpg but with extension '**.png**'?

With thumbnail this doesn't happen, the thumbnail is format 'jpg' with extension .jpeg.

kamil4:
Fixes #375 

* Use the source format for small/medium with GD
* Clean up redundant/confusing calls in Imagick: `setImageFormat` had no effect; the output format is determined by the output file name extension.

### DIFF
--- a/app/Image/GdHandler.php
+++ b/app/Image/GdHandler.php
@@ -53,7 +53,20 @@ class GdHandler implements ImageHandlerInterface
 		}
 
 		imagecopyresampled($image, $sourceImg, 0, 0, 0, 0, $newWidth, $newHeight, $width, $height);
-		imagejpeg($image, $destination, $this->compressionQuality);
+
+		switch ($mime) {
+			case IMAGETYPE_JPEG:
+			case IMAGETYPE_JPEG2000:
+				imagejpeg($image, $destination, $this->compressionQuality);
+				break;
+			case IMAGETYPE_PNG:
+				imagepng($image, $destination);
+				break;
+			case IMAGETYPE_GIF:
+				imagegif($image, $destination);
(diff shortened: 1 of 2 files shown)


app/Image/GdHandler.php
```diff
@@ -53,7 +53,20 @@ public function scale(
...
 		imagecopyresampled($image, $sourceImg, 0, 0, 0, 0, $newWidth, $newHeight, $width, $height);
-		imagejpeg($image, $destination, $this->compressionQuality);
+
+		switch ($mime) {
+			case IMAGETYPE_JPEG:
+			case IMAGETYPE_JPEG2000:
+				imagejpeg($image, $destination, $this->compressionQuality);
+				break;
+			case IMAGETYPE_PNG:
+				imagepng($image, $destination);
```
ildyria: you are missing the compression here: 
https://www.php.net/manual/en/function.imagepng.php

kamil4: What value would you like me to provide though? According to the manual, _The default (-1) uses the zlib compression default_, so I figured I didn't need to provide anything. And the config we use for JPEGs is in a different range (0-100 vs 0-10 here).

ildyria: Well, divide my 10 and round. :) What's the problem ? :smile:

d7415: Also as it's lossless, it has quite a different meaning.

PR #377, opened 2019-10-23
Pass numbers to the front end as strings

kamil4:
Fixes https://github.com/LycheeOrg/Lychee-front/issues/161

The issue there was the identity operator (`===`) resolving to `false` when one side was an integer and the other was a string. It could happen because we were being sloppy on the server side, especially with IDs, sometimes returning them as integers, sometimes as strings. This PR attempts to fix up all such cases that I could find, turning them into strings, which as a bonus should avoid any potential 32-bit issues as well.

### DIFF
--- a/app/Http/Controllers/AlbumController.php
+++ b/app/Http/Controllers/AlbumController.php
@@ -155,10 +155,10 @@ class AlbumController extends Controller
 		$return['photos'] = $this->albumFunctions->photos($photos_sql, $full_photo);
 
 		$return['id'] = $request['albumID'];
-		$return['num'] = count($return['photos']);
+		$return['num'] = strval(count($return['photos']));
 
 		// finalize the loop
-		if ($return['num'] === 0) {
+		if ($return['num'] === '0') {
 			$return['photos'] = false;
 		}
 
--- a/app/ModelFunctions/AlbumFunctions.php
+++ b/app/ModelFunctions/AlbumFunctions.php
@@ -342,7 +342,7 @@ class AlbumFunctions
 			'thumbs' => array(),
 			'thumbs2x' => array(),
 			'types' => array(),
-			'num' => $photos_sql->count(),
+			'num' => strval($photos_sql->count()),
 		);
 
 		/*
--- a/app/Photo.php
+++ b/app/Photo.php
@@ -170,13 +170,13 @@ class Photo extends Model
 		$photo = array();
 
 		// Set unchanged attributes
-		$photo['id'] = $this->id;
+		$photo['id'] = strval($this->id);
 		$photo['title'] = $this->title;
 		$photo['tags'] = $this->tags;
 		$photo['star'] = $this->star == 1 ? '1' : '0';
-		$photo['album'] = $this->album_id;
-		$photo['width'] = $this->width;
-		$photo['height'] = $this->height;
+		$photo['album'] = $this->album_id !== null ? strval($this->album_id) : null;
+		$photo['width'] = strval($this->width);
+		$photo['height'] = strval($this->height);
 		$photo['type'] = $this->type;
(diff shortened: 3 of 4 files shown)


app/Http/Controllers/AlbumController.php
```diff
@@ -155,10 +155,10 @@ public function get(Request $request)
 		$return['photos'] = $this->albumFunctions->photos($photos_sql, $full_photo);
 
 		$return['id'] = $request['albumID'];
-		$return['num'] = count($return['photos']);
+		$return['num'] = strval(count($return['photos']));
```
ildyria: why ? Semantically this one does not make sense.

app/Http/Controllers/AlbumController.php
```diff
@@ -155,10 +155,10 @@ public function get(Request $request)
...
 
 		// finalize the loop
-		if ($return['num'] === 0) {
+		if ($return['num'] === '0') {
```
ildyria: Idem

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -342,7 +342,7 @@ public function genSmartAlbumsThumbs(array &$return, Builder $photos_sql, string
 			'thumbs' => array(),
 			'thumbs2x' => array(),
 			'types' => array(),
-			'num' => $photos_sql->count(),
+			'num' => strval($photos_sql->count()),
```
ildyria: item.

app/Photo.php
```diff
@@ -170,13 +170,13 @@ public function prepareData()
...
-		$photo['id'] = $this->id;
+		$photo['id'] = strval($this->id);
 		$photo['title'] = $this->title;
 		$photo['tags'] = $this->tags;
 		$photo['star'] = $this->star == 1 ? '1' : '0';
-		$photo['album'] = $this->album_id;
-		$photo['width'] = $this->width;
-		$photo['height'] = $this->height;
+		$photo['album'] = $this->album_id !== null ? strval($this->album_id) : null;
+		$photo['width'] = strval($this->width);
```
ildyria: given that they are used as strings in the front-end sure. :)
Does this interfere with the Justified Layout ??

app/Photo.php
```diff
@@ -170,13 +170,13 @@ public function prepareData()
...
+		$photo['height'] = strval($this->height);
```
ildyria: given that they are used as strings in the front-end sure. :)
Does this interfere with the Justified Layout ??

d7415:
I'm mostly a fan of treating numbers as numbers, but consistency is important, and it's not like we use ID as an int once it's generated. It also removes any requirement for special handling of leading zeroes.

I agree with @ildyria though, that anything used as a number in the front end should really be left as such unless we have a good reason.

kamil4:
You are correct that only the integer IDs were found to be a problem in practice so I could've stopped there.

I converted `width` and `height` to strings as well because that's how they are sent in v3 and I figured it was better to be consistent (they were the only int fields in their JSON objects; everything else seems to be sent as strings, including 0/1 flags, ISO values, latitude/longitude, etc.). So, yes, they work fine as strings with the different layouts that we have.

On closer inspection, v3 sends `num` as an integer just like v4, so I may have been overzealous there :wink:. I converted it to a string because it was the last remaining non-string field in the whole response so, again, I figured it was better to be consistent. It doesn't seem to cause any problems.

Let me know if this response changed your minds. I have no problem reverting the non-essential changes, of course.

d7415:
Consistency with v3 is a good enough reason for me. If we ever release v4 we could revisit some day.

ildyria:
If this is consistent with v3, then it is fine. However, once the v4 is out and we drop the v3, I think it would be nice to double check everything. :)

PR #398, opened 2019-12-01
fix: add an existing table check when migration tables

YoshinoriN:
## Description

`pages` and `page_contents` tables migration are failed if I do `artisan migrate rollback` once.

## Solution

Added conditional operator to check these tables already exists or not before do migrate these tables.


ildyria:
Thank you. :)

PR #402, opened 2019-12-05
update php version of composer

YoshinoriN:
The Laravel 6.x requires PHP `7.2.0`.

> [Server Requirements](https://laravel.com/docs/6.x/installation#server-requirements)

IMHO should update the PHP version of the composer to `7.2.0`

Thanks :)


ildyria:
Indeed :laughing:

Issue #378, opened 2019-06-15
Live photo


Aymenel:
Сould you implement view live photo from Apple?

kamil4:
FYI, I moved this issue to the Lychee-Laravel repo so that it doesn't get lost...

Regarding the exiftool support, I just added some comments about it at https://github.com/LycheeOrg/Lychee-Laravel/issues/363#issuecomment-545507833.

tmp-hallenser:
Hi,

I’ve just pushed a first version supporting Live Photos (Apple Live Photos and Google Motion Photos) to my repository (branches support-live-photo)
* Backend: https://github.com/tmp-hallenser/Lychee-Laravel
* Frontend: https://github.com/tmp-hallenser/Lychee-front

When installing it, you need to run npm install && npm run compile in the frontend and composer install in the backend to update files.

Upload:
* Apple Live Photos: Upload jpg and mov after each other
* Google Motion Photos: Normal upload of jpg, video is part of jpg file

Notes:
* High resolution video might challenge the browser
* Package for viewing only supports mov files, hence the current version converts the mp4 file from a Google Motion Photo to mov while keeping codecs, resolution etc. (should be very fast)
* we need exiftool to extract the relevant exif information
* it’s just a first version with likely some room for improvements and bugfixes

File samples:
* Google Motion Photo:
1. https://github.com/joemck/ExtractMotionPhotos/issues/1#issuecomment-397476658 
2. https://photos.google.com/share/AF1QipO2_gTkgT1QwgYaWCTowaN6d2Cb5rvyJU10cjAdSU9Ao8v9Ec-r1v1cKdWEx6PNqg/photo/AF1QipP3IopzoNWtRutf8O0b0V_Tm83GIg_MdgnEWVMA?key=WEdYT3BMNFZGdUlwQ0l6aEdFT1UwVlg2LUZESDhn
* Apple Live Photo:
1. https://developer.apple.com/live-photos/assets/boy.jpg and https://developer.apple.com/live-photos/assets/boy.mov
2. https://developer.apple.com/live-photos/assets/girl.jpg and https://developer.apple.com/live-photos/assets/girl.mov

tmp-hallenser:
I'll close the issue since we have a first version of live photo implemented.

jscmidt:
How do I upload Live Photos? When I upload images via Safari on iPhone, only the JPEG part is uploaded…

PR #401, opened 2019-12-04
Support for exiftool

tmp-hallenser:
This PR adds support to use Exiftool to extract metadata from images and videos.

1. Adds a new config field 'has_exiftool'
* field initially set to null
* first call determines if exiftool is available 

2. Make use of library PHPExif/php-exif (https://github.com/PHPExif/php-exif/) as a wrapper for Exiftool, FFmpeg and native extraction
* the current latest official version does not support FFmpeg and has other limitations which are currently needed in Lychee
* I enhanced the library such that it's on par with todays exif features of Lychee and created an PR to the official repo
* Suggestion until PR is approved:
(A) Use my fork of the library
(B) Transfer my fork to LycheeOrg and use this one (suggested)

PR #404, opened 2019-12-08
Support live photo

tmp-hallenser:
This PR adresses #378.

**DB changes:**
* new col 'LivePhotoURL' storing the URL to the video
* new col 'LivePhotoContentID' storing the contentID to match photo and video for Apple Live Photos (contentID is the unique identifier)

**Logic:**
1. Apple Live Photo
- Sequential upload of video and photo
- Matching based on LivePhotoContentID (exiftool mandatory to extract this)
- Eventually only a single db entry

2. Google Motion Photo
- Upload of the JPG (which contains the video)
- extraction of the video from the JPG file & convert it into a mov container (no recoding, just switch of container)
- Eventually only a single db entry


d7415:
>     * extraction of the video from the JPG file & convert it into a mov container (no recoding, just switch of container)

Why mov? To keep it similar to the iOS version?

tmp-hallenser:
> > ```
> > * extraction of the video from the JPG file & convert it into a mov container (no recoding, just switch of container)
> > ```
> 
> 
> Why mov? To keep it similar to the iOS version?

The npm package for the frontend (livephotokit, https://www.npmjs.com/package/livephotoskit) only supports mov. I did quite some research, but it seems to be the only package which is still maintained and working on current devices. Of course, we could build our own version supporting, but the benefit would be imho minimal.

Since we're keeping the same video and audio codec, it's just a different container and the conversion should be very fast (+ typically max 3 seconds of video)

ildyria:
I will have a look at this PR once #401 is merged as the diff with master is quite big and redundant.

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -638,3 +755,34 @@ public function getValidExtensions(): array
 		return $this->validExtensions;
 	}
 }
+
+// Class for FFMpeg to convert files to mov format
+class MOVFormat extends FFMpeg\Format\Video\DefaultVideo
```
ildyria: Why is this class not in another file ? :confused:

PR #407, opened 2019-12-15
fix update_check_every_days value in DB from 'Config' to 'config'

YoshinoriN:
## Description

Current configuration definition value in DB (`configs.cat`) has two values `config` and `Config`.

In 'Full Settings' appear two 'Config' sections caused by this diff.

config-gui

## Fix

Update `Config` to `config` for `update_check_every_days`.

after2

Thank you.


ildyria:
Yeah I needed to fix that one at some point. ^^

Issue #305, opened 2019-07-13
Download button does not work with Android Chrome


Vorlif:
### Detailed description of the problem
If you try to download a picture with the `Download` button under Android Chrome you will be redirected to the overview and nothing happens.

No problems with Firefox for Android. Therefore, probably more of a frontend problem?

### Steps to reproduce the issue
**Steps to reproduce the behavior:**
1. Open an album
2. Open an image
3. Open the *three-point-menu*
4. Press the `Download` button

### Output of the diagnostics
```text
Diagnostics
-----------
Warning: Dropbox import not working. No property for dropboxKey.

System Information
------------------
Lychee-front Version:      3.2.16
Lychee Version (git):      7ccd77f (master) - Up to date.
DB Version:                040000
System:                    Linux
PHP Version:               7.3
PostgreSQL Version:        PostgreSQL 11.4 (Debian 11.4-1.pgdg90+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516, 64-bit

Lychee total space:        225.87 MB
Upload folder space:       169.20 MB
System total space:        55.27 GB
System free space:         48.26 GB (87%)

Imagick:                   1
Imagick Active:            1
Imagick Version:           1687
GD Version:                2.2.5

Config Information
------------------
version:                   040000
checkForUpdates:           1
layout:                    1
image_overlay:             1
image_overlay_type:        exif
full_photo:                1
Mod_Frame:                 0
Mod_Frame_refresh:         30000
landing_twitter:           
landing_instagram:         
landing_youtube:           
landing_page_enable:       0
compression_quality:       94
site_title:                Lychee v4
landing_owner:             Photographen
sortingPhotos_col:         takestamp
sortingPhotos_order:       ASC
sortingAlbums_col:         description
sortingAlbums_order:       DESC
imagick:                   1
skipDuplicates:            0
small_max_width:           0
small_max_height:          360
medium_max_width:          1920
medium_max_height:         1080
default_license:           none
deleteImported:            1
landing_subtitle:          Cat, Dogs & Humans Photography
landing_title:             Photographen
landing_background:        dist/cat.jpg
thumb_2x:                  1
small_2x:                  0
medium_2x:                 0
site_copyright_begin:      2019
site_copyright_end:        2019
additional_footer_text:    
display_social_in_gallery: 0
public_recent:             0
recent_age:                1
public_starred:            0
[...]

PR #408, opened 2019-12-15
support cache busting feature for JavaScript and CSS

YoshinoriN:
## Description

If `js` of `css` updated user has to clear their browser cache.
Or webserver owner has to configure their web server's cache settings just like #305 .

This PR adds UnixTimeStamp suffix to `js` and `css` just like `main.js?1576386855`

## Implementation

* Created `Assets\Helpers` class. And add `static function` it name is `cacheBusting`.
    * This function require `filePath`.
    * If file exists return `filePath` with UnixTimeStamp.
    * If file does not exist return `filePath` without change.
* Wrap `js` and `css` in `*.blade.php` with `cacheBusting` function.

## Question

* Is test needed? Which directory should I add UnitTest?

## Screenshot

JavaScript with UnixTimeStamp

jstimestamp

CSS with UnixTimeStamp

csstimestamp

### DIFF
--- a/app/Assets/Helpers.php
+++ b/app/Assets/Helpers.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace App\Assets;
+
+use Illuminate\Support\Facades\File;
+
+class Helpers
+{
+	/**
+	 * Add UnixTimeStamp to file path suffix.
+	 *
+	 * @param string $filePath
+	 *
+	 * @return string
+	 */
+	public static function cacheBusting(string $filePath): string
+	{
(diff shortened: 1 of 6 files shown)


ildyria:
> Is test needed? Which directory should I add UnitTest?

If possible but I'm not sure how to test that one...
Test are placed into the ... `tests/Feature` folder. Take example on the other test files to have an idea of how it works. It is actually pretty simple.

YoshinoriN:
@ildyria 
Thank you :)
I will try later.

Issue #413, opened 2019-12-24
EXIF tool addition breaks importing PNG and GIFs


alex-phillips:
### Detailed description of the problem
Unless I'm missing a package (I've installed php-exif and exiftool), I'm now getting a 500 error when importing PNG or GIF files. I've tried both with has_exiftool = 0 and = 1.

Error when `has_exiftool=0`:
```
(1/1) Error  Call to a member function getMimeType() on bool
--
in Extractor.php line 87
at Extractor->extract('/app/lychee/public/uploads/big/b38c37a22a3dea48ba0f556f1052c568.gif', 'image/gif')in PhotoFunctions.php line 331
at PhotoFunctions->add(array('name' => 'XMHK3LQ.gif', 'type' => 'image/gif', 'tmp_name' => '/tmp/phpneoMff'), '0')in PhotoController.php line 173
at PhotoController->add(object(Request))
at call_user_func_array(array(object(PhotoController), 'add'), array(object(Request)))in Controller.php line 54
at Controller->callAction('add', array(object(Request)))in ControllerDispatcher.php line 45
at ControllerDispatcher->dispatch(object(Route), object(PhotoController), 'add')in Route.php line 219
at Route->runController()in Route.php line 176
at Route->run()in Router.php line 681
at Router->Illuminate\Routing\{closure}(object(Request))in Pipeline.php line 130
at Pipeline->Illuminate\Pipeline\{closure}(object(Request))in UploadCheck.php line 54
at UploadCheck->handle(object(Request), object(Closure))in Pipeline.php line 171
at Pipeline->Illuminate\Pipeline\{closure}(object(Request))in SubstituteBindings.php line 41
at SubstituteBindings->handle(object(Request), object(Closure))in Pipeline.php line 171
at Pipeline->Illuminate\Pipeline\{closure}(object(Request))in VerifyCsrfToken.php line 76
at VerifyCsrfToken->handle(object(Request), object(Closure))in VerifyCsrfToken.php line 60
at VerifyCsrfToken->handle(object(Request), object(Closure))in Pipeline.php line 171
at Pipeline->Illuminate\Pipeline\{closure}(object(Request))in AuthenticateSession.php line 39
at AuthenticateSession->handle(object(Request), object(Closure))in Pipeline.php line 171
at Pipeline->Illuminate\Pipeline\{closure}(object(Request))in ShareErrorsFromSession.php line 49
[...]

PR #414, opened 2019-12-24
Fix issue 413

tmp-hallenser:
Fix for issue #413

### DIFF
--- a/app/Metadata/Extractor.php
+++ b/app/Metadata/Extractor.php
@@ -3,6 +3,7 @@
 namespace App\Metadata;
 
 use App\Configs;
+use App\Logs;
 use PHPExif\Reader\Reader;
 
 class Extractor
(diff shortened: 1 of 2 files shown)


app/Metadata/Extractor.php
```diff
@@ -105,6 +105,17 @@ public function extract(string $filename, string $type): array
 		$metadata['livePhotoContentID'] = ($exif->getContentIdentifier() !== false) ? $exif->getContentIdentifier() : null;
 		$metadata['MicroVideoOffset'] = ($exif->getMicroVideoOffset() !== false) ? $exif->getMicroVideoOffset() : null;
 
+		// We need to make sure, takestamp is between '1970-01-01 00:00:01' UTC to '2038-01-19 03:14:07' UTC.
+		// We set value to null in case we're out of bounds
+		if ($metadata['takestamp'] !== null) {
+			$min_date = new \DateTime('1970-01-01 00:00:01', new \DateTimeZone('UTC'));
+			$max_date = new \DateTime('2038-01-19 03:14:07', new \DateTimeZone('UTC'));
+			$takestamp = new \DateTime($metadata['takestamp']);
+			if ($takestamp < $min_date || $takestamp > $max_date) {
```
ildyria: I would add a Log line for 2038. Even though I doubt we will still be using that piece of software in 19 years from now, better be safe. ^^

Issue #409, opened 2019-12-16
has_exiftool not updating in DB


psjbeisler:
### Steps to reproduce the issue

Clean checkout of #401 ;
With postgreqsl database ;
`artisan migrate` Nothing to migrate. ;
`Error: Wrong property for has_exiftool in database, expected 0 or 1, got `
`has_exiftool:              ` (blank)

### Output of the diagnostics
    System Information
    ------------------
    Lychee-front Version:      3.2.16
    Lychee Version (git):      c08d927 (master) - Data not in Cache
    DB Version:                040000
    System:                    Linux
    PHP Version:               7.4
    PostgreSQL Version:        PostgreSQL 11.2 on x86_64-pc-linux-musl, compiled by gcc (Alpine 8.3.0) 8.3.0, 64-bit
    
    Lychee total space:        1.90 GB
    Upload folder space:       1.84 GB
    System total space:        48.99 GB
    System free space:         28.89 GB (58%)
    
    Imagick:                   1
    Imagick Active:            1
    Imagick Version:           1690
    GD Version:                bundled (2.1.0 compatible)

### Browser and system
Firefox / Linux

tmp-hallenser:
What does `php artisan migrate:status` say? Did you try to initiate the migration manually?

avalx:
The same happened on my installation. I ran the migration manually and everything looks OK:

`# php artisan migrate
PHP Notice:  date_default_timezone_set(): Timezone ID 'CEST' is invalid in /var/www/html/Lychee/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/LoadConfiguration.php on line 49
Migrating: 2019_12_02_2100_config_exiftool
Migrated:  2019_12_02_2100_config_exiftool (0.01 seconds)
Migrating: 2019_12_15_1000_config_check_update_every_cat_fix
Migrated:  2019_12_15_1000_config_check_update_every_cat_fix (0.01 seconds)
`
And status is also fine:

`| Yes  | 2019_12_02_2100_config_exiftool                   | 4     |
| Yes  | 2019_12_15_1000_config_check_update_every_cat_fix | 4     |
+------+---------------------------------------------------+-------+
`
So I changed the value of `has_exiftool` in the database to 1 (is this OK?) and now the error has gone away.

psjbeisler:
```
| Yes  | 2019_10_07_0900_config_map_include_sub_albums     | 3     |
[...]

PR #415, opened 2019-12-25
Updated type of has_exiftool to ternary

tmp-hallenser:
#409 has been fixed by #411 by whitelisting sanity checks for getting config values (setting still caused an error). This PR now removed again the whitelisting and introduces a proper solution changing type of has_exiftool to ternary (0,1,2).

The config variable has_exiftool was implemented as a bool and null was used in cases, test for exiftool has not yet been performed. This causes issues during sanity checks (getting config values and setting config values).

Solution: Move to ternary with the following values:
0 = Exiftool not available
1 = Exiftool is available
2 = Not yet tested

### DIFF
--- a/database/migrations/2019_12_25_0600_config_exiftool_ternary.php
+++ b/database/migrations/2019_12_25_0600_config_exiftool_ternary.php
@@ -21,11 +21,26 @@ class ConfigExiftoolTernary extends Migration
 			define('TERNARY', '0|1|2');
 		}
 
+		// Let's run the check for exiftool right here
+		$status = 0;
+		$output = '';
+		$has_exiftool = 2; // not set
+		try {
+			exec('which exiftool 2>&1 > /dev/null', $output, $status);
+			if ($status != 0) {
+				$has_exiftool = 0; // false
+			} else {
+				$has_exiftool = 1; // true
+			}
+		} catch (\Exception $e) {
+			// let's do nothing
+		}
+
 		if (Schema::hasTable('configs')) {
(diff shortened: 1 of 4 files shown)


database/migrations/2019_12_25_0600_config_exiftool_ternary.php
```diff
@@ -0,0 +1,55 @@
...
+	{
+		if (!defined('BOOL')) {
+			define('BOOL', '0|1');
+		}
+		if (!defined('TERNARY')) {
+			define('TERNARY', '0|1|2');
+		}
+
+		if (Schema::hasTable('configs')) {
+			Configs::where('key', '=', 'has_exiftool')
```
ildyria: I would also add something in the like:
```php
->whereNull('value')
```

but I think in the current state, this is also fine.

ildyria: One could ask why we don't do the 
```php
			exec('which exiftool 2>&1 > /dev/null', $output, $status);
```
check in the migration too. :)

tmp-hallenser: > I would also add something in the like:
> 
> ```
> ->whereNull('value')
> ```
> 
> but I think in the current state, this is also fine.

We're moving from bool to ternary -> imho, we should always overwrite to be on the safe side

tmp-hallenser:
Added two more fixes.
@ildyria please have a second look and merge if you're okay.

PR #424, opened 2020-01-04
Improve video handling in case of no ffmpeg

tmp-hallenser:
Video upload failed in case ffmpeg was not installed.

* Added config field has_ffmpeg (0 = no ffmpeg; 1 = ffmpeg available; 2 = not tested)
* Metadata extraction first trys ffmpeg, then exiftool and uses php native tools as fallback (which only extracts mime type and filesize)

### DIFF
--- a/app/Configs.php
+++ b/app/Configs.php
@@ -278,6 +278,40 @@ class Configs extends Model
 		return $has_exiftool;
 	}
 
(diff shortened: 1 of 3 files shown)


d7415:
Looks like it failed on style.

PR #427, opened 2020-01-06
Sync from CLI

tmp-hallenser:
This PR adds a CLI command to sync a folder (similar to Lychee-Sync)

ImportController is adjusted to be callable from CLI.

Support for some for video formats is added.

### DIFF
--- a/app/Console/Commands/sync.php
+++ b/app/Console/Commands/sync.php
@@ -1,0 +1,92 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Http\Controllers\ImportController;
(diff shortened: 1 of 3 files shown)


ildyria:
Can you check how this compares with #237 ? :)

tmp-hallenser:
> Can you check how this compares with #237 ? :)

It's similar. #237 imports photos using the Lychee settings (e.g. for delete imported files, duplicates etc). This one is meant for syncing, i.e. never deletes imported files and always skips duplicates.

One could keep import and sync separate, but could also merge them.

PR #434, opened 2020-02-08
removing php7.2 support, moving to bionic (18.4) instead of xenial (16.4)'

ildyria:
https://www.php.net/supported-versions.php

7.2 is end of life, Security fixes only. It will be completely unsupported by 30 Nov 2020

By switching from Xenial to Bionic, I hope to get some speed on the already installed php versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,13 @@
 os:
   - linux
 
-dist: xenial
+dist: bionic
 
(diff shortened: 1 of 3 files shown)


d7415:
No arguments. May be worth updating README (Still says 7.1) and checking the wiki.

PR #445, opened 2020-02-24
Fix photo isDuplicate detection

pchampio:
fixes https://github.com/LycheeOrg/Lychee-Laravel/issues/444

I'm not a PHP developer, If you wish to add unit test for this, you can edit this PR.

### DIFF
--- a/app/Photo.php
+++ b/app/Photo.php
@@ -179,11 +179,12 @@ class Photo extends Model
 	 */
 	public function isDuplicate(string $checksum, $photoID = null)
 	{
-		$sql = $this->where('checksum', '=', $checksum)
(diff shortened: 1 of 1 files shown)


ildyria:
Can you test with this request ? This would make the code more readable with respect to the `OR`/`AND` priority.
```php
		$sql = $sql->where(function ($q)  use ($checksum)  {
			$q->where('checksum', '=', $checksum)
				->orWhere('livePhotoChecksum', '=', $checksum) });
		if (isset($photoID)) {
			$sql = $sql->where('id', '<>', $photoID);
		}
```

pchampio:
Yes, It works:
select * from "photos" where ("checksum" = ? or "livePhotoChecksum" = ?) and "id" <> ?

PR #449, opened 2020-03-05
Fix installer

ildyria:
Previous installer required the use of `exec`, we solve this problem and allow the user to directly apply migrations.

This PR should bring the User Experience of the installation very close to what Lychee v3 was:
download the zip, extract and set up.


app/ControllerFunctions/Install/RequirementsChecker.php
```diff
@@ -28,21 +28,28 @@ public function check(array $requirements)
...
 							$results['errors'] = true;
+							// @codeCoverageIgnoreEnd
 						}
 					}
 
 					if ($this->checkExec()) {
 						$results['requirements'][$type]['Php exec() available'] = true;
+					} else {
+						// @codeCoverageIgnoreStart
+						$results['requirements'][$type]['Php exec() not available (optional)'] = false;
```
d7415: This is familiar :P But I'm confused that they're different strings/indexes *and* different values? i.e. `{ yes = true; } else { no = false; }`

app/Http/Middleware/InstalledCheck.php
```diff
@@ -0,0 +1,47 @@
...
+		}
+
+		// base safety
+		if (file_exists(base_path('installed.log'))) {
+			return ToHome::go();
+		}
+
+		// This is the second safety:
+		// Assume you do a "git pull" but forget to do the migration,
+		// the installed.log will not be created!!!
```
d7415: :+1:

app/ControllerFunctions/Install/RequirementsChecker.php
d7415: Ah. Because it just lists through and shows `true` as a tick and `false` as a cross. I remember.

ildyria:
> As usual, untested but looks good.

https://www.youtube.com/watch?v=rR4n-0KYeKQ

:laughing:

Issue #451, opened 2020-03-06
Unable to import some JPGs


gardiol:
### Detailed description of the problem

Try to import attached file fails. Only error displayed is the classic red error line on top of the page (Loading failed unknown error...). Nothing appears in the logs too!

I have a lot of these files. They seems to be perfectly fine opening with any image program. Also, they come from the same camera, one before and one after imports correctly, not this one (and a few others too.. but i am going to upload just a couple)

### Steps to reproduce the issue

Import attached file

### Output of the diagnostics (Settings => Diagnostics)

Diagnostics
    -----------
    Info: Latest version of PHP is 7.4
    Warning: Dropbox import not working. dropbox_key is empty.
    Warning: Pictures that are rotated lose their metadata! Please install Imagick to avoid that.

    System Information
    ------------------
    Lychee-front Version:      3.2.16
    Lychee Version (git):      No git data found. Probably installed from release or could not read .git
    DB Version:                040000
    System:                    Linux
    PHP Version:               7.2
    MySQL Version:             10.2.29-MariaDB-log
    
    Lychee total space:        70.40 MB
    Upload folder space:       43.00 B
    System total space:        411.20 GB
    System free space:         52.69 GB (12%)
    
    Imagick:                   -
    Imagick Active:            1
    Imagick Version:           -
    GD Version:                bundled (2.1.0 compatible)

    Config Information
    ------------------
    version:                   040000
    check_for_updates:         0
    sorting_Photos_col:        title
    sorting_Photos_order:      ASC
    sorting_Albums_col:        title
    sorting_Albums_order:      DESC
    imagick:                   1
    skip_duplicates:           0
    small_max_width:           0
    small_max_height:          360
    medium_max_width:          1920
    medium_max_height:         1080
    lang:                      it
    layout:                    1
    image_overlay:             1
    image_overlay_type:        desc
    default_license:           reserved
    compression_quality:       90
    full_photo:                1
    delete_imported:           0
    Mod_Frame:                 1
    Mod_Frame_refresh:         30
    thumb_2x:                  1
    small_2x:                  1
    medium_2x:                 1
    landing_page_enable:       1
    landing_owner:             Willy Gardiol
    landing_title:             Willy Gardiol Photos
    landing_subtitle:          Foto di Willy Gardiol
  landing_facebook:          
    landing_flickr:            
    landing_twitter:           
    landing_instagram:         
    landing_youtube:           
    landing_background:        dist/cat.jpg
[...]

PR #457, opened 2020-03-15
Decode GPS data into location name

tmp-hallenser:
This PR decodes GPS data (latitude and longitude) into location name, i.e. street, city, country etc.

It works as follows:
* table 'photos' gets a new column 'location'
* during upload, latitude and longitude is decoded into street, city, country etc. using OpenStreetmap (can be deactivated in settings, default is not enabled)
* OpenStreetmap only allows 1 request per second, hence we limit request rate is limited to 1 request per second
* Requests are cached for speedup and reduce call to OpenStreetmap (Default: using filesystem as cache, can be deactivated or set to in-memory; the latter only works by invoking decoding through separate command - see below - since memory is cleared after every photo)
* Decoded location is shown below map, but can be deactivated for all or just for public access (needs separate PR for frontend)
* Location field is added to columns used by search function

Using `php artisan lychee:decode_GPS_locations`, a decoding of existing GPS data can be triggered. In-memory caching is supported in this case.


ildyria:
Can you explain your choices for the cache system? I don't really understand that part.
Laravel comes already with a cache system: https://laravel.com/docs/7.x/cache and to me it feels you are adding another system on top of it (or am I missing something?).

Also why do you need `uploads/cache` to be publically available? 🤔

tmp-hallenser:
> Can you explain your choices for the cache system? I don't really understand that part.
> Laravel comes already with a cache system: https://laravel.com/docs/7.x/cache and to me it feels you are adding another system on top of it (or am I missing something?).

Happy to use the provided cache, but I didn't found a way to do it. We have the following requirements:
1. We need a http adapter to work
2. It should support caching to reduce impact of 3.
3. It must support request throttling to limit calls to Openstreetmap to 1 call per second


Please correct me on the following thoughts
re 1. Larval does not bring a native http client, but recommends `guzzlehttp/guzzle` (https://laravel.com/docs/7.x/http-client). I used `php-http/guzzle6-adapter` as a http adapter for the afore mentioned. 
re 2. the Geodecoder already offers a provider for caching (https://github.com/geocoder-php/cache-provider), but it need an actual caching implementation below. The caching implementation needs to be a PSR-6 compatible cache driver - an this is where I didn't find a solution. 

There is `madewithlove/illuminate-psr-cache-bridge`, but working only with Larval 5. I did not succeed including the larval cache - happy to get pointed into the right direction. 


re 3. We now need some throttling, which can be done by using `spatie/guzzle-rate-limiter-middleware` (used) or `hamburgscleanest/guzzle-advanced-throttle`. 
> Also why do you need `uploads/cache` to be publically available? 🤔

No, it does not need to be public (sorry for the perhaps stupid question - how do I make it non-public?)

ildyria:
@tmp-hallenser I cleaned the cache part. :) It should now directly use the one provided by Laravel. :)

Can you double check I did not break anything?

By the way, why do we need that? 
```
"nyholm/psr7": "^1.2",
```

tmp-hallenser:
> @tmp-hallenser I cleaned the cache part. :) It should now directly use the one provided by Laravel. :)
> 
> Can you double check I did not break anything?
> 
> By the way, why do we need that?
> 
> ```
> "nyholm/psr7": "^1.2",
> ```

Good point - not needed since we're using guzzle.

tmp-hallenser:
First of all - many thanks for your push!

Unfortunately, it does not work for me.

I needed to install `php-http/message` to fix the first issue. Now, I'm stuck with the following error:
```
   TypeError 

  Argument 2 passed to Geocoder\Provider\Cache\ProviderCache::__construct() must implement interface Psr\SimpleCache\CacheInterface, instance of Symfony\Component\Cache\Adapter\Psr16Adapter given, called in /html/Lychee-Laravel/app/Metadata/Geodecoder.php on line 45

  at vendor/geocoder-php/cache-provider/ProviderCache.php:48
    44|      * @param Provider       $realProvider
    45|      * @param CacheInterface $cache
    46|      * @param int            $lifetime
    47|      */
  > 48|     final public function __construct(Provider $realProvider, CacheInterface $cache, int $lifetime = null)
    49|     {
    50|         $this->realProvider = $realProvider;
    51|         $this->cache = $cache;
    52|         $this->lifetime = $lifetime;

  1   app/Metadata/Geodecoder.php:45
      Geocoder\Provider\Cache\ProviderCache::__construct(Object(Geocoder\Provider\Nominatim\Nominatim), Object(Symfony\Component\Cache\Adapter\Psr16Adapter))

  2   app/Console/Commands/decode_GPS_locations.php:69
      App\Metadata\Geodecoder::getGeocoderProvider()

```

ildyria:
Can you give me a picture for which has data to decode so I can try to fix that ? :)

tmp-hallenser:
There are two ways:

1. Just upload an image (w/o gps data) and insert some random (more or less, e.g. 2.294481 and 48.858370) coordinates into the database, e.g. using phpmyadmin, and call `php artisan lychee:decode_GPS_locations` afterwards.

2. Upload images with gps data, e.g. use the ones provided in #451 (they both should contain GPS data); Or use `php artisan lychee:sync /folder/to/sync`

Thanks for you effort!

ildyria:
@tmp-hallenser and now ? :smiley:

ildyria:
> I needed to install php-http/message to fix the first issue. Now, I'm stuck with the following error:

weird.

tmp-hallenser:
> php-http/message

still need it

public/dist/view.js
```diff
@@ -2156,6 +2188,7 @@ lychee.locale = {
 	'SET_LICENSE': 'Set License',
 	'SET_OVERLAY_TYPE': 'Set Overlay',
 	'SET_MAP_PROVIDER': 'Set OpenStreetMap tiles provider',
+	'SET_LOCATION_DECODING_CACHING_TYPE': 'Set location decoder caching',
```
d7415, line 2191: See below...

app/Metadata/Geodecoder.php
```diff
@@ -0,0 +1,97 @@
...
+
+		$httpClient = new \GuzzleHttp\Client([
+			'handler' => $stack,
+			'timeout' => Configs::get_value('location_decoding_timeout'),
+		]);
+
+		$httpAdapter = new \Http\Adapter\Guzzle6\Client($httpClient);
+
+		$cachedProvider = null;
+		// $caching_type = Configs::get_value('location_decoding_caching_type');
```
d7415, line 38: I can't see this config, but there does seem to be language. Is this a `TODO` line?

app/Metadata/Geodecoder.php
```diff
@@ -0,0 +1,97 @@
...
+		$stack->push(RateLimiterMiddleware::perSecond(1));
+
+		$httpClient = new \GuzzleHttp\Client([
+			'handler' => $stack,
+			'timeout' => Configs::get_value('location_decoding_timeout'),
+		]);
+
+		$httpAdapter = new \Http\Adapter\Guzzle6\Client($httpClient);
+
+		$cachedProvider = null;
```
d7415, line 37: Unused.

app/Metadata/Geodecoder.php
```diff
@@ -0,0 +1,97 @@
...
+		// $caching_type = Configs::get_value('location_decoding_caching_type');
+
+		$provider = new Nominatim($httpAdapter, 'https://nominatim.openstreetmap.org', config('app.name'));
+
+		return new ProviderCache($provider, app('cache.store'));
+	}
+
+	/**
+	 * Decode GPS coordinates into location.
```
d7415, line 46: Should this have a slightly different description? e.g. "Wrapper for...."?

config/cache.php
```diff
@@ -12,7 +12,9 @@
 	| using this caching library. This connection is used when another is
 	| not explicitly specified when executing a given caching function.
 	|
-	| Supported: "apc", "array", "database", "file", "memcached", "redis"
+
+	| Supported: "apc", "array", "database", "file",
+	|			"memcached", "redis", "dynamodb"
```
d7415, line 17: Not a big issue, but this line wrapping seems weird.

app/Metadata/Geodecoder.php
ildyria: fixed.

app/Metadata/Geodecoder.php
ildyria: actually that was in the first implementation when there was multiple cache set-up. Not the case anymore.

public/dist/view.js
ildyria: See below...

app/Metadata/Geodecoder.php
ildyria: fixed.

config/cache.php
ildyria: that is actually the file from the lastest Laravel release. :)

PR #458, opened 2020-03-17
Unhide SL options and raw_formats in Diagnostics.

d7415:
I'm not clear on why the SL options were hidden. If nothing else they're very useful for diagnosing issues here. I'm similarly unsure why `raw_formats` was hidden, but I feel less strongly about it being shown.

For once I have actually tested this.

### DIFF
--- a/database/migrations/2020_03_17_200000_unhide_configs.php
+++ b/database/migrations/2020_03_17_200000_unhide_configs.php
@@ -1,0 +1,40 @@
+<?php
+
+/** @noinspection PhpUndefinedClassInspection */
+use App\Configs;
+use Illuminate\Database\Migrations\Migration;
+
+class UnhideConfigs extends Migration
+{
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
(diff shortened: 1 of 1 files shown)


ildyria:
I disagree strongly for the lifetime of the links and for the raw format.

Lifetime : you reveal how often you refresh those links. Not cool, in a context of hot linking etc.
Raw format: you reveal which files extension can be uploaded without processing from the server.
This may lead to someone adding `.php` in their list and allow the upload of such script. I'm not found of having this information visible, even in a debug part.

d7415:
re: lifetime sure. It's not that useful debugging and I can see some logic there (though limited)
re: `raw_format` I guess. Again not major for debugging - I just saw it and figured it wasn't that sensitive. The php example implies much bigger problems, and as a plus, might give us the opportunity to explain their error ;) But I'll remove it.

d7415:
@ildyria Are you ok with this version?

d7415:
That was quick! Thanks.

PR #490, opened 2020-04-20
lots of refactoring

### LINKED ISSUE #485: [Enhancement] Display version
If one go to Diagnostic currently, this is what we see:
```
    System Information
    --------------
    Lychee-front Version:      4.0.0
    Lychee Version (git):      49d5862 (master) - Up to date (3 minutes ago).
    DB Version:                040001
    System:                    Linux
    PHP Version:               7.4
    MySQL Version:             10.3.22-MariaDB-1
    
    Imagick:                   1
    Imagick Active:            1
    Imagick Version:           1690
    GD Version:                2.2.5
```

The problem is when doing the update from the GUI the Lychee-front version is not correct anymore (even if up to date) because we do not update the sub module.

The second problem is in the case of the release, the Lychee-front version is not available.  
We should hide this (currently it is asking to do `git submodule init ; git submodule update`).

The third problem is in the case of the release, we should make the distincting between the version used as released (get the data from `version.md`) and the version used when using `git`.

ildyria:
fixes #485

Moving things around, _Slimify_ controllers.

Codecov will decrease for now, I will fix the part which are not covered.


app/ControllerFunctions/Update/Check.php
```diff
@@ -0,0 +1,128 @@
...
+		} else {
+			return 'Already up to date';
+		}
+	}
+
+	/**
+	 * Check for updates, returns the code
+	 * 0 - Not Master
+	 * 1 - Not in cache
+	 * 1 - Up to date
```
d7415, line 107: I'm guessing we don't care which of these has happened?

app/Exceptions/GitNotAvailable.php
```diff
@@ -0,0 +1,16 @@
...
+use Exception;
+use Throwable;
+
+class GitNotAvailable extends Exception
+{
+	public function __construct(
+		$code = 0,
+		Throwable $previous = null
+	) {
+		parent::__construct('../.git (and subdirectories) are not executable, check the permissions.', $code, $previous);
```
d7415, line 14: Should this read `(and/or its subdirectories)`? I think it gets the message across either way though.

app/Http/Controllers/DiagnosticsController.php
```diff
@@ -43,28 +42,44 @@ class DiagnosticsController extends Controller
...
+		$this->lycheeVersion = $lycheeVersion;
 		$this->sessionFunctions = $sessionFunctions;
 		$this->diskUsage = $diskUsage;
+		$this->checkUpdate = $checkUpdate;
+
+		$this->versions = $this->lycheeVersion->get();
+	}
+
+	/**
+	 * Return the padded string to $pad_length.
```
d7415: `$pad_length` looks like 27?

app/Metadata/GitHubFunctions.php
```diff
@@ -207,25 +207,6 @@ public function get_behind_text()
...
-	 * @return string
-	 */
-	public function get_info()
-	{
-		$branch = $this->get_current_branch();
-		$head = $this->get_current_commit();
-		if ($head == false || $branch == false) {
-			// when going through CI, .git exists...
-			// @codeCoverageIgnoreStart
-			return 'No git data found. Probably installed from release or could not read .git';
```
d7415, line 222: Note mostly for myself - we don't need this because we have access to a release version now.

public/dist/version.md
```diff
@@ -0,0 +1 @@
+{"version":"4.0.1","commit":"7c9762c"}
```
d7415: `*.md` normally indicates markdown rather than json...

d7415 (approved):
I haven't tested it live, and haven't followed every path, but it looks sane. A few minor comments/queries.

app/ControllerFunctions/Update/Check.php
ildyria: Yes because :
1- will show "Check for update"
2- will show "Apply update.

app/Http/Controllers/DiagnosticsController.php
ildyria: hahaha left over from previous code ;)

public/dist/version.md
ildyria: :'D

PR #499, opened 2020-04-23
remove Lychee-front version number, add more Diagnostic informations

### LINKED ISSUE #498: Do we need Lychee-front version ?
Currently we are giving three version numbers:

- Lychee-front
- Lychee files
- Lychee DB

Do we really need Lychee-front as this is "packaged" with the Lychee files number?

ildyria:
- fixes #498 
- add more debug information:
  - which composer install was used ?
  - what is the value of `APP_ENV` (eg `production`) ? 
  - what is the value of `APP_DEBUG` (`true` or `false`) ?


app/Http/Controllers/DiagnosticsController.php
```diff
@@ -189,9 +190,13 @@ public function get_errors()
 			}
 		}
 
-		$version_num = implode('.', array_map('intval', str_split($settings['version'], 2)));
-		if ($this->lycheeVersion->isRelease && $version_num < $this->versions['Lychee']) {
-			$errors[] = 'Error: Database is behind file versions. Please apply the migration.';
+		if (isset($settings['version'])) {
+			$version_num = implode('.', array_map('intval', str_split($settings['version'], 2)));
+			if ($this->lycheeVersion->isRelease && $version_num < $this->versions['Lychee']) {
+				$errors[] = 'Error: Database is behind file versions. Please apply the migration.';
```
d7415: Existing wording, but I wonder if it's worth mentioning `php artisan migrate` here?

I've also wondered whether it would be worth having the migration part of the updater available in release versions...

app/Http/Controllers/DiagnosticsController.php
```diff
@@ -189,9 +190,13 @@ public function get_errors()
...
+			}
+		} else {
+			$errors[] = 'Error: Version number not found in database. Please apply the migrations.';
```
d7415: As above.

app/Http/Controllers/DiagnosticsController.php
ildyria: I was also considering adding something to apply migrations from the web.
Especially for those with only FTP access.

app/Metadata/LycheeVersion.php
```diff
@@ -29,6 +43,15 @@ private function fetchReleaseInfo()
 		return !file_exists(base_path('.git'));
 	}
 
+	/**
+	 * Returns true if we are using the release channel
+	 * Returns false if we are using the git channel.
+	 */
+	private function fetchComposerInfo()
+	{
+		return file_exists(base_path('vendor/bin/phpunit'));
```
d7415, line 54: Could this just be on line 34? Does it need its own function?

ildyria: It is just for caching data. :x Sure could be refactored. Not a strong opinion on that one.

d7415: So long as there's a reason, I'm happy.

ildyria:
Finally !

app/ControllerFunctions/Update/Check.php
```diff
@@ -106,23 +132,38 @@ public function getText()
...
 	public function getCode()
 	{
-		$update = $this->canUpdateBool();
+		if ($this->lycheeVersion->isRelease) {
+			// @codeCoverageIgnoreStart
+			$versions = $this->lycheeVersion->get();
+
+			return 3 * intval($versions['DB']['version'] < $versions['Lychee']['version']);
+		// @codeCoverageIgnoreEnd
+		} else {
```
d7415: It's not a big deal, but we shouldn't need an `else` if there's a return. I'm only mentioning it because there are a lot of layers here.

app/Http/Controllers/DiagnosticsController.php
```diff
@@ -366,12 +364,11 @@ public function get_config()
...
-	 * should be used for AJAX request.
+	 * Return the requested information.
 	 *
 	 * @return array
 	 */
-	public function get()
+	private function get_data()
 	{
 		$errors = $this->get_errors();
 		$infos = ['You must be logged to see this.'];
```
d7415: Not really in scope for the PR, but it feels like an `else` would be cleaner here?

d7415 (approved):
Only comments I had were style choices. Untested, as usual. Looks sane!

PR #512, opened 2020-04-28
Fix a few issues.

ildyria:


ildyria:
This PR will also allow `php artisan lychee:Diagnostics` to work even without database connection.

app/Configs.php
```diff
@@ -259,13 +259,19 @@ public static function hasExiftool()
...
+					self::set('has_exiftool', 1);
+					$has_exiftool = true;
+				}
+			} catch (Exception $e) {
 				self::set('has_exiftool', 0);
 				$has_exiftool = false;
-			} else {
-				self::set('has_exiftool', 1);
-				$has_exiftool = true;
+				Logs::warning(__METHOD__, __LINE__, 'exec is disabled, set has_exiftool to 0.');
```
d7415: Maybe `has_exiftool set to 0` or `has_exiftool has been set to 0` to clarify that this isn't an instruction.

app/Configs.php
```diff
@@ -290,13 +296,19 @@ public static function hasFFmpeg()
...
+					self::set('has_ffmpeg', 1);
+					$has_ffmpeg = true;
+				}
+			} catch (Exception $e) {
 				self::set('has_ffmpeg', 0);
 				$has_ffmpeg = false;
-			} else {
-				self::set('has_ffmpeg', 1);
-				$has_ffmpeg = true;
+				Logs::warning(__METHOD__, __LINE__, 'exec is disabled, set_ffmpeg to 0.');
```
d7415: As above.

app/Http/Controllers/DiagnosticsController.php
```diff
@@ -134,8 +134,22 @@ public function get_errors()
...
+			['mysql', 'mysqli'],
+			['mysql', 'pdo_mysql'],
+			['pgsql', 'pgsql'],
+			['pgsql', 'pdo_pgsql'],
+			['sqlite', 'sqlite3'],
+		];
+		try {
+			foreach ($db_possibilities as $db_possibility) {
+				if (DB::getDriverName() == $db_possibility[0] && !extension_loaded($db_possibility[1])) {
+					$errors[] = 'Error: PHP ' . $db_possibility[1] . ' extension not activated';
```
d7415: For clarity, maybe this? Might need line wrapping.
```suggestion
					$errors[] = 'Error: ' . $db_possibility[0] . ' db driver selected and PHP ' . $db_possibility[1] . ' extension not activated';
```

d7415 (dismissed):
Some wording suggestions, but the logic looks sound. Good thinking!

app/Http/Controllers/DiagnosticsController.php
d7415: I don't think I'd looked at these properly before. Maybe I should use them more often for the easy-commit, especially as we squash-merge.

PR #516, opened 2020-04-29
improve codecov: use their interface

ildyria:
Minor, no need for review.
Will merge if passes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -95,7 +95,7 @@ script:
   # - php artisan migrate:rollback
 
 after_success:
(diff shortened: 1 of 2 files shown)


d7415:
If we're piping their script in, is there a need to update the version in the repo (rather than deleting it)?

ildyria:
> If we're piping their script in, is there a need to update the version in the repo (rather than deleting it)?

I don't understand your question.

d7415:
> > If we're piping their script in, is there a need to update the version in the repo (rather than deleting it)?
> 
> I don't understand your question.

Doesn't apply any more. in the first commit you had Travis fetching the codecov script with curl, but also updated our copy. Now it's just using our copy.

PR #532, opened 2020-05-03
Feature/orientation

alex-phillips:
Goes with matching PR in the Lychee-front repo.

If we fail to auto-rotate an image (read-only filesystem, example: when syncing with symlinks from a read-only mount), we need to store the orientation EXIF data so that the frontend can account for it. Otherwise images are displayed sideways.

### DIFF
--- a/app/ModelFunctions/PhotoFunctions.php
+++ b/app/ModelFunctions/PhotoFunctions.php
@@ -381,13 +381,6 @@ class PhotoFunctions
 						$photo->height = $rotation['height'];
 					} else {
 						$photo->orientation = $info['orientation'];
-
-						if ($photo->orientation >= 5) {
-							$tempWidth = $photo->width;
-							$tempHeight = $photo->height;
-							$photo->width = $tempHeight;
-							$photo->height = $tempWidth;
-						}
 					}
 				}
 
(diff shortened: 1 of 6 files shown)


app/ModelFunctions/PhotoFunctions.php
```diff
@@ -481,7 +484,7 @@ public function extractVideoFrame(Photo $photo): string
 		$frame = $video->frame(FFMpeg\Coordinate\TimeCode::fromSeconds($photo->aperture / 2));
 
 		$tmp = tempnam(sys_get_temp_dir(), 'lychee');
-		Logs::notice(__METHOD__, __LINE__, 'Saving frame to ' . $tmp);
+		Nnotice(__METHOD__, __LINE__, 'Saving frame to ' . $tmp);
```
ildyria: ???

ildyria (commented):
Looks pretty simple to me.
How does this interact with https://github.com/LycheeOrg/php-exif ? :)

app/ModelFunctions/PhotoFunctions.php
alex-phillips: Whoops, this was a mistake!

alex-phillips:
@ildyria PHPExif already has support for orientation, so I was just pulling the value out. I can supply images with orientation EXIF metadata set if you'd like to test.

ildyria:
@alex-phillips can you have a look at https://github.com/LycheeOrg/Lychee/pull/538 to see if this fixes your problem ?

ildyria:
Fixed with #538

PR #540, opened 2020-05-09
Origin/fix 539

### LINKED ISSUE #539: Invalid exif geolocation causes db error
### Error when upload photos with invalid geo info
During upload, I noticed some photos produced errors ("something went wrong, error22003). After some digging in the log files this turned out to be due to invalid geo info. In my case, for whatever reason, the longitude was set to -4367743116.3897 which resulted in: "error","App\\ModelFunctions\\PhotoFunctions::save","692","Something went wrong, error 22003, SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'longitude' at row 1

I realize that this is caused by an external problem, but in my opinion, it would be preferable to just disable the respective fields automatically instead of requiring the user to sift through photos individually to remove exif data that is broken anyway.

tmp-hallenser:
Fix #539 
* Add boundary check for latitude & longitude
* Add boundary check for altitude
* Add boundary check for imgDirection

### DIFF
--- a/app/Metadata/Extractor.php
+++ b/app/Metadata/Extractor.php
@@ -138,40 +138,39 @@ class Extractor
 				Logs::notice(__METHOD__, __LINE__, 'Takestamp (' . $takestamp->format('Y-m-d H:i:s') . ') out of bounds (needs to be between 1970-01-01 00:00:01 and 2038-01-19 03:14:07)');
 			}
 		}
-		
+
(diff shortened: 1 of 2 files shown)


csaxx:
Thanks for including these fixes so quickly. I was just looking over the changes and I think there might be a small error in the altitude check:

if ($altitude < -999999.9999 || $imgDirection > 999999.9999) {

ildyria (approved):
LGTM.

PR #541, opened 2020-05-10
create a default database for sqlite

ildyria:


d7415:
I know it's an empty file, but feels messy when so few users use sqlite. *shrug*

(I don't have a better solution)

ildyria:
It is to prevent issues such as "I switched to SQLite but I get an error DB not found" to be opened.
Just to prevent some user to shoot themselves in the foot.

PR #543, opened 2020-05-11
handle composer not installed

ildyria:


bootstrap/simple-errors.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+function displaySimpleError($tite, $code, $message)
+{
+	$error_display = file_get_contents(__DIR__ . '/../simple_error_template.html');
+	$replacing = array(
+		'$tite' => $title,
```
d7415: Typo?

ildyria: Damn Sonarcloud is pretty good too. :sweat_smile:

index.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+require __DIR__ . '/bootstrap/simple-errors.php';
+
+displaySimpleError('ROOT', 403, '<span class="important">This is the root directory and it MUST NOT BE PUBLICALLY ACCESSIBLE.</span><br>
```
d7415, line 5: Am I correct that the 403 is just for show? :(

ildyria: what do you mean ?

d7415: Does it return HTTP 403 somehow?

d7415: ( https://www.php.net/manual/en/function.http-response-code.php )

ildyria: done.

PR #551, opened 2020-05-12
Rss2

### LINKED ISSUE #39: Rss feed
Add a module that provide a rss feed of the "x" last pictures added (they need to be publicly available). :)

ildyria:
fixes #39

Add Rss feed to the public pictures uploaded over the last 7 days.

This is disabled by default.


app/Http/Controllers/RSSController.php
```diff
@@ -0,0 +1,83 @@
...
+	 */
+	public function getRSS()
+	{
+		if (Configs::get_value('rss_enable', '0') != '1') {
+			abort(404);
+		}
+
+		$photos = Photo::with('album', 'owner')
+		->where('created_at', '>=', Carbon::now()->subDays(intval(Configs::get_value('rss_recent_days', '7')))
+		->toDateTimeString())
```
d7415: I don't know if the style tests would get upset, but an extra indentation here would be nice.

database/migrations/2020_05_12_114228_rss.php
```diff
@@ -0,0 +1,46 @@
...
+	{
+		defined('BOOL') or define('BOOL', '0|1');
+		defined('INT') or define('INT', 'int');
+
+		DB::table('configs')->insert([
+			[
+				'key' => 'rss_enable',
+				'value' => '0',
+				'confidentiality' => 0,
+				'cat' => 'Mod RSS',
```
d7415, line 23: Why not `'RSS'`?

d7415 (changes requested):
Looks ok. A couple of comments. Nothing major.

database/migrations/2020_05_12_114228_rss.php
ildyria: Because all the additions in the config starts with "Mod" :stuck_out_tongue: 
Mod Frame, etc...

d7415: It's not the first time I've thought about it... It makes them sound like add-ons rather than features.

app/Http/Controllers/RSSController.php
```diff
@@ -0,0 +1,83 @@
...
+		}
+
+		$photos = Photo::with('album', 'owner')
+			->where('created_at', '>=', Carbon::now()->subDays(intval(Configs::get_value('rss_recent_days', '7')))
+			->toDateTimeString())
+			->where(function ($q) {
+				$q->whereIn('album_id',
+					$this->albumFunctions->getPublicAlbums())
+					->orWhere('public', '=', '1');
+			})->get();
```
d7415: It may be worth adding a limit here for when someone has added a lot of photos (or uses a big `rss_recent_days`).

PR #553, opened 2020-05-13
Improve hooks + install them at start up.

ildyria:
we directly copy the hooks in `.git/hooks`.

This will make sure that composer is run at every pull.
This will also set up php-cs-fixer to run on each php file before a commit.

### DIFF
--- a/post-merge
+++ b/post-merge
@@ -22,10 +22,10 @@ echo "\n${GREEN}post merge hook start${NO_COLOR}\n"
 PHPUNIT="vendor/bin/php-cs-fixer"
 
 if [ -x "$PHPUNIT" ]; then
-    echo "Dev mode detected"
+    echo "\n${ORANGE}Dev mode detected${NO_COLOR}\n"
     composer install --no-suggest
 else
-    echo "Dist mode detected"
+    echo "\n${ORANGE}--no-dev mode detected${NO_COLOR}\n"
     composer install --no-dev --prefer-dist --no-suggest
 fi
 php artisan migrate
--- a/post-merge
+++ b/post-merge
@@ -1,5 +1,21 @@
 #!/bin/sh
-echo "post merge hook start"
+NO_COLOR="\033[0m"
+RED="\033[38;5;009m"
+GREEN="\033[38;5;010m"
+YELLOW="\033[38;5;011m"
+ORANGE="\033[38;5;214m"
+LIGHTPURPLE="\033[38;5;177m"
+PURPLE="\033[38;5;135m"
+CYAN="\033[38;5;014m"
+LIGHTBLUE="\033[38;5;39m"
+BLUE="\033[38;5;75m"
+DARKBLUE="\033[38;5;33m"
+LIGHTGRAY="\033[38;5;252m"
+DARKGRAY="\033[38;5;242m"
+BRIGHTRED="\033[91m"
+BOLD="\033[1m"
+
+echo "${GREEN}post merge hook start${NO_COLOR}"
 
 # we check if vendor/bin/phpunit is present.
 # If yes this means we are in dev mode.
@@ -15,4 +31,4 @@ fi
 php artisan migrate
 
 
-echo "post merge hook start"
+echo "${RED}post merge hook finish${NO_COLOR}"
--- a/post-merge
+++ b/post-merge
@@ -1,19 +1,7 @@
 #!/bin/sh
 NO_COLOR="\033[0m"
-RED="\033[38;5;009m"
 GREEN="\033[38;5;010m"
-YELLOW="\033[38;5;011m"
 ORANGE="\033[38;5;214m"
-LIGHTPURPLE="\033[38;5;177m"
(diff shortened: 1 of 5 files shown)


post-merge
```diff
@@ -1,3 +1,34 @@
...
+LIGHTPURPLE="\033[38;5;177m"
+PURPLE="\033[38;5;135m"
+CYAN="\033[38;5;014m"
+LIGHTBLUE="\033[38;5;39m"
+BLUE="\033[38;5;75m"
+DARKBLUE="\033[38;5;33m"
+LIGHTGRAY="\033[38;5;252m"
+DARKGRAY="\033[38;5;242m"
+BRIGHTRED="\033[91m"
+BOLD="\033[1m"
```
d7415: I don't think we should be cluttering the scripts with things we're not using.

post-merge
```diff
@@ -1,3 +1,34 @@
...
+
+echo "\n${GREEN}post merge hook start${NO_COLOR}\n"
```
d7415, line 6: We could add a test for an untracked file e.g. `NO_AUTO_COMPOSER` here to disable this functionality and report "skipped" instead of "start".

post-merge
```diff
@@ -1,3 +1,34 @@
...
+    echo "\n${ORANGE}Dev mode detected${NO_COLOR}\n"
+    composer install --no-suggest
+else
+    echo "\n${ORANGE}--no-dev mode detected${NO_COLOR}\n"
+    composer install --no-dev --prefer-dist --no-suggest
+fi
 php artisan migrate
+
+
+echo "\n${RED}post merge hook finish${NO_COLOR}\n"
```
d7415: Why red?

If you want to test for non-zero return codes and show red/green that might work.

pre-commit
```diff
@@ -1,22 +1,25 @@
...
+LIGHTPURPLE="\033[38;5;177m"
+PURPLE="\033[38;5;135m"
+CYAN="\033[38;5;014m"
+LIGHTBLUE="\033[38;5;39m"
+BLUE="\033[38;5;75m"
+DARKBLUE="\033[38;5;33m"
+LIGHTGRAY="\033[38;5;252m"
+DARKGRAY="\033[38;5;242m"
+BRIGHTRED="\033[91m"
+BOLD="\033[1m"
```
d7415: Same comment about clutter.

pre-commit
```diff
@@ -25,9 +28,8 @@ else
...
     echo "Please install php-cs-fixer, e.g.:"
     echo ""
-    echo "  composer install"
+    echo "${YELLOW}  composer install${NO_COLOR}"
     echo ""
 fi
 
-cd $CURRENT_DIRECTORY;
-echo "pre commit hook finish"
+echo "\n${RED}pre commit hook finish${NO_COLOR}\n"
```
d7415: Same comment about red.

pre-commit
```diff
@@ -25,9 +28,8 @@ else
     echo ""
     echo "Please install php-cs-fixer, e.g.:"
     echo ""
-    echo "  composer install"
+    echo "${YELLOW}  composer install${NO_COLOR}"
```
d7415: Why yellow?

d7415 (changes requested):
There's a lot of colour, which is probably sensible. We're not testing for support, but realistically I doubt that's a problem.

It should be possible for a user to disable `post-merge` (if not both) without their own branch. At the moment they can't even comment it out or make it non-executable because we overwrite on every composer run.

This doesn't affect me (I either want this or an not running composer at all), but I don't think we should be forcing users that way - I don't mind changing the default, but if we do there should be a way out.

pre-commit
ildyria: why not ? :laughing:

d7415: If the text was yellow as a warning, sure. But it makes the "do this" part look like a warning rather than an instruction.

pre-commit
```diff
@@ -26,10 +14,10 @@ if [ -x "$PHP_CS_FIXER" ]; then
     done
 else
     echo ""
-    echo "Please install php-cs-fixer, e.g.:"
+    echo "${YELLOW}Please install php-cs-fixer, e.g.:${NO_COLOR}"
     echo ""
-    echo "${YELLOW}  composer install${NO_COLOR}"
```
d7415: You can make that yellow *too* if you prefer, but this is good :)

d7415 (approved):
:+1:

post-merge
```diff
@@ -1,3 +1,25 @@
...
+GREEN="\033[38;5;010m"
+ORANGE="\033[38;5;214m"
+
+echo "\n${GREEN}post merge hook start${NO_COLOR}\n"
+
+# we check if vendor/bin/phpunit is present.
+# If yes this means we are in dev mode.
+PHPUNIT="vendor/bin/phpunit"
+
+if [ -x ".NO_AUTO_COMPOSER_MIGRATE" ]; then
```
d7415, line 12: Oops.

PR #570, opened 2020-05-16
prepare ground work for new smart albums

ildyria:
Rename smart albums... please tell me if I missed some.


app/Http/Controllers/AlbumController.php
```diff
@@ -225,14 +199,14 @@ public function getPositionData(Request $request)
 
 				$full_photo = $album->full_photo_visible();
```
d7415:
```suggestion
				$full_photo = $album->is_full_photo_visible();
```
?

app/Album.php
```diff
@@ -64,20 +62,27 @@
...
+		'downloadable' => 'int',
+		'share_button_visible' => 'int',
+	];
+
+	/**
+	 * Define if a content is accessible or not.
+	 *
+	 * @Var bool
+	 */
+	public $content_accessible = true;
```
d7415: Presumably this is going to be expanded into a proper function at some point? At the moment it doesn't seem to be referenced anywhere.

app/Album.php
```diff
@@ -89,6 +94,14 @@ public function photos()
 		return $this->hasMany('App\Photo', 'album_id', 'id');
 	}
 
+	/**
+	 * Return the list of photos.
+	 */
+	public function get_photos()
+	{
+		return $this->photos();
```
d7415, line 95: For consistency between normal and smart albums?

app/Album.php
```diff
@@ -146,239 +163,87 @@ public function full_photo_visible()
...
+	 * @return string|null
 	 */
-	public function is_downloadable()
+	public function str_parent_id()
 	{
-		if ($this->public) {
-			return $this->downloadable == 1;
-		} else {
-			return Configs::get_value('downloadable', '0') === '1';
+		if ($this->parent_id !== null) {
```
d7415: This logic is the reverse of the next two (they test for ` == null`, this tests for ` !== null`). They should probably be the same. Or they could be ternary on-liners? I'm not sure what's cleanest/clearest.

```php
return ($this->parent_id == null ? '' : strval($this->parent_id))
```

app/Album.php
```diff
@@ -146,239 +163,87 @@ public function full_photo_visible()
...
 		}
-
-		return $no_error;
 	}
 
 	/**
-	 * Recalculate takestamps of all albums in the database.
-	 * This is expensive and not normally necessary so we only use it
-	 * during migration.
+	 * Return the Album license or the default one.a1.
```
d7415:
```suggestion
	 * Return the Album license or the default one.
```

app/Exceptions/RequestAdminDataException.php
```diff
@@ -0,0 +1,16 @@
...
+
+use Exception;
+use Throwable;
+
+class RequestAdminDataException extends Exception
+{
+	public function __construct(
+		$code = 0,
+		Throwable $previous = null
+	) {
```
d7415, line 13: Aside: Something about this style choice irks me. I don't know why. (No reason to change it, but while I'm commenting anyway :P)

app/Http/Controllers/AlbumController.php
```diff
@@ -261,13 +218,10 @@ public function getPublic(Request $request)
...
-				return 'false';
-			case 's':
-				return 'false';
-			case 'r':
-				return 'false';
-			case '0':
+			case 'starred':
+			case 'public':
+			case 'recent':
+			case 'unsorted':
```
d7415, line 225: :+1:

app/Http/Controllers/AlbumController.php
```diff
@@ -310,15 +264,11 @@ public function setTitle(Request $request)
 
 		$albums = Album::whereIn('id', explode(',', $request['albumIDs']))->get();
 
-		if ($albums == null) {
-			return 'false';
-		}
-
 		$no_error = false;
```
d7415: Technically, is this not `true` if there *is* an error?

app/Http/Controllers/AlbumController.php
```diff
@@ -504,7 +457,7 @@ public function delete(Request $request)
 			$no_error &= $album->delete();
 
 			if ($parentAlbum !== null) {
-				$no_error &= $parentAlbum->update_takestamps([$minTS, $maxTS], false);
+				$no_error &= AlbumUpdate::update_takestamps($parentAlbum, [$minTS, $maxTS], false);
```
d7415, line 461: I'm going to assume there's a reason it's better for this to live outside the album object.

app/Http/Controllers/AlbumController.php
```diff
@@ -777,13 +730,17 @@ public function getArchive(Request $request)
 
 				$compress_album = function ($photos_sql, $dir, &$dirs, $parent_dir, $album) use (&$zip, $badChars, &$compress_album) {
 					if ($album !== null) {
-						if (!$this->sessionFunctions->is_current_user($album->owner_id) &&
-						!$album->is_downloadable()) {
+						if (
+							!$this->sessionFunctions->is_current_user($album->owner_id) &&
+							!$album->is_downloadable()
+						) {
```
d7415: Is it just me, or do the conditions look like code when styled this way?

app/Http/Controllers/PhotoController.php
```diff
@@ -411,8 +419,11 @@ public function setLicense(Request $request)
...
 		if (!$found) {
-			Logs::error(__METHOD__, __LINE__,
-				'wrong kind of license: ' . $request['license']);
+			Logs::error(
+				__METHOD__,
+				__LINE__,
+				'wrong kind of license: ' . $request['license']
+			);
 
 			return Response::error('wrong kind of license!');
```
d7415: I know this was an existing message, but would "License not recognised" be clearer here?

app/ModelFunctions/AlbumActions/Cast.php
```diff
@@ -0,0 +1,93 @@
...
+	public static function toArray(Album $album): array
+	{
+		return [
+			'id' => strval($album->id),
+			'title' => $album->title,
+			'public' => strval($album->public),
+			'full_photo' => Helpers::str_of_bool($album->is_full_photo_visible()),
+			'visible' => strval($album->visible_hidden),
+			'parent_id' => $album->str_parent_id(),
+			'description' => strval($album->description),
```
d7415, line 30: Why is this `strval()`'d?

app/ModelFunctions/AlbumActions/Cast.php
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	/**
+	 * Set up the wrap arround of the photos if setting is true and if there are enough pictures.
+	 */
+	public static function wrapAroundPhotos(array &$return_photos): void
+	{
+		$photo_counter = count($return_photos);
+
+		if ($photo_counter > 0 && Configs::get_value('photos_wraparound', '1') === '1') {
```
d7415: Again existing code so probably out of scope, but do we want to wrap around for single-photo albums? I can't think of a way in which that would be useful.

Actually, it looks like we're not adding them for single-photo, but we're still executing this block. In theory changing this to `$photo_counter > 0` would remove the need for the `!==` block below.

app/ModelFunctions/AlbumActions/Cast.php
```diff
@@ -0,0 +1,93 @@
...
+
+			if ($lastElementId !== $firstElementId) {
+				$return_photos[$photo_counter - 1]['nextPhoto'] = $firstElementId;
+				$return_photos[0]['previousPhoto'] = $lastElementId;
+			}
+		}
+	}
+
+	/**
+	 * Given an Album, return the thumbs with respect to it's direct pictures (no recursion).
```
d7415: Typo and clarification.
```suggestion
	 * Given an Album, return the thumbs of its first 3 pictures (excluding subalbums).
```

app/ModelFunctions/AlbumActions/UpdateTakestamps.php
```diff
@@ -0,0 +1,164 @@
...
+			return $collect
+				->concat([$_album->id])
+				->concat(self::get_all_sub_albums_id($_album));
+		}, new Collection());
+	}
+
+	/**
+	 * Go through each sub album and update the minimum and maximum takestamp of the pictures.
+	 * This is expensive and not normally necessary so we only use it
+	 * during migration.
```
d7415: Wrapping discrepancy - Should we wrap at the long or short line?

app/ModelFunctions/AlbumActions/UpdateTakestamps.php
```diff
@@ -0,0 +1,164 @@
...
+	 */
+	public function update_min_max_takestamp(Album $album)
+	{
+		$album_list = self::get_all_sub_albums_id($album, [$album->id]);
+
+		$min = Photo::whereIn('album_id', $album_list)->min('takestamp');
+		$max = Photo::whereIn('album_id', $album_list)->max('takestamp');
+		$album->min_takestamp = $min;
+		$album->max_takestamp = $max;
```
d7415: Would
```suggestion
		$album->min_takestamp = Photo::whereIn('album_id', $album_list)->min('takestamp');
		$album->max_takestamp = Photo::whereIn('album_id', $album_list)->max('takestamp');
```
be better?

app/ModelFunctions/AlbumActions/UpdateTakestamps.php
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	/**
+	 * Update album's min_takestamp and max_takestamp based on changes made
+	 * to the album content.  If needed, recursively updates parent album(s).
```
d7415:
```suggestion
	 * to the album content.  If needed, recursively update parent album(s).
```

app/ModelFunctions/AlbumActions/UpdateTakestamps.php
```diff
@@ -0,0 +1,164 @@
...
+	 * @param bool  $adding     :     true if adding new content, false if removing
+	 *
+	 * @return bool: true if successful
+	 */
+	public static function update_takestamps(Album $album, array $takestamps, bool $adding)
+	{
+		// Begin by calculating min and max takestamps from the array.
+		// The array may contain null values, which is why we can't use the
+		// built-in min() function for this (it will always return null if
+		// present).  For consistency, we don't use the built-in max()
```
d7415, line 60: Potentially we could use `array_filter()`, but I doubt that would significantly improve performance, if at all. This shouldn't be very performance sensitive anyway. I'd be tempted to use built-in `max()` though, especially as the "consistent" version includes an extra null check that we don't really need. Unless we're likely to mess with this later?

app/ModelFunctions/AlbumActions/UpdateTakestamps.php
```diff
@@ -0,0 +1,164 @@
...
+			if ($takestamp !== null) {
+				if ($minTS === null || $minTS > $takestamp) {
+					$minTS = $takestamp;
+				}
+				if ($maxTS === null || $maxTS < $takestamp) {
+					$maxTS = $takestamp;
+				}
+			}
+		}
+		if ($minTS === null || $maxTS === null) {
```
d7415: I can't decide about this one. The logic is sound (if either is null we don't want to mess with it), but if one is null, both *should* be null. Maybe just...
```suggestion
		}
		// If either minTS or maxTS is null, both should be null so we don't need to update anything.
		if ($minTS === null || $maxTS === null) {
```

That's really bad diffing, and I haven't checked the comment length.

app/ModelFunctions/AlbumActions/UpdateTakestamps.php
```diff
@@ -0,0 +1,164 @@
...
+					$album->max_takestamp = $max_photos;
+				} else {
+					$album->max_takestamp = $max_albums;
+				}
+				$changed = true;
+			}
+		}
+
+		if ($changed) {
+			$no_error &= $album->save();
```
d7415, line 133: This could just be `=`, but I guess it makes sense for consistency.

app/ModelRessources/AlbumRessources.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace App\ModelRessources;
+
+use App\Album;
+use Illuminate\Support\Facades\Hash;
+
+class AlbumRessources
+{
+	/**
+	 * Given a password, check if it matches albums password.
```
d7415:
```suggestion
	 * Given a password, check if it matches the album's password.
```

app/ModelRessources/AlbumRessources.php
```diff
@@ -0,0 +1,22 @@
...
+	 *
+	 * @param string $password
+	 *
+	 * @return bool returns when album is public
```
d7415:
```suggestion
	 * @return bool returns true if password is correct or no album password is set
```

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -86,7 +96,7 @@ public function create(string $title, int $parent_id, int $user_id): Album
 		if ($parent !== null) {
 			$album->parent_id = $parent->id;
 
-			// Admin can add subalbums to other users' albums.  Make sure that
+			// Admin can add retSubAlbums to other users' albums.  Make sure that
```
d7415: I think this was in error?

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -123,61 +133,73 @@ public function create(string $title, int $parent_id, int $user_id): Album
...
 	}
 
-	/**
-	 * get the thumbs of an album.
-	 *
-	 * @param array $return
-	 * @param array $album_list
-	 */
-	public function gen_thumbs(array &$return, $album_list)
+	public function get_thumbs_album($album, array $previousThumbsId): BaseCollection
```
d7415: To match the calling function and to make it clearer that there may be more than one.
```suggestion
	public function get_thumbs_album($album, array $previousThumbIDs): BaseCollection
```

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -123,61 +133,73 @@ public function create(string $title, int $parent_id, int $user_id): Album
...
-			foreach ($return['albums'] as &$album) {
-				$childThumbIDs = array_merge($childThumbIDs, $album['thumbIDs']);
-				unset($album['thumbIDs']);
-			}
-		}
-
-		$photos = Photo::whereIn('album_id', $album_list)
-			->orWhereIn('id', $childThumbIDs)
+		$photos = Photo::where('album_id', $album->id)
+			->orWhereIn('id', $previousThumbsId)
```
d7415: To match above
```suggestion
			->orWhereIn('id', $previousThumbIDs)
```

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -123,61 +133,73 @@ public function create(string $title, int $parent_id, int $user_id): Album
...
-			$k++;
-		}
+
+		// php7.4: return $photos->map(fn ($photo) => PhotoCast::toThumb($photo, $this->symLinkFunctions));
+		return $photos->map(function ($photo) {
+			return PhotoCast::toThumb($photo, $this->symLinkFunctions);
+		});
+	}
+
+	public function get_thumbs_reduction(Album $album, BaseCollection $previous): BaseCollection
```
d7415: Is there a plan for using these functions differently? Otherwise `get_thumbs_album()` and `get_thumbs_reduction()` may as well just be part of `get_thumbs()`.

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -302,112 +318,65 @@ public function prepare_albums(?Collection $albums)
...
-				}
+		$albumIDs = new Collection();
+		$kinds = ['albums', 'shared_albums'];
+
+		foreach ($kinds as $kind) {
+			if ($toplevel[$kind] !== null) {
+				$toplevel[$kind]->each(function ($album) use (&$albumIDs) {
+					if ($this->readAccessFunctions->album($album) === 1) {
+						$albumIDs->push($album->id);
+						$albumIDs = $albumIDs->concat($this->get_sub_albums_id($album));
```
d7415, line 377: I don't think it matters, but it looks like this will include sub albums that the user doesn't have access to

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -417,129 +386,98 @@ public function getPublicAlbums($toplevel = null)
...
+		$smartAlbums[] = new UnsortedAlbum($this, $this->sessionFunctions);
+		$smartAlbums[] = new StarredAlbum($this, $this->sessionFunctions);
+		$smartAlbums[] = new PublicAlbum($this, $this->sessionFunctions);
+		$smartAlbums[] = new RecentAlbum($this, $this->sessionFunctions);
+
+		$can_see_smart = $this->sessionFunctions->is_logged_in() && $this->sessionFunctions->can_upload();
+
+		foreach ($smartAlbums as $smartAlbum) {
+			if ($can_see_smart || $smartAlbum->is_public()) {
+				$publicAlbums = $publicAlbums ?? $this->getPublicAlbumsId($toplevel);
```
d7415, line 411: Is there a scope issue if this line is outside the loop?

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -417,129 +386,98 @@ public function getPublicAlbums($toplevel = null)
...
-				 */
-				$photos_sql = Photo::select_recent(Photo::whereIn('album_id', $publicAlbums))->limit(3);
-				$this->genSmartAlbumsThumbs($return, $photos_sql, 'recent');
-			}
+		return $return;
+	}
 
-			return $return;
+	/**
+	 * Given a query, depending of the sort collumn, we do it in the query or on the collection.
```
d7415:
```suggestion
	 * Given a query, depending on the sort column, we do it in the query or on the collection.
```

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -417,129 +386,98 @@ public function getPublicAlbums($toplevel = null)
...
 	/**
+	 * ! may be memory intensive
+	 * TODO: Add recursion limit back.
+	 *
 	 * Recursively returns the tree structure of albums.
 	 *
 	 * @param Album $album
 	 * @param $username : speed optimization to avoid an extra query,
-	 * taking advantage of the fact that subalbums inherit parent's owner
+	 * taking advantage of the fact that retSubAlbums inherit parent's owner
```
d7415:
```suggestion
	 * taking advantage of the fact that subalbums inherit parent's owner
```

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -417,129 +386,98 @@ public function getPublicAlbums($toplevel = null)
...
-						// to get accurate thumbs info though.
-						$subSubAlbums = [$subAlbum->id];
-						$this->get_sub_albums($subSubAlbums, $subAlbum);
-						$this->gen_thumbs($subAlbumData, $subSubAlbums);
-						$subAlbumData['has_albums'] = count($subSubAlbums) > 1 ? '1' : '0';
+						$_album->content_accessible = false;
 					}
+					// when we generate the thumbs, we need to know whether that content is visible or not.
+					return new Collection([$_album, $collected]);
 				}
```
d7415, line 473: It might be cleaner to repeat `return new Collection([$_album, $collected]);` than to repeat the test. I don't feel especially strongly about it though.

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -550,60 +488,68 @@ public function get_albums(Album $album, $username, $recursionLimit = 0): array
...
+	 * ! Memory intensive
+	 * ! lots of SQL query.
+	 *
 	 * Recursively go through each sub album and build a list of them.
-	 * Unlike Album::get_all_sub_albums(), this function follows access
-	 * checks and skips hidden subalbums.  The optional third argument, if
-	 * true, will result in password-protected albums being included (but not
-	 * their content).
+	 * Unlike AlbumActions\UpdateTakestamps::get_all_sub_albums_id(),
+	 * this function follows access checks and skips hidden retSubAlbums.
```
d7415:
```suggestion
	 * this function follows access checks and skips hidden subalbums.
```

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -550,60 +488,68 @@ public function get_albums(Album $album, $username, $recursionLimit = 0): array
...
+	 * The optional third argument, if true, will result in password-protected
+	 * albums being included (but not their content).
 	 *
 	 * @param array $return
 	 * @param Album $parentAlbum
```
d7415, line 512: This parameter no longer exists.
```suggestion
```

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -550,60 +488,68 @@ public function get_albums(Album $album, $username, $recursionLimit = 0): array
...
+	 * ! lots of SQL query.
+	 *
 	 * Recursively go through each sub album and build a list of them.
-	 * Unlike Album::get_all_sub_albums(), this function follows access
-	 * checks and skips hidden subalbums.  The optional third argument, if
-	 * true, will result in password-protected albums being included (but not
-	 * their content).
+	 * Unlike AlbumActions\UpdateTakestamps::get_all_sub_albums_id(),
+	 * this function follows access checks and skips hidden retSubAlbums.
+	 * The optional third argument, if true, will result in password-protected
```
d7415:
```suggestion
	 * The optional second argument, if true, will result in password-protected
```

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -550,60 +488,68 @@ public function get_albums(Album $album, $username, $recursionLimit = 0): array
...
+	}
+
+	/**
+	 * ! Memory intensive.
+	 *
+	 * Same as above but only with the ID
+	 *
+	 * Recursively go through each sub album and build a list of them.
+	 * Unlike Album::get_all_sub_albums(), this function follows access
+	 * checks and skips hidden retSubAlbums.  The optional third argument, if
```
d7415:
```suggestion
	 * checks and skips hidden subalbums.  The optional second argument, if
```

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -550,60 +488,68 @@ public function get_albums(Album $album, $username, $recursionLimit = 0): array
...
+	 * true, will result in password-protected albums being included (but not
+	 * their content).
+	 *
+	 * @param array $return
+	 * @param Album $parentAlbum
```
d7415, line 541:
```suggestion
```

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -550,60 +488,68 @@ public function get_albums(Album $album, $username, $recursionLimit = 0): array
...
+				$collect = $collect->concat([$_album, $collected]);
 			}
-		}
+		}, new Collection());
+	}
+
+	/**
+	 * ! Memory intensive.
+	 *
+	 * Same as above but only with the ID
```
d7415:
```suggestion
	 * Same as above but only returns the album IDs
```

app/ModelFunctions/PhotoActions/Cast.php
```diff
@@ -0,0 +1,41 @@
...
+namespace App\ModelFunctions\PhotoActions;
+
+use App\ModelFunctions\SymLinkFunctions;
+use App\Photo;
+use Illuminate\Support\Facades\Storage;
+
+class Cast
+{
+	/**
+	 * Returns album-attributes into a front-end friendly format. Note that some attributes remain unchanged.
```
d7415:
```suggestion
	 * Returns photo attributes into a front-end friendly format. Note that some attributes remain unchanged.
```
?

app/ModelFunctions/PhotoActions/Thumb.php
```diff
@@ -0,0 +1,45 @@
...
+	public function set_thumb2x(): void
+	{
+		$this->thumb2x = Helpers::ex2x($this->thumb);
+	}
+
+	public function from_photo(Photo $photo): void
+	{
+		$this->types = $photo->type;
+		$this->thumbIDs = $photo->id;
+		$this->thumbs = Storage::url('thumb/' . $photo->thumb);
```
d7415:
```suggestion
		$this->thumb = Storage::url('thumb/' . $photo->thumb);
```
?

Also, does this need any symlink handling?

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -145,17 +147,18 @@ public function file_type($file, string $extension)
...
+	 * @param array      $file
+	 * @param int|string $albumID_in
+	 * @param bool       $delete_imported
+	 * @param bool       $force_skip_duplicates
+	 * @param bool       $resync_metadata
 	 *
 	 * @return string|false ID of the added photo
 	 */
-	public function add(array $file, $albumID_in = 0, $delete_imported = false, $force_skip_duplicates = false, $resync_metadata = false)
+	public function add(array $file, $albumID_in = 0, bool $delete_imported = false, bool $force_skip_duplicates = false, $resync_metadata = false)
```
d7415:
```suggestion
	public function add(array $file, $albumID_in = 0, bool $delete_imported = false, bool $force_skip_duplicates = false, bool $resync_metadata = false)
```

app/ModelFunctions/PhotoFunctions.php
```diff
@@ -169,33 +172,24 @@ public function add(array $file, $albumID_in = 0, $delete_imported = false, $for
...
 
-				// r for recent
-				// 0 for unsorted
-			case '0':
-			case 'r':
-				$public = 0;
-				$star = 0;
-				$albumID = null;
+			case 'unsorted':
+			case '0': // root
```
d7415, line 189: Does this handle (or need to handle) integer 0?

d7415 (changes requested):
Mostly looks good. Untested. Some queries, some typos, etc.

app/Album.php
ildyria: yes.

app/Album.php
ildyria: removed.

app/Album.php
ildyria: Moving to ternary.

app/Album.php
ildyria: fixed.

app/Http/Controllers/AlbumController.php
ildyria: This is just extracting the action from the model to make the model lighter. Also those actions would not be used on SmartAlbums which extends Album.

app/ModelFunctions/AlbumActions/Cast.php
ildyria: because description is nullable:
```php
			$table->text('description')->nullable();
```

app/ModelFunctions/AlbumActions/Cast.php
ildyria: fixed.

app/ModelFunctions/AlbumActions/UpdateTakestamps.php
ildyria:
```php
max(null, 0) = null
max(0, null) = 0
```
I would rather avoid max. :/

app/ModelFunctions/AlbumFunctions.php
ildyria: True. But then I fear that the function becomes a bit too big to be easy to understand as we are using a functional right fold construction. :confused:

app/ModelFunctions/AlbumFunctions.php
ildyria: Actually no. Because in `get_sub_albums_id()` we also check if `$this->readAccessFunctions->album($subalbum) === 1`

app/ModelFunctions/AlbumFunctions.php
ildyria: Not really. It is a small optimization. So we don't do all the heavy query if none of the smart albums are displayed.

app/ModelFunctions/PhotoFunctions.php
ildyria: nope does not need to. Inputs are seen as string.

ildyria (commented):
Fixed.

app/ModelFunctions/AlbumFunctions.php
d7415: ~21 lines? I'm going to pretend I understand what you said at the end there :P

Not a strong opinion, I just find it harder to follow code that jumps around for no clear reason. If there's a reason, that's fine.

d7415 (approved):

Issue #575, opened 2020-05-18
[Enhancement] Support XMP sidecar files


alex-phillips:
I'm managing all of my photos through Digikam and have opted to not write / directly modify my images and instead write all metadata to XMP sidecar files. Would be great if lychee, when running `lychee:sync` could identify these files and use the metadata from them instead of the image (or use the image metadata and overwrite values from the XMP file if the field exists in both).

I'd be happy to help implement, but wanted to be sure this was of interest. Could also be controlled via an advanced config setting.

alex-phillips:
I would also want XMP keywords to be converted into 'tags' in lychee so photos can be imported and already be tagged based on their metadata.

tmp-hallenser:
I'd support that (although I'd add an option if image metadata or xmp data has higher priority)

ildyria:
I support that too. Unfortunately I am lacking the time to implement it.

alex-phillips:
@ildyria Exiftool actually already supports XMP files, so reading these in and merging the data turned out to be super simple: https://github.com/LycheeOrg/Lychee/pull/577

PR #577, opened 2020-05-19
Feature/xmp sidecars

alex-phillips:
This reads in XMP sidecar files (if they exist). Thankfully, `exiftool` supports reading in sidecars, so we can use the same technique we're using to read the files. We merge both file and sidecar metadata, taking priority based on user settings (default to prefer image metadata). Resolves feature request https://github.com/LycheeOrg/Lychee/issues/575

### DIFF
--- a/app/Configs.php
+++ b/app/Configs.php
@@ -320,14 +320,6 @@ class Configs extends Model
 	}
 
 	/**
-	 * @return bool returns the sidecare metadata preference setting
-	 */
-	public static function preferSidecarMetadata()
-	{
-		return intval(self::get_value('prefer_sidecar_metadata')) === 1;
-	}
-
-	/**
 	 * Define scopes.
 	 */
 
--- a/app/Configs.php
+++ b/app/Configs.php
@@ -320,6 +320,14 @@ class Configs extends Model
 	}
 
 	/**
+	 * @return bool returns the sidecare metadata preference setting
+	 */
+	public static function preferSidecarMetadata()
+	{
+		return intval(self::get_value('prefer_sidecar_metadata')) === 1;
+	}
+
+	/**
 	 * Define scopes.
 	 */
 
--- a/app/Metadata/Extractor.php
+++ b/app/Metadata/Extractor.php
@@ -104,8 +104,8 @@ class Extractor
 			$exif = $reader->read($filename);
 
 			$realFile = readlink($filename) ?: $filename;
-			if (Configs::hasExiftool() && file_exists("$realFile.xmp")) {
-				$sidecarData = $reader->read("$realFile.xmp")->getData();
+			if (Configs::hasExiftool() && file_exists($realFile . '.xmp')) {
+				$sidecarData = $reader->read($realFile . '.xmp')->getData();
 			}
 		} catch (\Exception $e) {
 			// Use Php native tools
@@ -114,7 +114,7 @@ class Extractor
 			$exif = $reader->read($filename);
 		}
 
-		if (Configs::preferSidecarMetadata()) {
+		if (Configs::get_value('prefer_available_xmp_metadata', '0') == '1') {
 			$exif->setData(array_merge($exif->getData(), $sidecarData));
 		} else {
 			$exif->setData(array_merge($sidecarData, $exif->getData()));
--- a/app/Metadata/Extractor.php
+++ b/app/Metadata/Extractor.php
@@ -98,14 +98,15 @@ class Extractor
 
 		// Attempt to get sidecar metadata if it exists, make sure to check 'real' path in case of symlinks
 		$sidecarData = [];
-		$realFile = readlink($filename) ?: $filename;
-		if (file_exists("$realFile.xmp")) {
-			$sidecarData = $reader->read("$realFile.xmp")->getData();
-		}
 
 		try {
 			// this can throw an exception in the case of Exiftool adapter!
 			$exif = $reader->read($filename);
+
+			$realFile = readlink($filename) ?: $filename;
+			if (Configs::hasExiftool() && file_exists("$realFile.xmp")) {
+				$sidecarData = $reader->read("$realFile.xmp")->getData();
+			}
 		} catch (\Exception $e) {
 			// Use Php native tools
 			Logs::error(__METHOD__, __LINE__, $e->getMessage());
--- a/app/Metadata/Extractor.php
+++ b/app/Metadata/Extractor.php
@@ -98,7 +98,14 @@ class Extractor
 
 		try {
 			// this can throw an exception in the case of Exiftool adapter!
-			$exif = $reader->read($filename);
+
+			// Get the 'real' file in case we are importing via symlinks
+			$realFile = readlink($filename) ?: $filename;
+			if (file_exists("$realFile.xmp")) {
+				$exif = $reader->read("$realFile.xmp");
+			} else {
+				$exif = $reader->read($filename);
+			}
 		} catch (\Exception $e) {
 			// Use Php native tools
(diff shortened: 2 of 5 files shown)


tmp-hallenser:
Could you share an example of an xmp file? I was wondering if the filename is the only way to determine the match between image and xmp file?

app/Configs.php
```diff
@@ -319,6 +319,14 @@ public static function hasFFmpeg()
 		return $has_ffmpeg;
 	}
 
+	/**
```
ildyria: I don't think this is really needed as this is called only once.

app/Metadata/Extractor.php
```diff
@@ -96,15 +96,30 @@ public function extract(string $filename, string $type): array
...
+				$sidecarData = $reader->read("$realFile.xmp")->getData();
+			}
 		} catch (\Exception $e) {
 			// Use Php native tools
 			Logs::error(__METHOD__, __LINE__, $e->getMessage());
 			$reader = Reader::factory(Reader::TYPE_NATIVE);
 			$exif = $reader->read($filename);
 		}
+
+		if (Configs::preferSidecarMetadata()) {
```
ildyria: Here you can instead use:
```php
Configs::get_value('prefer_sidecar_metadata', '0') == '1'
```

app/Metadata/Extractor.php
```diff
@@ -96,15 +96,30 @@ public function extract(string $filename, string $type): array
...
+		// Attempt to get sidecar metadata if it exists, make sure to check 'real' path in case of symlinks
+		$sidecarData = [];
+
 		try {
 			// this can throw an exception in the case of Exiftool adapter!
 			$exif = $reader->read($filename);
+
+			$realFile = readlink($filename) ?: $filename;
+			if (Configs::hasExiftool() && file_exists("$realFile.xmp")) {
+				$sidecarData = $reader->read("$realFile.xmp")->getData();
```
ildyria: I don't really like the syntax `"$var.something"`, we tend to stick with `$var . ".something"` as this is slightly more readable (also syntax coloring wise).

ildyria (changes requested):
I think I would rather use as setting name  `prefer_available_xmp_metadata` instead of  `prefer_sidecar_metadata` as I feel this is clearer (avoid the question what are sidecar metadata which to be honest had no knowledge of before).

ildyria:
Requested opinion from @d7415 and @tmp-hallenser , should they wish to give them. :+1:

alex-phillips:
@ildyria I modified the PR per your requested changes

ildyria (approved):
LGTM.

alex-phillips:
@tmp-hallenser I believe so. Here's an XMP file that I added the tags `test` and `test2` to. The file it belongs to is called `IMG_0377.PNG` and the XMP file is `IMG_0377.PNG.xmp`. I don't see any specific identifiers inside the XMP file.

```<?xpacket begin="﻿" id="W5M0MpCehiHzreSzNTczkc9d"?>
<x:xmpmeta xmlns:x="adobe:ns:meta/" x:xmptk="XMP Core 4.4.0-Exiv2">
 <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
  <rdf:Description rdf:about=""
    xmlns:mwg-rs="http://www.metadataworkinggroup.com/schemas/regions/"
    xmlns:MP="http://ns.microsoft.com/photo/1.2/"
    xmlns:MPRI="http://ns.microsoft.com/photo/1.2/t/RegionInfo#"
    xmlns:digiKam="http://www.digikam.org/ns/1.0/"
    xmlns:MicrosoftPhoto="http://ns.microsoft.com/photo/1.0/"
    xmlns:lr="http://ns.adobe.com/lightroom/1.0/"
    xmlns:mediapro="http://ns.iview-multimedia.com/mediapro/1.0/"
    xmlns:acdsee="http://ns.acdsee.com/iptc/1.0/"
    xmlns:dc="http://purl.org/dc/elements/1.1/"
   acdsee:categories="&lt;Categories&gt;&lt;Category Assigned=&quot;1&quot;&gt;test&lt;Category Assigned=&quot;1&quot;&gt;test2&lt;/Category&gt;&lt;/Category&gt;&lt;/Categories&gt;">
   <mwg-rs:Regions rdf:parseType="Resource">
    <mwg-rs:RegionList>
     <rdf:Bag/>
    </mwg-rs:RegionList>
   </mwg-rs:Regions>
   <MP:RegionInfo rdf:parseType="Resource">
    <MPRI:Regions>
     <rdf:Bag/>
    </MPRI:Regions>
   </MP:RegionInfo>
   <digiKam:TagsList>
    <rdf:Seq>
     <rdf:li>test</rdf:li>
     <rdf:li>test/test2</rdf:li>
    </rdf:Seq>
   </digiKam:TagsList>
   <MicrosoftPhoto:LastKeywordXMP>
    <rdf:Bag>
     <rdf:li>test</rdf:li>
     <rdf:li>test/test2</rdf:li>
    </rdf:Bag>
   </MicrosoftPhoto:LastKeywordXMP>
   <lr:hierarchicalSubject>
    <rdf:Bag>
     <rdf:li>test</rdf:li>
     <rdf:li>test|test2</rdf:li>
    </rdf:Bag>
   </lr:hierarchicalSubject>
   <mediapro:CatalogSets>
    <rdf:Bag>
     <rdf:li>test</rdf:li>
     <rdf:li>test|test2</rdf:li>
    </rdf:Bag>
   </mediapro:CatalogSets>
   <dc:subject>
    <rdf:Bag>
     <rdf:li>test</rdf:li>
     <rdf:li>test2</rdf:li>
    </rdf:Bag>
   </dc:subject>
  </rdf:Description>
 </rdf:RDF>
</x:xmpmeta>
<?xpacket end="w"?>

```

tmp-hallenser:
Too bad, I was hoping to have some additional identifier which would allow us to also upload xmp files via the UI.

app/Metadata/Extractor.php
```diff
@@ -96,15 +96,30 @@ public function extract(string $filename, string $type): array
...
 		}
 
+		// Attempt to get sidecar metadata if it exists, make sure to check 'real' path in case of symlinks
+		$sidecarData = [];
+
 		try {
 			// this can throw an exception in the case of Exiftool adapter!
 			$exif = $reader->read($filename);
+
+			$realFile = readlink($filename) ?: $filename;
```
tmp-hallenser, line 108: What's `?:` doing?

ildyria: it is very similar to `??` but will test on the boolean truth of the value rather than on the isset.
```php
$value ?: $default;
// is equivalent to
!$value ? $value : $default;
// is not equivalent to
(isset($value) && !$value) ? $value : $default;
```
also:
```php
$a = ''; // or 0 or false

$b = $a ?? 'a'; 
// $b is '' or 0 or false

$c = $a ?: 'a';
// $c is 'a'
```

alex-phillips: Right, basically "use the value from 'readline' if available, otherwise use $filename".

ildyria: Note that readlink "Returns the contents of the symbolic link path or FALSE on error." so `??` would not work here.

tmp-hallenser: I'd vote to add a short comment on the line explaining what the notation does to make it easier to read for everyone not being a PHP guru :)

alex-phillips:
Anything else I need to tackle before this gets merged?

ildyria:
Just waiting for @tmp-hallenser, see if he is happy, otherwise I would say good for merge. :)

tmp-hallenser (approved):

PR #579, opened 2020-05-19
Feature/ghostbust dead symlinks

### LINKED ISSUE #578: [Enhancement] Ghostbuster command to clean up dead symlinks
I'm importing my library via symlinks and the `lychee:sync` command but it doesn't look like the sync command removes photos if the image from the source directory is no longer there. I would opt to extend (or implement a new command) that scans the "big" directory for dead symlinks and cleans them up and removes them from the database.

I'm happy to implement this if it is something that will be accepted.

alex-phillips:
Resolves https://github.com/LycheeOrg/Lychee/issues/578

A couple of caveats here: I had to manually delete the 'big' file path even though we are calling `->delete()` on the photo because inside the `predelete` method, `Storage::exists` actually returns `false` if you pass in a dead symlink. So a predelete and delete is not enough here, have to manually unlink the dead symlink itself.

Something to consider: should this be an additional flag we pass into the ghostbuster command? I'm thinking in the event that images are mounted on an external drive, NFS, whatever and are 'temporarily' unavailable. Just something to keep in mind.

### DIFF
--- a/app/Console/Commands/Ghostbuster.php
+++ b/app/Console/Commands/Ghostbuster.php
@@ -27,7 +27,7 @@ class Ghostbuster extends Command
 	 *
 	 * @var string
 	 */
-	protected $signature = 'lychee:ghostbuster {dryrun=1 : Dry Run default is True}';
+	protected $signature = 'lychee:ghostbuster {removeDeadSymLinks=0 : Remove Photos with dead symlinks} {dryrun=1 : Dry Run default is True}';
 
 	/**
 	 * The console command description.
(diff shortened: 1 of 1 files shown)


app/Console/Commands/Ghostbuster.php
```diff
@@ -114,6 +117,22 @@ public function handle()
...
 					}
 				}
+
+				if ($isDeadSymlink) {
+					foreach ($photos as $photo) {
+						if ($dryrun) {
+							$this->line(str_pad($photo->url, 50) . $this->col->red(' photo will be removed') . '.');
+						} else {
+							// Laravel apparently doesn't thing dead symlinks 'exist', so manually remove the original here.
+							unlink("$path/$url");
```
ildyria:
```suggestion
							unlink($path . '/' . $url);
```

app/Console/Commands/Ghostbuster.php
```diff
@@ -114,6 +117,22 @@ public function handle()
 						}
 					}
 				}
+
+				if ($isDeadSymlink) {
+					foreach ($photos as $photo) {
+						if ($dryrun) {
+							$this->line(str_pad($photo->url, 50) . $this->col->red(' photo will be removed') . '.');
+						} else {
+							// Laravel apparently doesn't thing dead symlinks 'exist', so manually remove the original here.
```
ildyria:
```suggestion
							// Laravel apparently doesn't think dead symlinks 'exist', so manually remove the original here.
```

ildyria (approved):
Not tested. LGTM (minus the small editions).

ildyria:
> should this be an additional flag we pass into the ghostbuster command? I'm thinking in the event that images are mounted on an external drive, NFS, whatever and are 'temporarily' unavailable. Just something to keep in mind.

I think this would be a good idea. This would allow the use of the command to remove ghosts files (which is usually safe) while retaining the ability to not modify the database (for the photos).

alex-phillips:
Changes made. Thanks!

alex-phillips:
Anything else I need to tackle before this gets merged?

PR #585, opened 2020-05-22
Fix #584

### LINKED ISSUE #584: [Enhancement] Default to RSS 2.0 to support enclosure tag
I'm glad to see feeds in the last release!

The feed defaults to Atom 1.0, which doesn't support linking media. What do you think about defaulting to RSS 2.0 which would allow the actual image to be sent along in the feed? It would take a couple template tweaks in the `laravel-feed` views and a couple lines in the controller file.

bennettscience:
`laravel-feed` defaults to Atom 1.0 feeds, which do not support media attachments. Since this is a feed for photos, switching to RSS 2.0 as the default allows items in the feed to have an `enclosure` tag which sends the photo to RSS aggregators.

This update creates local versions of the feed templates from the vendor folder. Users could modify these further, adding more metadata (icons, etc).

The `enclosure` tag requires length and mime parameters which are retrieved using Laravel's built-in file information methods.

### DIFF
--- a/app/Http/Controllers/RSSController.php
+++ b/app/Http/Controllers/RSSController.php
@@ -71,6 +71,8 @@ class RSSController extends Controller
 				$id = 'view?p=' . $photo_model->id;
 			}
 
+			$photo['url'] = $photo['url'] ?: $photo['medium2x'] ?: $photo['medium'];
+
 			$length = \File::size($photo['url']);
 			$mime_type = \File::mimeType($photo['url']);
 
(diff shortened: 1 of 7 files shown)


ildyria:
You forgot to change that file: https://github.com/LycheeOrg/Lychee/blob/master/tests/Feature/RSSTest.php

ildyria:
I fixed the url and rebased on master to fix the build by travis.

bennettscience:
I actually ran into another issue that I need to commit. The enclosure fails when images in the album aren't available for download, throwing a 500 error. I have a fix that handles that case and I'll get it pushed up later.

ildyria:
@bennettscience you mean this : https://github.com/LycheeOrg/Lychee/pull/585/commits/af62a71f0774b34bf6d652ac3ebec8b48ae84115 ? :p

PR #587, opened 2020-05-22
use exif reader for video xmp files

alex-phillips:
We can't assume that the reader is already an EXIF reader since the file might be a video. Create a new reader here specifically for the XMP sidecar.


tmp-hallenser (approved):

alex-phillips:
sorry, made an additional accidental commit here (meant for my other PR) which is why I had to do a force push. It's still the original changes in the PR though. Sorry for the false alarm!

ildyria (approved):

PR #574, opened 2020-05-18
Fix support of raw files

tmp-hallenser:

PR #588, opened 2020-05-22
add an option to the sync command to 're-sync' images that already exist

alex-phillips:
Since XMP sidecars are now allowed, images with the same checksum will not get updated metadata if the sidecar changes but the checksum remains the same. This adds an optional flag to the `lychee:sync` command that forces existing images with no change in checksum to get updated metadata from XMP files (if they exist). It only updates the image if the metadata read in differs from the metadata that the image aleady has.


app/ModelFunctions/PhotoFunctions.php
```diff
@@ -787,4 +794,37 @@ public function getValidExtensions(): array
...
+	 *
+	 * @return void
+	 */
+	private function getFileMetadata($file, $path, $kind, $mimeType, $extension)
+	{
+		if ($kind == 'raw') {
+			$info = $this->metadataExtractor->bare();
+			$this->metadataExtractor->size($info, $path);
+			$this->metadataExtractor->validate($info);
+			$info['type'] = 'raw';
```
tmp-hallenser: #574 changes support of raw files, i.e. metadata is also extracted for raw files if possible. See https://github.com/LycheeOrg/Lychee/blob/32b460e93e56b50e0d49c25e4cf7a0e12968e282/app/ModelFunctions/PhotoFunctions.php#L286-L289

alex-phillips:
It looks like this may have been an issue downloading EXIFtool and not the code changes?

app/ModelFunctions/PhotoFunctions.php
alex-phillips: @tmp-hallenser I adjusted my code to match the https://github.com/LycheeOrg/Lychee/pull/574

ildyria:
> It looks like this may have been an issue downloading EXIFtool and not the code changes?

Yes. I'm currently doing a PR to fix that one.

PR #597, opened 2020-05-24
modified logic so we can preserve the image EXIF data if sidecar fails

alex-phillips:

PR #598, opened 2020-05-24
need to ignore MimeType, not 'type'

alex-phillips:
My previous PR was unsetting `type`, but it's actually `MimeType` that we need to unset here.


alex-phillips:
Closing this, going to just include this change in PR https://github.com/LycheeOrg/Lychee/pull/597

PR #599, opened 2020-05-24
Added method to get the full path of albums

alex-phillips:
Specifically in the sharing screen, when albums are sometimes named the same (if they are organized by Year and Month), then it is impossible to tell which album you are actually sharing. This adds the ability to get the album's "full path" and sends it down for the sharing settings.

This is fast in my library, but I don't have a large amount of nested albums. This may be inefficient and I'm open to alternatives. I also thought about adding a new column on the database for "full_path" but am unsure of how best to approach it so it is backwards compatible. Can a migration be created to retroactively alter existing records in the database?

### DIFF
--- a/app/Http/Controllers/SharingController.php
+++ b/app/Http/Controllers/SharingController.php
@@ -41,10 +41,17 @@ class SharingController extends Controller
 				->join('albums', 'album_id', 'albums.id')
 				->orderBy('title', 'ASC')
 				->orderBy('username', 'ASC')
-				->get();
+				->get()
+				->each(function (&$s) {
+					$album = Album::find($s->album_id);
+					$s->title = $album->getFullPath();
+				});
 
-			$albums = Album::select(['id', 'title'])->orderBy('title', 'ASC')
-				->get();
+			$albums = Album::select(['id', 'title', 'parent_id'])->orderBy('title', 'ASC')
+				->get()->each(function (&$album) {
+					$album->title = $album->getFullPath();
+				});
+
 			$users = User::select(['id', 'username'])
 				->orderBy('username', 'ASC')->get();
 		} else {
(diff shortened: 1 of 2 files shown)


alex-phillips:
Some screenshots to help explain. This first one is without these code changes, as you can see there are many albums named "01" which is January for every year album I have:
image

This is with the changes, you can easily see the full path of the albums:
image

app/Http/Controllers/SharingController.php
```diff
@@ -41,10 +41,17 @@ public function listSharing()
 				->join('albums', 'album_id', 'albums.id')
 				->orderBy('title', 'ASC')
 				->orderBy('username', 'ASC')
-				->get();
+				->get()
+				->each(function (&$s) {
+					$album = Album::find($s->album_id);
```
ildyria: This Album::find() is triggering me as we already have most of the data required to apply the getFullPath().
We are just missing `parent_id`.
If we also do a select on `parent_id`, I think it should be possible to apply directly getFullPath() on the resulting object.
This would remove all the calls to `Album::find($s->album_id)`.

Idem for the query in line 68.

ildyria (changes requested):
I find this PR really useful.

I think the Query can be optimized as currently for n shared albums, we are going to query 2n times (in addition the going down the tree to build the path).

app/Http/Controllers/SharingController.php
alex-phillips: It doesn't seem like we're getting an `Album` object back from that `DB::table` call, it's a collection of arrays with properties. What we can do is include the `title` and `parent_id` and start building the paths from there, but currently we'd still need an album object to call `getFullPath`. Unless make that method a static and allow it to accept a stdObj with `title` and `parent_id`? Thoughts?

ildyria: > Unless make that method a static and allow it to accept a stdObj with `title` and `parent_id`? Thoughts?

That would be my go to. :|

alex-phillips: @ildyria I pushed a change that makes `getFullPath` a public static and removes the `Album::find` from the sharing controller.

ildyria (approved):

PR #616, opened 2020-06-10
Merging PHP side of things

gardiol:
As known improvements, this code will rotate all images, while it is probably better to regenrate the various sizes from the big one.

Also, this code will not work without imagemagick, this could be improved somehow.


app/Http/Controllers/PhotoEditorController.php
```diff
@@ -0,0 +1,104 @@
...
+					$x_pos = strpos($photo->{$img_type}, 'x');
+					$old_w = substr($photo->{$img_type}, 0, $x_pos);
+					$old_h = substr($photo->{$img_type}, $x_pos + 1);
+					$photo->{$img_type} = $old_h . 'x' . $old_w;
+					$save_photo = true;
+				}
+			} else {
+				$filename = $photo->thumbUrl;
+			}
+			if (($split = strpos($pathType, '2')) !== false) {
```
ildyria: can't you just use `$img_type` here and then on line 70 remove the `strtolower()` ?

app/Http/Controllers/PhotoEditorController.php
```diff
@@ -0,0 +1,104 @@
...
+			return 'false';
+		}
+
+		$request->validate([
+			'photoID' => 'string|required',
+			'direction' => 'integer|required',
+		]);
+
+		$photo = Photo::find($request['photoID']);
+		$direction = $request['direction'];
```
ildyria, line 45: You need to test that direction is either `-1` or `1`, you can actually do that in the `->validate([..])`.

Have a look at this: https://laravel.com/docs/7.x/validation#rule-in Note that it will require you to add a `use ...` at the top to get the `Rule` class in.

app/Http/Controllers/PhotoEditorController.php
```diff
@@ -0,0 +1,104 @@
...
+				}
+			} else {
+				$filename = $photo->thumbUrl;
+			}
+			if (($split = strpos($pathType, '2')) !== false) {
+				$pathType = substr($pathType, 0, $split);
+			}
+			$uploadFolder = Storage::path(strtolower($pathType) . '/');
+			$img_path = $uploadFolder . $photo->url;
+			if (strpos($img_type, '2x') > 0) {
```
ildyria: you are using this test and something similar in line 67. You should just declare a boolean at the beginning of the loop to tell if you are working with a 2x or not. :)

app/Http/Controllers/PhotoEditorController.php
```diff
@@ -0,0 +1,104 @@
...
+				$filename = preg_replace('/^(.*)\.(.*)$/', '\1@2x.\2', $filename);
+			}
+
+			// Rotate the image
+			$img_path = $uploadFolder . $filename;
+			if (file_exists($img_path)) {
```
ildyria, line 100: How does this works in the case of a import by Symlink ?

app/Http/Controllers/PhotoEditorController.php
```diff
@@ -0,0 +1,104 @@
...
+				$image = new \Imagick();
+				$image->readImage($img_path);
+
+				if ($direction == 1) {
+					$image->rotateImage(new \ImagickPixel(), 90);
+				} elseif ($direction == -1) {
```
ildyria: If you test against `-1`, `1` earlier then the `elseif` just becomes a `else` :)

app/Http/Controllers/PhotoEditorController.php
```diff
@@ -0,0 +1,104 @@
...
+	 *
+	 * @return tring
+	 */
+	public function rotate(Request $request)
+	{
+		// Safety check...
+		if (!Configs::get_value('editor_enabled', '0')) {
+			return 'false';
+		}
+
```
ildyria, line 32: As another safety check, verify that imagick is enabled too.

app/Locale/English.php
```diff
@@ -414,8 +414,11 @@ public static function get_locale()
 			'PHOTO_THUMB_HIDPI' => 'Square thumb HiDPI',
 			'PHOTO_LIVE_VIDEO' => 'Video part of live-photo',
 			'PHOTO_VIEW' => 'Lychee Photo View:',
+
+			'PHOTO_EDIT_ROTATECWISE' => 'Rotate clockwise',
+			'PHOTO_EDIT_ROTATECCWISE' => 'Rotate counter-clockwise',
 		];
```
ildyria, line 420: You need to add these 3 lines to all the other languages files. We do not ask you to translate it, just to put it there in plain English, someone will take care of the translation later.

ildyria (changes requested):
Nice first PR. :+1: 

Please find some changes requested. :smiley:

app/Locale/English.php
ildyria: (This is one of the reason my the CI will fail).

app/Http/Controllers/PhotoEditorController.php
gardiol: Yes, leftovers from some copied code.

app/Http/Controllers/PhotoEditorController.php
gardiol: Yes, you are right, more checks are needed.

app/Http/Controllers/PhotoEditorController.php
gardiol: This can be improved overall, yes

app/Http/Controllers/PhotoEditorController.php
gardiol: I have no idea what you are taling about?

app/Http/Controllers/PhotoEditorController.php
gardiol: of course

app/Http/Controllers/PhotoEditorController.php
gardiol: I will, can you point me to how to do it?

app/Locale/English.php
gardiol: ok!

app/Http/Controllers/PhotoEditorController.php
ildyria: There are two (three) ways to import photos:
- you upload them
- you import them from some place on the server. Depending of the server settings, it will either move them to the big folder OR it will create a symbolic link to the said photo.

It the later case, this may be a problem. There you basically need to decide for the behavior: 
  - Refuse the rotation if the image is a symbolic link (is probably the safest approach)
  - or rotate the original (less safe, will most likely fail if you only have read access to the folder).

As a result, it might be interesting to test if the `'big/' . $filename` is a symbolic link or an actual image. If it is a link, return 'false' and do nothing.

This will be handy: https://www.php.net/manual/en/function.is-link.php

app/Http/Controllers/PhotoEditorController.php
ildyria: https://github.com/LycheeOrg/Lychee/blob/master/app/Configs.php#L234 :)

app/Http/Controllers/PhotoEditorController.php
gardiol: Ok, added the check

app/Http/Controllers/PhotoEditorController.php
gardiol: done

app/Http/Controllers/PhotoEditorController.php
```diff
@@ -45,13 +49,29 @@ public function rotate(Request $request)
 			return false;
```
ildyria: false needs to be a string.
```suggestion
			return 'false';
```

gardiol:
I committed on my fork all the changes... What now?

app/Http/Controllers/PhotoEditorController.php
```diff
@@ -45,13 +49,29 @@ public function rotate(Request $request)
...
 		}
 
+		// direction is valid?
+		if (($direction != 1) && ($direction != -1)) {
+			Logs::error(__METHOD__, __LINE__, 'Direction must be 1 or -1');
+
+			return false;
```
ildyria: false needs to be a string.
```suggestion
			return 'false';
```

app/Http/Controllers/PhotoEditorController.php
```diff
@@ -45,13 +49,29 @@ public function rotate(Request $request)
...
+		}
+
+		// Abort on symlinks to avoid messing with originals linked
+		if (is_link(Storage::path('big/') . $photo->url)) {
+			Logs::error(__METHOD__, __LINE__, 'Synlinked images cannot be rotated');
+
+			return false;
```
ildyria: false needs to be a string.
```suggestion
			return 'false';
```

ildyria (changes requested):
Small fixes required. :+1:

gardiol:
Fixed and committed

ildyria (approved):

PR #644, opened 2020-07-01
Add Cloudron Installation Method

nebulade:
We have packaged Lychee since a long time and keept it up-to-date automatically for our users.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -46,7 +46,7 @@ Alternatively, an official Docker image can be found at [LycheeOrg/Lychee-Docker
 
 Cloudron is a complete solution for running apps on your server and keeping them up-to-date and secure.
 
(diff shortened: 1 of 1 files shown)


d7415:
I'm not sure about this. If it were to be included, I think it needs to be very clear that it's 3rd party and not supported by LycheeOrg. Then again, for that reason I'd lean against mentioning it in the first place. Happy to see what others think though.

nebulade:
I understand the concern. If it helps, we kept the package up-to-date since a long time now and also usually act as the first level support to triage issues with the app.

Would you actually prefer to put the packaging metadata into a subfolder of the repo itself?

ildyria:
Added to our documentation: https://lycheeorg.github.io/docs/

Source code here: https://github.com/LycheeOrg/LycheeOrg.github.io

nebulade:
Thanks, this looks great!

PR #656, opened 2020-07-09
Support fire tv

tmp-hallenser:
This PR adds basic support to view Lychee on a FireTV as a WebApp.

Changes:
* We need to detect if the client is a TV as the page delivered needs to change.
* UI is simplified and button are disabled using a couple of switches
* Navigation on the screen is done by setting 'tabindex'; We add a parameter 'data-tabindex' to be able to enable/disable tabbing through certain elements (change for blades)
* No all navigation elements can be accessed as some are not needed (e.g. download button etc.)


vorlif:
Hey, can you briefly explain how I can link/open lychee as a web app under fire tv?

d7415:
> Hey, can you briefly explain how I can link/open lychee as a web app under fire tv?

It sounds like @tmp-hallenser might have some experience, but at a glance, https://developer.amazon.com/docs/fire-tv/getting-started-with-web-apps.html

tmp-hallenser:
Install the Web App Tester (https://www.amazon.de/Amazon-Digital-Services-Inc-Tester/dp/B00DZ3I1W8) and enter your Lychee URL. It's so far just a first shot leaving some room for improvements.

vorlif:
Thanks, I will try it.

Issue #630, opened 2020-06-22
[Enhancement] Optimise uploaded images


mxmehl:
It would be useful to have tools like optipng run over newly uploaded images, or offer this as an option for all existing images in the database. In our tests at Free Software Foundation Europe, we've experienced an average space saving of 8-15% with the same quality.

The reason why this should be in Lychee directly is that the checksum of the pictures are in the database, and when running these tools on the files directly, the sums would not match any more.

Downside: the optimisiation might take some seconds depending on the file type and size. That is why this option should be called upon explicit request by the admin, and/or configurably during upload/import.

d7415:
I can see some benefit, but I'm mostly of the opinion that processing should be done before a photo hits Lychee.

tmp-hallenser:
The idea is interesting. Let me share my view:
We save 2 different kinds of image files within Lychee: (1) The original file and (2) scaled (and cropped) versions for lightbox and thumbnail.
(1) I'd vote against manipulating the original file since I'd like to retrieve the original file when downloading (file is only delivered to browser if user wants to download the original file)
(2) These files are delivered to the browser when viewing the album/photo. If some package delivers smaller files at same size, it's worth a thought.

@mxmehl Do you have benchmarks comparing imagick with optipng (and jpg equivalent tools)?

mxmehl:
I agree that the original photo should not be touched.

Regarding benchmarks, just an example:

URL: https://pics.fsfe.org/uploads/medium/69a7e6a2a2e36b14fced8e1ab3044b66.png

```
$ optipng 69a7e6a2a2e36b14fced8e1ab3044b66.png 
** Processing: 69a7e6a2a2e36b14fced8e1ab3044b66.png
800x432 pixels, 4x8 bits/pixel, RGB+alpha
Reducing image to 3x8 bits/pixel, RGB
Input IDAT size = 142128 bytes
Input file size = 142233 bytes

Trying:
  zc = 9  zm = 8  zs = 0  f = 5		IDAT size = 126969
  zc = 9  zm = 8  zs = 1  f = 5		IDAT size = 122776
                               
Selecting parameters:
  zc = 9  zm = 8  zs = 1  f = 5		IDAT size = 122776

Output IDAT size = 122776 bytes (19352 bytes decrease)
Output file size = 122833 bytes (19400 bytes = 13.64% decrease)
```

There are many other tools which work for JPG and PNG alike, e.g. https://github.com/siiptuo/pio

[...]

PR #659, opened 2020-07-11
Add functionality to apply lossless image optimization

tmp-hallenser:
Implement lossless image optimization (#630) 
Uploaded original photo is not touched, only medium, small and thumbnails.

Requirements: jpegoptim, optipng, pngquant, gifsicle, webp


ildyria (commented):
LGTM.

Do we want to add warnings if the executables are not available ? 
Or in the diagnostic page, if optimized is set to 1 but the softwares are not there?

config/image-optimizer.php
```diff
@@ -0,0 +1,64 @@
...
+    /*
+     * The maximum time in seconds each optimizer is allowed to run separately.
+     */
+    'timeout' => 10,
+
+    /*
+     * If set to `true` all output of the optimizer binaries will be appended to the default log.
+     * You can also set this to a class that implements `Psr\Log\LoggerInterface`.
+     */
+    'log_optimizer_activity' => false,
```
tmp-hallenser: @ildyria Can we plug the Lychee Logging into here?

tmp-hallenser:
> LGTM.
> 
> Do we want to add warnings if the executables are not available ?
> Or in the diagnostic page, if optimized is set to 1 but the softwares are not there?

Good idea. The library already provides some logging, we need to specify a logging interface (type `Psr\Log\LoggerInterface`) -> can we use the Lychee Logging?

config/image-optimizer.php
ildyria: I don't think we have something which implements Psr\Log\LoggerInterface but that can probably be easilly done. :)

ildyria:
What I am also worried, is, if you enable the optimizer but the software are not here, does it raises exceptions or does it just append to the logs ?

ildyria:
@tmp-hallenser can you double check ?

tmp-hallenser:
works, thanks. Nice job with the logger.

Issue #622, opened 2020-06-17
[Enhancement] Add more licenses


mxmehl:
We started to use Lychee but noticed that only a handful of licenses are supported. For instance, some of our existing (and surely future) images are under CC-BY-2.0. We cannot label it as CC-BY-4.0 single-handedly though.

Is there a possibility to add more licenses? I do not expect Lychee to support [hundreds](https://spdx.org/licenses/), but perhaps one could add them in the admin panel manually?

Thanks for making this software even better!

ildyria:
I don't have problem for this to be added. However most of us won't have the time to work on that one.

PR #660, opened 2020-07-11
Additional licences

tmp-hallenser:
Add the different versions of CC licenses (#622).


ildyria:
Don't forget the default licenses in the config. ;) 
Also you have default licenses for albums and you have the "accepted values" in the diagnostic check.

tmp-hallenser:
> Don't forget the default licenses in the config. ;)
> Also you have default licenses for albums and you have the "accepted values" in the diagnostic check.

Perhaps it's already too late, but where did I miss something?

ildyria:
https://github.com/LycheeOrg/Lychee/blob/master/app/Configs.php#L60
I thought you may want to make a "license" specific check. :)

Well I just had a look at : https://github.com/LycheeOrg/Lychee/blob/master/database/migrations/2019_09_28_171753_config_fix.php#L233
And I was expecting a check against licenses possibilities, but it is just "string required".

So indeed nothing needed more. :)

tmp-hallenser:
> https://github.com/LycheeOrg/Lychee/blob/master/app/Configs.php#L60
> I thought you may want to make a "license" specific check. :)
> 
> Well I just had a look at : https://github.com/LycheeOrg/Lychee/blob/master/database/migrations/2019_09_28_171753_config_fix.php#L233
> And I was expecting a check against licenses possibilities, but it is just "string required".
> 
> So indeed nothing needed more. :)

please review.

ildyria (approved):

PR #676, opened 2020-07-25
Mark update_min_max_takestamp as static.

Diablofan:
I encountered an issue while initially trying to migrate my Lychee v3 install to v4 that complained that `update_min_max_takestamp` should not be called statically. Looks like it actually is safe to call statically, it just wasn't marked `static`. Adding this made the error go away when I reran the migrations from scratch.


ildyria:
Thank you ! I probably missed that one.

PR #680, opened 2020-07-30
Distinguish UTC vs local video takestamps

### LINKED ISSUE #675: Wrong takestamp on some videos
i cant get the correct takestamps on videos other than quicktime. I have tried to change timezone at app.php and also at mysql and restart after that but nothing changes, i guess i'm missing a point. Also i have tried to run "php artisan config:cache" after timezone change.

As i detect quicktime videos have correct time zones but the others are on zulu timezone i'm on +3 so it has to be 12:49.19 in this example:
`object(PHPExif\Exif)#1884 (2) {
  ["data":protected]=>
  array(10) {
    ["width"]=>
    int(1920)
    ["height"]=>
    int(1080)
    ["framerate"]=>
    float(29.982983550766)
    ["duration"]=>
    string(9) "29.383333"
    ["creationdate"]=>
    object(DateTime)#1885 (3) {
      ["date"]=>
      string(26) "2017-08-20 16:34:45.000000"
      ["timezone_type"]=>
      int(1)
      ["timezone"]=>
      string(6) "+03:00"
    }
    ["make"]=>
    string(5) "Apple"
    ["camera"]=>
    string(8) "iPhone 6"
    ["FileName"]=>
    string(12) "IMG_7763.MOV"
    ["FileSize"]=>
    string(8) "56472993"
    ["MimeType"]=>
    string(15) "video/quicktime"

`
`object(PHPExif\Exif)#1883 (2) {
  ["data":protected]=>
  array(11) {
    ["width"]=>
    int(1920)
    ["height"]=>
    int(1080)
    ["framerate"]=>
    float(29.879816737124)
    ["duration"]=>
    string(8) "4.117000"
    ["creationdate"]=>
    object(DateTime)#1882 (3) {
      ["date"]=>
      string(26) "2016-07-24 09:49:19.000000"
      ["timezone_type"]=>
      int(2)
      ["timezone"]=>
      string(1) "Z"
    }
    ["latitude"]=>
    string(8) "+39.9837"
    ["longitude"]=>
    string(9) "+032.8752"
    ["gps"]=>
    string(18) "+39.9837,+032.8752"
    ["FileName"]=>
    string(23) "VID_20160724_124914.mp4"
    ["FileSize"]=>
    string(8) "10722475"
    ["MimeType"]=>
    string(9) "video/mp4"
  }
`

kamil4:
Fixes #675.

This adds a heuristic to determine whether the takestamp embedded in a video file is in UTC or in local time. The heuristic is based on the filename extension. There is a new config variable, `local_takestamp_video_formats`, that provides a list the extensions for files that should be treated as having a takestamp in the local time zone. It defaults to AVI and MOV, as those are the two most commonly used in digital cameras, which typically use the local time zone. For files that don't match (such as MP4 files from Android phones) the takestamp is considered to be in UTC and is converted at import time to the local time zone of the server.

While working on this I noticed several bugs that this PR also fixes. In particular:
* MPG files were treated as image files because at least on my system that's what the MIME says (`image/x-tga`)?!

* `file_type` method in `PhotoFunctions` was fundamentally broken; it would return `photo` for video files. Fixing its twisted logic required a complete refactoring.

* With that fixed, I was able to replace all the ad-hoc tests for video files based on MIME, which didn't work for MPG files (see above).

* WMV videos misbehaved when using the Exiftool metadata extractor because of conflicting MIME data (Exiftool identifies them as `video/x-ms-wmv`). There were many other issues when using Exiftool instead of FFmpeg for video files' metadata, which were addressed in https://github.com/LycheeOrg/php-exif/pull/22 (now merged).

* If FFmpeg is disabled in the config, we shouldn't try to use it to for anything, including extracting a video frame for the thumb. I fixed that particular use case but it appears that there may be more left in the new live photos support? I'm not familiar with that feature so I didn't touch it. @tmp-hallenser?

### DIFF
--- a/app/Metadata/Extractor.php
+++ b/app/Metadata/Extractor.php
@@ -64,11 +64,11 @@ class Extractor
 	 * Extracts metadata from an image file.
 	 *
 	 * @param string $filename
-	 * @param  string mime type
+	 * @param  string file kind
 	 *
 	 * @return array
 	 */
-	public function extract(string $filename, string $type): array
+	public function extract(string $filename, string $kind): array
 	{
 		$reader = null;
 
@@ -82,7 +82,7 @@ class Extractor
 			$is_raw = true;
 		}
 
-		if (strpos($type, 'video') !== 0) {
+		if ($kind !== 'video') {
 			// It's a photo
 			if (Configs::hasExiftool()) {
 				// reader with Exiftool adapter
@@ -164,7 +164,6 @@ class Extractor
 		$metadata['make'] = ($exif->getMake() !== false) ? $exif->getMake() : '';
 		$metadata['model'] = ($exif->getCamera() !== false) ? $exif->getCamera() : '';
 		$metadata['shutter'] = ($exif->getExposure() !== false) ? $exif->getExposure() : '';
-		$metadata['takestamp'] = ($exif->getCreationDate() !== false) ? $exif->getCreationDate()->format('Y-m-d H:i:s') : null;
 		$metadata['lens'] = ($exif->getLens() !== false) ? $exif->getLens() : '';
 		$metadata['tags'] = ($exif->getKeywords() !== false) ? (is_array($exif->getKeywords()) ? implode(',', $exif->getKeywords()) : $exif->getKeywords()) : '';
 		$metadata['latitude'] = ($exif->getLatitude() !== false) ? $exif->getLatitude() : null;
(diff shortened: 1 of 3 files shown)


app/Metadata/Extractor.php
```diff
@@ -237,7 +272,7 @@ public function extract(string $filename, string $type): array
 			$metadata['position'] = implode(', ', $fields);
 		}
 
-		if (strpos($type, 'video') !== 0) {
```
ildyria, line 240: are you sure about that one ?

app/Metadata/Extractor.php
```diff
@@ -82,7 +82,7 @@ public function extract(string $filename, string $type): array
 			$is_raw = true;
 		}
 
-		if (strpos($type, 'video') !== 0) {
```
ildyria, line 85: are you sure about that one ?

kamil4: Pretty sure... It's because of those crazy MPG files being identified by `mime_content_type` as `image/x-tga`. So I added `image/x-tga` to `$validVideoTypes`, fixed `file_type` to do what it was actually supposed to (so that it actually returns `'video'` for video files), and then replaced all these ad-hoc tests by the more robust checks based on the value returned by `file_type`.

app/Metadata/Extractor.php
kamil4: See above...

ildyria: yeap I'm reading I'm reading. :) I just quickly commented. :p

app/Metadata/Extractor.php
```diff
@@ -175,16 +174,52 @@ public function extract(string $filename, string $type): array
...
+			//
+			// Other possible approaches would include deriving local time from
+			// the file name or from other objects in the same album, as well
+			// as extracting the time zone from the location data if present.
+			if ($kind === 'video') {
+				$locals = strtolower(Configs::get_value('local_takestamp_video_formats', ''));
+				if (!in_array(strtolower($extension), explode('|', $locals), true)) {
+					// This is a video format where we expect the takestamp
+					// to be provided in UTC .
+					if ($takestamp->getTimezone()->getName() === date_default_timezone_get()) {
```
tmp-hallenser, line 194: Can you help me on this one? I was assuming that the time zone is set to 'Zulu' if it's not set in the exifdata.

tmp-hallenser: Without having tested it, I think it would screw up my videos shot with my iPhone in my time zone, wouldn't it?
The iPhone shoots mov files and adds the correct time zone (which is equal to my server time zone).

kamil4: This is indeed a tricky part of the code that involves certain subtleties that I didn't discuss in the comments.

Depending on the file type and the metadata extractor used, the timestamp they output either will have a `Z` at the end or not. There is no clear-cut rule; e.g., FFprobe adds a `Z` for mov and mp4 files but Exiftool does not, though both do it for webm files and neither does it for avi files.

So when the extracted timestamp string gets passed to `DateTime`, it will either get interpreted as UTC time (if it had a `Z` at the end) or as a local time (if it didn't). It's that second case that the code is dealing with in the line you questioned.

Now the second question you asked is what if the video file _had_ the time zone embedded, as mov files from iPhones do? If the time zone on the server is the same as in your phone, wouldn't that wrongly trigger the above condition?

I worried about that possibility as well but testing shows that it works fine. That's because `date_default_timezone_get` returns strings such as `UTC` or, in my case, `America/Chicago`, whereas `getTimezone()->getName()` for iPhone files returns `-05:00`. So even though they indicate the same time zone, the strings are different, the comparison fails, and we don't do anything.

Now, given the length of this explanation, I think you will understand why I didn't put this in the comments :smile:.

tmp-hallenser: omg. If that's the case, then I'm fine

ildyria (approved):

PR #683, opened 2020-08-03
skip live video files from cleanup

alex-phillips:
I noticed that after running `lychee:ghostbuster` command, my live images would no longer play and their download link would 404. Ensure that we skip them if they exist in the db as a valid file for live photos.


tmp-hallenser:
LGTM

ildyria (approved):

PR #691, opened 2020-08-18
Fix password protection of albums

kamil4:
So we were leaking data behind password-protected albums like a sieve. It was leaking out through album thumbs, public search, public smart albums, and probably also the frame.

The most important fix is to `AlbumsFunctions->getPublicAlbumsId()`, which is used all over the place. It now, by default, does not return the IDs of password-protected albums, since we can't access them anyway. That closed most of the leaks.

Fixing the thumb leaking required additional checks in `AlbumFunctions->get_thumbs()` and `AlbumsFunctions->prepare_albums`. I also made use of `content_accessible` which was set but wasn't used.

`SearchController` then needed to be fixed so that it _would_ include password-protected albums (but not their contents). This required additions to `AlbumsFunctions->getPublicAlbumsId()` and `AlbumsFunctions->get_children()`.

Finally, I removed the pulling of unsorted public photos into public smart albums. In my opinion, it's a bit dangerous, as unsorted photos are effectively hidden so they should be accessible only to those who know the URL...

### DIFF
--- a/app/ModelFunctions/AlbumsFunctions.php
+++ b/app/ModelFunctions/AlbumsFunctions.php
@@ -72,9 +72,11 @@ class AlbumsFunctions
 				$album_array['owner'] = $albums[$key]->owner->username;
 			}
 
-			$thumbs = $this->albumFunctions->get_thumbs($albums[$key], $children[$key]);
-			$this->albumFunctions->set_thumbs($album_array, $thumbs);
-			$this->albumFunctions->set_thumbs_children($album_array['albums'], $thumbs[1]);
+			if ($this->readAccessFunctions->album($albums[$key]) === 1) {
+				$thumbs = $this->albumFunctions->get_thumbs($albums[$key], $children[$key]);
+				$this->albumFunctions->set_thumbs($album_array, $thumbs);
+				$this->albumFunctions->set_thumbs_children($album_array['albums'], $thumbs[1]);
+			}
 
 			// Add to return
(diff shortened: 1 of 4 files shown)


ildyria (commented):
Looks pretty good to me. I will need to test with a large DB it to be extra sure. :)

ildyria (approved):

app/ModelFunctions/AlbumsFunctions.php
```diff
@@ -72,9 +72,11 @@ public function prepare_albums(BaseCollection $albums, BaseCollection $children)
 				$album_array['owner'] = $albums[$key]->owner->username;
 			}
 
-			$thumbs = $this->albumFunctions->get_thumbs($albums[$key], $children[$key]);
-			$this->albumFunctions->set_thumbs($album_array, $thumbs);
-			$this->albumFunctions->set_thumbs_children($album_array['albums'], $thumbs[1]);
+			if ($this->readAccessFunctions->album($albums[$key]) === 1) {
```
pchampio, line 75: I actually liked the thumb leakage.
Removing this if check enables me to have both a thumb and an password protected album.

I know my demand is quite special.
Can we add an 'advanced setting' that allows for this use-case?

Thanks!

kamil4: LOL. If ever https://xkcd.com/1172/ was more true :wink:...

PR #704, opened 2020-08-27
Feature/#48 album by tags

Aleshus:
Added support for tag albums


ildyria:
@Aleshus thank you for your PR. Can you do `make formatting` at the base of the repository so that it unifies the style. :+1:

Aleshus:
> @Aleshus thank you for your PR. Can you do `make formatting` at the base of the repository so that it unifies the style. 👍

Ok, done, thanks..

app/ModelFunctions/AlbumsFunctions.php
```diff
@@ -164,17 +168,17 @@ private function flatMap_id(BaseCollection $subAlbums): BaseCollection
 	 */
 	public function getPublicAlbumsId($toplevel = null, $children = null): BaseCollection
 	{
+		$albumIDs = new BaseCollection();
```
ildyria, line 175: nicely spotted.

ildyria (commented):
Please double check my changes. :)

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -476,4 +476,9 @@ public function get_sub_albums(Album $album, $includePassProtected = false): Bas
 			return $e->isEmpty();
 		});
 	}
+
+	public static function is_tag_album(Album $album): bool
```
ildyria, line 484: Not sure if this needs to be static anymore.

app/Http/Controllers/AlbumController.php
```diff
@@ -87,6 +91,28 @@ public function add(Request $request)
...
+	 */
+	public function addByTags(Request $request)
+	{
+		$request->validate([
+			'title' => 'string|required|max:100',
+			'tags' => 'string',
+		]);
+
+		$album = new Album();
+		$album->id = Helpers::generateID();
```
kamil4: You don't need these two lines. `AlbumFunctions.createTagAlbum` takes care of that.

app/Http/Controllers/AlbumController.php
```diff
@@ -213,6 +202,27 @@ public function getPositionData(Request $request)
...
 	}
 
+	/**
+	 * @param string $albumID
+	 *
+	 * @return Album|SmartAlbum
+	 */
+	public function getAlbum(string $albumID): Album
+	{
+		// TODO: improve here.
```
kamil4: @ildyria Cryptic as always :wink:.

app/Http/Controllers/AlbumController.php
```diff
@@ -371,6 +381,39 @@ public function setDescription(Request $request)
...
+	 *
+	 * @param Request $request
+	 *
+	 * @return bool|string
+	 */
+	public function setShowTags(Request $request)
+	{
+		$request->validate([
+			'albumID' => 'integer|required',
+			'show_tags' => 'string|required|max:1000|min:1',
```
kamil4, line 389: Is there a reason to have a max? While `1000` seems like a lot, I wonder if an arbitrary limit is necessary at all?

app/ModelFunctions/AlbumFunctions.php
```diff
@@ -66,28 +66,78 @@ public function __construct(SessionFunctions $sessionFunctions, ReadAccessFuncti
...
+
+		return $this->store_album($album);
 	}
 
 	/**
 	 * Create a new album from a title and optional parent_id.
 	 *
 	 * @param string $title
 	 * @param int    $parent_id
+	 * @param array  $tags
```
kamil4: Change of plans here? I don't see that argument being added to the actual function below...

app/SmartAlbums/TagAlbum.php
```diff
@@ -0,0 +1,34 @@
...
+class TagAlbum extends SmartAlbum
+{
+	public function get_title()
+	{
+		return $this->title;
+	}
+
+	public function get_photos(): Builder
+	{
+		// /!\ this may reveal private picture.
```
kamil4: @ildyria Are you sure? I don't see the code doing anything different from Starred or Recent. In particular, `filter` is being applied at the end. We should test, of course...

kamil4 (changes requested):
Github's change line count made me apprehensive about reviewing this; thankfully, most of it was the front end and the languages...

kamil4 (approved):
I pushed additional commits that:
* revert the `is_smart_album` changes. We need it to trigger only for the base smart albums in the upload check middleware; otherwise everybody could modify/delete the tagged smart albums.
* refactor album download code. It didn't work for tagged smart album and was using obsolete functionality for other albums anyway so it was due for an overhaul.
* I verified that @ildyria's concern about `TagAlbum\get_photos` leaking private pictures was unfounded.

This should be good to go now!

PR #727, opened 2020-09-15
Check the provided password against all albums

### LINKED ISSUE #724: [Enhancement] - heirarchical password check for password protected albums
At present - if you use password protection together with sub-albums then you can either:

Set the password on just the parent - in this case - you can see the thumbs of the sub-albums when you are showing the parent and can navigate in - but - if you head to the url of a sub-album directly then its open

Set the password on the parent and all sub-albums - in this case - the sub-albums are not open via direct link - but - if you start at the top and work down - you can only see padlock icons and for each new sub-album you have to enter the password again.

It would be nice to be able to check up the tree so that a sub-album was protected by its parent's (or higher in the tree) settings.

kamil4:
Fixes #724

### DIFF
--- a/app/Http/Controllers/AlbumController.php
+++ b/app/Http/Controllers/AlbumController.php
@@ -252,7 +252,20 @@ class AlbumController extends Controller
 						return 'true';
 					}
 					if ($album->password == '' || Hash::check($request['password'], $album->password)) {
-						$this->sessionFunctions->add_visible_album($album->id);
+						// We add all the albums that the password unlocks so
+						// that the user is not repeatedly asked to enter the
+						// password as they browse through the hierarchy.  This
+						// should be safe as the list of such albums is not
+						// exposed to the user and is considered as the last
+						// access check criteria.
+						$albums = Album::get();
+						$albumIDs = [];
+						foreach ($albums as $album) {
+							if ($album->password !== '' && Hash::check($request['password'], $album->password)) {
+								$albumIDs[] = $album->id;
+							}
+						}
+						$this->sessionFunctions->add_visible_albums($albumIDs);
 
 						return 'true';
 					}
--- a/app/Http/Controllers/AlbumController.php
+++ b/app/Http/Controllers/AlbumController.php
@@ -251,7 +251,7 @@ class AlbumController extends Controller
 					if ($this->sessionFunctions->has_visible_album($album->id)) {
 						return 'true';
 					}
-					if ($album->password == '' || Hash::check($request['password'], $album->password)) {
+					if (Hash::check($request['password'], $album->password)) {
 						// We add all the albums that the password unlocks so
 						// that the user is not repeatedly asked to enter the
 						// password as they browse through the hierarchy.  This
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/AlbumController.php
```diff
@@ -251,8 +251,21 @@ public function getPublic(Request $request)
 					if ($this->sessionFunctions->has_visible_album($album->id)) {
 						return 'true';
 					}
-					if ($album->password == '' || Hash::check($request['password'], $album->password)) {
```
d7415, line 254: I take it the `$album->password == '' || ` part is handled elsewhere so not required here?

kamil4: Yeah, exactly six lines earlier...

d7415 (approved):
Makes sense to me.

app/Http/Controllers/AlbumController.php
```diff
@@ -251,8 +251,21 @@ public function getPublic(Request $request)
...
-						$this->sessionFunctions->add_visible_album($album->id);
+					if (Hash::check($request['password'], $album->password)) {
+						// We add all the albums that the password unlocks so
+						// that the user is not repeatedly asked to enter the
+						// password as they browse through the hierarchy.  This
+						// should be safe as the list of such albums is not
+						// exposed to the user and is considered as the last
+						// access check criteria.
+						$albums = Album::get();
```
ildyria: This should be refined with a `where('password','!=','')`

app/Http/Controllers/AlbumController.php
```diff
@@ -251,8 +251,21 @@ public function getPublic(Request $request)
...
+						$albumIDs = [];
+						foreach ($albums as $album) {
+							if ($album->password !== '' && Hash::check($request['password'], $album->password)) {
```
ildyria: `$album->password !== ''` is now useless once the where condition is applied to the Query.

ildyria (changes requested):
some refinement to avoid looping over ALL albums.

app/Http/Controllers/AlbumController.php
kamil4: Good point, I don't know what I was thinking when I wrote this...

kamil4:
Should be good to go now...

app/Http/Controllers/AlbumController.php
```diff
@@ -251,8 +251,21 @@ public function getPublic(Request $request)
...
-					if ($album->password == '' || Hash::check($request['password'], $album->password)) {
-						$this->sessionFunctions->add_visible_album($album->id);
+					if (Hash::check($request['password'], $album->password)) {
+						// We add all the albums that the password unlocks so
+						// that the user is not repeatedly asked to enter the
+						// password as they browse through the hierarchy.  This
+						// should be safe as the list of such albums is not
+						// exposed to the user and is considered as the last
+						// access check criteria.
+						$albums = Album::whereNotNull('password')->get();
```
ildyria:
```suggestion
						$albums = Album::whereNotNull('password')->where('password', '!=', '')->get();
```

ildyria: just to be sure 😄

PR #745, opened 2020-10-01
Update album takestamps

### LINKED ISSUE #738: [Enhancement] Sort order of albums: Ignore images without timestamps
Hello,

I sort my albums by `max_takestamp DESC`. Unfortunately I have a few albums with pictures that have no timestamp. These albums are always displayed at the first position. It would be nice if the images without timestamps were ignored in the sorting order.

kamil4:
Fixes #738.

We did not update the album min/max takestamps after a successful extraction.

I ended up rewriting most of its logic since it made no sense to me...

### DIFF
--- a/app/Console/Commands/Takedate.php
+++ b/app/Console/Commands/Takedate.php
@@ -4,7 +4,6 @@ namespace App\Console\Commands;
 
 use App\Metadata\Extractor;
 use App\ModelFunctions\AlbumActions\UpdateTakestamps as AlbumUpdate;
-use App\ModelFunctions\PhotoFunctions;
 use App\Models\Photo;
 use Illuminate\Console\Command;
 use Storage;
@@ -26,11 +25,6 @@ class Takedate extends Command
 	protected $description = 'Make sure takedate is correct';
 
 	/**
-	 * @var PhotoFunctions
-	 */
-	private $photoFunctions;
-
-	/**
 	 * @var Extractor
 	 */
 	private $metadataExtractor;
@@ -38,14 +32,12 @@ class Takedate extends Command
 	/**
 	 * Create a new command instance.
 	 *
-	 * @param PhotoFunctions $photoFunctions
-	 * @param Extractor      $metadataExtractor
+	 * @param Extractor $metadataExtractor
 	 */
-	public function __construct(PhotoFunctions $photoFunctions, Extractor $metadataExtractor)
+	public function __construct(Extractor $metadataExtractor)
 	{
 		parent::__construct();
 
-		$this->photoFunctions = $photoFunctions;
 		$this->metadataExtractor = $metadataExtractor;
 	}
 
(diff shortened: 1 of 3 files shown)


ildyria (approved):
LGTM (not tested)

app/Console/Commands/Takedate.php
```diff
@@ -73,13 +73,14 @@ public function handle()
 			$url = Storage::path('big/' . $photo->url);
 			if (file_exists($url)) {
 				$info = $this->metadataExtractor->extract($url, $photo->type);
-				if ($photo->takestamp == '') {
+				if ($info['takestamp'] != null) {
 					$photo->takestamp = $info['takestamp'];
-				}
-				if ($photo->save()) {
-					$this->line($i . ': Takestamp updated for ' . $photo->title);
+					if ($photo->save()) {
```
d7415, line 70: Do we need handling if this fails?

app/Console/Commands/Takedate.php
```diff
@@ -73,13 +73,14 @@ public function handle()
...
+						$this->line($i . ': Takestamp updated for ' . $photo->title);
+					}
+					AlbumUpdate::update_takestamps($photo->album, [$photo->takestamp], true);
```
d7415: Would this be better before the `}`, so that it only updates if the photo update was successful?

d7415 (commented):
Seems to make sense. A couple of queries.

kamil4:
Thanks for your comments, @d7415. They inspired me to have another critical look at this part of the code and prompted a more thorough rewrite of not just `Takedate.php` but also `ExifLens.php` and `VideoData.php`. All three now feature a more sensible SQL query and updated diagnostics which makes it clear if something was actually updated or not.

@ildyria, @d7415, the latest commit significantly expands the scope of this PR so would you mind making another review round?

app/Console/Commands/VideoData.php
```diff
@@ -98,41 +97,47 @@ public function handle()
...
 					$photo->focal = $info['focal'];
+					$updated = true;
 				}
-				if ($info['aperture'] !== '') {
+				if ($photo->aperture == '' && $info['aperture'] !== '') {
 					$photo->aperture = $info['aperture'];
+					$updated = true;
 				}
-				if ($info['takestamp'] !== null) {
-					$photo->takestamp = $info['takestamp'];
```
d7415, line 115: Presumably this is handled elsewhere so was redundant?

app/Console/Commands/VideoData.php
```diff
@@ -67,7 +66,7 @@ public function handle()
 			)
 		);
 
-		$photos = Photo::where('type', 'like', 'video/%')
+		$photos = Photo::whereIn('type', $this->photoFunctions->getValidVideoTypes())
```
d7415, line 69: Less permissive, but IIRC more in line with how we handle this elsewhere, so makes sense.

d7415 (approved):
Makes sense to me (untested, as usual)

app/Console/Commands/VideoData.php
kamil4: We have a separate `lychee:takedate` command specifically for that. Removing it from here makes it more consistent with `lychee:exif_lens` which doesn't do it.

app/Console/Commands/VideoData.php
kamil4: Actually, check: https://github.com/LycheeOrg/Lychee/blob/2b0664a09dea2ca9533c1879bffad7c97b36c519/app/ModelFunctions/PhotoFunctions.php#L51-L62
As you can see, one video type (MPG) is sometimes recognized as `image/x-tga`, so this change actually makes it more inclusive...

ildyria (approved):
LGTM

ildyria (approved):

PR #750, opened 2020-10-06
Add links to first section

chrissearle:
Added links to the top (help me I'm lost) section - these are the links I keep having to go lookup from lower down in the file or via the org repo list. Just a suggestion :)

PR #752, opened 2020-10-07
Add some more shields. Remove I'm Lost section. Make logo clickable.

d7415:
Possible alternative to #750.

Thoughts?


d7415:
It may be easier to view [here](https://github.com/LycheeOrg/Lychee/blob/readmelinks/readme.md).

PR #789, opened 2020-11-11
Update RedirectController.php

pchampio:
If a password is pass as a get param to `/r/{album}` / `/r/{album}/{photo}`. Unlocks all the albums that the password unlocks.

Implemented in this manner in order for:
`https://domain.com/r/recent?password=my_password` to work!
companion lychee_front PR: https://github.com/LycheeOrg/Lychee-front/pull/229

### DIFF
--- a/app/Http/Controllers/RedirectController.php
+++ b/app/Http/Controllers/RedirectController.php
@@ -32,8 +32,12 @@ class RedirectController extends Controller
 	 */
 	public function album(Request $request, $albumid)
 	{
-		if ($request['password'] != '' && Configs::get_value('unlock_password_photos_with_url_param', '0') === '1') {
-			$this->albumsFunctions->unlockAlbums($request['password']);
+		if ($request['password'] != '') {
+			if (Configs::get_value('unlock_password_photos_with_url_param', '0') === '1') {
+				$this->albumsFunctions->unlockAllAlbums($request['password']);
+			} else {
+				$this->albumsFunctions->unlockAlbum($albumid, $request['password']);
+			}
 		}
 
 		return redirect('gallery#' . $albumid);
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/RedirectController.php
```diff
@@ -4,21 +4,71 @@
...
+                return redirect('gallery#' . $albumid . '/' . $photoid);
+        }
+
+        /**
+         * Provided an password, add all the albums that the password unlocks
+         *
+         * @param string $password
+         */
+        private function unlockPasswordAlbum(string $password) {
+                $albums = Album::whereNotNull('password')->where('password', '!=', '')->get();
```
ildyria: No, this is way too dangerous.
You know the album ID, thus you should query the album, check if the password is Ok,

And if it is a the correct password, then we also do the check for the other albums.

It is a bit slower but this increases the difficulty of unlocking all the albums by bruteforce.

ildyria (changes requested):
aside from the check which is a bit too wide, this looks fine. :+1:

app/Http/Controllers/RedirectController.php
pchampio: Given a photo uploaded to a password protected album.
I want this photo to be available for people opening the smart album `recent` with the following link: `https://domain.com/r/recent?password=my_password`.
How should I implement this?

ildyria: So I assume that your `recent` smart album is public ?

pchampio: Yes

ildyria: Just to clarify, you want to add a password to your `recent` folder or you want your `recent` folder to display recent pictures from the "_unlocked_" folders?

pchampio: I want the `recent` smart album to display recent pictures from the "unlocked" folders (and public photos).
And, I only want provide one share link to the `recent` smart album.

I don't want to have a password to the `recent` folder. But if a `?password=` is provided to the `recent` smart album, then unlock the photos.

pchampio: > `recent` folder to display recent pictures from the "unlocked" folders?

`recent` smart album currently works this way, if you unlock a password protected album in a given session. Then when this session browse the `recent` smart album, he will see the pictures from the "unlocked" folders in the `recent` album.

ildyria: Hooo I forgot about that.

ildyria (approved):
LGTM

PR #799, opened 2020-11-23
PHP 7.3 is dead, long live PHP 7.4!

ildyria:


pchampio:
As a reminder, LycheeOrg/Lychee-Docker uses php-fpm7.3.
https://github.com/LycheeOrg/Lychee-Docker/blob/50442d3a520e2d1a86c741183a1e3e641caec363/entrypoint.sh#L120

ildyria:
@pchampio thank you for pointing that one out! :scream: 

Well, it would be a good idea to fix it @d7415  :smile:

d7415 (approved):
Untested, but looks good. Most of the PHP changes seem to be the ones you prepared in advance.

The Docker images don't support this yet as they haven't been built. But they also don't contain this for the same reason so that's ok.

Lychee-Docker master branch is ready, so the next build will support this.

Merge when ready!

PR #810, opened 2020-12-07
fix(rss): avoid display feed link in HTML if RSS option is disabled

yoshinorin:
Hi, I found the title issue :)

## Description

Currently, lychee always displays a link for feed on HTML even if `rss_enable` option is false.

```
<link rel="alternate" type="application/atom+xml" href="http://localhsot:90/feed" title="Latest pictures">
```

## Implementation

*  IndexController, PageController, and ViewController checks the `rss_enable` setting and return `true/false` to `head.blade.php`.
* `head.blade.php` receive `true/false` from them.
    * if true, displays a link feed.
    * if false, does not display a link feed.


ildyria (approved):

ildyria:
Thank you. :)

PR #819, opened 2020-12-16
Add admin passwordless auth, bump 4.0.9 to force migration

ildyria:
:warning: this PR changes completely the admin authentication system. We now have a user with ID 0 instead of using the config database. This was necessary to enable the use of Yubikey and similar for authentication.

Currently:

- only available for the admin user
- new tab available: U2F
- To login, press the secret key `K`, the password-less query will ask you to authentify.

Left to be done:

- some translations
- UI to select user to login with
- Also add the possibility to use it as 2FA
- make the possibility to force other users to use 2FA


ildyria:
Once this is merged, I will bump the version number to 4.1 as this is a major change in the way we handle admin user account.

app/ControllerFunctions/WebAuth/VerifyAuthentication.php
```diff
@@ -0,0 +1,84 @@
...
+	 * @param array $credentials
+	 *
+	 * @return \Illuminate\Contracts\Auth\Authenticatable|\DarkGhostHunter\Larapass\Contracts\WebAuthnAuthenticatable|null
+	 */
+	protected function getUserFromCredentials(array $credentials)
+	{
+		// We will try to ask the User Provider for any user for the given credentials.
+		// If there is one, we will then return an array of credentials ID that the
+		// authenticator may use to sign the subsequent challenge by the server.
+		if ($this->isSignedChallenge($credentials) && $id = $this->binaryID($credentials['rawId'])) {
```
kamil4: Whoa! And I thought I was the reckless one here :wink:. Could you wrap that assignment of `$id` in brackets and add an explicit test against `null` so that it's clear that this is not a typo?

app/Legacy/Legacy.php
```diff
@@ -0,0 +1,70 @@
...
+
+namespace App\Legacy;
+
+use App\Models\Configs;
+use App\Models\Logs;
+use Illuminate\Support\Facades\Hash;
+use Illuminate\Support\Facades\Session;
+
+/**
+ * Stuff we need to delete in the future.
```
kamil4, line 11: Exactly, I've been wondering all along why we have this `App\Legacy` at all? Can't we clean this up during the database migration on version update and not have any of this code in the server at all?

app/Legacy/Legacy.php
```diff
@@ -0,0 +1,70 @@
...
+{
+	public static function resetAdmin(): void
+	{
+		Configs::where('key', '=', 'username')->orWhere('key', '=', 'password')->update(['value' => '']);
+	}
+
+	public static function SetPassword($request)
+	{
+		$configs = Configs::get();
+		if (Configs::get('version', '040000') < '040008') {
```
kamil4, line 23: So is this all just in case the database migration wasn't run for some reason?

app/ModelFunctions/SessionFunctions.php
```diff
@@ -176,17 +166,16 @@ public function log_as_user(string $username, string $password, string $ip)
 	 */
 	public function log_as_admin(string $username, string $password, string $ip)
```
kamil4, line 167: So do we actually need a separate `log_as_admin` at this point? It looks like `log_as_user` could easily handle it... Or is it because of the Legacy stuff?

resources/views/includes/head.blade.php
```diff
@@ -17,4 +17,6 @@
   @include('feed::links')
 @endif
 
+<script src="dist/Larapass.js"></script>
```
kamil4, line 20: Why isn't this compiled in with the rest of the front end?

database/migrations/2020_12_12_203153_migrate_admin_user.php
```diff
@@ -0,0 +1,37 @@
...
+
+class MigrateAdminUser extends Migration
+{
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
```
kamil4, line 15: Why doesn't this also remove `username` and `password` from Config?

kamil4 (approved):
My meta-comment is that this whole U2F feature flies way above my head. I have no idea how that part works in detail or how I would even use it.

Having said that, the concept of making the admin a regular user in the Users table seems like a good idea to me. I'm a little surprised though that the code doesn't completely rip out the "old way" but rather seems to implement the new one alongside the old one. Why? Is it a temporary safety measure during the transition?

Also, I don't know if it's just me, but I've never heard the "U2F" term before, yet it looks like we are putting that very term in a prominent place in the GUI. I'm just wondering how much confusion/questions it will result in?

app/Legacy/Legacy.php
ildyria: This stuff is for the users who do not use the command line.

We now have a check which has a look at the file version and compare it to the database version. If the database is behind the file, a request for migration is prompted. However this request requires admin privilege (thus admin login), as a result, we still need to support the old admin login so that the migration can be properly applied.

app/Legacy/Legacy.php
ildyria: Yes: e.g. if you don't have command line access.

app/ModelFunctions/SessionFunctions.php
ildyria: To log as admin you need:
- hash of username
- hash of password

If we didn't need the hash of username, then yes those two could be merged.

resources/views/includes/head.blade.php
ildyria: It doesn't work if it is compiled. :(

database/migrations/2020_12_12_203153_migrate_admin_user.php
ildyria: Failsafe. Just in case something goes wrong, it is easier to keep the data than to recreate them. We could delete them at some point.

ildyria:
> My meta-comment is that this whole U2F feature flies way above my head. I have no idea how that part works in detail or how I would even use it.

Tl;DR: This PR allows you to login on Lychee with a fingerprint reader rather than having to type your password.

Issue #524, opened 2020-05-01
[Enhancement] Possible age Verification 


fotocreativo:
Hello, everyone. I hope I'm in the right place with my question. Is there a way to add an age verification to a specific folder that you choose in Lychee? Since I am a photographer who also takes cgi pictures, which are not suitable for young people. Even if these pictures are harmless, the German law is very restrictive concerning such sites. A help would be a small query in Lychee v4.0.2 which verifies the age. And a hint that you could enter manually. An imprint / privacy policy that I create as a html page and then link to it would be great if you could implement that. Many thanks to the community who continue to develop Lychee.

ildyria:
@fotocreativo Can you check if this https://github.com/LycheeOrg/Lychee/pull/822 would solve your issue  or does it need to be an explicit age verification (which we all know can just be by passed by a 15 yo kid).

ildyria:
This should be fixed with https://github.com/LycheeOrg/Lychee/commit/27e4428786c2cacface06c8986d62db02ee0dfaf

It is not on the full website but per album.

PR #822, opened 2020-12-18
Add a new kind of albums: Sensitive - bump to 4.0.10 for migration.

ildyria:
related https://github.com/LycheeOrg/Lychee/issues/524

Not sure if this will fix the problem completely though.

In my case, I had to open the admin view while my parents were behind my back, having folders such as "Boudoir" etc can be tricky to explain.

Thus this PR implements NSFW albums.
- adds a new "flag" so that such folders can be recognized at first glance while being logged in.
- adds a key-bind `H` who hides all folders previously marked as sensitive.
- adds a pink star in the header to notify the user is in a Sensitive folder.
- Changing this property is either done on clicking on the star or toggling it in the visibility parameters.
- makes it possible to hide Sensitive folder by default.

**Important note: if a sensitive folder is set as public, it will be revealed by `H`.** If you do not wish such public folder to be visible, set it as _hidden_.


app/ControllerFunctions/ReadAccessFunctions.php
```diff
@@ -53,7 +53,7 @@ public function album($album, bool $obeyHidden = false): int
 
 		if (
 			$album->public != 1 ||
-			($obeyHidden && $album->visible_hidden !== 1)
+			($obeyHidden && $album->viewable !== 1)
```
kamil4, line 56: Thank you! I've always found the name of this field confusing (although personally I probably would've preferred if it was called simply `visible`, but what you have is definitely better than what it was, so...).

app/Http/Controllers/AlbumController.php
```diff
@@ -600,6 +602,25 @@ public function move(AlbumIDsRequest $request)
 		return $no_error ? 'true' : 'false';
 	}
 
+	/**
+	 * Set if a photo is a favorite.
```
kamil4: Comment needs updating (or not? :stuck_out_tongue_winking_eye:).

app/Http/Controllers/AlbumController.php
```diff
@@ -600,6 +602,25 @@ public function move(AlbumIDsRequest $request)
...
+	 * @return string
+	 */
+	public function setNSFW(Request $request)
+	{
+		$request->validate([
+			'albumID' => 'required|string',
+		]);
+
+		$album = Album::where('id', $request['albumID'])->first();
+		$album->nsfw = ($album->nsfw != 1) ? 1 : 0;
```
kamil4, line 619: So it basically toggles the current state? I must say that I'm not fond of the stateful nature of this API though I'm guessing you took inspiration from existing ones that are coded the same way? Personally I would prefer if the front end simply sent the new state and the server just set it; fewer surprises that way...

app/Http/Kernel.php
```diff
@@ -69,8 +69,8 @@ class Kernel extends HttpKernel
 	 * @var array
 	 */
 	protected $routeMiddleware = [
-		'auth' => \Illuminate\Auth\Middleware\Authenticate::class,
-		'auth.basic' => \Illuminate\Auth\Middleware\AuthenticateWithBasicAuth::class,
+		// 'auth' => \Illuminate\Auth\Middleware\Authenticate::class,
+		// 'auth.basic' => \Illuminate\Auth\Middleware\AuthenticateWithBasicAuth::class,
```
kamil4, line 73: Didn't you just enable this code in #819? I'm surprised to see it go away so quickly :smiley:.

app/Locale/Lang.php
```diff
@@ -2,17 +2,19 @@
...
 		$return = [];
 		$list_lang = scandir(__DIR__);
 		for ($i = 0; $i < count($list_lang); $i++) {
-			if ($list_lang[$i] != '.' &&
+			if (
+				$list_lang[$i] != '.' &&
 				$list_lang[$i] != '..' &&
-				$list_lang[$i] != 'Lang.php' &&
-				$list_lang[$i] != 'LangInterface.php'
+				$list_lang[$i] != 'Lang.php'
```
kamil4: So _this_ is why the server craps out whenever I have a locale file open in `vi` (which creates a temporary file in the same directory named, e.g., `.English.php.swp`)?! I've always wondered, but never enough to investigate...

OK, so this question is unrelated to this PR, but is there a reason why we don't first filter out any name that doesn't end with the `.php` extension? Things would be a lot more robust then...

app/Models/User.php
```diff
@@ -91,6 +91,11 @@ public function shared()
 		return $this->belongsToMany('App\Models\Album', 'user_album', 'user_id', 'album_id');
 	}
 
+	public function get_username(): string
+	{
+		return $this->is_admin() ? 'Admin' : $this->username;
```
kamil4, line 96: What's going on here? I thought that the admin account could be called anything one wants?

kamil4 (approved):
I've got some comments scattered around but they are mostly genuine questions...

app/ControllerFunctions/ReadAccessFunctions.php
ildyria: yeah but `visible` is a reserved key word by Laravel, just like `hidden` by the way. ^^"

app/Http/Controllers/AlbumController.php
ildyria: With livewire this will be fixed once for all. :)

app/Http/Kernel.php
ildyria: yeah, mistakes were made. ^^"

app/Locale/Lang.php
ildyria: sure. :)

app/Models/User.php
ildyria: When you return the album, you have the owner username: `$album->owner->username`.
Now with the admin as `user(id = 0)`, the value of username is fixed to the bcrypt of the admin username (just like previously in the config table). As a result, `$album->owner->username` was leaking the admin username hash.

app/Http/Controllers/AlbumController.php
ildyria: oupsie

app/ControllerFunctions/ReadAccessFunctions.php
kamil4: Bummer...

app/Models/User.php
kamil4: Right... So here's a question: is it really important to bcrypt the admin username? Why not make it consistent for all users and store the username verbatim?

ildyria: It is not really important to be honest, it somewhat served as a second password.

It is just that applying the migration is going to be "difficult" as we would need to have code that manage the migration (once again in the LEGACY module) and we would still need to manage the case where the migration is not done yet.

Or we could just tell all our users _"screw you, please reset your admin username/password"_, but I'm not sure that would be appreciated with regard to Public Relations. :sweat_smile:

PR #856, opened 2021-01-09
Major fixes

ildyria:
* Fix subalbums thumbnail in a hidden album (in case of directly accessed)
* Remove `min_takestamp`, `max_takestamp`, they are now directly generated via SQL sub query.


kamil4 (approved):
Looks good! Let's merge it into `new-front-end` and do some practical testing of that branch!

PR #863, opened 2021-01-12
Report too large requests to the frontend

kamil4:
As suggested in #858, this adds additional diagnostics during failed photo uploads due to too small `upload_max_filesize` or `post_max_size`.


kamil4:
Never mind. The failures from the testsuite made me realize the error of my way. This should be mostly a frontend affair. I'll submit a new PR, against the refactored server.

PR #865, opened 2021-01-12
Make Photo:add validation error easier to discern

kamil4:
This is a second try, once I realized that most of the problem is in the frontend.

As suggested in #858, this adds additional diagnostics during failed photo uploads due to too small `upload_max_filesize` or `post_max_size`.

The server needs a tweak because by default a failed validation (which is what happens if a file size exceeds `upload_max_filesize`) results in a redirect so normally we never even see the error code in the frontend.


ildyria (approved):
Looks good to me.

ildyria (approved):

PR #870, opened 2021-01-15
fix sysdate for import_via_symlink

evoludolab:
with import_via_symlink enabled set sysdate to timestamp of original photo (instead of the time of the symlink, i.e. the time of the import)

### DIFF
--- a/app/Models/Extensions/PhotoCast.php
+++ b/app/Models/Extensions/PhotoCast.php
@@ -153,7 +153,12 @@ trait PhotoCast
 		$return['thumb2x'] = ($this->thumb2x == '1') ? Storage::url('thumb/' . Helpers::ex2x($this->thumbUrl)) : '';
 
 		$path_prefix = ($this->type == 'raw') ? 'raw/' : 'big/';
-		$return['url'] = Storage::url($path_prefix . $this->url);
+		$filename = Storage::url($path_prefix . $this->url);
+		$return['url'] = $filename;
+		// With import_via_symlink enabled sysdate is import date; change to original photo timestamp
+		if (is_link($filename)) {
+			$return['sysdate'] = date('d F Y \a\\t H:i', filemtime(readlink($filename)));
+		}
 
 		if ($this->livePhotoUrl !== '' && $this->livePhotoUrl !== null) {
 			$return['livePhotoUrl'] = Storage::url('big/' . $this->livePhotoUrl);


kamil4:
Could you explain the motivation for this change? As far as I know, we always set `sysdate` to the time a given object was created in Lychee (album creation time or photo upload/import time). I don't believe that we ever pay any attention to the timestamps of imported files (we extract the takestamps from the metadata embedded inside the photos and videos, such as EXIF). So what's the reason to do it differently when `import_via_symlink` is enabled?

evoludolab:
I have a number of old and/or edited photos that lost the EXIF info at some point. However, the timestamps of the files are fairly accurate. Without EXIF info the date range displayed on album covers as well as when hovering over images inside an album are the import dates, which are quite meaningless. This small fix prevent this and returns the timestamp of the original photo. Of course this is only makes sense with ``import_via_symlink`` because otherwise the import date and the file creation date should be the same.

As an alternative, I had been playing with checking whether ``takedate`` is set. If not, set it to the timestamp of the original photo. Using ``sysdate`` results in a slightly smaller change in the code but has the same effect. If this is more in line with the Lychee design principles I would be happy to provide the few lines of code.

If there is interest, I am working on a related change to Lychee-front for small changes to the album title as well as the overlay information on photos.

kamil4:
Personally I would prefer if, for photos without EXIF, the file timestamp was extracted at import time and stored in `takedate`, and I think this should be an optional feature guarded by a new config variable in the advanced settings. But let's see what other developers think. @ildyria, @d7415?

evoludolab:
I agree that my suggestion is a patch and a cleaner solution would be to digest the timestamp info at import time. However, changes of that scope are way beyond my current understanding of the Lychee code, while the proposed fix is simple, unintrusive and works ;-)

Thinking a little further, maybe I should have a closer look at the console utilities and Takedate.php, in particular. This should allow to address my issue in a persistent manner and within a scope that seems more manageable.

ildyria:
> Personally I would prefer if, for photos without EXIF, the file timestamp was extracted at import time and stored in `takedate`, and I think this should be an optional feature guarded by a new config variable in the advanced settings. But let's see what other developers think. @ildyria, @d7415?

I am not favorable in changing the value of `takedate` as when `takedate` is set we display the small camera in the front end to "certify" the date of the picture.

(TBH, I don't really care about the value of sysdate, I don't use it).

With regard to that patch, I would be way more in favor of the following option:

> Thinking a little further, maybe I should have a closer look at the console utilities and Takedate.php, in particular. This should allow to address my issue in a persistent manner and within a scope that seems more manageable.

One reason to go for the command approach is it limits the number of file queries: imagine you have an album with 200 pictures without `takedate`, suddenly you have to query 200 pictures, checking if it is a symlink, following the link...
This is not a good solution in that regard as it will most certainly induce a slowdown. Having this part of the Takedate command is IMHO a way better approach (even though in this case, it is the sysdate that would be updated).

kamil4:
> One reason to go for the command approach is it limits the number of file queries: imagine you have an album with 200 pictures without `takedate`, suddenly you have to query 200 pictures, checking if it is a symlink, following the link...
> This is not a good solution in that regard as it will most certainly induce a slowdown. Having this part of the Takedate command is IMHO a way better approach (even though in this case, it is the sysdate that would be updated).

I totally agree with you on that. That's why I was advocating doing it at import time and storing the result in the database. Takedate command vs import -- that's just an implementation detail; in my opinion it should be both.

> I am not favorable in changing the value of `takedate` as when `takedate` is set we display the small camera in the front end to "certify" the date of the picture.
> 
> (TBH, I don't really care about the value of sysdate, I don't use it).

Keep in mind that `sysdate` is not a separate column in the database; it's just a (IMO confusing) name we use when sending the creation (i.e., upload) time to the frontend. And upload time is a well-defined concept that is valid for _all_ photos, so I really don't think it's a good idea to mess with it (in particular, we do display it as _Uploaded_ in the sidebar).

We have three timestamps in `photos` records: `created_at`, `updated_at`, and `takestamp`. Unless we add a fourth one, I really don't see where else we could store this than in `takestamp`. I don't consider this value "certified" in any way; the little camera picture we display is simply there to indicate that it's the take date rather than upload date (which we display if the take date is missing). In fact, at the very top of my TODO list for new features is adding the UI to edit `takestamp`, because it's frequently missing or wrong (lost during edit, wrong time zone, etc), and we have been asked by numerous users to provide this ability.

kamil4:
> Keep in mind that `sysdate` is not a separate column in the database; it's just a (IMO confusing) name we use when sending the creation (i.e., upload) time to the frontend. And upload time is a well-defined concept that is valid for _all_ photos, so I really don't think it's a good idea to mess with it (in particular, we do display it as _Uploaded_ in the sidebar).

I've thought about this some more and I'm no longer feeling that strong about it. For uploaded photos `created_at` is the upload time, but for imported ones I can see it both ways: either keep it as "time the photo was imported into Lychee" (as it is now) or read the timestamp of the file being imported and use that as the "upload time". We _can_ set `created_at`, correct? (I know that Laravel uses it internally)

ildyria:
> We can set `created_at`, correct? (I know that Laravel uses it internally)

After creation for sure. During creation I don't know (and didn't bother looking TBH).

In the case of symlinks, I feel that created at can be the timestamp of the file, no strong opinion on that one. In the case of an uploaded file, created at will be the timestamp of the file as the file is created on the server...

evoludolab:
Thank you very much for the insightful discussion. I close this pull request in favour of #874 which sets the `create_at` entry through a new option for the `Takedate` command line utility. In order to display the dates consistently in the frontend I have additionally adjusted the formatting of `sysdate` to be the same as that of `takedate`.

PR #873, opened 2021-01-15
more refactoring

ildyria:
Just a few more stuff.


app/Actions/Import/Exec.php
```diff
@@ -0,0 +1,273 @@
...
+	/**
+	 * @param string $pattern
+	 * @param string $filename
+	 *
+	 * @return bool
+	 */
+	private function check_file_matches_pattern(string $pattern, string $filename)
+	{
+		// This function checks if the given filename matches the pattern allowing for
+		// star als wildcard (as in *.jpg)
```
kamil4: Typo: `als` -> `as`

app/Actions/Import/Exec.php
```diff
@@ -0,0 +1,273 @@
...
+			$filesCount++;
+			// It is possible to move a file because of directory permissions but
+			// the file may still be unreadable by the user
+			if (!is_readable($file)) {
+				$this->status_update('Problem: ' . $file . ': Could not read file');
+				Logs::error(__METHOD__, __LINE__, 'Could not read file or directory (' . $file . ')');
+				continue;
+			}
+			$extension = Helpers::getExtension($file, true);
+			$raw_formats = strtolower(Configs::get_value('raw_formats', ''));
```
kamil4: I don't think this needs to be looked up in every new iteration?

app/Actions/Import/FromServer.php
```diff
@@ -0,0 +1,74 @@
...
+
+	public function __construct(Exec $exec)
+	{
+		$this->exec = $exec;
+	}
+
+	public function do($validated)
+	{
+		if (isset($validated['delete_imported'])) {
+			$this->exec->delete_imported = $validated['delete_imported'] === '1';
```
kamil4: Add parentheses? (also two lines below)

app/Actions/RSS/Generate.php
```diff
@@ -0,0 +1,97 @@
...
+	public function do()
+	{
+		$publicIds = resolve(PublicIds::class)->getPublicAlbumsId();
+		$rss_recent = intval(Configs::get_value('rss_recent_days', '7'));
+		$rss_max = Configs::get_Value('rss_max_items', '100');
+		$nowMinus = Carbon::now()->subDays($rss_recent)->toDateTimeString();
+
+		$photos = Photo::with('album', 'owner')
+			->where('created_at', '>=', $nowMinus)
+			->where(fn ($q) => $q->whereIn('album_id', $publicIds)->orWhere('public', '=', '1'))
```
kamil4, line 93: What's the reason behind this additional `orWhere`? Won't it expose the content of hidden or password-protected albums?

app/Actions/Import/Exec.php
```diff
@@ -0,0 +1,273 @@
...
+			// Folder
+			$album = null;
+			if ($this->force_skip_duplicates || Configs::get_value('skip_duplicates', '0') === '1') {
+				$album = Album::where('parent_id', '=', $albumID == 0 ? null : $albumID)
+					->where('title', '=', basename($dir))
+					->get()
+					->first();
+			}
+			if ($album === null) {
+				$create = resolve(AlbumCreate::class);
```
kamil4: This fails; `use` missing (and now that the use of `Photo\Create` is in another file, I guess it no longer needs to be renamed to `AlbumCreate`?).

app/Actions/Photo/Rotate.php
```diff
@@ -0,0 +1,112 @@
...
+			Logs::error(__METHOD__, __LINE__, 'imagick is disabled.');
+
+			return false;
+			// @codeCoverageIgnoreEnd
+		}
+
+		// Abort on symlinks to avoid messing with originals linked
+		if (is_link(Storage::path('big/') . $photo->url)) {
+			// @codeCoverageIgnoreStart
+			Logs::error(__METHOD__, __LINE__, 'Synlinked images cannot be rotated');
```
kamil4: Typo; `Synlinked` -> `Symlinked`

kamil4 (changes requested):
I reviewed it unasked since I want to make changes to some of this functionality (Import, Rotate) and I'd rather do it to the refactored versions :smiley:.

I spotted one regression: importing folders with subdirectories fails again.

Also, there's a suspicious-looking query in the RSS code but that was there before as well...

app/Actions/Import/Exec.php
ildyria: https://translate.google.nl/?sl=nl&tl=en&text=als&op=translate :nerd_face:

app/Actions/Import/Exec.php
ildyria: Actually, Configs has a singleton approach.

ildyria:
@kamil4 please double check.

cleytontiagosorinho (approved):
Ajuda nessa questão

app/Actions/Import/Exec.php
kamil4: Oh, I know, I did actually learn Dutch during my 5+ years in NL. Near the end my English even acquired Dutch accent :smiley:.

app/Actions/Photo/Rotate.php
```diff
@@ -39,7 +39,7 @@ private function check(Photo $photo, int $direction): bool
 		// Abort on symlinks to avoid messing with originals linked
 		if (is_link(Storage::path('big/') . $photo->url)) {
 			// @codeCoverageIgnoreStart
-			Logs::error(__METHOD__, __LINE__, 'Synlinked images cannot be rotated');
+			Logs::error(__METHOD__, __LINE__, 'SynLinked images cannot be rotated');
```
kamil4: Sy**m**linked :smile:.

kamil4 (approved):
OK, importing nested folders works now.

I should point out that I tested only the import via GUI; I have no experience with using the command line for that.

Also, there's still the bug that I reported last night on Gitter, in that thumbs are missing for `Unsorted` or `Public` smart albums for non-admin users.

app/SmartAlbums/BareSmartAlbum.php
```diff
@@ -41,12 +41,18 @@ public function setAlbumIDs(BaseCollection $albumIds): void
...
+			return $query;
 		}
 
-		if (AccessControl::is_logged_in() && AccessControl::id() > 0) {
-			$query = $query->orWhere('owner_id', '=', AccessControl::id());
+		if (AccessControl::is_logged_in()) {
+			$query = $query->where('owner_id', '=', AccessControl::id())
+				->orWhere(
+					fn ($q) => $q->whereNotNull('album_id')
+						->whereIn('album_id', $this->albumIds)
```
kamil4, line 52: The `whereNotNull` seems redundant here given the `whereIn` right after it?

PR #874, opened 2021-01-16
update CLI Takedate

evoludolab:
* option --timestamp added to set create_at to timestamp of media file for media that lacks EXIF information
* use same formatting for 'sysdate' as for 'takedate'

For background see #870

### DIFF
--- a/app/Console/Commands/Takedate.php
+++ b/app/Console/Commands/Takedate.php
@@ -14,14 +14,15 @@ class Takedate extends Command
 	 *
 	 * @var string
 	 */
-	protected $signature = 'lychee:takedate {from=0 : from which do we start} {nb=5 : generate exif data if missing} {tm=600 : timeout time requirement}';
+	protected $signature = 'lychee:takedate {from=0 : from which do we start} {nb=5 : generate exif data if missing} {tm=600 : timeout time requirement}' .
+		   '{--timestamp : use timestamps of media files if exif data missing}';
 
 	/**
 	 * The console command description.
 	 *
 	 * @var string
 	 */
-	protected $description = 'Make sure takedate is correct';
+	protected $description = 'Update missing takedate entries';
 
 	/**
 	 * Execute the console command.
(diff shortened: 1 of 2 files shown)


kamil4 (approved):
This looks good. We should also provide this functionality during regular import, but that can be done in a separate PR.

app/Console/Commands/Takedate.php
```diff
@@ -45,22 +50,36 @@ public function handle(Extractor $metadataExtractor)
...
-			if (file_exists($url)) {
-				$info = $metadataExtractor->extract($url, $photo->type);
-				if ($info['takestamp'] != null) {
-					$photo->takestamp = $info['takestamp'];
-					if ($photo->save()) {
-						$this->line($i . ': Takestamp updated for ' . $photo->title);
-					} else {
-						$this->line($i . ': Failed to update takestamp for ' . $photo->title);
-					}
+			$i++;
```
ildyria, line 53: This `$i++` is not consistent with what is actually being done and is off by +1.
Just place it before each `continue` and at the end of the loop.

app/Console/Commands/Takedate.php
```diff
@@ -45,22 +50,36 @@ public function handle(Extractor $metadataExtractor)
...
-					$this->line($i . ': Could not get Takestamp data for ' . $photo->title . '.');
+					$this->line($i . ': Failed to update takestamp for ' . $photo->title);
 				}
+				continue;
+			}
+			if (!$timestamps) {
+				$this->line($i . ': Could not get Takestamp data for ' . $photo->title . '.');
+				continue;
+			}
+			$filename = $url;
```
ildyria: why is this needed? `$url` is not used anymore afterwards.

if you do not like `$url` then just rename all the previous `$url` before by `$filename`.

ildyria (changes requested):
Looks good.
Untested.

Please fix the two comments. :smiley:

ildyria (approved):

cleytontiagosorinho (approved):
Só os que compensa no momento

cleytontiagosorinho (approved):
As melhores opicoes

Issue #442, opened 2020-02-23
[Enhancement] choose image as album thumbnail


ezzra:
currently you can only change the album thumbnail image by setting the favorite star to images, but that should be independent from that. It would be necessary to create a new menu operation to change the album thumbnail, maybe even the two underlying thumbs.

ildyria:
That could be an interesting optimization. But the UI design (and implementation) of it is unsure.

gardiol:
I immagine it as a click you do on an image like for star, maybe in a differenc orner... You can set one or more and if you set more than one, it cycles.

d7415:
From #871 (@evoludolab):

> I would like to be able to select photos that are displayed as album covers. In particular, it would be nice to set the album cover by tagging one of its photos as the cover photo through the context menu. For albums that contain sub-albums, the sub-album-covers should also be available as potential covers for the parent album. This becomes particularly important if an album includes only sub-albums and no photos.
> 
> Maybe this could be implemented by leveraging the existing tags, e.g. by introducing an `AlbumCover`-tag. However, tagging multiple photos in one album could be an issue as well as if the tagged photo should serve as the cover of a parent album higher up in the hierarchy. Allowing users to override the default set by the owner of the album (if any) would be nice but certainly lower priority.
> 
> Since such an enhancement would possibly require changes to the backend and database structure, I post the feature request here rather than in Lychee-front.
> 
> Thanks for an impressive (and stunning) tool!

ildyria:
It would be better to actually add a column on the albums table to take care of such specification.
We only need to record the ID of a single picture as the 3 others behind are somewhat hidden anyway.

ildyria:
Additionally @evoludolab in the current implementation, you can chose the picture by starring it. :)

kamil4:
Starring is a very crude solution though. If more than one picture is starred, we can't pick their order. "Star just one per album" is a current workaround but it breaks down if there are subalbums with starred pictures.

I agree this needs to be stored in a new column in `albums`. Storing in `photo` wouldn't work for nested albums, and making it work for nested albums _is_ important (otherwise the current workaround is good enough).

Depending on how foolproof we wanted to make it, the implemention could quickly become messy though. We would need to check and possibly update that column on every deletion/move/merge operation. Or we do it quick-and-dirty and make it the owner's responsibility to update the cover if the photo is no longer in the album. Same with permissions ("if the owner selected a photo that isn't public, that's their choice; assume they know what they are doing").

[...]

PR #875, opened 2021-01-18
Custom album covers

evoludolab:
**Draft** pull request to implement setting custom cover images for albums, see #442

This pull request implements the back end for setting custom album covers. The corresponding pull request for the front end extends the context menu for photos and albums with _Set/Remove Album cover_ to set/remove custom album covers of the parent album, see LycheeOrg/Lychee-front#239

* add a `cover_id` column and `cover` relationship to Album
* thumb is now added in the `toReturnArray()` function which simplifies the code at multiple places
* the `cover` image is greedy fetched to lower the number of SQL queries.

Future work (for another PR):
- allow the selection of the ancestor for which the photo is to be set, for now this is only for the direct parent.
- allow the ability to clear the cover photo without having to click on a picture.

**IMPORTANT:**
* ~~⚠️ this draft uses `album.description` to store the information of custom covers. It requires a new column `cover` in the `album` table, which is beyond my pay grade.~~
* ~~SQL query in `AlbumController.php` should be reviewed (this is completely unfamiliar territory...)~~
* ~~something seems amiss with the gulpfile because iconic.svg does not get processed (the new icon needs to be manually added to resources/views/includes/svg.blade.php and resources/views/view.blade.php)~~
* ~~only English localization done~~
* ~~support for 3 distinct images for albums dropped (seems overkill given that 2 are barely visible); if this is acceptable some code cleaning and streamlining would be recommended~~


cleytontiagosorinho (approved):
Sim

ildyria:
Still WIP. Incoming changes. Please wait for #873 to be merged.

ildyria:
@evoludolab please check.

app/Http/Controllers/AlbumController.php
```diff
@@ -185,6 +187,22 @@ public function setShowTags(AlbumIDRequestInt $request, SetShowTags $setShowTags
...
+	 * Set cover image of the album.
+	 *
+	 * @param Request $request
+	 *
+	 * @return bool|string
+	 */
+	public function setCover(AlbumIDRequestInt $request, SetCover $setCover)
+	{
+		$request->validate([
+			'cover_id' => 'integer|nullable',
```
kamil4: Should this be an integer or a string? Come to think of it, why do we even have `AlbumIDRequestInt` and `AlbumIDRequest`? Shouldn't one be enough?

app/Http/Controllers/SearchController.php
```diff
@@ -138,13 +138,6 @@ function (Builder $query) use ($escaped_term) {
 				if (AccessControl::is_logged_in()) {
 					$album['owner'] = $album_model->owner->username;
 				}
-				if ($this->readAccessFunctions->album($album_model) === 1) {
```
kamil4: This is going to conflict with `more-refactoring`...

app/Models/Album.php
```diff
@@ -129,6 +129,16 @@ public function children()
 		return $this->hasMany('App\Models\Album', 'parent_id', 'id');
 	}
 
+	/**
+	 * Return the relationship between a sub album and its parent.
```
kamil4: Comment needs updating.

app/Models/Extensions/AlbumGetters.php
```diff
@@ -68,24 +68,30 @@ public function get_all_photos()
...
+			}
 
-		//? apply safety filter : Do not leak pictures which are not ours
-		$forbiddenID = resolve(PublicIds::class)->getNotAccessible();
-		if ($forbiddenID != null && !$forbiddenID->isEmpty()) {
-			$sql = $sql->whereNotIn('album_id', $forbiddenID);
+			$cover = $sql->orderBy('star', 'DESC')
+				->orderBy($sort_col, $sort_order)
+				->orderBy('photos.id', 'ASC')
+				->limit(1)
```
kamil4, line 93: So are we eliminating the two "behind" cover pictures with this PR, leaving just the frontmost one?

app/Models/Extensions/AlbumGetters.php
```diff
@@ -68,24 +68,30 @@ public function get_all_photos()
...
-		[$sort_col, $sort_order] = $this->get_sort();
+		if ($this->cover != null) {
+			$cover = $this->cover;
+		} else {
+			[$sort_col, $sort_order] = $this->get_sort();
+
+			$sql = $this->get_all_photos();
+
+			//? apply safety filter : Do not leak pictures which are not ours
+			$forbiddenID = resolve(PublicIds::class)->getNotAccessible();
```
kamil4, line 81: I'm guessing this might be responsible for the unrelated issue I'm seeing with thumbs of smart albums for non-admin users?

kamil4 (approved):
Looks good overall.

I still need to look at the front end, and I'd like to try it out...

app/Http/Controllers/AlbumController.php
ildyria: `AlbumIDRequestInt` ensure you are not doing it on a basic smart album.

app/Http/Controllers/SearchController.php
ildyria: yep. which is why I would like to merge the other one before.

app/Models/Album.php
ildyria: oupsie.

app/Models/Extensions/AlbumGetters.php
ildyria: somewhat yes. I don't have strong opinion on that one. Visually, it will just be the same picture displayed 3 times.

app/Models/Extensions/AlbumGetters.php
ildyria: Indeed.

ildyria: fixed.

app/Http/Controllers/SearchController.php
ildyria: fixed.

app/Models/Extensions/AlbumGetters.php
```diff
@@ -70,9 +71,12 @@ public function get_all_photos()
 
 	public function get_thumb(): ?Thumb
 	{
+		DebugBar::warning('get_thumb:' . $this->id);
```
kamil4: Leftover debug code :smile:.

evoludolab:
Wow this is cool! Unfortunately I have a day job and won't have time to check until later ☹️ The first impression is great and everything seems to work! The badge is nice but a tad intrusive - especially on sub albums. Maybe a tooltip would help to explain that this marks the cover image of the parent? Anyways, I'll test more thoroughly later. Thanks!

kamil4:
> The badge is nice but a tad intrusive - especially on sub albums.

That's why I tweaked it to be the rightmost one... As to it being intrusive, we use badges all over the place already, so adding one more doesn't seem like a big stretch...

> Maybe a tooltip would help to explain that this marks the cover image of the parent?

We don't currently have tooltips in that area of the UI -- I think we may be using them only in the menu bar? I suppose we could add some but I'm not sure how necessary it is in this case, as the badge appears immediately after the user sets a cover, so it should be pretty obvious what it indicates. Of course, after a while one might forget...

kamil4:
I'm wondering about the security implications here. Shouldn't we at least be checking that the new cover Photo belongs to the user? I think the upload middleware would automagically do it if the request field was called `photoID`, not the current `cover_id`...

ildyria:
> I'm wondering about the security implications here. Shouldn't we at least be checking that the new cover Photo belongs to the user? I think the upload middleware would automatically do it if the request field was called `photoID`, not the current `cover_id`...

Then we just need to update the Ajax query to send a photoId :)

kamil4:
Done...

evoludolab:
@kamil4 nothing against badges and I am clearly in favour of marking the cover - makes everything more transparent! What about adding a more discreet marker similar to albums that contain sub albums?

Concerning tooltips I was more thinking of visitors that wonder what the orange badge means. But that seems quite cosmetic and probably better suited for a separate discussion.

ildyria:
> Concerning tooltips I was more thinking of visitors that wonder what the orange badge means. But that seems quite cosmetic and probably better suited for a separate discussion.

Visitors do not see badges.

evoludolab:
> Visitors do not see badges.

@ildyria Oh, did not yet notice... In that case long live the orange badge!

evoludolab:
As far as I can tell everything works like a charm. I am looking forward to studying all your changes and improvements carefully to get a better understanding of Lychee. It's been a pleasure - thanks!

PR #832, opened 2020-12-26
Massive refactoring of backend: v4.2.0 + fixes #843 #846 #858

ildyria:
- start using Livewire for the front-end, for now accessible at `example.com/livewire` if enabled via `LIVEWIRE_ENABLED` in `.env`
- use Facade `AccessControl` to access `Session` information (basically home-brewed `Auth`)
- use Facade `Lang` to access `Lang` information (Easier to use through the blade template)
- heavy refactoring of the core, introducing more granularity:
  - Interfaces are Contracts
  - Group Factories
  - use `trait` on album for smaller dedicated operations
  - add Nested Set theory to Album to allow access to all descendants  

Will probably be a bump to version 4.2.0 to be honest given the amount of change applied.

Additionally fixes #843

PR #877, opened 2021-01-19
Various fixes

kamil4:
Various assorted fixes for old bugs discovered during the review of #832.

* Copying a photo to Unsorted didn't work (server hang).
* Deleting the (content of) Unsorted triggered server error due to missing `Photo::select_unsorted`. I tried to replace it with factories and such, but I couldn't figure it out, so I used the good old query builder. I'm sure @ildyria can do something better :smiley:.
* Album download has a custom filter that prevents the downloads of photos via smart albums if the actual album the photos are in is not downloadable. This needed to be expanded to handle tag albums, and to handle smart albums with multi-user content (e.g., Recent) correctly.

### DIFF
--- a/app/Actions/Album/Archive.php
+++ b/app/Actions/Album/Archive.php
@@ -141,12 +141,12 @@ class Archive extends Action
 		// cares in what order photos are zipped?
 
 		foreach ($photos as $photo) {
-			// For photos in public smart albums, skip the ones
-			// that are not downloadable based on their actual
-			// parent album.
+			// For photos in smart or tag albums, skip the ones that are not
+			// downloadable based on their actual parent album.  Photos from
+			// Unsorted are handled implicitly by the photo ownership test.
 			if (
-				$this->albumFactory->is_smart($albumID) && !AccessControl::is_logged_in() &&
-				$photo->album_id !== null && !$photo->album->is_downloadable()
+				$album->smart && !AccessControl::is_current_user($photo->owner_id) &&
+				!$photo->album->is_downloadable()
 			) {
 				continue;
 			}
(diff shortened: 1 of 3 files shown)


app/Actions/Album/Archive.php
```diff
@@ -141,12 +141,12 @@ private function compress_album($photos_sql, $dir_name, &$dirs, $parent_dir, $al
...
-			// that are not downloadable based on their actual
-			// parent album.
+			// For photos in smart or tag albums, skip the ones that are not
+			// downloadable based on their actual parent album.  Photos from
+			// Unsorted are handled implicitly by the photo ownership test.
 			if (
-				$this->albumFactory->is_smart($albumID) && !AccessControl::is_logged_in() &&
-				$photo->album_id !== null && !$photo->album->is_downloadable()
+				$album->smart && !AccessControl::is_current_user($photo->owner_id) &&
+				!$photo->album->is_downloadable()
```
ildyria: if you did not prefetch the `album` this will generate 1 query per picture.

Also, if `album_id` is `null`, this will crash because you will try to call `->is_downloadable()` on null.

ildyria (approved):
Looks okay. Not tested.
Please fix below comment.

app/Actions/Album/Archive.php
kamil4: You mean prefetching `$photo->album`? Yeah, I get that. But this is a condition that will rarely be checked so I don't think we should optimize for it.

I gave it some considerable thought last night and I convinced myself that `album_id` will never be `null` by the time the third case is reached. The reason is that in case of photos with `album_id = null`, the smart albums only show the ones belonging to the current user (or none in public mode). So any such photos will be caught by the second case (`AccessControl::is_current_user($photo->owner_id)`) and the third one will never be reached. This is captured in the comments (maybe in a little cryptic fashion :wink:).

kamil4: Having thought about it some more, I decided to handle the case of `album_id == null`. I don't think it can _currently_ happen, but why make the code needlessly fragile and subject to breaking after some seemingly unrelated future changes?

PR #885, opened 2021-01-25
Allow to share an album with users right from the album page

bastantoine:
This PR adds a new icon in the top bar, next to the album visibility option.
On click, it opens a modal with the list of users, asking to choose which one to share it with.

See LycheeOrg/Lychee-front#242 for the related code in the front


resources/views/livewire/header.blade.php
```diff
@@ -49,35 +49,38 @@
         <a class="button button--eye" id="button_visibility_album" title="{{ Lang::get('VISIBILITY_ALBUM') }}" data-tabindex="3">
             <svg class="iconic iconic--eye"><use xlink:href="#eye"></use></svg>
         </a>
-        <a class="button button--nsfw" id="button_nsfw_album" title="{{ Lang::get('ALBUM_MARK_NSFW') }}" data-tabindex="4">
+        <a class="button" id="button_sharing_album_users" title="{{ $locale['SHARING_ALBUM_USERS'] }}" data-tabindex="4">
```
ildyria: Instead of using `$locale` can you use the `Lang` facade (see below for the other translations).

ildyria (approved):
I haven't tested the front-end yet but server side it looks good.

Would be nice to get rid of the `$locale` variable and use the `Lang` facade instead. :)

resources/views/livewire/header.blade.php
bastantoine: Oh right, I simply did a copy-paste, without looking any further...

bastantoine: Should be fixed know.

d7415:
Just to consider - I remember us spending time a while back trying to minimise the number of icons on that toolbar for mobile use. I can't remember if we also came up with another solution.

bastantoine:
> Just to consider - I remember us spending time a while back trying to minimise the number of icons on that toolbar for mobile use. I can't remember if we also came up with another solution.

Indeed I made my tests on a 32" monitor so that wasn't really an issue, but I see what you meant when I had a look on my phone.

With all the icons there was almost no place to see the album name (see below, on an iPhone 11 Pro, so not the biggest phone there is, but definitively not the smaller).

IMG_9DAEDF2D2A8D-1

I thought of including the content of the new modal in the album visibility modal, but I thought that it was already well filled, and adding new content in it would make it quite a mess.

Perhaps it might be ok to have both modal (album visibility and people sharing) in the same one with collapsible sections. What do you say?

ildyria:
> Perhaps it might be ok to have both modal (album visibility and people sharing) in the same one with collapsible sections. What do you say?

I think it might be time to add an hamburger.

Note that we already have some kind of filtering implemented by @tmp-hallenser
see here:
- https://github.com/LycheeOrg/Lychee-front/blob/master/scripts/main/lychee.js#L50
- https://github.com/LycheeOrg/Lychee-front/blob/master/scripts/main/header.js#L358

That being said, I am quite in favor to have it as a full CSS burger depending on the width of the screen.

bastantoine:
> I think it might be time to add an hamburger.

Or (just though of that) a sidebar, like the one you have when you click the _"About"_ icon. In there you would have all settings related to the album: album visibility, NSFW and sharing with people.

About the NSFW, you would still see the ⚠️  icon in the top bar if the album has the NSFW mode activated.

ildyria:
> Or (just though of that) a sidebar, like the one you have when you click the _"About"_ icon. In there you would have all settings related to the album: album visibility, NSFW and sharing with people.

Totally fine with that proposition.

> About the NSFW, you would still see the  icon in the top bar if the album has the NSFW mode activated.

Obviously. Just like the star for pictures.

app/Locale/English.php
```diff
@@ -65,6 +65,11 @@ public function get_locale(): array
 			'FULLSCREEN_ENTER' => 'Enter Fullscreen',
 			'FULLSCREEN_EXIT' => 'Exit Fullscreen',
 
+			'SHARING_ALBUM_USERS' => 'Share this album with users',
+			'WAIT_FETCH_DATA' => 'Please wait while we get the data.',
+			'SHARING_ALBUM_USERS_NO_USERS' => "There's no user to share the album with.",
+			'SHARING_ALBUM_USERS_LONG_MESSAGE' => 'Select the users to share this album with them',
```
kamil4: A couple of comments about the language:
* Is there a reason why some of these strings end with a period (`.`) and some don't?
* Without checking how `WAIT_FETCH_DATA` is used, I would guess it should end with an ellipsis (`...`).
* `SHARING_ALBUM_USERS_NO_USERS` would probably look better in plural (`There are no users [...]`).
* Drop `them` from `SHARING_ALBUM_USERS_LONG_MESSAGE`.

kamil4 (changes requested):

app/Locale/English.php
bastantoine: > Is there a reason why some of these strings end with a period (.) and some don't?

Not at all, it just looked better  at the time

Should be good now

kamil4 (approved):
I'm OK with the server changes, I think. I still need to read through the front end and also test it. Tomorrow...

Issue #886, opened 2021-01-27
[Enhancement] Disable direction display (more general: ignore certain EXIF tags)


sjml:
I'm setting up a new gallery with a lot of images that are geotagged. What I'm discovering, though, is that while the GPS locations are accurate, the direction info usually is not. On top of that, the apps that I use tend to only expose correction of the latitude/longitude, not of the direction, so it's a difficult problem to fix for large numbers of photos. 

I'd be happy for the photo's map to simply not show the direction, but I don't see any simple way to disable it beyond hacking the JavaScript myself (which would not survive an upgrade). 

Would you accept a PR to expose something like "Ignore GPS direction" as a setting? 

Alternately, and maybe solving a more general problem: is there some way to create an exclusion list of EXIF tags so that the import process could treat the photo as if it does not have those tags?

EDIT: I have a PR ready to go that does the simpler version of this (just adding a config flag to not use the direction data in the embedded map), but not sure how you prefer to handle it since it's properly *two* PRs, one to this repo and another to `Lychee-front`. Let me know!

ildyria:
> Would you accept a PR to expose something like "Ignore GPS direction" as a setting?

I have nothing against such setting. :)

> Alternately, and maybe solving a more general problem: is there some way to create an exclusion list of EXIF tags so that the import process could treat the photo as if it does not have those tags?

This is an interesting alternative, though the implementation of such thing would be interesting.

> I have a PR ready to go that does the simpler version of this (just adding a config flag to not use the direction data in the embedded map), but not sure how you prefer to handle it since it's properly two PRs, one to this repo and another to Lychee-front. Let me know!

Usually the process is as follows: 
- we take the 2 PR (Lychee-front and Lychee),
- review them both,
- merge on Lychee-front
- switch Lychee-front to master, rebuild & commit on the PR.
- merge to Lychee master

sjml:
Great, I'll prep two PRs, then. Thanks! 

> This is an interesting alternative, though the implementation of such thing would be interesting.

(Yeah, I think "interesting" is the right choice of word there. :) While it might be worth doing in the long run, it's definitely beyond the scope of fixing this very simple problem.)

Issue #634, opened 2020-06-24
[Enhancement] Add Regression tests.


ildyria:
Add more tests here https://github.com/LycheeOrg/Lychee/tree/master/tests/Feature :

- [x] #631
- [x] #635
- [x] #734
- [x] #643 
- [ ] #669 
- [x] #673
- [x] #781

kamil4:
I would add #691 to the list. Password-protection is a tricky edge case, especially since it needs to be handled differently in regular browsing vs search.

ildyria:
added #734 to the list

nagmat84:
I have checked #635, #643 and #781. These bugs are definitely covered by the newly added authorization tests.

PR #887, opened 2021-01-27
adding backend setting to disable display of GPS direction on map

sjml:
This is the backend PR for issue #886. Adds a new configuration setting to control whether the embedded maps use the GPS direction data to display a facing cone. It's set to `true` by default so upgrades won't cause unexpected breaks to existing behavior. 

(Paired with LycheeOrg/Lychee-front#243)

### DIFF
--- a/app/Http/Controllers/Administration/SettingsController.php
+++ b/app/Http/Controllers/Administration/SettingsController.php
@@ -224,24 +224,6 @@ class SettingsController extends Controller
 	}
 
 	/**
-	 * Enable display of photos on map for public albums.
-	 *
-	 * @param Request $request
-	 *
-	 * @return string
-	 */
-	public function setMapDisplayPublic(Request $request)
-	{
-		$request->validate(['map_display_public' => 'required|string']);
-
-		if ($request['map_display_public'] == '1') {
-			return Configs::set('map_display_public', '1') ? 'true' : 'false';
-		}
-
-		return Configs::set('map_display_public', '0') ? 'true' : 'false';
-	}
-
-	/**
 	 * Enable display of photo direction on map.
 	 *
 	 * @param Request $request
--- a/app/Http/Controllers/Administration/SettingsController.php
+++ b/app/Http/Controllers/Administration/SettingsController.php
@@ -224,21 +224,21 @@ class SettingsController extends Controller
 	}
 
 	/**
-	 * Enable display of photo direction on map.
+	 * Enable display of photos on map for public albums.
 	 *
 	 * @param Request $request
 	 *
 	 * @return string
 	 */
-	public function setMapDisplayDirection(Request $request)
+	public function setMapDisplayPublic(Request $request)
 	{
-		$request->validate(['map_display_direction' => 'required|string']);
+		$request->validate(['map_display_public' => 'required|string']);
 
-		if ($request['map_display_direction'] == '1') {
-			return Configs::set('map_display_direction', '1') ? 'true' : 'false';
+		if ($request['map_display_public'] == '1') {
+			return Configs::set('map_display_public', '1') ? 'true' : 'false';
 		}
 
-		return Configs::set('map_display_direction', '0') ? 'true' : 'false';
+		return Configs::set('map_display_public', '0') ? 'true' : 'false';
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


sjml:
Ah, I see the CI failures. Will address shortly.

sjml:
Looks like things are good other than the code coverage. I don't see a spot in the existing tests where it would be obvious to plug in, though... how important is this? 

(Squashed commits.)

ildyria:
> Looks like things are good other than the code coverage. I don't see a spot in the existing tests where it would be obvious to plug in, though... how important is this?

Meh, not much to be honest. We need to refine our testing suite to have more depth and also take care of checking regressions. See #634 

> (Squashed commits.)

That was not necessary, we always do Squash-merge. :)

database/migrations/2021_01_27_085903_config_map_display_direction.php
```diff
@@ -0,0 +1,40 @@
...
+	 */
+	public function up()
+	{
+		defined('BOOL') or define('BOOL', '0|1');
+
+		DB::table('configs')->insert([
+			[
+				'key' => 'map_display_direction',
+				'value' => '1',
+				'confidentiality' => 1,
```
ildyria: I'm not sure if it is a confidentiality `1` or a `0`. It might actually be a `0` because maps may be accessible while not being logged in.

ildyria (approved):
Aside from the confidentiality, looks good.

sjml:
> Meh, not much to be honest. We need to refine our testing suite to have more depth and also take care of checking regressions. See #634

Test coverage is a constant uphill battle. Glad to see you all are tackling it, though! 

> [Squashing commits] was not necessary, we always do Squash-merge. :)

Ah, good to know. I've formed the habit from other repos I contribute to where they prefer the PR to do the squashing. :)

database/migrations/2021_01_27_085903_config_map_display_direction.php
sjml: I wasn't 100% clear on the confidentiality values — are they documented someplace? Happy to make this whatever you think it should be, obviously. Will push a change to make it 0.

ildyria: Not really documented aside from in the code...
https://github.com/LycheeOrg/Lychee/blob/master/app/Models/Configs.php#L244
https://github.com/LycheeOrg/Lychee/blob/master/app/ModelFunctions/ConfigFunctions.php

app/Http/Controllers/Administration/SettingsController.php
```diff
@@ -241,6 +241,24 @@ public function setMapDisplayPublic(Request $request)
...
 	}
 
+	/**
+	 * Enable display of photo direction on map.
+	 *
+	 * @param Request $request
+	 *
+	 * @return string
+	 */
+	public function setMapDisplayDirection(Request $request)
```
kamil4: Are you planning to add a GUI to the settings to control this? I don't see one in your front end PR so far, so I don't think you need this at all?

sjml: Ah yeah, that was leftover from when I was experimenting to see where things were set in the engine. I don't think this needs to be exposed in the GUI but obviously it could be if people think it would be useful. Should I remove the backend code then?

kamil4: Yeah, adding a GUI is a PITA and personally I wouldn't bother. Since this is effectively dead code, I would recommend removing it.

sjml: Done!

kamil4 (approved):

ildyria:
@sjml For your next PR, Please create a new branch rather than working on your master. :)
https://lycheeorg.github.io/docs/contributions.html#which-branch-to-use

That makes it easier for use to interact with your work as we just need to do:
```sh
git remote add sjml git@github.com:sjml/Lychee
git fetch sjml
git checkout branch-name-you-chose
#modify stuff
git add -u
git commit -m "modified"
git push sjml branch-name-you-chose
```
instead of
```sh
git remote add sjml git@github.com:sjml/Lychee
git fetch sjml
git checkout sjml/master
git checkout -b sjml/master
#modify stuff
git add -u
git commit -m "modified"
git push sjml sjml-master:master  # this here is harder to remember compared to the block above :)
```
The reason the push syntax is different is because of the name conflict between your master branch and our master branch.

sjml:
> For your next PR, Please create a new branch rather than working on your master

Aye, cap'n!

PR #888, opened 2021-01-27
Fix rotation

### LINKED ISSUE #741: Image rotation is fundamentally broken
I believe the interactive image rotation feature as currently implemented is fundamentally broken.

The underlying problem is that the server modifies the content of the images but does not change their URLs in the process. Since the original images are likely to be already cached by the client, inconsistencies are pretty much guaranteed.

The front end tries to work around these issues by attaching a random query to the image URL to force a reload. But it does so only for the image currently being displayed (we are in the photo view at the time) and for the parent album, and not fully correctly at that (HiDPI is not handled properly).

If the image being modified happens to be the album thumb, that thumb will not change. Worse, if the user browses through the images and albums back and forth, the random query will eventually get dropped from the image URL and we are back to square one.

I think the only way to make it work reliably is for the server to rename the image files while rotating them (if symlinking is enabled, we can get away with simply generating new symlinks though), and for the front end to force a full reload of the albums tree, as we already do for operations such as image addition or deletion.

There is also another minor bug in the implementation in that it doesn't correctly handle missing intermediate image sizes (those fields should be empty in the DB but after the rotation they show `x` I believe).

kamil4:
Fixes #741.

The corresponding frontend changes are in https://github.com/LycheeOrg/Lychee-front/pull/244.

This ended up being much larger and more complex than I anticipated (and it took like a week :disappointed:). Once I started fixing that part of the code I kept uncovering small bugs and inconsistencies. I like to think that I fixed all of them or that I didn't expose any additional ones that I didn't fix but I wouldn't bet my money on it...

Here are the main fixes to the image rotation code:
* The rotated images are written under new file names, to avoid any inconsistencies due to client caching.  This requires API changes as we now need to return a full updated photo
description to the caller.
* The bug that left an `x` in database columns for missing intermediate image sizes is fixed and there's a migration in place that fixes the database as well.
* Intermediate image sizes are not rotated but instead are regenerated from the rotated original. Rotating intermediate images was incorrect as it results in wrong image sizes (e.g., a `180x240` `medium` image will be `240x180` after rotation instead of the expected `320x240`).
* Fixed the hard dependency on imagick.  Instead implemented a generic image rotation capability that also works with GD.
* Added checks for live photos and raw files, which rotation does not support.
* The photo file size is updated after the rotation as it may have changed.
* Duplicates are updated as well.

Other fixes:
* `import_via_symlink` had a number of problems with image auto-rotation, mainly having to do with the same incorrect assumption about rotating intermediate image sizes post-generation.  Even though the original is not rotated, we still need to flip `width` and `height` in `Photo` as the front end uses those for layout.  Enabling it would also unnecessarily prevent auto-rotation during regular uploads
* `is_uploaded_file` stops working correctly after the file is moved so it's best to invoke it once and cache the result in `Create`.
* We were using `tempnam` incorrectly, resulting in leftover empty temporary files after video and raw imports.
* `autoRotate` was only using the first engine for some reason, unlike the other image handling methods.
* Symlinking was broken for raw files.

### DIFF
--- a/tests/Feature/PhotosTest.php
+++ b/tests/Feature/PhotosTest.php
@@ -35,7 +35,7 @@ class PhotosTest extends TestCase
 		$file = new UploadedFile(
 			'public/uploads/import/night.jpg',
 			'night.jpg',
-			'image/jpg',
+			'image/jpeg',
 			null,
 			true
 		);
--- a/app/Actions/Photo/Create.php
+++ b/app/Actions/Photo/Create.php
@@ -52,6 +52,8 @@ class Create
 	/** @var array */
 	public $info;
 	public $livePhotoPartner;
+	/** @var bool */
+	public $is_uploaded;
 
 	public function add(
 		array $file,
@@ -80,6 +82,7 @@ class Create
 
 		// Set paths
 		$this->tmp_name = $file['tmp_name'];
+		$this->is_uploaded = is_uploaded_file($file['tmp_name']);
 		$this->photo_Url = md5(microtime()) . $this->extension;
 		$this->path_prefix = ($this->kind != 'raw') ? 'big/' : 'raw/';
 		$this->path = Storage::path($this->path_prefix . $this->photo_Url);
@@ -126,7 +129,7 @@ class Create
 			$res = $this->save($this->photo);
 		}
 
-		if ($delete_imported && !is_uploaded_file($this->tmp_name) && ($exists || Configs::get_value('import_via_symlink', '0') !== '1')) {
+		if ($delete_imported && !$this->is_uploaded && ($exists || Configs::get_value('import_via_symlink', '0') !== '1')) {
 			@unlink($this->tmp_name);
 		}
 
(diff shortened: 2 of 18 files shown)


ildyria:
Can you fix the CI and code smell ? :D

kamil4:
Working on it. CI is giving me a _very_ weird error though...

tests/Feature/PhotosTest.php
```diff
@@ -35,7 +35,7 @@ public function testUpload()
 		$file = new UploadedFile(
 			'public/uploads/import/night.jpg',
 			'night.jpg',
-			'image/jpg',
```
ildyria, line 38: it is weird that you need to modify that one. :thinking:

ildyria (approved):

tests/Feature/PhotosTest.php
kamil4: I thought that may have been the problem but turns out it wasn't. It was a failing test leaving a leftover photo in the database and affecting the next case. Still, `image/jpg` is not a valid mime type. It should say `image/jpeg`. So I think it's still worth fixing.

kamil4:
I believe the remaining code smell to be a false positive.

PR #895, opened 2021-01-29
fixing default license display

sjml:
Looks like this was just an oversight or maybe some copy-paste casualty. Just need to make sure the `get_license()` function is getting called like it is for the albums, and not just using the straight property.

### DIFF
--- a/app/Actions/Photo/Prepare.php
+++ b/app/Actions/Photo/Prepare.php
@@ -47,6 +47,8 @@ class Prepare extends SymLinker
 			$return['share_button_visible'] = '1';
 		}
 
+		$return['license'] = $photo->get_license();
+
 		return $return;
 	}
 }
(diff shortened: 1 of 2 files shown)


ildyria:
This was not an oversight but a SQL optimization. But indeed we were missing the call on pictures.

From _gitter_ for the explanation.
On line 66, you have `toReturnArray()` in `App/Actions/Album/Photos.php`
if you add the get_license here it will generate a lot of additional SQL query
This is why it is better to modify `App/Actions/Photo/Prepare.php`
see what has been done with `public`.

sjml:
Ah, ok. Thanks for explaining. How can we avoid hitting the database, though, in Prepare? If both the photo and album have a license of `"none"`, won't it have to hit the configuration anyway to lookup the default?

ildyria:
In `Photo/Prepare` it does not matter because this is called only for 1 picture.

ildyria (approved):

PR #898, opened 2021-01-30
fix delete

### LINKED ISSUE #890: Why I deleted selected albums and the photos still in the recent Album
I tested it many times. This problem only happens in deleting Multi-selected albums.
Deleting albums one by one, photos in that album will be deleted. But when I deleted all albums together, lots of photos still in the recent Album. And I check "Photos" in the SQL, there also have the data of these photos.  And a lot of photos also still in public/uploads/big.

Browser DevTools Network     Album::delete Response is true.

Lychee logs show no error.

ildyria:
fixes #890 

should be fixed with this one. (hopefully)


kamil4 (approved):
Looks sane and appears to handle correctly the cases that were failing before.

PR #907, opened 2021-02-05
Use file checksum as file name prefix

### LINKED ISSUE #903: Upload images in batches will cause images to repeat
When images are uploaded in batches, using a browser There's a chance that two images will repeat. 
The two images are not the same images before uploading.

Upload images in batches

**Screenshots**

*(Settings => Diagnostics or https://example.com/Diagnostics or `php artisan lychee:diagnostics`)*
`Diagnostics
    -------

    System Information
    --------------
    Lychee Version (release):        4.2.1
    DB Version:                      4.2.1
    
    APP_DEBUG:                       true
    
    System:                          WINNT
    PHP Version:                     7.4
    Max uploaded file size:          10000M
    Max post size:                   10000M
    MySQL Version:                   5.7.26
    
    Imagick Version:                 1799
    GD Version:                      bundled (2.1.0 compatible)

    Config Information
    --------------
    version:                         040201
    sorting_Photos_col:              title
    sorting_Albums_order:            ASC
    image_overlay:                   1
    landing_owner:                   John Smith
    landing_title:                   John Smith
    landing_subtitle:                Cats, Dogs & Humans Photography
    landing_facebook:                https://www.facebook.com/JohnSmith
    landing_flickr:                  https://www.flickr.com/JohnSmith
    landing_twitter:                 https://www.twitter.com/JohnSmith
    landing_instagram:               https://instagram.com/JohnSmith
    landing_youtube:                 https://www.youtube.com/JohnSmith
    landing_background:              dist/cat.jpg
    site_title:                      Lychee v4
    site_copyright_begin:            2019
    site_copyright_end:              2019
    additional_footer_text:          
    has_exiftool:                    1
    has_ffmpeg:                      1
    nsfw_warning_admin:              0`
### Browser and system
Chrome

kamil4:
Fixes #903

At least on Windows `microtime()` may be updated infrequently, so with parallel uploads it's possible to get conflicts.


ildyria (approved):
Looks good, but do I'm wondering whether a file hash would be better. This would solve collision problem (and duplicates).

kamil4:
An interesting point. We do, of course, already calculate the file hash... The file names would be a little longer than the current ones (sha1 is 160 bits vs 128 bits for md5) but I don't think that's a problem?

What do you think, should we bite the bullet and just do it? Should we leave the existing files as-is or rename them? I would be inclined to keep them as-is; less chance of any problems and people won't complain of a sudden spike in the size of their incremental backups :smiley:.

kamil4:
Also, we would still need to generate different file names later if the user chooses to rotate an image. So _File name prefix is the original file hash_ would not be a permanent property...

ildyria:
> An interesting point. We do, of course, already calculate the file hash... The file names would be a little longer than the current ones (sha1 is 160 bits vs 128 bits for md5) but I don't think that's a problem?

Nah, you can also truncate the sha1 hash to 128 bits if you wish so...

>  Should we leave the existing files as-is or rename them?

I would leave them as is and not bother with them.

> Also, we would still need to generate different file names later if the user chooses to rotate an image. So File name prefix is the original file hash would not be a permanent property...

I don't really care about this "property", especially if you already assume from the beginning that the file name is not necessarily the hash of the file. :)

haivala:
This is pretty handy in something like this: https://hashids.org/

ildyria:
@kamil4 shall we merge it while waiting for using the hash of the files or would you rather do that now?
(there is no more permanent solution than a temporary fix...)

kamil4:
> (there is no more permanent solution than a temporary fix...)

Exactly what I'm afraid of. For that reason I would suggest that we _don't_ merge the current implementation. I'll work on it soon.

kamil4:
OK, here it is finally, an alternative implementation based on file checksums. It made the `Create` case much simpler, the `Rotate` case moderately more complex, and it turns out that I also had to increase the length of `thumbUrl` in the DB to make it work...

Should be ready for merging!

haivala:
> This is pretty handy in something like this: https://hashids.org/

no comments on this?

ildyria:
> This is pretty handy in something like this: [hashids.org](https://hashids.org/)

It is interesting, but that would only be useful for the displayed ID for e.g., the album (instead of having this number string). I am not opposed to such modification, I just don't have the time.

haivala:
> > This is pretty handy in something like this: [hashids.org](https://hashids.org/)
> 
> It is interesting, but that would only be useful for the displayed ID for e.g., the album (instead of having this number string). I am not opposed to such modification, I just don't have the time.

OK. 
I think it could be used for everything. Every install could have their own salt, characters can be chosen and the length of the string can be modified..

ildyria:
> > > This is pretty handy in something like this: [hashids.org](https://hashids.org/)
> > 
> > 
> > It is interesting, but that would only be useful for the displayed ID for e.g., the album (instead of having this number string). I am not opposed to such modification, I just don't have the time.
> 
> OK.
> I think it could be used for everything. Every install could have their own salt, characters can be chosen and the length of the string can be modified..

But this would increase quite a lot the code complexity on server side :(

ildyria (commented):
Why do you need to extend the `thumbUrl` field ?

Can't you just truncate to the appropriate length `substr($str , 0, 32)` will give you the same length as with a md5 hash while using sha1.

You do not need to worry about collision resistance as it will be pretty much the same as with md5.

ildyria:
> But this would increase quite a lot the code complexity on server side :(

@haivala I just realized what you meant: using hashid directly from the start. I would rather avoid that from the start. While it makes the id shorter, it does not bring any other benefits. On the other side I would be quite in favor to improve the folder structure of Lychee for the picture to something similar to what is used for the cache by laravel: `/public/uploads/big/0a/0a12335b9283523.jpg`
That way we do not have 12000 pictures in the same folder...

kamil4:
> Why do you need to extend the `thumbUrl` field ?
> 
> Can't you just truncate to the appropriate length `substr($str , 0, 32)` will give you the same length as with a md5 hash while using sha1.

My lizard brain objected to doing it that way :smiley:. I think it goes back to my undergrad days when in _Numerical methods_ I was taught to _never_ use just a subset of bits of the output of a PRNG (or was it only about the lower bits? My memory fades; it was over 20 years ago :wink:). Anyway, that advice was about crappy LCGs, not about crypto hashes, so I guess you are right...

> You do not need to worry about collision resistance as it will be pretty much the same as with md5.

Does that mean _good_ or _bad_? :wink:

ildyria:
> My lizard brain objected to doing it that way . I think it goes back to my undergrad days when in _Numerical methods_ I was taught to _never_ use just a subset of bits of the output of a PRNG (or was it only about the lower bits? My memory fades; it was over 20 years ago ). Anyway, that advice was about crappy LCGs, not about crypto hashes, so I guess you are right...

> > You do not need to worry about collision resistance as it will be pretty much the same as with md5.
> 
> Does that mean _good_ or _bad_?

Assuming a Random Oracle, getting a collision on a 128 bit hash requires about 2^64 operations... In comparison, the computational effort to find a collision in SHA1 was 2^63.1 compression call. [source](http://shattered.io/static/shattered.pdf). We do not aim to be cryptographically secure on that one call anyway, so not much to worry about.

Anyway, what is not advised is reducing an output to a modulo a number, for example, if you wish to get a proper random dice simulator, doing `cprng() mod 6` is not going to end well... :)

haivala:
> > But this would increase quite a lot the code complexity on server side :(
> 
> While it makes the id shorter, it does not bring any other benefits.

Except that every Lychee install would have their own hashes and you would not need to think about collisions.

ildyria:
> Except that every Lychee install would have their own hashes and you would not need to think about collisions.

Not really. The collision problem is only significant if you try to gain access to a server with respect to password etc.

The only mitigation that pepper adds is to avoid an attacker whom you already gave upload rights to actually upload a file which will overwrite one of your picture. Not much of a concern I would say.

We do not aim to use this hash in a cryptographic secure way (we were even using `md5` of `microtime()` before), so the use a salt/pepper is out of scope in my opinion.

haivala:
Hashids is not in any way secure (as it states at the site).
salt would only be used to make different hashes between two instances of lychee

Issue #880, opened 2021-01-23
Not possible to unselect "hidden" under sharing options


Sesshoumaru-sama:
### Steps to reproduce the issue
Uploading new photo.
Under Sharing I can make the photo public, but I can not modify the other sharing options - e.g. Hidden is checkey by default. The dialog tells me that the information is shown just as information and the actual values can be "fine-tuned" by using global setting, however there are no global settings to be found.

**Screenshots**
![sharingopts](https://user-images.githubusercontent.com/10712111/105607544-98d1c680-5d9f-11eb-8601-b1752f0c9b91.png)

### Browser and system
Lychee v4.0.1
Chrome

kamil4:
This is working as intended; it's just that you've stumbled on a rather obscure feature and probably _not_ what you were looking for. Allow me to explain:

The _recommended_ way of sharing your photos with the public in Lychee is to upload them into an album (created using `+` / `New Album`) and then to make that _album_ public using its sharing settings. You will notice that all the visibility options will then be available for you to fine-tune. All the photos inside the album will inherit the visibility settings of the album.

Now, Lychee does also have a somewhat obscure feature of making individual _photos_ public: either photos that are not in any album (those show under `Unsorted`) or that are in a currently private album. This is what you were trying to do. Such photos can be made public, but they will always be _hidden_, meaning that they will not be visible when browsing through the albums in public mode. One needs to know the URL of that individual photo in the gallery in order to see it.

The comment about _global settings_ refers to the settings available under `Settings` / `More`; see also https://lycheeorg.github.io/docs/settings.html. There are specific settings there, such as `full_photo`, `downloadable`, and `share_button_visible`, that determine the visibility settings of photos individually made public. There is no global setting for `hidden`; as already mentioned, that one is always true for such photos.

I hope this helps. I realize that our manual is currently lacking when it comes to user-facing features.

Sesshoumaru-sama:
I don't understand the beviour then.
I uploaded a photo and made it public. Then I created a Tagged Album, which is public, but still the image is not public visible when I log out. What am I doing wrong?

Photo:
![img1](https://user-images.githubusercontent.com/10712111/105634042-6a65f100-5e5c-11eb-85c0-e62403b492f0.png)

Under the Public Folder the image is listed, but still with the Orange Eye:
![img2](https://user-images.githubusercontent.com/10712111/105634068-9d0fe980-5e5c-11eb-9b9d-904fc6de68af.png)

[...]

PR #932, opened 2021-02-19
Public photos hidden

kamil4:
This is inspired by the discussion in #880.

It adds a new config variable `public_photos_hidden`, which defaults to `1` (preserving the current behavior). If set to `0`, photos individually made public (rather than through an album) will be included in the results of a public search or in public tagged albums. Basically, they will no longer be treated as hidden (even though they still won't be findable through browsing; we could address that as well by making the _Public_ smart album, well, public -- I welcome your input if we should, and if so, what to name the config variable to enabled that, other than `public_public` :wink:).

The PR also fixes an unrelated regression where we weren't setting `public` to `2` for photos in public albums belonging to the current user (including admin).


ildyria (approved):
Looks good to me.

d7415 (approved):
Looks ok to me.

PR #956, opened 2021-04-02
Fix .lycheeignore file for import

tmp-hallenser:
the file '.lycheeignore' was ignored, file was read correctly, but variable not updated


d7415 (approved):
Makes sense

PR #971, opened 2021-04-19
fix livewire front end

ildyria:
More work.


Lars-:
@ildyria Because of this, I can't do the migrations anymore... It says "The use statement with non-compound name 'Helpers' has no effect". It literally worked an hour ago, but I decided to start over since I can't get it to work behind a reverse proxy...

ildyria:
Oh I'm sorry. Let me fix this super quick.

ildyria:
@Lars- Done.

PR #974, opened 2021-04-23
Fixes

ildyria:
So let's try this again...


ildyria:
And also tested locally (migration v3 -> v4).

PR #990, opened 2021-05-02
Fix return value for empty array of photos

nagmat84:
The response type for an empty array of photos should not artificially set to "false" because it breaks type correctness for clients which use strong typing.


kamil4:
So the current web frontend is not affected by this change?

nagmat84:
I searched for all occurrences where the JS code of the web frontend access `photos`. In all cases it has a combined test or "double" test like `if( photos === false || photos.length == 0 ) ...` or `if( photos !== false ) photos.forEach( ... )`. So, IMHO the frontend is fine.

nagmat84:
Anything I need to do, to get this PR merged?

PR #991, opened 2021-05-02
Add attribute 'filesize_raw' to entity 'Photo'

nagmat84:
JSON API reports raw (i.e. unformatted) size of file. This is used by clients which cache responses in local (shadow) database and which convert the file size into human readable format (with units kB, MB, etc.) at their own GUI layer.

This PR is related to [PR #260](https://github.com/LycheeOrg/Lychee-front/pull/260/files) of repo [Lychee-front](https://github.com/LycheeOrg/Lychee-front).


ildyria:
When you add a migration, you should also add a version update. You can do that easily by doing:
`make gen_minor` and then commit the file changed.

We have a code that checks that the version in version.md is the same as the database number, and request the migration when not up-to-date. This helps us to guarantee that the files are not ahead of the database, otherwise this code may crash because the property does not exist yet in the model if the migration is not applied.

nagmat84:
My fault, I didn't know that. I'll fix that probably the day after tomorrow and then update this PR.

ildyria:
> My fault, I didn't know that. I'll fix that probably the day after tomorrow and then update this PR.

No worries, and its not your fault. :)

database/migrations/2021_05_02_174300_add_filesize_raw_col.php
```diff
@@ -0,0 +1,35 @@
...
+
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		Schema::table(self::TABLE_NAME, function (Blueprint $table) {
+			$table->unsignedInteger(self::COLUMN_NAME)->default(0)->after('size');
```
kamil4: It wouldn't be too expensive in terms of lines of code to invoke `filesize` for each photo right here, thus ensuring that the data is actually available. Or would it simply take too much time for people with tens of thousands of photos?

nagmat84: Good question, I have no clue how expensive this would be. The main reason why I didn't do it, that I did not know how to do it, i.e. how to receive the current entry from DB, construct the correct path of the photo, etc.

ildyria: I think it might be too expensive, especially for users that have something like 50k pictures, you might get hit my a script time-out...

> I didn't do it, that I did not know how to do it,

The `url` field gives you access to the name of the picture. Then you can just use the `Storage` facade to extract the path: `Storage::path('big/' . $url);`

nagmat84:
Sorry for the delay. I increased the version (as requested).

> It wouldn't be too expensive in terms of lines of code to invoke filesize for each photo right here, thus ensuring that the data is actually available. Or would it simply take too much time for people with tens of thousands of photos?
>> I think it might be too expensive, especially for users that have something like 50k pictures, you might get hit my a script time-out...

As @ildyria assumes that it might be too expensive to run `filesize` for each photo during database migration, I haven't changed that. It is still possible to create that data running `./artisan lychee:exif_lens` from CLI explicitly.

ildyria:
Actually after further though. Here is a better solution:

1. Convert all the values of size back to Byte (or KB see below). This is quite easy to do and significantly faster than re-parsing the files for a lower cost of accuracy.
2. See if it is possible to change the column type from string to (Big?) Integer (see discussion KB below).
3. Fix the front-end (or the cast) so that the string is generated properly.

The benefit of this method are:

1. @nagmat84 get raw format directly.
2. **We can use SUM directly on the SQL so that we have a rough idea of the space used by the gallery (for the *future* statistics page).**

@nagmat84 do you want to have the Data in RAW (bytes) or an KB approximation is close enough ? The reason for this question is that in the Database, this saves us 10 bits per row and remove the necessity of using BigInteger (Max int 32 bit is 2'147'483'647, so that would mean a 2 GB max file size... Even though it is more probable that the system crashes before...)

nagmat84:
> Actually after further though. Here is a better solution:
> 
>   [...]

I am actually already working on it.

> @nagmat84 do you want to have the Data in RAW (bytes) or an KB approximation is close enough ?

Raw bytes are better. Other GUI frameworks (like Android, Qt, etc.) already have nice built-in converters which take a file size in raw bytes and convert it to some human readable string. So the raw counter would be welcomed.

With respect to the current JS frontend, I would adopt the PHP method `\App\Metadata\Extractor#size` to the JS frontend. It is not localized or anything, but it is neither a regression compared to the current situation.

@ildyria Any better ideas?

ildyria:
Nope. Screw the 32 bits users with more than 2GB files. :laughing:

database/migrations/2021_05_02_174300_add_filesize_raw_col.php
nagmat84: A modified the migration such that it uses the stored size as an rough estimation for the new column

ildyria (approved):

Issue #259, opened 2019-05-23
[Enhancement] Counters for Number of Views


tobogdan:
As the title says, if it's possible to add a viewcount tracker for each item. 

On what I am hosting (on 4.0.0), we'd like to know how many times certain items were viewed, so having some tracking mechanism for it would be great.

timo-reymann:
Would it be a possibility to allow users injecting javascript? This way it would be pretty simple to use tools like Matomo or Google Analytics to get exactly that.

If thats an option would also take a look implementing it and creating a PR.

Sounds similar to https://github.com/LycheeOrg/Lychee/issues/348#issue-498453161

ildyria:
@timo-reymann the ability to inject JS is already implemented:
https://github.com/LycheeOrg/Lychee/pull/1697

timo-reymann:
@ildyria Ah alrighty, didnt see that. Unfortunately the CSP is not allowing matomo code execution never the less :/

ildyria:
You can fix that by changing this file: https://github.com/LycheeOrg/Lychee/blob/master/config/secure-headers.php
I think we would accept a Pull request providing an optional support for that.

PR #1003, opened 2021-05-14
JSON response reports different size variants of the same photo more consistently

nagmat84:
This PR is related to [PR #259](https://github.com/LycheeOrg/Lychee-front/pull/259) of repository [Lychee-front](https://github.com/LycheeOrg/Lychee-front).

In essence all the different size variants (except for the original variant) are reported more consistently by JSON. A JSON response to `/api/Photo::get` now looks like this

```javascript
{
  ...
  "sizeVariants": {
    ... 
    "small2x": {    // If the particular size variant exists
      "width": 960,
      "height": 720,
      "url": "some-url"
    },
    ...
    "medium": null, // If the particular size variant does not exist
    ...
  }
}
```

Instead of

```javascript
{
  ...
  "medium2x": "some-url",
  "medium2x_dim": "42x42"
  ...
}
```

Also, the thumbnail and the double of a thumbnail are reported the same way, incl. their dimensions. This has the following benefits:

 - Improved uniformity of the JSON interface; thumbnails are not handled differently in JSON and attribute names are similar (i.e. no `thumbURL` vs. `medium2x`)
 - No "magic constants" for dimensions hard-coded into the frontend; the front-end must not know that thumbnails are 200x200 (400x400, resp.), because they are reported by the server, too
 - No "stringification" of attributes; width and height are reported as integers in all cases not as a merged string `42x42`.

Changes under the hood:

 - No need for `parseInt( dimensionString, 10)` and `substr`, `subpos` in the front-end to re-split the string of dimensions
 - Database change: The attribute width/height are stored as independent integer columns, not as a string column
 - The method `\Models\Extensions\PhotoCast#prepareLocationData` which essentially was a duplicate of `\Models\Extensions\PhotoCast#toReturnArray` is gone
 - Introduced a lot of class constants to avoid using "magic" string literals (such as `'thumb2x'` and alike) throughout the PHP code


app/Actions/Album/Extensions/LocationData.php
```diff
@@ -21,7 +21,7 @@ public function photosLocationData(Builder $photos_sql)
 
 		$return_photos = [];
 		$photo_counter = 0;
-		$photos = $photos_sql->select('album_id', 'photos.id', 'latitude', 'longitude', 'small', 'small2x', 'takestamp', 'thumb2x', 'thumbUrl', 'photos.title', 'type', 'url')
```
ildyria, line 24: The point of this was to reduce the memory footprint of Lychee :)
But nevermind. :)

app/Models/Extensions/PhotoCast.php
```diff
@@ -15,6 +16,40 @@ trait PhotoCast
 	 */
 	public function toReturnArray(): array
 	{
+		if (strpos($this->type, 'video') === 0) {
```
ildyria: use `->is_video()` instead (there should be a function like that in the code base).
the `type` contains the mimetype data of the *photo* and there is unfortunately 1 exception.

app/Models/Extensions/PhotoCast.php
```diff
@@ -15,6 +16,40 @@ trait PhotoCast
...
+		$thumbFileName2x = $this->thumb2x === '1' ? Helpers::ex2x($this->thumbUrl) : null;
+
+		// The original size is not stored in this sub-array but on the root level of the JSON response
+		// TODO: Maybe harmonize and put original variant into this array, too? This would also avoid an ugly if branch in SymLink#override.
+		$sizeVariants = [
+			Photo::VARIANT_THUMB => $this->serializeSizeVariant(
+				Photo::VARIANT_THUMB, $this->thumbUrl, Photo::THUMBNAIL_DIM, Photo::THUMBNAIL_DIM
+			),
+			Photo::VARIANT_THUMB2X => $this->serializeSizeVariant(
+				Photo::VARIANT_THUMB, $thumbFileName2x, Photo::THUMBNAIL2X_DIM, Photo::THUMBNAIL2X_DIM
```
ildyria:
```suggestion
				Photo::VARIANT_THUMB2x, $thumbFileName2x, Photo::THUMBNAIL2X_DIM, Photo::THUMBNAIL2X_DIM
```

app/Models/Extensions/PhotoCast.php
```diff
@@ -23,166 +58,75 @@ public function toReturnArray(): array
...
-			$photoUrl = $this->url;
-		}
-		$photoUrl2x = ($photoUrl !== '') ? Helpers::ex2x($photoUrl) : '';
-
-		$sizes = [
-			'medium' => $photoUrl,
-			'medium2x' => $photoUrl2x,
-			'small' => $photoUrl,
-			'small2x' => $photoUrl2x,
+		return ($width === null || $height === null || $fileName === null || $fileName === '') ? null : [
```
ildyria: I would prefer the use of a `if` instead of a ternary here as this improve the readability (Nitpicking).

I have nothing against ternary in general. :)

app/Models/Photo.php
```diff
@@ -113,6 +117,29 @@ class Photo extends Model
...
+	const VARIANT_THUMB2X = 'thumb2x';
+	const VARIANT_SMALL = 'small';
+	const VARIANT_SMALL2X = 'small2x';
+	const VARIANT_MEDIUM = 'medium';
+	const VARIANT_MEDIUM2X = 'medium2x';
+	const VARIANT_ORIGINAL = 'original';
+
+	/**
+	 * Maps a size variant to the path prefix (directory) where the file for that size variant is stored.
+	 * Use this array to avoid the anti-pattern "magic constants" throughout the whole code.
```
ildyria, line 132: :+1:

app/View/Components/Photo.php
```diff
@@ -27,8 +27,8 @@ class Photo extends Component
 	public $srcset2x = '';
 
 	public $layout = false;
-	public $_w = 200;
-	public $_h = 200;
+	public int $_w = \App\Models\Photo::THUMBNAIL_DIM;
```
ildyria: you can do
```php
use App\Models\Photo as PhotoModel;
```
on the top of the file. This will alias the path (damn I really miss this when I do Java).

app/View/Components/Photo.php
```diff
@@ -65,6 +65,7 @@ public function __construct(array $data)
 		$dim2x = '';
 		$thumb2x = '';
 
+		// TODO: The class Photo for the database model does not anymore contain the attributes `small`, `small_dim`, etc. Probably this code needs some fix/refactoring, too. However, where is this method invoked and what is the structure of the passed `data` array? (Could find any invocation.)
```
ildyria, line 69: This part is Work in Progress. It is part of the Livewire side. no worries I'll fix it.

database/migrations/2021_05_13_140700_refactor_size_variants.php
```diff
@@ -0,0 +1,150 @@
...
+				self::MEDIUM_HEIGHT_COL_NAME => $mediumHeight,
+				self::MEDIUM2X_WIDTH_COL_NAME => $medium2xWidth,
+				self::MEDIUM2X_HEIGHT_COL_NAME => $medium2xHeight,
+			]);
+		}
+
+		DB::commit();
+
+		Schema::table(self::PHOTOS_TABLE_NAME, function (Blueprint $table) {
+			$table->dropColumn([
```
ildyria, line 73: I don't remember, but I think SQLite does not really like to drop multiple column in one go. I may be wrong. The CI should catch that one anyway.

database/migrations/2021_05_13_213244_bump_version040301.php
```diff
@@ -0,0 +1,27 @@
...
+class BumpVersion040301 extends Migration
+{
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		Configs::where('key', 'version')->update(['value' => '040301']);
```
ildyria, line 15: This probably will need to be a `040302` as there is the other PR pending. ;)

ildyria (changes requested):
I did not test it yet. Really damn good job. 
Minor fix requested. Now on my way to read the Lychee-front one.

app/Actions/Album/Extensions/LocationData.php
nagmat84: OK, I already assumed that. However, fetching the object partially from DB comes with to problems: a) We cannot use `toReturnArray` to serialize a Photo model into JSON. The current conversion routines which convert booleans into strings throw an exception on null values, if those attributes have not been fetched. b) Reporting partial object to the front-end ist difficult to handle for front-ends which don't directly use the response for displaying purpose, but feed the response into a DB for caching purposes and then drive the UI from that local shadow DB. Essentialy, the frontend would need to know, if a null value is null, because it is really null (also in the backend DB) and hence must be updated in the local shadow DB or if it is just a null value because it has not been fetched and thus has been left out of the reponse. Of course, there are solutions for that problem.

In essence, I decided to not bother with that problem right now given the fact that the PHP backend needs some major refactoring anyway. (Future outlook: Don't covert model objects into response arrays at some abitrary point in code, but only at the very end just before sending the reponse to the client.) In prospect of this refactoring it is even easier to have only one method (`toReturnArray`) instead of two which need attention. So I decided to take that efficiency penalty.

app/Models/Extensions/PhotoCast.php
nagmat84: Great, I'll fix that. I tried to stay at close at the old code as possible and the old code did it that way. But of course, using `isVideo` is much cleaner. I'll update this PR.

app/Models/Extensions/PhotoCast.php
nagmat84: Damn. Silly copy&paste error. A fix is on the way.

app/View/Components/Photo.php
nagmat84: 👍 I'll incorporate that into the update of this PR.

app/View/Components/Photo.php
nagmat84: 👍 Thx

database/migrations/2021_05_13_140700_refactor_size_variants.php
nagmat84: I don't know. My local testing environment uses sqlite and my local development environment uses PostgreSQL. I have run `make test` and `artisan db:migrate` in both environment several times and didn't encounter an error, but may be it depends on the version of sqlite and only a particular version has problems with that. But I can easily fix that in order to be on the save side.

database/migrations/2021_05_13_213244_bump_version040301.php
nagmat84: Whatever PR is merged first, I will re-base the other PR on the updated master and then re-run `make minor_version` again to avoid conflicts.

app/Models/Extensions/PhotoCast.php
ildyria: You can just use the "commit suggestion" button. :D

database/migrations/2021_05_13_140700_refactor_size_variants.php
ildyria: The CI will catch it anyway as we do migration and rollback. So if the CI passes, don't bother with it. :)

database/migrations/2021_05_13_213244_bump_version040301.php
ildyria: :+1:

app/Actions/Album/Extensions/LocationData.php
d7415: As a rule I don't think we should be sending everything anyway unless explicitly requested, and if that hurts the memory footprint then even more so. I suspect there are more users on low memory systems than writing custom front ends! But I don't know what a back end refactor would look like, and I'd probably rather send more over the wire than have the code look a mess. I'll trust @ildyria's judgement.

nagmat84: @d7415: In general, I would agree with you. However, there is so much duplication of data right now in the code, that this is the least problem. There is one approach, IHHO the _right_, approach to solve this (and other) issues. But this is a completely different story and requires much more refactoring. I will post some thoughts about that to Gitter soon. In essence, this would entirely drop this hand-crafted implementations of `toReturnArray` and use `toJSON` of the Laravel middleware. So at the moment, I would leave it as it is.

nagmat84:
Damn it. After I accepted your change request by simply clicking on "accept change request and re-commit", some tests fail. I need to look into that in more detail tomorrow.

app/Models/Extensions/PhotoCast.php
nagmat84: After I did, the automatic tests failed. I'll have to undo this "commit suggestion" and investigate more thoroughly. Actually it should not fail.

nagmat84: Problem solved. It was a simply case problem in the name of the constant.

app/Models/Extensions/PhotoCast.php
nagmat84: Fixed in commit c513d92d.

app/Models/Extensions/PhotoCast.php
nagmat84: Fixed in commit c513d92d.

app/View/Components/Photo.php
nagmat84: Fixed in commit c513d92d.

ildyria (approved):

PR #1016, opened 2021-05-24
Refactor timestamps

nagmat84:
This PR is related to [PR #262](https://github.com/LycheeOrg/Lychee-front/pull/262) of the repository [Lychee-front](https://github.com/LycheeOrg/Lychee-front).

In essence, all datetime strings of the REST API are serialized as ISO 8601 (aka JSON time) strings incl. time offsets. This also holds for the "min/max takedate" of the albums which has previously been sent as a localized string including the year and month only. Moreover, for the take date of photos the original timezone is reported to the client, too.

Under the hood a lot had to be re-worked (unfortunately this PR became longer than I wanted it to be):

 - At the database layer all datetime without timezone information are stored as UTC. To this end, the model class converts datetime object to UTC before they are de-hydrated into the DB and converts them back to the application's default timezone after hydration from the database, unless no other explicit timezone information is given
 - The datetime when a photo has been taken uses an extra column to store the original timezone and a custom Laravel cast which manages dehydration/hydration from/to the database.
 - The REST API has been cleaned up, the timedate attributes are not reported twice anymore (previously they were reported as a localized string and as a "raw" timestamps in seconds since epoch). In particular the previously returned "raw" timestamp was wrong anyway, because the backend used the wrong timezone to create it. This is fixed now.


app/Metadata/Extractor.php
```diff
@@ -182,60 +183,186 @@ public function extract(string $filename, string $kind): array
...
+					// has been recorded.
+					// So we don't need to do anything.
+					//
+					// The only known example are the mov files from Apple
+					// devices; the time zone will be formatted as "+01:00"
+					// so neither of the two conditions above should trigger.
 				}
 			}
-
-			// We need to make sure that the takestamp is between '1970-01-01 00:00:01' UTC and '2038-01-19 03:14:07' UTC.
```
ildyria, line 227: This part is required because some photos have the wrong exif data... :(
It was added to fix a bug (I don't remember which one but a `git blame` should give us this information).

ildyria: See here: https://github.com/LycheeOrg/Lychee/pull/414 and https://github.com/LycheeOrg/Lychee/issues/413

nagmat84: OK, I will check and if I know the reason why it was added, I will re-add it. However, what do we do with photos which truly have been taken before 1970 or after 2038?

ildyria: I'm not sure about the after 2038. For the before 1970, I'm pretty sure we will not have this information in a digital picture. ;)

ildyria (approved):
Aside from the issue mentioned, at least on reading it looks damn good!
Just double check with #413

app/Metadata/Extractor.php
nagmat84: I checked the issue, unfortunately the links to the test photos in the [post](https://github.com/LycheeOrg/Lychee/issues/413#issuecomment-568795209) are dead. I tried to create some own GIF and PNG with various values for the tags `dateTimeOrigiginal` and `xmp:dateTimeOrigiginal` but could not reproduce the error :-(

> For the before 1970, I'm pretty sure we will not have this information in a digital picture.

You may have. Some national archives digitize historic photos and set the EXIF data correctly.

ildyria: Hopefully @alex-phillips is still around. I will also try to poke him on discord. :+1:

nagmat84: Thanks. It would be really great to have this photos and even put them into the test base (if @alex-phillips agrees). This way we can be sure that there is no regression. If I get a link to those photos, I will write tests for them.

alex-phillips: @ildyria @nagmat84 Unfortunately, I don't think I was involved in the date change here, just reported the error and opened the issue. Here is one of the problematic images I referenced in the issue if it helps.

@tmp-hallenser may have more info, but my guess is that exif field needed to be checked, but not sure why the date bounds were added.

6a4bde8ad79732826bf1df9bd11bed4b

nagmat84: @alex-phillips Thank you for giving me one of the problematic images. Unfortunately, everything works smoothly and I cannot reproduce the error. However, I also noticed that the EXIF data for this image does not contain a tag `dateTimeOriginal` at all. This means, the removed code path (checking for boundaries between 1970 and 2038) would have never been taken anyway and are totally unaffected by this image. @alex-phillips Are you sure that this one really has been one of the problematic images?

@ildyria How do we proceed?

ildyria: If the bug is no-longer reproducible, I would just move forward.

nagmat84: Great. In that case I like to kindly ask @ildyria to also review the changes to the frontend and accept them if everything is fine. Then I can proceed with the merge.

PR #1017, opened 2021-05-29
Refactor timestamps anew

nagmat84:
This is a new attempt for the unfortunate, previous [PR #1016](https://github.com/LycheeOrg/Lychee/pull/1016).

Changes since the previous PR:

 - Migration now sensitive to different DB back-ends and does not convert times for MySQL (the migration is the part which has changed most since previous PR).
 - All model classes which use timestamps now store their timestamps in UTC, not only `Photo` and `Album` as discussed on Gitter
 - The class `PatchedBaseModel` became a trait `UseUTCTimes` because it is also required for the model `User` which does not directly extend from Eloquent's `Model` class
 - Times are now formatted acc. to PHP `DateTimeInterface::ATOM`and not `DateTimeInterface::ISO8601`, because the latter does not conform to ISO 8601 despite its name.


kamil4:
FYI, I'm away from my computer until Monday evening. Realistically, it probably won't be until Tuesday that I can check it out.

Kamil

nagmat84:
Closed this PR and created PR #1022 instead which is a branch from the Lychee repo.

PR #1020, opened 2021-05-31
Fix for issue 1019

nagmat84:
This bug fix solves to issue:

- For those users who haven't updated yet and are below version 4.3.1, the original migration file has been changed to create a `unsignedBigInt` right away in order to avoid migration errors for those who have files larger than >4GB in their database
- For those users who already have upgraded successfully, there is a new migration file which changes the column type from `integer` to `unsignedBigInt`such that those user won't experience errors in the future.

### DIFF
--- a/database/migrations/2021_05_02_174300_add_filesize_raw_col.php
+++ b/database/migrations/2021_05_02_174300_add_filesize_raw_col.php
@@ -20,7 +20,7 @@ class AddFileSizeRawCol extends Migration
 	public function up()
 	{
 		Schema::table(self::TABLE_NAME, function (Blueprint $table) {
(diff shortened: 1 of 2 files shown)


kamil4 (approved):
Looks sane. I ran it here (on a 64-bit system with MySQL) and it worked as expected though I don't have any sufficiently large files that would trigger the issue.

I'm kind-of freaked out that people put >2GB movie files in Lychee but I guess that's just me... :smile:

d7415:
> I'm kind-of freaked out that people put >2GB movie files in Lychee but I guess that's just me... smile

Not just you! But I'm not surprised, either.

PR #1022, opened 2021-06-01
Refactor timestamps anew

nagmat84:
This is a new attempt for the unfortunate, previous [PR #1016](https://github.com/LycheeOrg/Lychee/pull/1016).

Changes since the previous PR:

 - Migration now sensitive to different DB back-ends and does not convert times for MySQL (the migration is the part which has changed most since previous PR).
 - All model classes which use timestamps now store their timestamps in UTC, not only `Photo` and `Album` as discussed on Gitter
 - The class `PatchedBaseModel` became a trait `UseUTCTimes` because it is also required for the model `User` which does not directly extend from Eloquent's `Model` class
 - Times are now formatted acc. to PHP `DateTimeInterface::ATOM`and not `DateTimeInterface::ISO8601`, because the latter does not conform to ISO 8601 despite its name.


developer-notes.md
```diff
@@ -0,0 +1,135 @@
...
+For DBMS which support to set an explicit timezone for the connection
+between the DBMS server and the DBMS client of the application
+(e.g. MySQL and PostgreSQL), the timezone of the connection is set to "UTC".
+This setting resides in `./config/database.php` and should not be changed.
+The default timezone of the application as set in `./config/app.php` or
+`./.env` resp. is independent of the DBMS timezone and may be configured.
+At the level of the PHP application all date/time values are `Carbon` objects.
+Conversion of the timezone of a `Carbon` instance and UTC happens at the
+application layer during hydration/dehydration from/to the DBMS
+(that's where `\App\Models\PatchedBaseModel` comes into play).
```
kamil4: Out of date? I believe you use a trait now.

developer-notes.md
```diff
@@ -0,0 +1,135 @@
...
+    Basically, `timestamp` and `datetime_tz` should be considered to be
+    buggy.
+    
+# Information on Date/Time Attributes and How Date/Time Values are Handled by the Lychee Application
+
+## Summary
+
+All date/time values are stored at the DB back-end relative to UTC without
+explicit timezone information.
+for communication between the PHP application and the DB back-end all
```
kamil4: Line should start with uppercase.

developer-notes.md
```diff
@@ -0,0 +1,135 @@
...
+Let us ignore the fact that PostgreSQL surpasses any other DB with respect to range and precision for each type, we have the following mapping between the types with respect to functional behaviour (i.e. auto-conversion vs. no auto-conversion):
+
+     # | MySQL     | PostgreSQL                  | SQLite
+    ---+-----------+-----------------------------+----------
+     1 | DATETIME  | TIMESTAMP WITHOUT TIME ZONE | DATETIME
+     2 | TIMESTAMP | TIMESTAMP WITH TIME ZONE    | n/a
+
+### Conclusion
+
+The Lychee application uses option 1, i.e. `DATETIME` for MySQL and `TIMESTAMP WITHOUT TIME ZONE` for PosgreSQL for the simple reason that MySQL provides the larger range for that type.
```
kamil4: Typo: PosgreSQL -> PostgreSQL

developer-notes.md
```diff
@@ -0,0 +1,135 @@
...
+Otherwise, the application had to ensure that there are no date/time values before 1970 and after 2038 or MySQL will throw SQL exceptions during INSERT/UPDATE.
+The lack of auto-conversion by the DB for the chose option 1 is not a problem.
```
kamil4: Typo: chose -> chosen

database/migrations/2021_06_01_181900_refactor_timestamps_anew.php
```diff
@@ -0,0 +1,563 @@
...
+	 */
+	protected function fixPagesTable(): void
+	{
+		DB::table('pages')
+			->where('title', '=', 'gallery')
+			->where('link', '=', '/gallery')
+			->whereNull('created_at')
+			->update([
+				'created_at' => '2019-02-21 11:43:56',
+				'updated_at' => '2019-02-21 11:43:56',
```
kamil4: I realize I'm bikeshedding here but wouldn't it make more sense to simply set them to current time?

database/migrations/2021_06_01_181900_refactor_timestamps_anew.php
```diff
@@ -0,0 +1,563 @@
...
+		Schema::table($tableName, function (Blueprint $table) {
+			$table->renameColumn(self::CREATED_AT_COL_NAME, self::CREATED_AT_COL_NAME . '_tmp');
+		});
+		Schema::table($tableName, function (Blueprint $table) {
+			$table->renameColumn(self::UPDATED_AT_COL_NAME, self::UPDATED_AT_COL_NAME . '_tmp');
+		});
+		Schema::table($tableName, function (Blueprint $table) {
+			$table->timestamps();
+		});
+		$needsConversion = $this->needsConversionOnUpgrade();
```
kamil4: Shouldn't this be `needsConversionOnDowngrade`?

database/migrations/2021_06_01_181900_refactor_timestamps_anew.php
```diff
@@ -0,0 +1,563 @@
...
+	 */
+	protected function downgradePhotos(): void
+	{
+		Schema::table(self::PHOTOS_TABLE_NAME, function (Blueprint $table) {
+			$table->timestamp(
+				self::PHOTO_TAKESTAMP_COL_NAME,
+				self::DATETIME_PRECISION
+			)->nullable(true)->default(null);
+		});
+		$needsConversion = $this->needsConversionOnUpgrade();
```
kamil4: Same comment about `needsConversionOnDowngrade`...

kamil4 (changes requested):
And I thought that the _previous_ DB migration was a monster at ~300 lines :wink:. This one is almost double that! :astonished:

I tested it with MySQL, both up and down migration paths.

The up part is flawless, although it took over a minute on my relatively decent host. After adding some instrumentation I found that it was the conversion of `logs` that took forever. I guess I haven't cleaned those up in a while...

The down part works as well, but seemingly only accidentally? You'll see in my comments to the migration file that I identified two potential issues -- basically, `needsConversionOnDowngrade` is never invoked. However, that actually makes it work _correctly_ with MySQL. We _don't_ want to do the conversion -- because the session timezone is set to `+00:00`, we need to be storing UTC data in `timestamp` columns, so what we read from `datetime` columns is perfect as-is.

developer-notes.md
nagmat84: You're are right. I had to switch from a base class to a trait, because I needed that for the `User` model, too. I forgot to update the documentation. It's fixed now.

developer-notes.md
nagmat84: Fixed.

developer-notes.md
nagmat84: Fixed.

developer-notes.md
nagmat84: Fixed.

database/migrations/2021_06_01_181900_refactor_timestamps_anew.php
nagmat84: You are right. That's a copy-&-past error. As discussed on Gitter this accidentally did the right thing, because we must not use a different method during downgrade than during upgrade. Removed `needsConversionOnDowngrade` entirely and renamed `needsConversionOnUpgrade` to `needsConversion`.

database/migrations/2021_06_01_181900_refactor_timestamps_anew.php
nagmat84: See other comment. Is fixed now.

database/migrations/2021_06_01_181900_refactor_timestamps_anew.php
nagmat84: I don't mind. I thought that it would be more correct to use the date/time which would have been used by the original migration which created that entry, but missed to set the dates back then. But actually I don't care.

What speaks for using the current date/time?

kamil4: Yeah, ideally we would set it to the time when the original migration ran. But we don't know when it was. For "old" users it may have been 2019, but for users who are just starting, it will have been "five seconds ago".

kamil4 (approved):

ildyria (approved):
Code wise, it looks good. Now testing...

database/migrations/2021_06_01_181900_refactor_timestamps_anew.php
nagmat84: After I had to learn that there are actually people who use this feature and create more entries manually (--> @ildyria), I incorporated @kamil4 suggestion, because we cannot know when those entries were created.

PR #1028, opened 2021-06-06
Fix issue 989

### LINKED ISSUE #989: JSON response does not set attributes "max_takestamp" and "min_takestamp" for the current album of a resposne
The JSON response for `api/Album::get` always contains empty string for the attributes `max_takestamp` and `min_takestamp` for the album which has been queried. The attributes `max_takestamp` and `min_takestamp` are only filled with correct values if the album object is included in the array of sub-albums.

**Expected behaviour:** Every album entity should always correctly report those attributes, i.e. the attributes should always be set, if the album contains at least one photo.

**Example:** Assume we call `api/Album::get` for album ID "16178764072801" and the album contains a photo. Then the response looks like:
```json
    {
      "id": "16178764072801",
      "title": "Test Album With Sub-Albums And Photos",
      ...
      "min_takestamp": "",    /* Error: Always empty string although this album contains a photo */
      "max_takestamp": "",    /* Error: Same here, expected "Sep 2007", see photo below */
      ..
      "has_albums": "1",
      "albums": [
        {
          "id": "16178766430089",
          "title": "Test Sub-Album",
          ...
          "min_takestamp": "Oct 2010",  /* This is OK */
          "max_takestamp": "Dec 2015",  /* This is OK */
          ...
        }
      ],
      "photos": [
        {
          "id": "16178766639428",
          "title": "Rheinfall_tourism4",
          ...
          "takedate": "02 September 2007 at 12:44",
        }
      ],
      "num": "1"
    }
```
Also note, if we call `api/Album::get` for the sub-album above (with ID "16178766430089"), then the response for that album will also contains empty strings for `max_takestamp` and `min_takestamp` instead of "Oct 2010" and "Dec 2015".

**Background**

At the PHP side there is `App\Models\Extensions\AlbumQuery` which setups a SQL query that sets `max_takestamp` and `min_takestamp`. However, it seems as if this the method `AlbumQuery::initQuery()` is only invoked as part `App\Models\Extensions\AlbumGetters::get_children()`. But `AlbumQuery::initQuery()` is not called during `App\Factories\AlbumFactory::make(string $albumId)` which only invokes `Album::findOrFail($albumId)` and runs the default SQL query for an entity.

Unfortunately, I am less of a Laraval/Eloquent programmer to fix that issue myself. I am still try to figure out how all that pieces work together.

On top, the query which is created by `AlbumQuery::initQuery()` looks suspiciously to me
```php
    return self::with(['owner', 'cover'])->select('albums.*')
      ->addSelect([
        'max_takestamp' => Photo::leftJoin('albums as a', 'album_id', '=', 'a.id')
          ->select('takestamp')
          ->where('albums._lft', '<=', DB::raw('a._lft'))
          ->where('a._rgt', '<=', DB::raw('albums._rgt'))
          ->whereNotNull('takestamp')
          ->orderBy('takestamp', 'desc')
          ->limit(1),
        'min_takestamp' => Photo::leftJoin('albums as a', 'album_id', '=', 'a.id')
          ->select('takestamp')
          ->where('albums._lft', '<=', DB::raw('a._lft'))
          ->where('a._rgt', '<=', DB::raw('albums._rgt'))
          ->whereNotNull('takestamp')
          ->orderBy('takestamp', 'asc')
          ->limit(1),
      ]);
```
Why those additional WHERE-clauses (e.g. `albums._lft <= a._lft`, etc.)? We pick the album `a` whose id `a.id` equals the id of the outer album `album`. IMHO, these additional WHERE clauses should always be true, i.e. they are not harmful, but superfluous. But may be I don't get the idea of those clauses.

**Why does it matter**

If one implements a client which caches all JSON responses in a local database it is important that consistent results are always returned for the same entity. For the web frontend this does not matter. The web frontend does not cache any data and does not display the `min_takestamp` and `max_takestamp` of the current album those attributes might not be necessary. However, if the client uses the most recent response of an entity to refresh its locally cached copy of that entity, attributes must not be null if they are not truly null at the database.

nagmat84:
Fixes #989

To avoid the same bug to re-appear in the future, the necessity for a developer to explicitly remember to call `initQuery` on an `Album` has been removed. The developer cannot forget to query those attributes, because they will be registered when the model is booted.


database/migrations/2019_12_15_0700_add_share_button_visible_option.php
```diff
@@ -21,7 +21,9 @@ public function up()
 			$table->boolean('share_button_visible')->after('downloadable')->default(false);
 		});
 
-		Album::where('id', '>', 1)
+		Album::query()
+			->withoutGlobalScopes()
+			->where('id', '>', 1)
```
ildyria: Why is `where('id', '>', 1)` needed?  :thinking: 
isn't this always true?

nagmat84: I haven't spent a thought on that. I did not change that line. But probably you are right. It seems very superfluous.

Issue #1011, opened 2021-05-17
[Enhancement] Notifying user of new uploads


pokemonbattledroid:
I looked around but couldn't find anything regarding this.  Is there a way to implement a feature where if a shared album gets new photos/videos, it automatically emails the user to say there is new content?

ildyria:
Sure there is for sure a way to implement it. The question is more "_Is there someone to do it?_" :confused: We are kinda short staffed here but we do accept contributions [read more here](https://lycheeorg.github.io/docs/contributions.html).

Note this also would require users to add their mails address etc. (while right now we only store their username & password). So adding such a functionality would also be a significant refactoring also in the User section to add the mail management.

d7415:
If adding email notifications it might make sense to at least consider push notifications at the same time. Even if it's just generic webhook support which many of them support.

Other thoughts to consider:
- Should pictures imported from the CLI be included
- Should there be a notification for every photo? Or grouped together somehow? Or a cooldown between notifications?

pokemonbattledroid:
I wish I could contribute, but instead I'm just a user and it seems like I posted a question without any regard for the complexity around personal information, security, etc.  Sorry!!

I like the ideas though and notifications!

ildyria:
>  it seems like I posted a question without any regard for the complexity around personal information, security

Indeed the mail address is personal data and would also require us to add a GDPR warning.
This link actually highlight the unexpected added complexity of such request :sweat_smile: :
https://www.litmus.com/blog/5-things-you-must-know-about-email-consent-under-gdpr/

PR #1049, opened 2021-07-05
New Photos Email Notification

micahhenshaw:
Hi there, this PR is because I wanted to be able to receive an email when new photos were added to albums that have been shared with me. I think it would probably also address issue #1011. 

There is also an accompanying Front End PR.

How I have set this up is to use the laravel notification system to send via SMTP, so those env settings will need to be setup. Also, it is set to send out emails once a week via cron, so the laravel cron job would need to be setup. But neither is required so if people don't want to set those up then it won't effect the operation of Lychee.

There is an admin setting to toggle on the notification option, which enables a user level menu item to change their email. This is purely opt-in, so users can clear their email if they don't wish to receive the emails anymore.

The email that gets sent out has been branded lychee, and will group up the last weeks worth of notifications and send them all out in one email to each user that has an email setup.

The notifications will go to the site admin, album owner & anyone who the album is shared with, provided their email has been added and they were not the one adding the photo to the album in the first place.

Happy to make any changes you would like! Thanks!


ildyria:
Hi @micahhenshaw,

thank you for your PR.
As you noticed, I already added you as an external contributor so you can execute the CI more easily.

Can you fix the tests so they pass again ?

you can execute them locally with `make test`. Though I suggest you use for example sqlite for your test database if you don't want to mess up with your live one. :)

That being said, I skimmed through your work, and I have to say I am not familiar with the Notification and Mail system of Laravel, but if it works that all great.

I am really happy with the quality of your code and the efforts you took to follow our folder architecture. This is greatly appreciated.

Thank you for doing the pull request in the Front-end. I haven't had time to look at it yet but will do once this one is not failing. :)

app/Http/Controllers/Administration/UserController.php
```diff
@@ -65,4 +66,49 @@ public function create(Request $request, Create $create)
...
+
+	/**
+	 * Return the email address of a user.
+	 *
+	 * @return string
+	 */
+	public function getEmail()
+	{
+		$id = AccessControl::id();
+		$user = User::findOrFail($id);
```
ildyria: If I am not wrong, those two lines can be directly replaced with 
```php
$user = AccessControl::user();
```

database/migrations/2021_06_30_121651_add_email_to_users_table.php
```diff
@@ -0,0 +1,32 @@
...
+{
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		Schema::table('users', function (Blueprint $table) {
+			$table->char('email', 100)->after('password')->nullable($value = true);
```
ildyria: the `$value = true` is unnecessary. :)

database/migrations/2021_06_30_122229_config_new_photos_notification.php
```diff
@@ -0,0 +1,38 @@
...
+	 */
+	public function up()
+	{
+		defined('BOOL') or define('BOOL', '0|1');
+
+		DB::table('configs')->insert([
+			[
+				'key' => 'new_photos_notification',
+				'value' => '0',
+				'confidentiality' => 0,
```
ildyria, line 22: I think this should be of level 2 (logged in) if I am not wrong.

app/Console/Commands/PhotosAddedNotification.php
```diff
@@ -0,0 +1,82 @@
...
+					if ($photo && $photo->thumbUrl) {
+						if (!isset($photos[$photo->album_id])) {
+							$photos[$photo->album_id] = [
+								'name' => $photo->album->title,
+								'photos' => [],
+							];
+						}
+
+						$photos[$photo->album_id]['photos'][$photo->id] = [
+							'thumb' => config('app.url') . '/uploads/thumb/' . $photo->thumbUrl,
```
ildyria: Instead of using `config('app.url')` 
this should be using `Storage::url(...)`
There is no guarantee that the images are stored on the same server as Lychee in the future.
You may want to use a S3 bucket or something (when supported (hopefully)).

I would suggest to have a look at:
https://github.com/LycheeOrg/Lychee/blob/master/app/Models/Extensions/PhotoCast.php#L107

Further more, thumbs are only square, you may consider to send the small version which preserve aspect ratio. :)

database/migrations/2021_06_30_122229_config_new_photos_notification.php
micahhenshaw: Yeh, I did try level 2 originally, however it doesn't seem to be allowing the config option to be accessible from a logged in (restricted) user, which is used to show/hide the notification menu item if the setting is enabled or not. All related to the Front-End. I could be missing something?

ildyria: Ummmm what about level 1 then ? :s
I don't remember exactly which level is which access, this should probably be clarified in the docs (our fault) :)

micahhenshaw: You'll know it better than me :) but when I've looked through it, I could only see logged in users receiving public configs (https://github.com/LycheeOrg/Lychee/blob/master/app/ModelFunctions/ConfigFunctions.php#L51) and admins receiving the admin configs. Although there is a scope for it: https://github.com/LycheeOrg/Lychee/blob/master/app/Models/Configs.php#L261

app/Actions/User/Notify.php
```diff
@@ -0,0 +1,45 @@
...
+			if ($album_id) {
+				$album = Album::find($album_id);
+			} else {
+				$album = Album::find($request->album_id);
+			}
+
+			$album_users = $album->shared_with;
+
+			$owner = User::find($album->owner_id);
+			$album_users = $album_users->push($owner);
```
kamil4: Is this assignment needed at all? I.e., doesn't `push` simply modify `$album_users`?

app/Actions/User/Notify.php
```diff
@@ -0,0 +1,45 @@
...
+
+			if ($album->owner_id != 0) {
+				$admin = User::find(0);
+				$album_users = $album_users->push($admin);
```
kamil4: Same question as 4 lines earlier...

app/Console/Commands/PhotosAddedNotification.php
```diff
@@ -0,0 +1,85 @@
...
+								'name' => $photo->album->title,
+								'photos' => [],
+							];
+						}
+
+						logger(Storage::url(Photo::VARIANT_2_PATH_PREFIX[Photo::VARIANT_THUMB] . '/' . $photo->thumbUrl));
+
+						$photos[$photo->album_id]['photos'][$photo->id] = [
+							'thumb' => Storage::url(Photo::VARIANT_2_PATH_PREFIX[Photo::VARIANT_THUMB] . '/' . $photo->thumbUrl),
+							'link' => config('app.url') . '/r/' . $photo->album_id . '/' . $photo->id,
```
kamil4: What does this `/r/` do? Is that something to do with RSS?

app/Http/Controllers/Administration/UserController.php
```diff
@@ -65,4 +66,47 @@ public function create(Request $request, Create $create)
...
+			$request->validate([
+				'email' => 'email|max:100',
+			]);
+		}
+
+		$user = AccessControl::user();
+
+		$user->email = $request->email;
+
+		if ($request->email = '') {
```
kamil4: Shouldn't this be `==` or even `===`?

resources/views/emails/photos-added.blade.php
```diff
@@ -0,0 +1,34 @@
+@component('mail::layout')
```
kamil4, line 1: So does this file generate the emails that get sent? Are those HTML-only emails?

routes/web.php
```diff
@@ -114,6 +114,8 @@
 Route::post('/api/User::Save', [Administration\UserController::class, 'save'])->middleware('admin');
 Route::post('/api/User::Delete', [Administration\UserController::class, 'delete'])->middleware('admin');
 Route::post('/api/User::Create', [Administration\UserController::class, 'create'])->middleware('admin');
+Route::post('/api/User::UpdateEmail', [Administration\UserController::class, 'updateEmail'])->middleware('read');
+Route::post('/api/User::GetEmail', [Administration\UserController::class, 'getEmail'])->middleware('read');
```
kamil4: I don't think `read` is the correct middleware here... I would think it should be `login` for the second one, and as for the first one -- probably `login` as well? For the first one we should probably be checking if the account isn't restricted ("locked") inside `updateEmail` (or are you doing it already?).

app/Console/Commands/PhotosAddedNotification.php
ildyria: @kamil4 it is redirect. :)
https://github.com/LycheeOrg/Lychee/blob/master/routes/web.php#L41

It is something I added a while back because of Instagram.
If you share something like https://photography.example.com/gallery#1234567 Instagram will break the URL into 2, one will be the link to the site, the other one will be seen as an instagram Hashtag. Stupid but that is the solution I came up with.

resources/views/emails/photos-added.blade.php
ildyria: I missed that one! :scream:

micahhenshaw: Yes that will generate a html email, it has a text version, but its not very useful as the email is mostly just small images

app/Http/Controllers/Administration/UserController.php
micahhenshaw: Yeh, good eye, thanks

ildyria: I missed that one! :scream:

routes/web.php
micahhenshaw: I haven't got a check if a user is locked, but I can add that to the command to avoid sending to locked accounts & to the notify action to avoid creating notifications for locked accounts.

micahhenshaw: Ill change the middleware to `login` Users that have restricted accounts will still be able to receive the notification, so would still need access to update their email

app/Console/Commands/PhotosAddedNotification.php
kamil4: Yeah, in retrospect the use of `#` wasn't a very good idea I guess... I realize we can't just unilaterally change it as it would break existing URLs that existing users may rely on, but what about making it an option, one that is on by default for new installations?

app/Http/Controllers/Administration/UserController.php
kamil4: I was surprised to see it as I was sure one of the static checkers would flag it?

app/Console/Commands/PhotosAddedNotification.php
ildyria: The use of `#` was necessary because Lychee aimed to be a one page website. And yes I did not chose it, this is very **old** legacy. :|

(BTW this `#` access is dropped with livewire at the moment).

app/Http/Controllers/Administration/UserController.php
ildyria: I would have expected SonarCloud to flag that one indeed.

micahhenshaw:
@ildyria Is there anything else you would like changed on this PR? I believe the only thing that I haven't updated is the confidentiality level, where the configs don't seem to get passed to a regular user unless set at public level.

app/Actions/User/Notify.php
```diff
@@ -0,0 +1,45 @@
...
+use App\Notifications\PhotoAdded;
+use Illuminate\Support\Facades\Notification;
+
+class Notify
+{
+	public function do(Photo $request, $album_id = null)
+	{
+		$settings = Configs::get();
+
+		if ($settings['new_photos_notification']) {
```
ildyria: This and the two lines above need to be changed for:
```suggestion
		if (Configs::get_Value('new_photos_notification','1') == '1') {
```
This ensure that you don't re-query the config and this provides a default value.

app/Console/Commands/PhotosAddedNotification.php
```diff
@@ -0,0 +1,85 @@
...
+	/**
+	 * Execute the console command.
+	 *
+	 * @return int
+	 */
+	public function handle()
+	{
+		$settings = Configs::get();
+
+		if ($settings['new_photos_notification']) {
```
ildyria: Same as above, this and the two lines above need to be changed for:
```suggestion
		if (Configs::get_Value('new_photos_notification','1') == '1') {
```
This ensure that you don't re-query the config and this provides a default value.

app/Http/Controllers/Administration/SettingsController.php
```diff
@@ -300,6 +300,23 @@ public function setLocationShowPublic(Request $request)
...
+	 *
+	 * @return string
+	 */
+	public function setNewPhotosNotification(Request $request)
+	{
+		$request->validate(['new_photos_notification' => 'required|string']);
+
+		return Configs::set(
+			'new_photos_notification',
+			$request['new_photos_notification']
```
ildyria: Ideally you want to check that the `new_photo_notification` is either `1` or `0`.

ildyria (changes requested):
Minor changes requested and should be good to go.

ildyria:
Do note that once this is merged, we will most likely bump the version to 4.3.5 :+1:

app/Console/Commands/PhotosAddedNotification.php
micahhenshaw: I've gone with a default of 0 instead, hope that's ok.

micahhenshaw: Also added an extra bit around the link that it puts together. Found it would 404 if there was two slashes '//' after the config url, so checked if there was a trailing slash first and added one if there wasn't. Not sure if that is the best way to go about that, so let me know if you can think of a better way.

ildyria (approved):

Issue #1064, opened 2021-07-15
[Enhancement] Index/Sort/Filter by color


andjar:
A nice feature in [PhotoPrism](https://demo.photoprism.org/) is grouping by similarity and filtering by color. PhotoPrism employs TensorFlow, but I think these features are based on a simpler algorithm based on colors. The color indexing is described [here](https://docs.photoprism.org/developer-guide/metadata/colors/) and [here](https://github.com/photoprism/photoprism/issues/7). They use an [algorithm in Go](https://github.com/EdlinOrg/prominentcolor) but there is no need to re-invent the wheel as solutions already exist for php, eg. [color-extractor](https://github.com/thephpleague/color-extractor) or [color-thief](https://github.com/ksubileau/color-thief-php) ([demo](http://www.kevinsubileau.fr/projets/color-thief-php?utm_campaign=github&utm_term=color-thief-php_readme))

I think color-thief [looks amazing](http://www.kevinsubileau.fr/projets/color-thief-php?utm_campaign=github&utm_term=color-thief-php_readme), and not too complicated as it already uses php-GD or Imagick. If we can add color indexing a lot of exciting options emerge; eg. background colors based on the active image

ildyria:
Color extraction has been implemented.

PR #1067, opened 2021-07-22
feature:extract color and search

andjar:
I got some help to implement colorthief (see #1064 ); now, a palette is made for each photo, and presented under info, and one can search by clicking on a color. See lower right corner in the screenshot [painting: Livets dans by Edvard Munch]:

lychee_palette


ildyria:
Okay, **first of all, I love it.** :)

However I think it is better to wait a bit for this to be merged.
What I mean is that @nagmat84 is currently undergoing a significant refactoring of the code,
and merging after his changes makes more sense.

Now, I'm wondering whether it is better to have the color relationship as you implemented.
Or have directly a table `color palette` which associate the 5 colors in one go...

@nagmat84 what would be your take on this? 

Implementation wise, I would make use of a trait for the color palette extraction (idea of Single Responsibility Principle), and then have the Photo::create() call the function where you had your code originally.
This also gives a bit more modularity to the system.

Now, on the missing parts to add:

1. A CLI command to parse already existing pictures to fill up their color palette.
2. Double check what happens when you delete a picture...
3. What happens when you duplicate a picture ?

By using a color palette instead of a set of color, you allow a has_one photo -> palette and has_many palette -> photo without having to duplicate the colors.

4. Add a function call "Extract color palette" ? :)

app/Actions/Search/PhotoSearch.php
```diff
@@ -50,16 +50,25 @@ public function query(array $terms)
...
-					->orWhere('tags', 'like', '%' . $escaped_term . '%')
-					->orWhere('location', 'like', '%' . $escaped_term . '%')
-					->orWhere('model', 'like', '%' . $escaped_term . '%')
-					->orWhere('taken_at', 'like', '%' . $escaped_term . '%')
-			);
+			if (preg_match('/rgb\((\d{1,3}),(\d{1,3}),(\d{1,3})\)/', $escaped_term, $matchers) > 0) {
+				$r = $matchers[1];
+				$g = $matchers[2];
+				$b = $matchers[3];
+				$query->join('colors', 'photos.id', '=', 'colors.photo_id')
```
nagmat84, line 57: I would prefer using

    query->whereHas('colors', function (Builder $innerQuery) use ($r, $g, $b) {
      $innerQuery
        ->whereBetween('r', [$r - 10, $r + 10])
        ->whereBetween('r', [$g - 10, $g + 10])
        ->whereBetween('r', [$rb- 10, $b + 10]);
    })
  )

See [Laraval Docs - Querying Relationship Existence](https://laravel.com/docs/8.x/eloquent-relationships#querying-relationship-existence)

It works more on the object layer and is not directly coupled to the DB layer. But this is probably a matter of taste.

database/migrations/2021_07_21_085048_create_colors_table.php
```diff
@@ -0,0 +1,37 @@
...
+	 */
+	public function up()
+	{
+		Schema::create('colors', function (Blueprint $table) {
+			$table->bigIncrements('id');
+			$table->unsignedTinyInteger('r')->nullable();
+			$table->unsignedTinyInteger('g')->nullable();
+			$table->unsignedTinyInteger('b')->nullable();
+			$table->foreignId('photo_id')->constrained();
+			$table->unsignedTinyInteger('is_main')->default(0);
```
nagmat84, line 22: Why is that a tiny integer? Shouldn't it be a boolean? If it is not a boolean, then the name of the attribute should not start with `is_`.

database/migrations/2021_07_21_085048_create_colors_table.php
```diff
@@ -0,0 +1,37 @@
...
+			$table->timestamps();
```
nagmat84, line 23: Don't use `->timestamps`. It is broken. See also `developer-notes.md` in the root directory of the project. If you really need timestamps on the model, then create the database columns manually using the correct type and the model `Color` should also use the trait `UTCBasedTimes`.

However, I don't see a reason, why the model `Colors` needs timestamps on its own. It probably should inherit the timestamps from its parent `Photo` object and just touch the `Photo` object.

database/migrations/2021_07_21_085048_create_colors_table.php
```diff
@@ -0,0 +1,37 @@
...
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		Schema::create('colors', function (Blueprint $table) {
+			$table->bigIncrements('id');
+			$table->unsignedTinyInteger('r')->nullable();
```
nagmat84, line 18: Why are `r`, `g` and `b` nullable? If a color entry exists, then those values should be set, shouldn't they?

app/Models/Photo.php
```diff
@@ -347,4 +362,13 @@ public function withTags($tags)
 
 		return ($sql->count() == 0) ? false : $sql->first();
 	}
+
+	public static function boot()
+	{
+		parent::boot();
+
+		static::deleting(function ($photo) {
+			$photo->colors()->delete();
```
nagmat84, line 371: Actually, this is not necessary and much slower than letting the DB do it. Add `cascadeOnDelete` and `cascaseOnUpdate` to the table definition instead.

app/Models/Photo.php
```diff
@@ -65,6 +66,10 @@
  * @property string      $livePhotoChecksum
  * @property Album|null  $album
  * @property User        $owner
+ * @property string|null $palette
+ * @property int|null    $main_color_R
```
nagmat84, line 70: Where are the attributes `palette`, `main_color_R`, `main_color_G` and `main_color_B` defined?

They do not seem to be proper attributes of the table `photos`. At least, I did not find a migration which adds these columns to the `photos` table.

If they were "virtual" attributes, then their accessor functions are missing. I could not find them in this PR.

Also it should be correct snake case, i.e. `main_color_r` and so on. (Indeed, this is also a good indicator that these attributes actually do not exist and are not used at all, because otherwise the code would fail.)

database/migrations/2021_07_21_085048_create_colors_table.php
```diff
@@ -0,0 +1,37 @@
...
+			$table->unsignedTinyInteger('g')->nullable();
+			$table->unsignedTinyInteger('b')->nullable();
+			$table->foreignId('photo_id')->constrained();
```
nagmat84, line 21: Add `->onUpdateCascade()` and `->onDeleteCascade()` here. Then you can drop the `boot` method and the listener for the deletion event from the model `Photo`.

app/Models/Color.php
```diff
@@ -0,0 +1,30 @@
...
+{
+	use HasFactory;
+	protected $guarded = ['id'];
+
+	public function photo(): BelongsTo
+	{
+		return $this->belongsTo(Photo::class, 'photo_id', 'id');
+	}
+
+	public function is_main($query)
```
nagmat84, line 26: Where is that method called at all? If it is dead code, we should drop it.

Secondly, the method does not do anything with the current object (i.e. `$this`). If we really need it (I doubt it), then it should be a static function at least.

app/Models/Color.php
```diff
@@ -0,0 +1,30 @@
...
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+/**
+ * Class Color.
+ *
+ * @property int|null $r
+ * @property int|null $g
+ * @property int|null $b
+ */
+class Color extends Model
```
nagmat84, line 16: The following properties are missing:

    @property int    $id
    @property bool   $is_main
    @property Photo  photo

app/Models/Photo.php
```diff
@@ -171,6 +176,16 @@ public function owner(): BelongsTo
 		return $this->belongsTo('App\Models\User', 'owner_id', 'id');
 	}
 
+	/**
+	 * Return the relationship between a Photo and its Owner.
+	 *
```
nagmat84, line 181: The comment is probably a copy-&-paste error. That's not what the method does.

app/Actions/Photo/Create.php
```diff
@@ -118,6 +121,22 @@ public function add(
 			$res = $this->save($this->photo);
 		}
 
+		//extract the main color and palette
+		$palette = ColorThief::getPalette($this->path, 5);
+
+		if (!empty($palette)) {
+			foreach ($palette as $index => $value) {
```
nagmat84, line 128: Why do you need the additional `if`-clause? `foreach` alone should be fine. If `$palette` is empty, then `foreach` does not iterate.

app/Actions/Photo/Create.php
```diff
@@ -118,6 +121,22 @@ public function add(
...
+				$color = new Color([
+					'r' => $value[0],
+					'g' => $value[1],
+					'b' => $value[2],
+					'is_main' => ($index === 0 ? 1 : 0),
+					'photo_id' => $this->photo->id,
+				]);
```
nagmat84, line 135: You should not set the id of the foreign relation manually. `$this->photo->colors()->save($color)` (1 line below) takes care of that. Your can remove `'photo_id' => $this->photo-id'`.

app/Actions/Photo/Create.php
```diff
@@ -118,6 +121,22 @@ public function add(
...
+				$this->photo->colors()->save($color);
+			}
+		}
+
```
nagmat84, line 139: It was just yesterday, when I learned the hard way, that you should unset a cached relation after one has modified it by `->make` or `->save` due to an Eloquent bug. Otherwise bad things might happen. Hence, after the `foreach`-loop when you have finished with all `$this->photo->colors()->save()`, you need to add

    if ($this->photo->relationLoaded('colors')) {
	  $this->photo->unsetRelation('colors');
    }

app/Models/Extensions/PhotoCast.php
```diff
@@ -84,6 +84,7 @@ public function toReturnArray(): array
 			'taken_at_orig_tz' => $this->taken_at_orig_tz,
 			'license' => $this->license,
 			'sizeVariants' => $sizeVariants,
+			'colors' => $this->colors()->get(),
```
nagmat84, line 87: `$this->colors` is enough and more efficient. (Note that the missing parentheses `()` at the end are by intention.) This way you call the attribute `colors` (which is a `Collection` of `Color`) and not the function which always returns a `HasMany` object. The difference is that the attribute uses caching while `colors()->get()` always queries the database.

nagmat84 (changes requested):

nagmat84:
I love this feature, too. :heart_eyes:

> Now, I'm wondering whether it is better to have the color relationship as you implemented.
> Or have directly a table `color palette` which associate the 5 colors in one go...

I don't get your point. There is a table `colors`, isn't it? And what *5* colors are you referring to?

> Implementation wise, I would make use of a trait for the color palette extraction (idea of Single Responsibility Principle), and then have the Photo::create() call the function where you had your code originally.
> This also gives a bit more modularity to the system.

I disagee . Using a trait here is not the right approach. (In particular, a trait is not the solution to the single responsibility principle.) If you wanted to decouple the code from the method `Create::do` and move out the code to an independent class, then you would have to create some kind of `ColorPaletteExtractionService` (or something like that) in the same spirit as we already have a `MetadataExtractor`, get an instance of the `ColorPaletteExtractionService` where you need it and call it. But this seems a little bit over the edge for these extra 10 lines of code.

Anyway, I am really undecided, if the code may remain where it is right now or if I would prefer having a de-coupled `ColorPaletteExtractionService`. But please don't use a trait!

ildyria:
> I love this feature, too. 
> 
> > Now, I'm wondering whether it is better to have the color relationship as you implemented.
> > Or have directly a table `color palette` which associate the 5 colors in one go...
> 
> I don't get your point. There is a table `colors`, isn't it? And what _5_ colors are you referring to?

When you query a Color palette, you will get 5 colors in return which you will need to save.

In the current set up, you have 1 photo = 5 new color rows.

I am just wondering about the size explosion of such table and whether a different implementation
would change this.
I am currently thinking of:
1. ~~Having a table color_palette with basically 16 columns which incorporate the attributes:
`id, (rgb)1, (rgb)2, (rgb)3, (rgb)4, (rgb)5`
But this would complexity the search.~~ But this is a bad idea.
2. Adding an intermediate table "color_palette" which would link the picture to a set of 5 colors:
`palette_id, color_1_id, color_2_id, color_3_id, color_4_id, color_5_id`, where 1 is the main color.

As I see it, this second approach as multiple benefits, it remove the need to define a boolean for primary color within the color table (as it would be `color_1_id`).
Futhermore, you can do eager loading on the colors.
And finally, it solves the duplication problem: you do not need to duplicate the colors when you duplicate the picture and you move to a `has_one` relationship `color->palette` instead of a `has_many`.

In the end you would get: `Photo -> palette -> colors`.

I would be leaning towards such step.

> > Implementation wise, I would make use of a trait for the color palette extraction (idea of Single Responsibility Principle), and then have the Photo::create() call the function where you had your code originally.
> > This also gives a bit more modularity to the system.
> 
> I disagee . Using a trait here is not the right approach. (In particular, a trait is not the solution to the single responsibility principle.) If you wanted to decouple the code from the method `Create::do` and move out the code to an independent class, then you would have to create some kind of `ColorPaletteExtractionService` (or something like that) in the same spirit as we already have a `MetadataExtractor`, get an instance of the `ColorPaletteExtractionService` where you need it and call it. But this seems a little bit over the edge for these extra 10 lines of code.

My mistake, though I still think this codes need to be extracted, so it can be reused in CLI.
This code will be duplicated otherwise.
Furthermore it makes it less likely to conflict with your refactoring and easier to integrate with your strategies. :)

nagmat84:
> When you query a Color palette, you will get 5 colors in return which you will need to save.

Are we sure that this will ever be the case? But it's fine for me to have that as an assumption.
 
> In the current set up, you have 1 photo = 5 new color rows.
> I am just wondering about the size explosion of such table and whether a different implementation
> would change this.

I would not be worried about that. We also have 1 photo with 7 different size variants. I am confident that space should not be a problem.

> I am currently thinking of:
> 
>  1. ~Having a table color_palette with basically 16 columns which incorporate the attributes:
>     `id, (rgb)1, (rgb)2, (rgb)3, (rgb)4, (rgb)5`
>     But this would complexity the search.~ But this is a bad idea.

I agree. Bad idea.

>  2. Adding an intermediate table "color_palette" which would link the picture to a set of 5 colors:
>     `palette_id, color_1_id, color_2_id, color_3_id, color_4_id, color_5_id`, where 1 is the main color.
> 
> As I see it, this second approach as multiple benefits, it remove the need to define a boolean for primary color within the color table (as it would be `color_1_id`).

Fine for me.

> Furthermore, you can do eager loading on the colors.

You could also do that with the current approach. That's no different.

> And finally, it solves the duplication problem: you do not need to duplicate the colors when you duplicate the picture and you move to a `has_one` relationship `color->palette` instead of a `has_many`.

I believe this is a bad idea, because it does not reflect the true semantics and feels really awkward.

It appears to me as if you wanted to solve a different problem here but with the wrong approach. I agree that duplication of photos could be improved. The specific problem which you mention in this context regarding the colors is actually only one incarnation of a more general problem. We have the same problem with the size variants, the symbolic links and the access rights. Also we have the problem that we only must erase the binary photo files when the last database entry is erased.

Actually, the right approach to solve all these problems together would be to have an m:n relation between albums and photos using a `album_photo_associations` table. That way duplication of a photo would be a breeze and only require an additional small entry in the table `album_photo_associations` without duplication of any photo data (not only color data). I already have proposed that change some time back, but you did not like it. Anyway we should not discuss this problem here but take it to Gitter (if you want), because it is beyond the scope of this PR.

> In the end you would get: `Photo -> palette -> colors`. I would be leaning towards such step.

Understood, but surprisingly, see above. This actually introduces a m:n relationship but through the back door :-) and (IMHO) at the wrong place.

Either I would leave the current approach as it is (with the `is_main` flag) or use the following alternative approach. This alternative also removes the need for the `is_primary` flag and rules out any inconsistencies (which could happen, if the `is_primary` flag is not unique among all colors of the same photo):

 In pseudo UML

    +-----------------------+                       +----------------+
    |          Photo        |                       |      Color     |
    +-----------------------+                       +----------------+
    | id:           bigint  | (1)--<--owned-by--(1) | id:    bigint  |
    | primaryColor: Color   | (1)---has-->---(0..1) | photo: Photo   |
    | colors:       Color[] | (1)---has-->---(0..*) | r:     int     |
    +-----------------------+                       | g:     int     |
                                                    | b:     int     |
                                                    +----------------+

In (Postge)SQL

    CREATE TABLE photos(
      id                SERIAL PRIMARY KEY,
      primary_color_id  INT NULLABLE,
      FOREIGN KEY(primary_color_id) REFERENCES colors(id)
    );
    
    CREATE TABLE colors(
      id        SERIAL PRIMARY KEY,
      photo_id  INT NOT NULLABLE,
      r         SMALLINT NOT NULLABLE,
      g         SMALLINT NOT NULLABLE,
      b         SMALLINT NOT NULLABLE,
      FOREIGN KEY(photo_id) REFERENCES photos(id)
    );

In PHP with Eloquent

    /**
     * @property Color      primary_color
     * @property Collection colors
     */
    class Photo extends Model {
      public function primary_color(): HasOne {
        return $this->hasOne(Color::class, 'primary_color_id', 'id');
      }
      
      public function colors(): HasMany {
        return $this->hasMany(Color::class);
      }
    }
      
    /**
     * @property Photo photo
     */
    class Color extends Model {
      public function photo(): BelongsTo {
        return $this->belongsTo(Photo::class);
      }
    }

>> If you wanted to decouple the code from the method `Create::do` and move out the code to an independent class, then you would have to create some kind of `ColorPaletteExtractionService` (or something like that) in the same spirit as we already have a `MetadataExtractor`, get an instance of the `ColorPaletteExtractionService` where you need it and call it.
> 
> I still think this codes need to be extracted, so it can be reused in CLI. This code will be duplicated otherwise. Furthermore it makes it less likely to conflict with your refactoring and easier to integrate with your strategies. :)

OK, then lets move the code to a service class. However, I wonder if it should be its own service (something like `ColorPaletteExtractionService`) or just be integrated into the metadata extractor. In the end, the histogram is also some kind of metadata and I assume that we want to extract the histogram every time we also extract the other metadata.

scottlimmer:
What is the status of this PR? I'd like to implement #1064 but I don't want to step any any toes.

andjar:
Hi, I'm sorry that I didn't follow up the PR; I don't have capacity to look more into this, so please feel free to do what you like @scottlimmer. The code in this PR is working perfectly, but as you can see, there are still some issue. Feel free to copy what you want:)

nagmat84:
The status is the same as before

1. the requested changes should be implemented, and
2. it would be more easy to merge the other PRs (with the massive refactoring of the architecture) first and then rebase this PR on the new architecture

As the original submitter of this PR has never replied to the requested changes, I actually would be willing to take the effort to apply the  requested changes myself and also do the re-basing. I really like the feature and want to see it being integrated. However, first the other PRs must be merged.

ildyria:
> As the original submitter of this PR has never replied to the requested changes, I actually would be willing to take the effort to apply the requested changes myself and also do the re-basing. I really like the feature and want to see it being integrated. However, first the other PRs must be merged.

I have exactly the same opinion. :)

With regard to this:
> ```
> +-----------------------+                       +----------------+
> |          Photo        |                       |      Color     |
> +-----------------------+                       +----------------+
> | id:           bigint  | (1)--<--owned-by--(1) | id:    bigint  |
> | primaryColor: Color   | (1)---has-->---(0..1) | photo: Photo   |
> | colors:       Color[] | (1)---has-->---(0..*) | r:     int     |
> +-----------------------+                       | g:     int     |
>                                                 | b:     int     |
>                                                 +----------------+
> ```

I would remove the `photo` attribute and just make it a m:n relationship. :)
Using a pivot table may be necessary though. :|

scottlimmer:
@ildyria @nagmat84 Sounds like I'll leave this to you. 

Are all open issues impacted by pending refactoring PRs?

nagmat84:
> Are all open issues impacted by pending refactoring PRs?

It depends. Minor bug fixes and security issues will be integrated of course. Also, if a PR does not conflict too much with the refactoring, it will likely be integrated.

@scottlimmer: Do you have a specific issue in mind?

qwerty287:
Due to various refactorings and changes, this will need a complete rework (rewrite from scratch), so I'm going to close it.

PR #1071, opened 2021-08-01
Avoid git rebase conflicts during development on other branches

nagmat84:
As discussed on Gitter.

### DIFF
--- a/app/Metadata/GitHubFunctions.php
+++ b/app/Metadata/GitHubFunctions.php
@@ -57,7 +57,7 @@ class GitHubFunctions
 	 *
 	 * @return false|string
 	 */
-	public function get_current_branch(): string | false
+	public function get_current_branch()
 	{
 		// @codeCoverageIgnoreStart
 		$head_file = base_path('.git/HEAD');
@@ -74,7 +74,7 @@ class GitHubFunctions
 	 *
 	 * @return false|string
 	 */
-	public function get_current_commit(): string | false
+	public function get_current_commit()
 	{
 		$file = base_path('.git/refs/heads/' . $this->branch);
 		$head_ = file_get_contents($file);
@@ -91,7 +91,7 @@ class GitHubFunctions
 	 *
 	 * @throws NotInCacheException
 	 */
-	private function get_commits(bool $cached = true): bool | array
+	private function get_commits(bool $cached = true)
 	{
 		return $this->gitRequest->get_json($cached);
 	}
@@ -110,7 +110,7 @@ class GitHubFunctions
 	 * @throws NotInCacheException
 	 * @throws NotMasterException
 	 */
-	public function count_behind(bool $cached = true): bool | int
+	public function count_behind(bool $cached = true)
 	{
 		if ($this->branch != 'master') {
 			// @codeCoverageIgnoreStart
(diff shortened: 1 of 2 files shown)


nagmat84:
I don't understand why `codecov/patch` is unhappy.

ildyria (changes requested):
https://github.com/LycheeOrg/Lychee/blob/master/app/Metadata/GitHubFunctions.php

Most of this code is sort of duplicate of this one and does not require `exec()`.
It would require minor tweak to get rid of https://github.com/LycheeOrg/Lychee/blob/master/app/Metadata/GitHubFunctions.php#L72

ildyria:
> I don't understand why `codecov/patch` is unhappy.

code patch wants at least 63% coverage on any new code.
I am actually considering nuking this check.

nagmat84:
> https://github.com/LycheeOrg/Lychee/blob/master/app/Metadata/GitHubFunctions.php
> 
> Most of this code is sort of duplicate of this one and does not require `exec()`.
> It would require minor tweak to get rid of https://github.com/LycheeOrg/Lychee/blob/master/app/Metadata/GitHubFunctions.php#L72

Unfortunately, `GitHubFunctions::get_current_branch` claims to be on `master`, if unit tests are executed, even if the current branch is not `master`. So using that method does not help here, because it actually would not prevent the unlucky rebase.

IMHO, `GitHubFunctions::get_current_branch` should be modified to actually report the correct branch and not botch with the returned value, but I cannot foresee the implication of what would happen, if I changed that.

app/Metadata/GitHubFunctions.php
```diff
@@ -69,10 +59,6 @@ private function trim($commit_id)
 	 */
 	public function get_current_branch()
 	{
-		if (App::runningUnitTests()) {
```
nagmat84, line 72: Are we sure that omitting that condition has no negative impact?

[`GitHubFunctions::count_behind(bool)`](https://github.com/LycheeOrg/Lychee/blob/bc918a394a995a939b6e5651746895c464370a2d/app/Metadata/GitHubFunctions.php#L127) checks for the master branch. `count_behind` is called in `is_up_to_date` and I have not checked further where this method is used. Anyway, I can't oversee right now (and haven't check) which other methods further down the call graph rely on `get_current_branch` to fake the master branch during tests.

app/Metadata/GitHubFunctions.php
```diff
@@ -69,10 +59,6 @@ private function trim($commit_id)
 	 */
 	public function get_current_branch()
```
nagmat84, line 59: The PHPdoc comment needs updating. The last sentence

> Return master if it is CI

does not hold anymore.

nagmat84 (commented):
In general, it looks good to me. My only hesitation comes from leaving out that one if-branch (see review comment in the affected source code). I would feel more confident about that, if someone could investigate and confirm that removing that if-branch has no negative impact.

app/Metadata/GitHubFunctions.php
ildyria: Nah I don't think this impact in the count_behind. I would expect the tests to break otherwise. :)

nagmat84: OK, then let's be brave. If any bug occurs, we will fix it then.

app/Metadata/GitHubFunctions.php
nagmat84: ok, it is updated.

PR #1086, opened 2021-08-14
Update Sync.php

### LINKED ISSUE #1056: [Enhancement] cli sync from from server with option to delete
I am using the command: `php artisan lychee:sync`
part of the cron job it will run every day, the android phone will push the photos to a directory in the server, 
but there is no option to "delete original" as in the GUI

maybe there is different API call, or it not exists for the CLI

pchampio:
fixes: #1056

### DIFF
--- a/app/Console/Commands/Sync.php
+++ b/app/Console/Commands/Sync.php
@@ -15,7 +15,7 @@ class Sync extends Command
 	 *
 	 * @var string
 	 */
-	protected $signature = 'lychee:sync {dir : directory to sync} {--album_id=0 : Album ID to import to} {--owner_id=0 : Owner ID of imported photos} {--resync_metadata : Re-sync metadata of existing files}';
+	protected $signature = 'lychee:sync {dir : directory to sync} {--album_id=0 : Album ID to import to} {--owner_id=0 : Owner ID of imported photos} {--resync_metadata : Re-sync metadata of existing files}  {--delete_imported : Delete the original files} {--import_via_symlink= : Imports photos from via a symlink instead of copying the files} {--not_skip_duplicates : Don\'t Skip photos and albums if they already exist in the gallery}';
 
 	/**
 	 * The console command description.
@@ -31,7 +31,6 @@ class Sync extends Command
 	 */
 	public function handle(Exec $exec)
 	{
-		$this->info('Start syncing.');
 		$directory = $this->argument('dir');
 		$owner_id = (int) $this->option('owner_id'); // in case no ID provided -> import as root user
 		$album_id = (int) $this->option('album_id'); // in case no ID provided -> import to root folder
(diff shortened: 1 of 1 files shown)


app/Console/Commands/Sync.php
```diff
@@ -31,21 +31,37 @@ class Sync extends Command
...
 		$owner_id = (int) $this->option('owner_id'); // in case no ID provided -> import as root user
 		$album_id = (int) $this->option('album_id'); // in case no ID provided -> import to root folder
 
 		// Enable CLI formatting of status
 		$exec->statusCLIFormatting = true;
 		$exec->memCheck = false;
-		$exec->delete_imported = false; // we want to sync -> do not delete imported files
-		$exec->import_via_symlink = (Configs::get_value('import_via_symlink', '0') === '1');
-		$exec->skip_duplicates = true;
+		$exec->delete_imported = $this->option('delete_imported');
```
nagmat84: I recommend an improvement here in order to avoid coding the same error message "--delete_imported [...] and [...] are conflicting" twice.

First check `if ($this->option('import_via_symlink') === '1' && $this->option('delete_imported') === '1')` even before you set `$exec->delete_imported` (i.e. before line 41) and if the condition holds, then bail out with an error message. Then you can be sure, that both options are not set at the same time in the following. This also avoids the nested `if` below.

nagmat84 (commented):
Looks good to me. I recommend one minor improvement to make the code  more readable.

app/Console/Commands/Sync.php
pchampio: > Then you can be sure, that both options are not set at the same time in the following. 

If we only check for options with `if ($this->option('import_via_symlink') === '1' && $this->option('delete_imported'))`, how are we dealing with the Config setting `(Configs::get_value('import_via_symlink', '0') === '1') == true` and `$this->option('delete_imported')` ?
I've done it this way to make the sync backward compatible so that people using the config  `import_via_symlink === 1` don't have to add the `--import_via_symlink=1` flag.

nagmat84: Could you push your forked branch `patch-1` from the repository `pchampio/Lychee ` as a new branch to `LycheeOrg/Lychee`? However, I would recommend to rename the branch to something else than `patch-1`.

Then I could checkout the branch myself and push a commit to show what I mean. I think you don't get the idea.

ildyria:
Pushing to a new branch on LycheeOrg is not feasible due to access right. 

⁣Télécharger BlueMail pour Android ​

nagmat84:
> Pushing to a new branch on LycheeOrg is not feasible due to access right.

Then I will do it.

app/Console/Commands/Sync.php
nagmat84: I took the liberty to push a new commit. I even went beyond my original idea and I added the additional feature that the default values of the command line options are taken from the user configuration.

nagmat84: @pchampio What do you think?

ildyria (approved):
LGTM.

nagmat84:
LGTM, too. I would approve this PR, too, but it seems that I cannot as a push an own commit (?). I would recommend to wait another day to see whether @pchampio agrees with my changes. If we have no reply by then, we should merge this PR.

app/Console/Commands/Sync.php
pchampio: This is much better!
Thanks, I was looking for a way to set the default values of the command with reference to the configuration, but didn't found how to do it.

PR #1094, opened 2021-08-30
fix "General error: 1 Expression tree is too large (maximum depth 1000)"

imaben:
fix "General error: 1 Expression tree is too large (maximum depth 1000)"


ildyria (approved):

nagmat84:
Actually, this bug fix won't be necessary after the big refactoring will have been merged, because the buggy code won't exist anymore.

ildyria:
I was thinking about it too but was considering it more as a temporary fix. :)

PR #1105, opened 2021-09-24
Fix type issues related to thumbs

kamil4:
Fixes the issues with thumbs reported on gitter, which trigger primarily with symlinking enabled.

The change to `SymLink.php` is needed because in PHP `'a4995077a769bf97d0fc352083407df8.jpeg' == 0` evaluates to true...

### DIFF
--- a/app/Models/Extensions/PhotoCast.php
+++ b/app/Models/Extensions/PhotoCast.php
@@ -25,7 +25,7 @@ trait PhotoCast
 			$filename = $this->url;
 		}
 		$filename2x = ($filename !== '') ? Helpers::ex2x($filename) : '';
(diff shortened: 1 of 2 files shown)


ildyria (approved):

nagmat84 (approved):

nagmat84:
The joys of PHP. :-(

This particular type error would have been gone with my big PR. But I wonder how many other bugs of this kind will be introduced. 😱

ildyria:
> The joys of PHP. :-(
> 
> This particular type error would have been gone with my big PR. But I wonder how many other bugs of this kind will be introduced.

Fortunately https://kinsta.com/blog/php-8/#string-to-number-comparison :)

PR #1121, opened 2021-10-17
Fix lychee:video_data

gpotter2:
This PR fixes the `lychee:video_data` command.

```
Will attempt to generate up to 100 video thumbnails/metadata with a timeout of 600 seconds...
Processing XXX...
Extracted video frame for thumbnails

In ImageEditing.php line 170:

  Undefined property: App\Console\Commands\VideoData::$imageHandler
```

Are commands tested regularly? It would also appear that `lychee:generate_thumbs` is broken?


nagmat84:
> Are commands tested regularly? It would also appear that `lychee:generate_thumbs` is broken?

Unfortunately not. The unit tests do not cover everything, especially the CLI is mostly untested.

nagmat84:
PR looks good to me.

nagmat84 (approved):

PR #1133, opened 2021-10-30
fix multiple submission migration

### LINKED ISSUE #1033: [Enhancement] Migration update form - give user feedback on submit.
See : https://youtu.be/EN7nu8hrcN4?t=181

This **NEEDS** to be added on the "503: _migration required_" page to prevent the user from double clicking. Some migrations may take time, we do not want the user to launch multiple of those in parallel.

ildyria:
Fixes #1033

A bit of JS to prevent multiple submission of the migration form.


kamil4 (approved):
Untested but looks sane...

PR #1142, opened 2021-11-16
also support Caps on size format

ildyria:
Just found out that if we have
`Max uploaded file size:          500M`
Then this will still raise the warning:
`Warning: You may experience problems when uploading a large amount of photos. Take a look in the FAQ for details.`


d7415 (approved):
LGTM

nagmat84 (approved):
LGMT

kamil4 (approved):

PR #1144, opened 2021-11-18
Upgrade composer

ildyria:


nagmat84:
Thank you, @ildyria for this PR, but I would rather not use it. I would not merge this commits into this branch independently but rather from master. Otherwise we (or at least me) will end up in complete merge hell. I.e. if the commits have become part of master (some of them already are), I would merge back master into this branch.

I already do that from time to time.

Issue #1072, opened 2021-08-01
[Enhancement] Enforce PHP 8 ?


ildyria:
As I see it there are multiple reasons to push this forward.

1. PHP 8 provides improvements in term of type checking (such as union types)
2. PHP 8 provides improvements in term of speed (JIT compilation).
3. Our dependencies are already pushing towards PHP 8, e.g. Laravel-feed (thank you spatie...)
4. We will depreciate PHP 7.4 on November 28th, 2021; so in 3 months. 

More about PHP 8 : https://stitcher.io/blog/new-in-php-8

christianhorn:
I run Lychee with Debian stable, like probably many others too. Would be nice if php7 could stay supported while php8 is not in Debian stable.

kamil4:
My understanding is that we drop support for old PHP versions once the PHP project itself drops those versions. See: https://www.php.net/supported-versions.php

d7415:
Yeah. It's not great for stable distros like Debian that run older versions (with backported patches for security or major bugs), but we don't have resource to run multiple branches like that. If there was a security issue we might look into backporting a fix if there's demand, but otherwise older Lychee versions are available to download from [releases](https://github.com/LycheeOrg/Lychee/releases), check out as [tags](https://github.com/LycheeOrg/Lychee/tags) or pin as [docker images](https://github.com/LycheeOrg/Lychee-Docker#image-content) (though that's less relevant as they contain their own PHP).

We list any new PHP version requirements in the release notes.

christianhorn:
Thanks for the hint, was not aware of https://www.php.net/supported-versions.php . That currently shows php7.4 security supported until 28 Nov 2022, that is a relief. Original post looks like just 2 month's of support left. 
I'm happy with Nov 2022. Because Lychee uses to many modules (which I consider as further surface and pieces which need to have an eye on), I run it in a Debian stable container. Even Fedora stable (34 currently) does not come with php8.

d7415:
> Original post looks like just 2 month's of support left.

Yeah. For Lychee we generally use the "active support" window, rather than security.

nagmat84:
[...]

PR #1101, opened 2021-09-15
Refactor album model

nagmat84:
# Main achievements

1. **New architecture and class inheritance:**
   Previously, the class hierarchy was somewhat odd with `Album` being at the top of the hierarchy. In particular, the other album related classes inherited from `Album` even though the did not fulfill the "is-a" relationship. This resulted in a lot of peculiarities. This situation has been rectified.
   
   Basically the new architecture looks like that (simplified view):
   
   Models (intro)
   
    - `AbstractAlbum`: the top of the hierarchy; this class only provides attributes (e.g. a `title`) and methods (e.g. `photos()`) which are really supported by any kind of album
    - `BaseSmartAlbum`: the common interface for all smart albums which are "built-in" into the software; as a common feature they can only be instantiated once and *cannot* created, deleted, renamed etc.
    - `BaseAlbum`: the common interface for all kind of albums which can be persisted to and hydrated from DB
  
   The SQL layer uses the pattern _table-based inheritance_. This means there are three tables: `base_albums`, `albums` and `tag_albums`. The table `base_albums` has provides columns for all common attributes. Columns for specific attributes are part of the tables `albums` and `tag_album`. The table `base_albums` owns a traditional primary key and purports the ID of the albums. The table `albums` and `tag_albums` use a foreign key as their primary key. E.g. the table layout looks like this:
   
       CREATE TABLE `base_albums` (
         id SERIAL NOT NULL PRIMARY KEY,
         ...
       )
       
       CREATE TABLE `albums` (
         id BIGINT NOT NULL PRIMARY KEY,
         ...
         FOREIGN KEY id REFERENCES base_albums(id)
       )
      
       CREATE TABLE `tag_albums` (
         id BIGINT NOT NULL PRIMARY KEY,
         ...
         FOREIGN KEY id REFERENCES base_albums(id)
       )
   
   Unfortunately, Eloquent completely lacks the support for table-based inheritance (or support of any other kind of model inheritance :confused:). Normally, (if we had a more powerful ORM library), the common methods and attributes would be implemented by the class `BaseAlbum`. Moreover, Eloquent can only use one table per model and assumes that each model is directly extended from `Model`.
   
   With Eloquent we have to use a work-around like this:
   
   Models (full)
   
[...]

PR #1055, opened 2021-07-10
Refactor Model Architecture

nagmat84:
**Updated Summary Nov 28th 2021**

This is the merger of the previous PR "Refactor photo model" and [PR #1101 Refactor album model](https://github.com/LycheeOrg/Lychee/pull/1101). See PR #1101 for what has changed there.

**Old Summary (of "Refactor photo model" only)**

As previously discussed on Gitter this PR intends to rectify two aspects:

 - The way how the different size variants of a photo are managed has been cleaned up. The size variants now live in their own table `size_variants` and are independent of each other, i.e. each has its own file path, its own size attributes and so on. The previously awkward mutual dependencies are gone.
 - The hard-coded path and magic constants all throughout the code are gone. I.e. there should not be any code left which relies on the original size of an image being stored in a folder named `big` or that certain files have `@2` in their filename.

Unfortunately, the amount of things which needed to be reworked became much larger than expected. In detail the following aspected have been changed:

 - The way how files are named has been moved to a contract `SizeVariantNamingStrategy`. At the moment the only implementing class is `SizeVariantLegacyNamingStrategy` which mimics the old behaviour.
 - The creation of the size variants has been moved to a contract `SizeVariantFactory`. At the moment the only implementing class is `SizeVariantDefaultFactory` which mimics the old behaviour.
 - Removal of a lot of code duplication. The code which creates size variants only exists in `SizeVariantDefaultFactory`. Previously, the same code (with minor derivations) existed several times (for uploading/creating new photos for re-creating missing size variants by the console command, etc.)
 - Complete rework of the strategy classes for photo upload. Previously, there had been two strategies `StrategyPhoto` and `StrategyDuplicate`. The code to handle the upload of live photos had been scattered in between. As this part needed to be completely reworked anyway (due to the changes how the size variants are created), there are now four strategies which cleanly encapsulate the upload of live photos, too.
 - The way how photos (and size variants) are deleted has been changed. Previously, it was the responsibility of the caller to first call `predelete` on these entities in order to properly clean-up the actual files followed by `delete`. Now it is save to directly call `delete` and the classes will take care of proper deletion.
[...]

PR #1145, opened 2021-11-19
Upgrade composer and php version

ildyria:
TL;DR: bye bye PHP7.4, long live PHP8.0 (and 8.1)

Rationale of such change:
- [Php supported version](https://www.php.net/supported-versions.php): we want to **stay in the green zone**, anything orange is ignored.
- PHP 8.0 brings lots of improvements: stronger typing, JIT... read more [here](https://stitcher.io/blog/new-in-php-8)
- Spatie requires in all they updated packages to be in version 8.0.

Given that this is a breaking change, does it deserve a version on its own? _E.g._, 4.3.7


nagmat84:
I understand that Spatie somehow forces us to migrate to PHP 8.x (thank you Spatie for nothing).
But despite from that compelling reason I have very mixed feeling about a migration to PHP 8. Personally, I don't mind as I use Gentoo as my distro and I like to use the latest stable versions which are available upstream. However, I fear that many, many user use slower distros like Debian, Ubuntu etc. and thus are not ready to migrate to PHP 8. Also PHP 7.4 will still get security fixes until November next year.

But yes, as we require Spatie (and probably do not want to drop that), I see no other option than moving forward to PHP 8.

ildyria:
Honestly, it is not that painful to upgrade to PHP 8 on Debian. I followed these guides and it was pretty quick:
https://www.howtoforge.com/how-to-install-php-8-on-debian-11/

The pain comes mostly from user who do not want to add the official stable PHP ppa to their desktop. :|

kamil4:
I haven't looked at the code changes yet but I would strongly recommend that we push 4.3.6 out the door first before merging this. Basically, let's give our users one more stable 7.4-compatible release and _then_ break things for them :smiley:.

app/Models/Album.php
```diff
@@ -26,27 +26,27 @@
...
+ * @property int         $viewable
+ * @property int         $downloadable
+ * @property int         $share_button_visible
+ * @property string|null $password
+ * @property string      $license
+ * @property bool        $smart
+ * @property text        $showtags
+ * @property Carbon      $created_at
+ * @property Carbon      $updated_at
  * @property Collection[Album] $children
```
nagmat84: The correct PHPdoc syntax for generics are the usual acute parentheses `<>`. This means `Collection<Album>` is correct, but `Collection[Album]` is wrong.

I know that this bug has not been introduced by this PR but is an old one. However, as you correct so many types with this PR, I thought it would be nice to fix this, too, on a global level.

Same syntax can be used for arrays, i.e. `array<Album>` is equivalent to `Album[]`, but `array[Album]` is wrong. However in case of traditional arrays, I would keep the shorter `Album[]`

nagmat84 (approved):
LGTM. One minor comment for improvement which can either be incorporated in this PR or postponed to the future.

nagmat84:
> I haven't looked at the code changes yet but I would strongly recommend that we push 4.3.6 out the door first before merging this. Basically, let's give our users one more stable 7.4-compatible release and _then_ break things for them smiley.

I lost track of what 4.3.6 shall going to be. Does 4.3.6 also include the big refactoring #1055 and #1101 or not? So do you plan with

 1. 4.3.6 some minor bug fixes
 2. 4.4.0 migration to PHP 8
 3. 4.5.0 big refactoring

or

 1. 4.3.6 some minor bug fixes
 2. 4.4.0 big refactoring
 3. 4.5.0 migration to PHP 8

I actually have no favorite, either way is fine for me, but I wonder if anybody has an opinion about that.

kamil4:
I vote for 4.3._whatever_ to be just fixes. No big PR merges, no php7 depreciation. The only PR that I still want to see merged into it is https://github.com/LycheeOrg/Lychee-front/pull/278 as it addresses a number of recent regressions in the front end.

Once that's out, I think we should merge everything else that's in the pipeline, before making another public release. So both #1055/#1101 and #1145, as far as I'm concerned, and the order doesn't matter...

Come to think of it, I need to update my php to version 8 as well...

ildyria:
> I lost track of what 4.3.6 shall going to be. Does 4.3.6 also include the big refactoring #1055 and #1101 or not?

**No!** :smile: 

If we look at the [Release notes](https://lycheeorg.github.io/docs/releases.html#v408), a year ago version 4.0.8 had a breaking change requiring PHP 7.4 and the official change for PHP 7.4 was the 4.1.

> So do you plan with
> 
> 1. 4.3.6 some minor bug fixes
> 2. 4.4.0 migration to PHP 8
> 3. 4.5.0 big refactoring

I would be in favor of that solution.

_4.3.6_ : includes bunch of fixes & last _freeze_ before PHP 8.
_4.4.0_ : migration to PHP 8 with breaking backward compatibility with PHP 7.4 (it might still work, but no longer supported :no_entry_sign: ). Why before the big refactoring ? because it is simple & easy, pretty much 0 change.
_4.5.0_ : big refactoring. :+1:

nagmat84:
> > 1. 4.3.6 some minor bug fixes
> > 2. 4.4.0 migration to PHP 8
> > 3. 4.5.0 big refactoring
>
> I would be in favor of that solution.

Fine with me. :+1: 

> migration to PHP 8 with breaking backward compatibility with PHP 7.4 (it might still work, but no longer supported)

Please note that PHP 7.x might not still work, it definitely won't work anymore. Your PR does not only move forward some version numbers, but your PR also makes use of union types and public constants. These are two completely new language constructs and PHP 7.x will fail with a fatal parsing error. So the question whether PHP 7.x may still work can ultimately denied.

I am fine with that but I wanted to explicitly point to that, so that nobody has false expectations.

nagmat84:
Fixes #1072. #1072 can be closed upon merge of this PR.

app/Actions/Albums/Top.php
```diff
@@ -35,7 +35,7 @@ public function __construct()
 	 * Note: the array may include password-protected albums that are not
 	 * accessible (but are visible).
 	 *
-	 * @return array[Collection[Album]]
+	 * @return array[Collection<Album>]
```
nagmat84, line 38: `array<Collection<Album>>`

app/Actions/Albums/Extensions/PublicIds.php
```diff
@@ -64,7 +64,7 @@ private function init(): Builder
 	 * Return a collection of Album that are not directly accessible by visibility criteria
 	 * ! we do not include password protected albums from other users.
 	 *
-	 * @return BaseCollection[(_lft, _rgt)]
+	 * @return BaseCollection<(_lft, _rgt)>
```
nagmat84: I know, this is nitpicking. given the fact that this method will be gone soon. But the correct PHPdoc syntax for associative arrays uses curly braces and an UML-like style for properties.

nagmat84 (approved):

app/Actions/Albums/Top.php
ildyria: Have you ever tried the "insert a suggestion" of Github ? :D

nagmat84: Honestly, I did not know that this feature exists. :see_no_evil: I will use it from now on.

app/Actions/Albums/Extensions/PublicIds.php
nagmat84:
```suggestion
	 * @return BaseCollection<array{_lft: int, _rgt: int}>
```

nagmat84 (approved):

kamil4 (approved):

Issue #1149, opened 2021-11-23
Failed to import photo via link


nebulade:
### Detailed description of the problem
While trying to update our app package for [Cloudron](https://www.cloudron.io/store/com.electerious.lychee.cloudronapp.html) to Lychee version 4.3.6, importing a photo from a link fails with:
```
2021-11-23 16:05:58 -- error   -- App\Actions\Import\FromUrl::do -- 54 -- Could not copy file (https://www.w3.org/MarkUp/Test/xhtml-print/20050519/tests/jpeg420exif.jpg) to temp-folder (/app/code/public/uploads/import/jpeg420exif.jpg)
```
This works strangely in version 4.3.4. The folder at `/app/code/public/uploads/import/` is writeable by www-data (apache user)
Uploading photos also works fine.

I am no laravel expert but it seems to be using the copy() from laravel, however I don't know how to get more error information besides what is paste above. Does anyone have any idea how to debug this further? Very happy to provider more information if I know what and how.

### Output of the diagnostics
```
# php artisan lychee:diagnosti

Diagnostics
-----------
Warning: Dropbox import not working. dropbox_key is empty.
Info: Latest version of PHP is 8

System Information
------------------
Lychee Version (release):        4.3.6
DB Version:                      4.3.6

composer install:                --no-dev
APP_ENV:                         production
APP_DEBUG:                       true

System:                          Linux
PHP Version:                     7.4
Max uploaded file size:          2M
Max post size:                   8M
MySQL Version:                   8.0.23-0ubuntu0.20.04.1

Imagick:                         1
Imagick Active:                  1
Imagick Version:                 1690
GD Version:                      2.2.5

Config Information
------------------
version:                         040306
check_for_updates:               0
sorting_Photos_col:              taken_at
sorting_Photos_order:            ASC
sorting_Albums_col:              max_taken_at
sorting_Albums_order:            ASC
imagick:                         1
skip_duplicates:                 0
small_max_width:                 0
small_max_height:                360
medium_max_width:                1920
medium_max_height:               1080
lang:                            en
layout:                          1
image_overlay_type:              desc
default_license:                 none
compression_quality:             90
full_photo:                      1
delete_imported:                 0
Mod_Frame:                       1
Mod_Frame_refresh:               30
thumb_2x:                        1
small_2x:                        1
medium_2x:                       1
landing_page_enable:             0
landing_owner:                   John Smith
landing_title:                   John Smith
[...]

PR #1152, opened 2021-11-25
Fix diagnostics

ildyria:
More diagnostics. :)
Should help with #1149

### DIFF
--- a/app/Actions/Diagnostics/Checks/IniSettingsCheck.php
+++ b/app/Actions/Diagnostics/Checks/IniSettingsCheck.php
@@ -17,15 +17,12 @@ class IniSettingsCheck implements DiagnosticCheckInterface
 		$size = intval($size);
 
 		switch ($last) {
-			case 'G':
 			case 'g':
 				$size *= 1024;
 				// no break
-			case 'M':
 			case 'm':
 				$size *= 1024;
 				// no break
-			case 'K':
 			case 'k':
 				$size *= 1024;
 		}
--- a/app/Actions/Diagnostics/Checks/IniSettingsCheck.php
+++ b/app/Actions/Diagnostics/Checks/IniSettingsCheck.php
@@ -46,12 +46,12 @@ class IniSettingsCheck implements DiagnosticCheckInterface
 			$this->convert_size(ini_get('post_max_size')) < $this->convert_size('100M')
 		) {
 			$errors[]
-				= 'Warning: You may experience problems when uploading a photos of large size. Take a look in the FAQ for details.';
+				= 'Warning: You may experience problems when uploading a photo of large size. Take a look in the FAQ for details.';
 		}
 		$max_execution_time = intval(ini_get('max_execution_time'));
 		if (0 < $max_execution_time && $max_execution_time < 200) {
 			$errors[]
-				= 'Warning: You may experience problems when uploading a large amount of photos. Take a look in the FAQ for details.';
+				= 'Warning: You may experience problems when uploading a photo of large size or handling many/large albums. Take a look in the FAQ for details.';
 		}
 		if (empty(ini_get('allow_url_fopen'))) {
 			$errors[]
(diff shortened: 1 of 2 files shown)


app/Actions/Diagnostics/Checks/IniSettingsCheck.php
```diff
@@ -51,9 +48,8 @@ public function check(array &$errors): void
 			$errors[]
 				= 'Warning: You may experience problems when uploading a photos of large size. Take a look in the FAQ for details.';
 		}
-		if (
-			intval(ini_get('max_execution_time')) < 200
-		) {
+		$max_execution_time = intval(ini_get('max_execution_time'));
+		if (0 < $max_execution_time && $max_execution_time < 200) {
 			$errors[]
 				= 'Warning: You may experience problems when uploading a large amount of photos. Take a look in the FAQ for details.';
```
kamil4: I know this isn't one of the lines you changed, but I have a problem with it. Basically, the diagnostics is bogus. The value of `max_execution_time` is irrelevant to the success of uploading multiple photos. Photos are uploaded one-per-API-call. It used to matter for _import_, but these days we manually reset it for every imported photo so it's irrelevant there as well.

Basically, as far as photos are concerned, the value needs to be set to something long enough to process a _single_ photo. I would imagine that for any remotely sensible server, `30` would be enough, though I guess for some low-end RPi more may be needed?

The limit is more likely to be hit if somebody has thousands of albums than during photo upload/import...

ildyria: > 30 would be enough, though I guess for some low-end RPi more may be needed?

Or... very very big pictures... :see_no_evil:

app/Actions/Diagnostics/Checks/IniSettingsCheck.php
```diff
@@ -77,5 +73,10 @@ public function check(array &$errors): void
 			$errors[]
 				= 'Warning: exec function has been disabled. You may experience some error 500, please report them to us.';
 		}
+
+		if (empty(ini_get('user_agent'))) {
+			$errors[]
+				= 'Warning: php_agent is not set. You may experiences problems when importing images via URL.';
```
nagmat84: Proposal for better message: "Warning: user_agent for PHP is not set. .....".
The configuration option `user_agent` should be named correctly.

nagmat84 (changes requested):
LGTM. One minor proposal. I have no opinion about the discussion which time limit suffices as execution time limit, so I am out.

app/Actions/Diagnostics/Checks/IniSettingsCheck.php
ildyria:
```suggestion
				= 'Warning: php_agent for PHP is not set. You may experiences problems when importing images via URL.';
```

ildyria:
```
```suggestion
```
is the way to go @nagmat84  ;)

nagmat84:
```suggestion
				= 'Warning: user_agent for PHP is not set. You may experiences problems when importing images via URL.';
```

I added my counter-suggestion ;-) I wanted to say that the name of the config option is `user_agent` not `php_agent`.

app/Actions/Diagnostics/Checks/IniSettingsCheck.php
```diff
@@ -77,5 +73,10 @@ public function check(array &$errors): void
 			$errors[]
 				= 'Warning: exec function has been disabled. You may experience some error 500, please report them to us.';
 		}
+
+		if (empty(ini_get('user_agent'))) {
+			$errors[]
+				= 'Warning: user_agent for PHP is not set. You may experiences problems when importing images via URL.';
```
d7415:
```suggestion
				= 'Warning: user_agent for PHP is not set. You may experience problems when importing images via URL.';
```

nagmat84 (approved):

app/Actions/Diagnostics/Checks/IniSettingsCheck.php
nagmat84: Is this conversation resolved? Can we merge?

nagmat84 (approved):

kamil4 (approved):

PR #1153, opened 2021-11-26
Update Exec.php

corrilan:
Update to allow multiple paths to be specified during import from server, separated by comma.


ildyria:
Seems like you managed to get a Pull Request going. If you have any feedback with regard to https://lycheeorg.github.io/docs/contributions.html#illustrated-example please do tell so we can improve.

app/Actions/Import/Exec.php
```diff
@@ -165,128 +165,134 @@ private function memWarningCheck()
...
-		foreach ($files as $file) {
-			// re-read session in case cancelling import was requested
-			session()->start();
-			if (Session::has('cancel')) {
-				Session::forget('cancel');
-				$this->status_error($origPath, 'Import cancelled');
-				Logs::warning(__METHOD__, __LINE__, 'Import cancelled');
-
+		
+		$paths = explode(',', $path_list);
```
nagmat84:
```suggestion
		$paths = explode(';', $path_list);
```

As @ildyria said, `;` would be a much more common choice.

app/Actions/Import/Exec.php
```diff
@@ -165,128 +165,134 @@ private function memWarningCheck()
 	 * @throws ImagickException
 	 */
 	public function do(
-		string $path,
+		string $path_list,
```
nagmat84:
```suggestion
		array $paths,
```

or

```suggestion
		string $path,
```

I would like to see an proper array of paths here. "Exploding" a list of paths into an array (as below) should - if at all - happen outside of the method. Alternatively, the method might stay as is and be called iteratively for each path by the caller (this is by `FromServer`).

On top, I would actually go one step further and say it is the responsibility of the front-end to already pass a proper JSON array of paths to be imported and not a string separated by semicolons. I.e. the `ImportServerRequest` should not expect a string for `path` but an array for `paths`. This will help to implement other clients in the long run. Please stop the "stringification" of arguments.

nagmat84 (changes requested):
Please ensure that the client (CLI or web front-end) already passes a correct array of paths to be imported and not a single string separated by semicolons. There might be cases where the semicolon is part of a path and thus must not be used as separator.

app/Actions/Import/Exec.php
corrilan: Hi @nagmat84, @ildyria originally suggested a comma: "Comma separated is usually a more natural option."

nagmat84: OK, maybe, then I got that wrong. Anyway, I would expect either an array as input parameter or even call the method iteratively with a single path. See my  comment below wrt. to line 168.

ildyria: As per [this](https://superuser.com/a/748264), I would say screw you if you use comma or semi-colon in your path. :smile:

corrilan: Thanks for your input folks, I will get onto this but may be a while, will chip away at weekends, got hands full with a newborn at moment - Lychee is proving great for sharing the photos with the family.

d7415: @ildyria Awful source aside, well designed tools allow escaping.

ildyria:
I did a path on that with associated PR: https://github.com/LycheeOrg/Lychee-front/pull/308

app/Actions/Import/Exec.php
ildyria: Fixed.

app/Locale/German.php
```diff
@@ -445,8 +445,8 @@ public function get_locale(): array
 			'UPLOAD_IMPORT' => 'Importieren',
 			'UPLOAD_IMPORT_SERVER' => 'Importieren von Server',
 			'UPLOAD_IMPORT_SERVER_FOLD' => 'Der Ordner ist leer oder enthält keine lesbaren Dateien zum Verarbeiten. Schauen Sie bitte ins Protokoll (Einstellungen/Protokoll ansehen).',
-			'UPLOAD_IMPORT_SERVER_INSTR' => 'Diese Aktion wird alle Fotos, Ordner und Unterordner importieren, die sich in folgendem Verzeichnis befinden.',
-			'UPLOAD_ABSOLUTE_PATH' => 'Absoluter Pfad zum Verzeichnis',
+			'UPLOAD_IMPORT_SERVER_INSTR' => 'Import all photos, folders and sub-folders located in the folders with the following absolute paths (on server). Paths are space separated, use \\ to escape a space in a path.',
+			'UPLOAD_ABSOLUTE_PATH' => 'Absolute path to directories, space separated',
```
qwerty287:
```suggestion
			'UPLOAD_IMPORT_SERVER_INSTR' => 'Importiert alle Fotos, Ordner und Unterordernern in den folgenden absoluten Pfaden (auf dem Server). Mehrere Pfade können mit Leerzeichen getrennt werden, mit \\ können Sie ein Leerzeichen im Pfad verwenden.',
			'UPLOAD_ABSOLUTE_PATH' => 'Absolute Pfade zu Verzeichnissen, mit Leerzeichen getrennt',
```

app/Actions/Import/FromServer.php
```diff
@@ -25,10 +25,12 @@ public function do(string $path, ?Album $album, ImportMode $importMode): Streame
...
 		$response->headers->set('X-Accel-Buffering', 'no');
-		$response->setCallback(function () use ($path, $album, $exec) {
+		$response->setCallback(function () use ($paths, $album, $exec) {
 			// Surround the response by `[]` to make it a valid JSON array.
 			echo '[';
-			$exec->do($path, $album);
+			foreach ($paths as $path) {
+				$exec->do($path, $album);
+			}
 			echo ']';
```
qwerty287, line 34: Why do we echo brackets here to get a valid JSON response? Just asking, because it's a bit weird to print something in the action

nagmat84: Because it is a streamed response. For a streamed response STDOUT is redirected to the body of the HTTP response. The `echo` and the `$exec->do` are part of an anonymous function which generates the stream. The brackets `[` and `]` are the beginning and end of a JSON array. `$exec->do` prints output as well and creates the entries within the JSON array.

Somewhere else (I believe it is in the action handler for uploads) it is a bit more explicit. There we have a static function instead of an anonymous function which is called `streamGenerator`. Maybe we should do that here as well.

nagmat84 (approved):
Untested, but looks sane.

ildyria:
I have not tested it either. :innocent:

app/Locale/German.php
nagmat84:
```suggestion
			'UPLOAD_IMPORT_SERVER_INSTR' => 'Importiert alle Fotos, Ordner und Unterordner in den folgenden absoluten Pfaden (auf dem Server). Mehrere Pfade können mit Leerzeichen getrennt werden; mit \\ können Sie ein Leerzeichen im Pfad verwenden.',
			'UPLOAD_ABSOLUTE_PATH' => 'Absolute Pfade zu Verzeichnissen, mit Leerzeichen getrennt',
```

I corrected the spelling mistake "Unterordernern" --> "Unterordner".

Moreover, I recognized that we use the formal form to address a user "Sie" (for the Frenchies: vous) instead of the informal form "Du" ("tu"). I know we are very inconsistent about that. This inconsistency has been bothering me for quite a while now, but I haven't felt like fixing that on a global scale.

But we should establish a rule what we want to use and then stick to it:

 - Option 1: Don't address the user directly at all and resort to different alternatives like passive voice, etc. This avoids the problem for languages which have more than a single "you" for everything. Moreover we avoid the problem of deciding whether the user is male or female. (Again for languages which differentiates this aspect.) Something like
   
   > Do you want to delete the photo? This cannot be undone.
   
   would become
   
   > Shall the photo be deleted? This cannot be undone.
   
   Pro: No problem with "Du"/"tu" vs. "Sie"/"vous"; no problem with gender. Con: Passive voice sometimes leads the less legible phrases.

 - Option 2a: Directly address the user in a formal way. Use "Sie"/"vous". 
 
 - Option 2b: Directly address the user in an informal way. Use "Du"/"tu".

nagmat84:
@ildyria Before hitting the "apply" button the next time, read the comment below the suggestion ;-) https://github.com/LycheeOrg/Lychee/pull/1153#discussion_r931281253 The first suggestion contained a spelling error.

app/Locale/German.php
ildyria: > Pro: No problem with "Du"/"tu" vs. "Sie"/"vous"; no problem with gender. Con: Passive voice sometimes leads the less legible phrases.

In the same style, my PhD Supervisor wrote his full PhD in passive voice (:scream:). Personally, I went for the "we" as _the authors of the publications_ rather than "we" as _the reader and the author going through the manuscript_.

That being said, I would be going for option 2. Most confirmation windows are directed at the user. e.g. in Debian when doing `apt dist-upgrade`: `Do you want to continue? [Y/n]`

nagmat84: That is fine for English. But there was an option 2a vs. 2b. Formal or informal?

ildyria: FYI, the French translations are making use of the Formal form.

qwerty287: German uses formal too (at least the manpages of GNU coreutils).

ildyria:
Need to merge Lychee-front (waiting for other PR before) and then we can merge this.

PR #1157, opened 2021-12-03
bump to version 4.4.0 + remove unnecessary php-intl requirement

ildyria:


app/Actions/Diagnostics/Checks/PHPVersionCheck.php
```diff
@@ -49,7 +44,7 @@ public function check(array &$errors): void
 		}
 
 		// Extensions
-		$extensions = ['session', 'exif', 'mbstring', 'gd', 'PDO', 'json', 'zip', 'intl'];
+		$extensions = ['session', 'exif', 'mbstring', 'gd', 'PDO', 'json', 'zip'];
```
nagmat84, line 47: Why has this become unnecessary?

nagmat84 (approved):

app/Actions/Diagnostics/Checks/PHPVersionCheck.php
ildyria: it was removed from the Laravel required package for installation.

PR #1172, opened 2021-12-16
Some fixes

nagmat84:
**Enforce consistent SQL modes for MySQL/MariaDB**

I stumbled across this bug in combination with our admin user. By default, MySQL does not like primary IDs which equal 0.

**Config option `user_agent` is also set for CLI commands now**

Another bug I found while testing and playing around with my huge PR.


ildyria (approved):
LGTM

PR #1194, opened 2022-01-14
Fixes #1192

### LINKED ISSUE #1192: Updating breaks Lychee
Updating breaks Lychee. ( thank $deity for backups and databasedumps)

1. log on to (working) Lychee
2. Click on 'update'
3. After updating there is an error about missing admin in the database. please run:  php artisan lychee:reset_admin
4. Error page:database out of date.

**Screenshots**
*If applicable, add screenshots to help explain your problem.*

### Output of the diagnostics 

Diagnostics
-----------

System Information
------------------
Lychee Version (git):            master (a7de020) - Up to date (13 hours ago).
DB Version:                      4.4.0

APP_ENV:                         local

System:                          FreeBSD
PHP Version:                     8
Max uploaded file size:          100M
Max post size:                   100M
Max execution time:              0
MySQL Version:                   8.0.27

Imagick Version:                 1692
GD Version:                      2.3.1

Config Information
------------------
version:                         040400
check_for_updates:               1
sorting_Photos_col:              taken_at
sorting_Albums_order:            DESC
skip_duplicates:                 0
image_overlay_type:              date
landing_page_enable:             1
landing_owner:                   Marc
landing_title:                   Marc
landing_subtitle:                Why stop now?
landing_facebook:
landing_flickr:
landing_twitter:
landing_instagram:
landing_youtube:
landing_background:              dist/finland.jpg
site_title:                      fotos
site_copyright_begin:            1984
site_copyright_end:              2021
additional_footer_text:          
map_provider:                    OpenStreetMap.org
has_exiftool:                    1
location_show:                   0
upload_processing_limit:         4

### Browser and system

Browser: Firefox 96.0

nagmat84:


ildyria:
Wondering also about adding a diagnostic check.

nagmat84:
> Wondering also about adding a diagnostic check.

Feel free to do it. :stuck_out_tongue_closed_eyes: :smirk:

But actually, this diagnostic won't be required after my PR. This PR also adds a ton of foreign constraints to the DB which will prevent those inconsistencies in the future on the DB level.

ildyria:
gimme a sec.

kamil4 (approved):
Untested but looks good!

nagmat84:
@ildyria Do you still want to add your diagnostic output? It probably won't serve a purpose in the future with proper foreign constraints in place.

Otherwise feel free to merge this PR without me. I have to deal with some family matter now.

kamil4 (commented):
@ildyria I looked over your additions but I don't understand what's going on with the versions?
- Why is MissingUserCheck now only done for `>= 4.5.0` DB versions when before it was done unconditionally?
- Personally I would _not_ restrict CorruptedDbCheck to `< 4.5.0`. Doesn't this limit its usefulness effectively to zero? I get @nagmat84's point that the DB is _supposed to_ be enforcing these constraints now, but that's why this is called `CorruptedDbCheck` and why it's under `Diagnostics`. We _want_ to double-check; that's the whole point. Maybe the DB engine is crap. Maybe the user was mucking with the DB in the console a month ago and doesn't even remember anymore what they did and why. If the test is quick and is not on the critical path, it's worth doing.

nagmat84:
>     * Personally I would _not_ restrict CorruptedDbCheck to `< 4.5.0`. Doesn't this limit its usefulness effectively to zero?

Yes it does. But that is the reason, why I would completely drop these checks entirely. They do not serve any useful purpose.

In #1192, the migration fails because we had [crappy code like this](https://github.com/LycheeOrg/Lychee/blob/a7de0208e5c26a4849a4fc21d3729edd782501bb/app/Http/Controllers/Administration/UserController.php#L36-L49) in the past and no foreign key constraints on the DB level which prevented it. Hence, it was possible even by Lychee's own code to break the DB consistency.

The migration introduces these foreign key constraints ([see here](https://github.com/LycheeOrg/Lychee/blob/32c374db4505bb44bf5b357ee952034ad271515c/database/migrations/2021_12_04_181200_refactor_models.php#L520)) now and that is why the migration may fail. But this is basically a one-time problem.

These foreign key constrains also enforce that we write proper code which deals with relations correctly, i.e. code like the above will fail in the future. During beta-testing of the PR @kamil4 had found such a bug in the code which prevented the deletion of photos which were being used as covers of albums [as we discussed on Gitter](https://gitter.im/LycheeOrg/Developers?at=61d34b6e9b470f38974b1f72). The foreign key constraints did their job and denied an action which would have broken the consistency.

> I get @nagmat84's point that the DB is _supposed to_ be enforcing these constraints now, [...] We _want_ to double-check; that's the whole point. [...] Maybe the user was mucking with the DB in the console a month ago and doesn't even remember anymore what they did and why.

That won't be possible anymore neither. If the user issues a `DELETE`/`UPDATE`/`INSERT` statement which breaks DB consistency, then the DB will emit a constraint violation and not execute the query. So the user can't break the consistency. That is the whole point of having constraints on the DB level.

Of course, the user might intentionally break things by altering the DB scheme and dropping constraints, i.e. by running a command like `ALTER TABLE xxx DROP CONSTRAINT yyy`, but this is simply hostile. We cannot protect against a user which alters the DB scheme, e.g. dropping columns or whatever.

> We _want_ to double-check; that's the whole point. Maybe the DB engine is crap.

We must assume that the DB engine is not. Otherwise we had to check a lot of other things, too.

Now, let's assume that we _hypothetically_ want to double check these things. Then we should ask ourself against what adverse action we want to protect ourselves. A hostile user which alters the DB scheme? A crappy DB engine which is buggy? In this case I do not understand why the newly written check `Actions/Diagnostics/Checks/CorruptedDbCheck.php` only checks for missing users. That is completely inconsequential and will only puzzle future developers (and ourselves).

If we wanted to double check, then we had to check _all_ relations like I do in the migration. If crappy DB engines and hostile user are a real concern, then we also had to check if all non-null columns are really non-null. Probably there is much more.

I guess `Actions/Diagnostics/Checks/CorruptedDbCheck.php` only focuses on missing users, because we had crappy code [crappy code like this](https://github.com/LycheeOrg/Lychee/blob/a7de0208e5c26a4849a4fc21d3729edd782501bb/app/Http/Controllers/Administration/UserController.php#L36-L49) in the past and hence someone came up with `Actions/Diagnostics/Checks/MissingUserCheck.php` as a cheap work-around. And now this check has been taken as the basis for `CorruptedDbCheck.php`. At least the latter should not be called `CorruptedDbCheck.php`, if it only checks for missing users. The name of the check promises more than it actually does. At least merge it with `MissingUserCheck.php` into one.

But honestly, just nuke both. It is insane and will only confuse everybody in the future who does not know the history of that code.

ildyria:
> Why is MissingUserCheck now only done for >= 4.5.0 DB versions when before it was done unconditionally?

@kamil4 because this check is broken if your database is in version 4.4.0. The table name changed, thus if you try `artisan lychee:diagnostic` before the migration, the code will fail.

I don't mind refactoring this code, I just strongly believe those checks needs to be done either in Diagnostics or before the migration (and before graceful failure).

nagmat84:
> I just strongly believe those checks needs to be done either in Diagnostics or before the migration (and before graceful failure).

Why? The checks in the migration are done _before_ the migration is executed. I don't think that anybody will have a look in the Diagnostic before he/she runs the migration. That's why the migration must do it. Moreover, your diagnostic only will run the check for version >= 4.5.0, i.e. after the migration has been run, so it will be too late. Although, if the whole intent of this checks is to ensure a successful migration, then _all_ relations must be checked (as in the migration). This diagnostic does not make any sense and it will be completely obsolete in the future. We should not create dead code.

I don't understand why you have problems with a graceful failure of the migration.

ildyria:
I reread your code @nagmat84 I'm satisfied. I nuked both checks. :)

nagmat84:
@kami4 You might to review the new changes as well.

I contemplated about what the most frequent DB problem might be and what most user will do about the errors. Hence, I decided it might be nice to have a automatic correction for the obvious low-hanging fruits.

Probably, the most common error is a missing user. In the past, the old code simply deleted users without taking care, if there were any albums owned by the user or any shares for this user. Hence, many DB might be corrupt without the user having done anything wrong. These users should not be bothered with fixing there DB manually.

Another problematic category are orphaned albums. If we asked the users to repair there database manually, most users would probably either delete the infringing album or make it a top-level album. However, I don't assume that any user will take care to properly update the `_lft` and `_rgt` values of the nested tree set and thus the DB will still be in an inconsistent state. Fixing the tree can only be done properly, if automated.

Having automatic fixes will also make the web-based updater more robust.

kamil4:
I haven't tested it but I like the idea behind the new code. I too had doubts about the original message ("fix it manually"); I thought that perhaps we should simply tell people to reach out to us on Gitter. Fixing easy problems automatically is much better!

I still think it would've been a good idea to do a DB consistency test as part of the diagnostics, especially since we request that users run diagnostics as part of reporting an issue, so having sanity tests there helps. I agree that it should've been a more comprehensive test, at least like the one from the migration. The arguments against doing it strike me as... ideological (for lack of a better word). Creating a provably perfect DB schema only helps so much when we have no control over what users can do. Yes, foreign constraints should prevent inconsistencies, but can't foreign constraints be temporarily disabled with a simple DB console command? As to calling such actions "hostile", it reminds me of that old saying: _Never attribute to malice that which is adequately explained by stupidity_. Imagine the following scenario: A user will want to do something that is not supported by the UI. They will have just enough knowledge to access the DB directly but thanks to the foreign constraints the DB engine will respond with an error message. The user won't even understand the message; however, they will copy-paste it into Google and the first search hit will have the "advice" to force it through by disabling the constraints. Boom! They end up with an inconsistent DB. They may not even notice any problems until months later when they try to, I don't know, move an affected album around or something, and the operation craps out in a mysterious way. Sure, a diagnostics test would not _prevent_ any of that but at least it will give _us_ tools to diagnose things if the user complains.

Anyway, it's water under the bridge since the code is gone and I don't feel like doing it myself...

PR #1196, opened 2022-01-15
Add indices to speed up query for album covers

nagmat84:
These missing indices should rectify the performance regression @ildyria has observed. If @ildyria can confirm that these indices do the trick, we should be able to merge.

### DIFF
--- a/database/migrations/2021_12_04_181200_refactor_models.php
+++ b/database/migrations/2021_12_04_181200_refactor_models.php
@@ -718,6 +718,14 @@ class RefactorModels extends Migration
 			$table->index(['album_id', 'created_at']);
 			$table->index(['album_id', 'is_starred']);
 			$table->index(['album_id', 'is_public']);
+			$table->index(['album_id', 'type']);
+			// These indices are needed to efficiently retrieve the covers of
+			// albums acc. to different sorting criteria
+			// Note, that covers are always sorted acc. to `is_starred` first.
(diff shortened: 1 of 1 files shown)


nagmat84:
@ildyria If you have already run the migration and don't want to rollback and re-run again, you may also just add the indices manually and check if the performance problem is gone:

    CREATE INDEX photos_album_id_type_index ON photos (album_id, `type`);
    CREATE INDEX photos_album_id_is_starred_created_at_index ON photos (album_id, is_starred, created_at);
    CREATE INDEX photos_album_id_is_starred_taken_at_index ON photos (album_id, is_starred, taken_at);
    CREATE INDEX photos_album_id_is_starred_is_public_index ON photos (album_id, is_starred, is_public);
    CREATE INDEX photos_album_id_is_starred_type_index ON photos (album_id, is_starred, `type`);

kamil4:
Yeah, I sort-of wonder about how reasonable it is to modify existing migrations for code that's already been merged. I think we should create a new one...

nagmat84:
@kamil4 you should also review [LycheeOrg #47](https://github.com/LycheeOrg/LycheeOrg.github.io/pull/47). It gives an explanation.

The main problem is that you cannot add missing indices to an existing table programmatically. You must re-create the whole table. As this is the `photos` table, this means to re-create `albums` (because there are foreign constraints), this means to re-create ... I think you see the problem now.

kamil4:
Yikes!

Issue #1200, opened 2022-01-17
Cannot import files with "0000:00:00 00:00:00" exif dates


dcarbone:
### Detailed description of the problem [REQUIRED]
When a camera's battery dies, some set their internal clock to `0000:00:00 00:00:00`, and leave it there until it is manually set by it's operator.

In my exact example, there are three different exif fields this is present in:

* `CreateDate`
* `DateTimeOriginal`
* `ModifyDate`

Example extract using `exiftool`:

```
$ exiftool -T -CreateDate -DateTimeOriginal -ModifyDate DSCN2129.JPG 
0000:00:00 00:00:00	0000:00:00 00:00:00	0000:00:00 00:00:00
```

This particular camera is and older Nikon Coolpix, but I'm sure a similar issue impacts others.

What results is the following useless datetime value for the `taken_at` column: `-0001-11-30 00:00:00`, and MariaDB bombs when attempting to set this value.

I have had a quick look at the code in `Extractor::extract` starting at and following line `221`, and while it's too late for me to bother trying to be parse the logic and / or be clever, I may be able to do so tomorrow.

Perhaps even something as lazy as `if 0000:00:00 00:00:00 seen set to epoch` would suffice.

nagmat84:
Thanks for the bug report! Nicely spotted.

I assume it should also break on PostgreSQL, because `0000-00-00 00:00:00` is simply no valid time and I wonder how `Carbon` (the PHP date/time class) behaves, especially when it comes to timezone conversion.

> Perhaps even something as lazy as if `0000:00:00 00:00:00` seen set to epoch would suffice.

In principle yes, but don't use epoch as a sentinal value. Epoch is a perfectly valid time. Instead, use `null`.

All these date attributes are optional and hence nullable. I believe it should be somewhere in the `MetaExtractor` (or similar): instead of returning the parsed datetime as a `Carbon` instance, return `null`. The string equality test should take place even before the string gets fed into the `Carbon` parsing function.

nagmat84:
But as I said, we are always open to new contributors and PRs. The project is here https://github.com/LycheeOrg/php-exif. Feel free to code!

PR #1201, opened 2022-01-17
Add check for non-viable exif dates

### LINKED ISSUE #1267: Invalid datetime format: 1292 Incorrect datetime value: '-0001-11-30 00:00:00.000000' for column `lychee1`.`photos`.`taken_at` at row 1
I am unable to upload the following image. I can see this is because the image does not have a valid date. I'm happy for there to be a default such as epoch zero, or todays date.
P1010089

### Full output from laravel.log
```
[2022-04-11 12:15:10] production.ERROR: SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect datetime value: '-0001-11-30 00:00:00.000000' for column `lychee1`.`photos`.`taken_at` at row 1 (SQL: insert into `photos` (`tags`, `updated_at`, `created_at`, `title`, `type`, `checksum`, `original_checksum`, `iso`, `aperture`, `make`, `model`, `shutter`, `focal`, `taken_at`, `taken_at_orig_tz`, `is_public`, `is_starred`, `album_id`, `owner_id`, `id`, `legacy_id`) values (, 2022-04-11 12:15:10.158385, 2022-04-11 12:15:10.158385, P1010089, image/jpeg, 69a1451436cfe8115604229f34c8cb9babefa0b2, 69a1451436cfe8115604229f34c8cb9babefa0b2, 100, f/2.0, OLYMPUS OPTICAL CO.,LTD, C2040Z, 1/125 s, 12 mm, -0001-11-30 00:00:00.000000, UTC, 0, 0, UED9jVdEx_Go1PjnOhcQs0q0, 0, 3RGm5bp3s4wacOsLtB9zqMMV, 16496793101607)) {"exception":"[object] (Illuminate\\Database\\QueryException(code: 22007): SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect datetime value: '-0001-11-30 00:00:00.000000' for column `lychee1`.`photos`.`taken_at` at row 1 (SQL: insert into `photos` (`tags`, `updated_at`, `created_at`, `title`, `type`, `checksum`, `original_checksum`, `iso`, `aperture`, `make`, `model`, `shutter`, `focal`, `taken_at`, `taken_at_orig_tz`, `is_public`, `is_starred`, `album_id`, `owner_id`, `id`, `legacy_id`) values (, 2022-04-11 12:15:10.158385, 2022-04-11 12:15:10.158385, P1010089, image/jpeg, 69a1451436cfe8115604229f34c8cb9babefa0b2, 69a1451436cfe8115604229f34c8cb9babefa0b2, 100, f/2.0, OLYMPUS OPTICAL CO.,LTD, C2040Z, 1/125 s, 12 mm, -0001-11-30 00:00:00.000000, UTC, 0, 0, UED9jVdEx_Go1PjnOhcQs0q0, 0, 3RGm5bp3s4wacOsLtB9zqMMV, 16496793101607)) at /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Connection.php:712)
[stacktrace]
#0 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Connection.php(672): Illuminate\\Database\\Connection->runQueryCallback()
#1 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Connection.php(502): Illuminate\\Database\\Connection->run()
#2 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Connection.php(454): Illuminate\\Database\\Connection->statement()
#3 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(2980): Illuminate\\Database\\Connection->insert()
#4 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(1657): Illuminate\\Database\\Query\\Builder->insert()
#5 /usr/local/lychee/app/Models/Extensions/HasRandomIDAndLegacyTimeBasedID.php(95): Illuminate\\Database\\Eloquent\\Builder->__call()
#6 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(994): App\\Models\\Photo->performInsert()
#7 /usr/local/lychee/app/Actions/Photo/Strategies/AddStandaloneStrategy.php(45): Illuminate\\Database\\Eloquent\\Model->save()
#8 /usr/local/lychee/app/Actions/Photo/Create.php(104): App\\Actions\\Photo\\Strategies\\AddStandaloneStrategy->do()
#9 /usr/local/lychee/app/Http/Controllers/PhotoController.php(105): App\\Actions\\Photo\\Create->add()
#10 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Controller.php(54): App\\Http\\Controllers\\PhotoController->add()
#11 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php(45): Illuminate\\Routing\\Controller->callAction()
#12 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Route.php(262): Illuminate\\Routing\\ControllerDispatcher->dispatch()
#13 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Route.php(205): Illuminate\\Routing\\Route->runController()
#14 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Router.php(721): Illuminate\\Routing\\Route->run()
#15 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(128): Illuminate\\Routing\\Router->Illuminate\\Routing\\{closure}()
#16 /usr/local/lychee/app/Http/Middleware/UploadCheck.php(44): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#17 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): App\\Http\\Middleware\\UploadCheck->handle()
#18 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Middleware/SubstituteBindings.php(50): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#19 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Routing\\Middleware\\SubstituteBindings->handle()
#20 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php(78): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#21 /usr/local/lychee/app/Http/Middleware/VerifyCsrfToken.php(61): Illuminate\\Foundation\\Http\\Middleware\\VerifyCsrfToken->handle()
#22 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): App\\Http\\Middleware\\VerifyCsrfToken->handle()
#23 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Session/Middleware/AuthenticateSession.php(39): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#24 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Session\\Middleware\\AuthenticateSession->handle()
#25 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/View/Middleware/ShareErrorsFromSession.php(49): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#26 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\View\\Middleware\\ShareErrorsFromSession->handle()
#27 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Session/Middleware/StartSession.php(121): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#28 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Session/Middleware/StartSession.php(64): Illuminate\\Session\\Middleware\\StartSession->handleStatefulRequest()
#29 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Session\\Middleware\\StartSession->handle()
#30 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Cookie/Middleware/AddQueuedCookiesToResponse.php(37): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#31 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Cookie\\Middleware\\AddQueuedCookiesToResponse->handle()
#32 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Cookie/Middleware/EncryptCookies.php(67): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#33 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Cookie\\Middleware\\EncryptCookies->handle()
#34 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(103): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#35 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Router.php(723): Illuminate\\Pipeline\\Pipeline->then()
#36 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Router.php(698): Illuminate\\Routing\\Router->runRouteWithinStack()
#37 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Router.php(662): Illuminate\\Routing\\Router->runRoute()
#38 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Router.php(651): Illuminate\\Routing\\Router->dispatchToRoute()
#39 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php(167): Illuminate\\Routing\\Router->dispatch()
#40 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(128): Illuminate\\Foundation\\Http\\Kernel->Illuminate\\Foundation\\Http\\{closure}()
#41 /usr/local/lychee/vendor/livewire/livewire/src/DisableBrowserCache.php(19): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#42 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Livewire\\DisableBrowserCache->handle()
#43 /usr/local/lychee/vendor/bepsvpt/secure-headers/src/SecureHeadersMiddleware.php(21): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#44 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Bepsvpt\\SecureHeaders\\SecureHeadersMiddleware->handle()
#45 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/TransformsRequest.php(21): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#46 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/ConvertEmptyStringsToNull.php(31): Illuminate\\Foundation\\Http\\Middleware\\TransformsRequest->handle()
#47 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Foundation\\Http\\Middleware\\ConvertEmptyStringsToNull->handle()
#48 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/TransformsRequest.php(21): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#49 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php(40): Illuminate\\Foundation\\Http\\Middleware\\TransformsRequest->handle()
#50 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Foundation\\Http\\Middleware\\TrimStrings->handle()
#51 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/ValidatePostSize.php(27): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#52 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Foundation\\Http\\Middleware\\ValidatePostSize->handle()
#53 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php(86): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#54 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Foundation\\Http\\Middleware\\PreventRequestsDuringMaintenance->handle()
#55 /usr/local/lychee/vendor/fideloper/proxy/src/TrustProxies.php(57): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#56 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Fideloper\\Proxy\\TrustProxies->handle()
#57 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(103): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#58 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php(142): Illuminate\\Pipeline\\Pipeline->then()
#59 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php(111): Illuminate\\Foundation\\Http\\Kernel->sendRequestThroughRouter()
#60 /usr/local/lychee/public/index.php(74): Illuminate\\Foundation\\Http\\Kernel->handle()
#61 {main}

[previous exception] [object] (PDOException(code: 22007): SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect datetime value: '-0001-11-30 00:00:00.000000' for column `lychee1`.`photos`.`taken_at` at row 1 at /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Connection.php:501)
[stacktrace]
#0 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Connection.php(501): PDOStatement->execute()
#1 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Connection.php(705): Illuminate\\Database\\Connection->Illuminate\\Database\\{closure}()
#2 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Connection.php(672): Illuminate\\Database\\Connection->runQueryCallback()
#3 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Connection.php(502): Illuminate\\Database\\Connection->run()
#4 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Connection.php(454): Illuminate\\Database\\Connection->statement()
#5 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(2980): Illuminate\\Database\\Connection->insert()
#6 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(1657): Illuminate\\Database\\Query\\Builder->insert()
#7 /usr/local/lychee/app/Models/Extensions/HasRandomIDAndLegacyTimeBasedID.php(95): Illuminate\\Database\\Eloquent\\Builder->__call()
#8 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(994): App\\Models\\Photo->performInsert()
#9 /usr/local/lychee/app/Actions/Photo/Strategies/AddStandaloneStrategy.php(45): Illuminate\\Database\\Eloquent\\Model->save()
#10 /usr/local/lychee/app/Actions/Photo/Create.php(104): App\\Actions\\Photo\\Strategies\\AddStandaloneStrategy->do()
#11 /usr/local/lychee/app/Http/Controllers/PhotoController.php(105): App\\Actions\\Photo\\Create->add()
#12 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Controller.php(54): App\\Http\\Controllers\\PhotoController->add()
#13 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php(45): Illuminate\\Routing\\Controller->callAction()
#14 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Route.php(262): Illuminate\\Routing\\ControllerDispatcher->dispatch()
#15 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Route.php(205): Illuminate\\Routing\\Route->runController()
#16 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Router.php(721): Illuminate\\Routing\\Route->run()
#17 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(128): Illuminate\\Routing\\Router->Illuminate\\Routing\\{closure}()
#18 /usr/local/lychee/app/Http/Middleware/UploadCheck.php(44): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#19 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): App\\Http\\Middleware\\UploadCheck->handle()
#20 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Middleware/SubstituteBindings.php(50): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#21 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Routing\\Middleware\\SubstituteBindings->handle()
#22 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php(78): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#23 /usr/local/lychee/app/Http/Middleware/VerifyCsrfToken.php(61): Illuminate\\Foundation\\Http\\Middleware\\VerifyCsrfToken->handle()
#24 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): App\\Http\\Middleware\\VerifyCsrfToken->handle()
#25 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Session/Middleware/AuthenticateSession.php(39): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#26 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Session\\Middleware\\AuthenticateSession->handle()
#27 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/View/Middleware/ShareErrorsFromSession.php(49): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#28 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\View\\Middleware\\ShareErrorsFromSession->handle()
#29 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Session/Middleware/StartSession.php(121): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#30 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Session/Middleware/StartSession.php(64): Illuminate\\Session\\Middleware\\StartSession->handleStatefulRequest()
#31 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Session\\Middleware\\StartSession->handle()
#32 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Cookie/Middleware/AddQueuedCookiesToResponse.php(37): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#33 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Cookie\\Middleware\\AddQueuedCookiesToResponse->handle()
#34 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Cookie/Middleware/EncryptCookies.php(67): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#35 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Cookie\\Middleware\\EncryptCookies->handle()
#36 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(103): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#37 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Router.php(723): Illuminate\\Pipeline\\Pipeline->then()
#38 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Router.php(698): Illuminate\\Routing\\Router->runRouteWithinStack()
#39 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Router.php(662): Illuminate\\Routing\\Router->runRoute()
#40 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Routing/Router.php(651): Illuminate\\Routing\\Router->dispatchToRoute()
#41 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php(167): Illuminate\\Routing\\Router->dispatch()
#42 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(128): Illuminate\\Foundation\\Http\\Kernel->Illuminate\\Foundation\\Http\\{closure}()
#43 /usr/local/lychee/vendor/livewire/livewire/src/DisableBrowserCache.php(19): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#44 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Livewire\\DisableBrowserCache->handle()
#45 /usr/local/lychee/vendor/bepsvpt/secure-headers/src/SecureHeadersMiddleware.php(21): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#46 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Bepsvpt\\SecureHeaders\\SecureHeadersMiddleware->handle()
#47 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/TransformsRequest.php(21): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#48 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/ConvertEmptyStringsToNull.php(31): Illuminate\\Foundation\\Http\\Middleware\\TransformsRequest->handle()
#49 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Foundation\\Http\\Middleware\\ConvertEmptyStringsToNull->handle()
#50 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/TransformsRequest.php(21): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#51 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php(40): Illuminate\\Foundation\\Http\\Middleware\\TransformsRequest->handle()
#52 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Foundation\\Http\\Middleware\\TrimStrings->handle()
#53 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/ValidatePostSize.php(27): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#54 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Foundation\\Http\\Middleware\\ValidatePostSize->handle()
#55 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php(86): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#56 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\\Foundation\\Http\\Middleware\\PreventRequestsDuringMaintenance->handle()
#57 /usr/local/lychee/vendor/fideloper/proxy/src/TrustProxies.php(57): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#58 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Fideloper\\Proxy\\TrustProxies->handle()
#59 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(103): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#60 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php(142): Illuminate\\Pipeline\\Pipeline->then()
#61 /usr/local/lychee/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php(111): Illuminate\\Foundation\\Http\\Kernel->sendRequestThroughRouter()
#62 /usr/local/lychee/public/index.php(74): Illuminate\\Foundation\\Http\\Kernel->handle()
#63 {main}
"}
```

### Browser and system
- PHP 8.0.17
- Lychee 4.5.0
- MariaDB 10.5.8
- Alpine Linux 3.15
- sql_mode =  STRICT_TRANS_TABLES,​ERROR_FOR_DIVISION_BY_ZERO,​NO_AUTO_CREATE_USER,​NO_ENGINE_SUBSTITUTION

dcarbone:
- Addresses #1200
- Add test and assets for zero and minimum-viable exif date values
- Add check to ensure "taken_at" value is usable


app/Metadata/Extractor.php
```diff
@@ -10,6 +10,15 @@
...
 class Extractor
 {
+	/*
+	 * https://github.com/LycheeOrg/Lychee/issues/1200
+	 *
+	 * Minimum possible recordable date value.
+	 *
+	 * Equates to 0000:01:01 12:00:00
+	 */
+	private const MINIMUM_VALID_DATE_UTC = -62167176000;
```
nagmat84, line 20: Why is this the minimum valid UTC date? What is about dates BC? Sure, there were no digital cameras but please consider digitized drawings with artificially created EXIF data which points to the time when the original drawing has presumably be created. This is what some national archives do.

Please provide a rational for this number.

Also note, that this number cannot be represented by 32bit and we still have users with 32bit installations. This code won't work for them.

app/Metadata/Extractor.php
```diff
@@ -477,4 +486,22 @@ public function extract(string $fullPath, string $kind): array
 
 		return $metadata;
 	}
+
+	/**
+	 * https://github.com/LycheeOrg/Lychee/issues/1200.
+	 *
+	 * Ensures the provided DateTime value is functional
+	 *
+	 * @param bool|\DateTime $d
```
nagmat84, line 495:
```suggestion
	 * @param \DateTime $d
```

app/Metadata/Extractor.php
```diff
@@ -477,4 +486,22 @@ public function extract(string $fullPath, string $kind): array
...
+	 *
+	 * @return bool
+	 */
+	protected function date_is_valid(bool|\DateTime $d): bool
```
nagmat84, line 499:
```suggestion
	protected function date_is_valid(\DateTime $d): bool
```

app/Metadata/Extractor.php
```diff
@@ -477,4 +486,22 @@ public function extract(string $fullPath, string $kind): array
...
+	{
+		if (false === $d) {
```
nagmat84, line 501: Drop the condition.

app/Metadata/Extractor.php
```diff
@@ -10,6 +10,15 @@
 
 class Extractor
 {
+	/*
+	 * https://github.com/LycheeOrg/Lychee/issues/1200
```
nagmat84, line 14: Please remove that first line. The first line is used by PHPdoc to generate a summary and by IDEs to generate a tooltip. A reference to an issue which will (hopefully) be closed, is not helpful.

app/Metadata/Extractor.php
```diff
@@ -477,4 +486,22 @@ public function extract(string $fullPath, string $kind): array
 
 		return $metadata;
 	}
+
+	/**
+	 * https://github.com/LycheeOrg/Lychee/issues/1200.
```
nagmat84, line 491: Please remove, see above.

app/Metadata/Extractor.php
```diff
@@ -477,4 +486,22 @@ public function extract(string $fullPath, string $kind): array
...
+	 *
+	 * @return bool
+	 */
+	protected function date_is_valid(bool|\DateTime $d): bool
+	{
+		if (false === $d) {
+			return false;
+		}
+
+		return self::MINIMUM_VALID_DATE_UTC <= intval($d->format('U'));
```
nagmat84, line 505: That is the condition I feared above.

While I understand that `0000-00-00 00:00:00` is not a valid date, because there is no zero month and day, I wonder why a valid `\DateTime` should print such a date when converted to a string. When I first read the issue #1200, I thought that maybe the EXIF extractor returned a ill-formatted string `0000-00-00 00:00:00` if this was the EXIF data. But here we already have a valid `\DateTime` instance.

Also please don't use the conversion chain `\DateTime` --> `string` --> `int`. This is just unnecessary and bad. It will even break on 32bit systems after 2038. If we really needed to enforce a lower bound on the timestamp, then use the appropriate methods of the `\DateTime` interface for comparison. You may even use the normal mathematical operators `<`, `<=`, `>=` and `>` for comparison between two `\DateTime` objects, because these operators are overloaded for `\DateTime`. (Yes, PHP allows operator overloading just like C++).

While I believe you that this might solve the bug for you, I am still not convinced that this is the least invasive solution.

nagmat84 (changes requested):

app/Metadata/Extractor.php
dcarbone: Couple of things:

1. The minimum possible valid date the `CreateDate`, `ModifyDate`, and `DateTimeOriginal` exif fields can go is `0000:01:01 00:00:00`.  They are not intended to handle BC dates.
2. If we were worried about being able to handle BC dates, that should be a different unit of work that looks at the XMP fields defined here: https://exiftool.org/TagNames/XMP.html#iptcExt
    * http://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#circa-date-created for some more reading
3. Good point, it should be a string:
    ```
    $ uname -m
    armv7l
    $ php -r 'var_dump(PHP_INT_MAX, date_create_from_format("U", "-62167176000"));'
    int(2147483647)
    object(DateTime)#1 (3) {
      ["date"]=>
      string(26) "0000-01-01 12:00:00.000000"
      ["timezone_type"]=>
      int(1)
      ["timezone"]=>
      string(6) "+00:00"
    }
    ```
4. Zero and negative year values (indicating a Gregorian BC year) is not supported by Postgres when inserting into a timestamp field using an iso 8601 string value without first wrapping the value in a `to_timestamp` function call, as such:
    ```
    # select to_timestamp('-0001-11-30', 'YYYY');
              to_timestamp        
     ---------------------------
     0001-01-01 00:00:00+14 BC
    ```
    * This is from a test container where I set the TZ to `UTC+14` to see how it behaved
    * As an example, this works:
    ```
    # insert into photos (id, legacy_id, title, type, checksum, original_checksum, created_at, updated_at, taken_at) values (1, 1, 'test1', 'photo', 'asdf', 'asdf', now(), now(), to_timestamp('-0001-01-01', 'YYYY-MM-DD bc'));
    INSERT 0 1
    ```
    * But this doesn't:
    ```
    # insert into photos (id, legacy_id, title, type, checksum, original_checksum, created_at, updated_at, taken_at) values (1, 1, 'test1', 'photo', 'asdf', 'asdf', now(), now(), '-0001-01-01');
    ERROR:  invalid input syntax for type timestamp: "-0001-01-01"
    LINE 1: ..., 'test1', 'photo', 'asdf', 'asdf', now(), now(), '-0001-01-..
    ```
    * There has been some [recent discussion](https://www.postgresql.org/message-id/9q0367bkf9b.fsf%40gmx.us) on supporting `+/-` prefix on on number values in pgsql iso 8601 parsing, so maybe this will change in the future (but I doubt it).
5. I chose that value after some experimentation 
6. Zero year values _are_ supported by MariaDB's timestamp columns, but negative values are not.

I ultimately chose that value after some local experimentation with sqlite and a MariaDB container, however after the [initial test failed](https://github.com/LycheeOrg/Lychee/runs/4844977600?check_suite_focus=true#step:17:75) I tried to quickly update before I had to head out to run some errands.

I've since spent some additional time testing locally with pgsql, and it seems the minimum possible iso 8601 date value pgsql will accept is `0001:01:01 00:00:00`.

app/Metadata/Extractor.php
dcarbone: I originally intended for this function call to be used for both `false` and value checking, but I ended up being pressed for time and didn't get around to it before I had to leave.  Unless there is a reason to explicitly check for `false` _and_ execute the call, I'd prefer to just execute the call for both scenarios.

app/Metadata/Extractor.php
dcarbone: I can see an argument for moving it to a lower position, however it is very common to annotate a given bit of logic with a link to the issue it stemmed from, as it provides an easy way to gain some additional context about why something has been implemented as it has without having to copy the entire context in-code.

Assuming such context is desired.  If not, I can remove it.

app/Metadata/Extractor.php
dcarbone: So technically speaking exif values like that are just ascii fields.  Without some kind of external control, the value can be whatever you want.   The Nikon in question, and I assume most / all Nikons of the era and possibly other cameras, have an invalid default value.

[\DateTime will happily consume a 0000:00:00 value](https://3v4l.org/SU8lg), and _technically_ it is responsible for further mutating the value to the one seen in DB errors.

> Also please don't use the conversion chain \DateTime --> string --> int. This is just unnecessary and bad.

I beg to differ.  In my opinion, an integer representing a Unix timestamp is the absolute easiest and least error-prone way to programmatically represent time.  However, this is not my project and I don't really feel like arguing on this point.

Additionally, post 2038 32bit systems will have a lot more to worry about than an exif field sanity check :)

I have a test setup for pgsql and mariadb locally now, so I will make some updates tonight or tomorrow.

dcarbone:
The crucial piece here is to determine the actual lowest usable timestamp value across all supported database types.  I will come back with a better solution to this soon.

app/Metadata/Extractor.php
nagmat84: I don't argue against context, but the context should be in-place.

app/Metadata/Extractor.php
nagmat84: Your argument suffers from the implicit assumption that one has to use ISO 8601 strings.

Only because a specific DB (e.g. PostgreSQL) does not support to insert a date like "-0009-14-01 17:34:00" with a negative date part does not mean that these dates aren't supported at all. I am even not sure if this counts as a valid ISO 8601 string. The lowest by PostgreSQL supported `DateTime` is something around 4000 BC. You yourself gave an example how to insert a "negative" date into PostgreSQL: you have to use the suffix BC instead of a prefix sign operator.

app/Metadata/Extractor.php
nagmat84: Please don't. Using union types is bad in most cases and in this specific case it is not necessary at all. If the caller has not a valid `\DateaTime` object, the caller should already know that is obviously not a valid datetime.

The only reason why you added `bool` here is by co-incidents, due to a bad design decision of the exif extractor to return `false` on error instead of throwing a proper exception. This may change in the future (when I will find the time).

app/Metadata/Extractor.php
nagmat84: > Additionally, post 2038 32bit systems will have a lot more to worry about than an exif field sanity check :)

This doesn't mean that we need to contribute to that problem.

>> Also please don't use the conversion chain \DateTime --> string --> int. This is just unnecessary and bad.

> I beg to differ. In my opinion, an integer [...] is the absolute easiest and least error-prone way to programmatically represent time.

This doesn't impu that you have to convert it to a string first.

> \DateTime will happily consume a 0000:00:00 value, and technically it is responsible for further mutating the value to the one seen in DB errors.

As you can see from your own example, DateTime accepts that input, but already normalizes it to the nearest valid value and stores that value internally.

That is exactly what I assumed in the first place. If we already have a DateTime object, the test is too late, because anything which can be represented by DateTime will already be a valid date. So the problem must be fixed at some complete different place, i.e. inside the exif extractor _before_ the DateTime object is even constructed and we still have access to the original ASCII string from the uninterpreted, raw EXIF data.

ildyria:
While I like the idea of the fix, I would have the tendency to believe it should be placed in: https://github.com/LycheeOrg/php-exif

nagmat84:
> While I like the idea of the fix, I would have the tendency to believe it should be placed in: https://github.com/LycheeOrg/php-exif

Exactly! Actually, there are two different aspects to be considered and they require to be solved at different places:

a) ASCII strings in the Exif data which represent invalid dates, such as `0000-00-00 00:00:00`

These issues must be handled _before_ LycheeOrg/php-exif the string is parsed into a `DateTime` object. As the example shows, a `DateTime` object always encodes a valid datetime. Anything else would also have surprised me. If the constructor of `DateTime` gets an invalid string, e.g. 2021-02-30` it will be rounded/normalized to the nearest valid date like `2021-03-02`. But after that is impossible to tell what the original date was. So these things must be handled inside LycheeOrg/php-exif

b) DateTime values which are generally valid but out of the scope which a specific DBMS can handle. 

This is a completely different pair of shoes and should not be handled on the level of the EXIF data extractor. The EXIF data extractor should be agnostic of the latter usage of the data.

This issue should be handled by the conversion routine which persists the DateTime value to the DBMS. In particular, the valid range may depend on the DBMS.

This is also the right place to do it, because not every instance of a DateTime object which is persisted to the DBMS originates from EXIF data.

app/Metadata/Extractor.php
dcarbone: Couple of things:

1. As I mentioned, MariaDB also cannot accept negative ISO 8601 values.
2. I would be all for modifying the serialization of timestamp columns to be something a bit more flexible, however there are a few things to consider here:
    1. As I mentioned above, The exif date fields _will never be negative_
    1. This issue is _purely_ to prevent improper interpolation of exif date values 
    1. I think it may make sense to increase the # of date fields represented in the DB so Lychee can handle those extended exif date fields I mentioned above.

app/Metadata/Extractor.php
dcarbone: Personally I think it being `false or \DateTime` is fine, as that is exactly how the upstream PHP `\DateTime` creation mechanisms behave, so I don't think that is worthy of a change.

Ultimately, though, either way works.

dcarbone:
@nagmat84: I agree with these arguments.  I'll work on something there.

dcarbone:
For what its worth, handling "BC" date values in PHP is a bit tricky.  We'll probably have to end up using the [IntlGregorianCalendar](https://www.php.net/manual/en/class.intlgregoriancalendar.php) suite of classes.

I propose that handling of BC dates, and altering the serialization to be something other than the default as done by Eloquent should be handled outside of this ticket as it will probably be more invasive.

stephen-hill:
Is there any indication on when this will be merged? As an outside observer, it appears like this issue has stalled due to various issues.

Would it be possible to implement a simpler method so that #1200 is resolved quicker?

nagmat84:
Unfortunately, I cannot give you any indication. Lychee is mostly developed by only a team of 4 volunteers who all have another fulltime job professionally. Sorry, that I cannot tell you something more optimistic.

But the issue is a good first issue, because the problem is quite obvious and straightforward to solve, nonetheless a little bit tedious.


The main problem is that there is not a quick and simple solution which works reliably. I totally agree that Lychee could be made more robust and fault-tolarent with respect to broken timestamps. But strictly spoken the problem is with the photos not Lychee. A "trivial" solution always bears the risk that errors which should better be reported a hidden away, too. So I rather prefer Lychee to properly fail for erroneous input than accidentally and silently "repairing" broken inputs in a wrong way.

stephen-hill:
@nagmat84 I understand. Thanks for the status update.

I totally understand the need for an elegant solution, but I was wondering if something as simple as the following would be an acceptable compromise?

```php
if ($date === '0000-00-00 00:00:00') {
  // Insert NULL into the database
}
```

nagmat84:
> but I was wondering if something as simple as the following would be an acceptable compromise?
> 
> ```php
> if ($date === '0000-00-00 00:00:00') {
>   // Insert NULL into the database
> }
> ```

Exactly this does not work, because there is simply no string variable `$date` which could be compared to '0000-00-00 00:00:00'. Within Lychee all dates are handled as instances of `Carbon` and this class never generates invalid dates like this. If you have a `Carbon` instance, then it always represents a valid date.

The problem must be fixed in php-exif, before the broken input is parsed by Carbon.

kamil4:
@stephen-hill Please see https://github.com/LycheeOrg/php-exif/pull/38. You can simply copy the contents of https://raw.githubusercontent.com/LycheeOrg/php-exif/0f00cfb8ecf1a39418d56a577ad84a1a1653dc99/lib/PHPExif/Mapper/Native.php and https://raw.githubusercontent.com/LycheeOrg/php-exif/0f00cfb8ecf1a39418d56a577ad84a1a1653dc99/lib/PHPExif/Mapper/Exiftool.php to your `Lychee/vendor/lychee-org/php-exif/lib/PHPExif/Mapper/` folder and that should do it.

kamil4:
Closing this as we instead fixed this bug in our dependency, php-exif: https://github.com/LycheeOrg/php-exif/pull/38.

stephen-hill:
@kamil4 and @nagmat84 

I just want to say thank you for resolving this issue.

PR #1210, opened 2022-01-31
Fixes 1208

nagmat84:
This PR fixes the `takedate` console command.


nagmat84:
Rebasing problems. Actually only a single file should have changed, but git shows 25. I'll give it another try tomorrow.

PR #1212, opened 2022-02-01
Fixes #1208

### LINKED ISSUE #1208: Complete the update of 'app/Console/Commands/Takedate.php' to the refactored timestamps

[Takedate.php](https://github.com/LycheeOrg/Lychee/blob/master/app/Console/Commands/Takedate.php) is not in all areas up2date with the latest timestamps implementation of Lychee. I.e. [takestamp](https://github.com/LycheeOrg/Lychee/blob/master/app/Console/Commands/Takedate.php#L71-L72) is no longer available as it seemed to be replaced with `taken_at` during the last refactoring. I guess at least in this line the last update was incomplete.

Please update - caused by https://github.com/LycheeOrg/Lychee/issues/1206 I have a huge amount of pictures to be updated with the correct timestamp information ;-)

nagmat84:
Fixes #1208 and syncs @kamil4's latest changes to the front-end while we are on it.


app/Console/Commands/Takedate.php
```diff
@@ -29,80 +38,140 @@ class Takedate extends Command
...
+			// The latter is actually an independent operation of the first.
+			if (!$useFileTime) {
+				$this->printError('Failed to get takestamp data for photo "' . $photo->title . '" (ID=' . $photo->id . ').');
 				continue;
 			}
 			if (is_link($fullPath)) {
 				$fullPath = readlink($fullPath);
 			}
 			$created_at = filemtime($fullPath);
 			if ($created_at == $photo->created_at->timestamp) {
```
kamil4: We don't need to compare the timezones here like you do for `taken_at`? earlier

ildyria (approved):
LGTM.

app/Console/Commands/Takedate.php
```diff
@@ -29,80 +38,140 @@ class Takedate extends Command
...
+					$this->printError('Failed to update takestamp for photo "' . $photo->title . '" (ID=' . $photo->id . ').');
 				}
 				continue;
 			}
-			if (!$timestamps) {
-				$this->line($i . ': Failed to get Takestamp data for ' . $photo->title . '.');
+
+			// TODO: The description of this command's signature suggests something different than we do here.
+			// The description suggests that we would use the file date as a fall back for `taken_at`, if EXIF data is not available.
+			// However, we set `created_at` (aka upload time) here.
```
kamil4: That is surprising, although now that you mention it, it does seem to ring a bell. @ildyria, was it you who wanted it to work that way?

kamil4 (approved):
Untested but looks sane.

app/Console/Commands/Takedate.php
nagmat84: `created_at` (and `updated_at`) only store the datetime in UTC without any timezone indication. So it is sufficient to compare the timestamp. This is also the only information we get from the filesystem anyway.

We only additionally store the original timezone in `taken_at_orig_tz` alongside `taken_at` in UTC so that we can present the user a datetime which matches the user's expectation when they look at the photo (otherwise a photo taken during broad daylight in Australia would show a midnight timestamp).

app/Console/Commands/Takedate.php
nagmat84: Also keep in mind that we call `touch` on the media file during import and explicitly set the creation time of the file to the extracted `taken_at` attribute of the EXIF data. In this case, the reported creation time of the file does not match the upload time but indeed the takedate. However, if we cannot extract the EXIF information upon import, then we don't `touch` the media file and the creation date of the file remains equal to the upload time.

Considering that I am not sure if we can use the creation time of the file at all. We simply don't know if the creation time of the file has been faked to equal the extracted takedate or if the creation time of the file equals the import time :-(

_Question:_ Why do we `touch` the media file during import at all? If we did not do that we would at least know that the creation time of the file is always the upload time.

Also I am not sure if I like to  fiddle with the `created_at` attribute at all. Again, I am thinking about mobile clients which cash a shadow database for offline usage. The columns `created_at` and `updated_at` are used to calculate a 3-way merge if both the master and the shadow database were modified during an offline period. Fiddling around with these "internal", "technical" attributes might have consequences which I cannot asses, but it feels wrong to me.

nagmat84: @ildyria Could you say something about this part? I would like to know, if I should do something about this before I merge the PR or not.

ildyria: the `touch` part is Legacy from Tobias:
https://github.com/LycheeOrg/Lychee-v3/blob/master/php/Modules/Photo.php#L256
or https://github.com/electerious/Lychee/blob/master/php/Modules/Photo.php#L223 in the original repo.
I am totally fine with removing it.

nagmat84: If nobody disagrees, I will do the following:

 * Drop the `touch` upon uploading. This way the timestamp of the file always consistently match the upload time. In particular, `touch` won't work anyway for Flysystem disks hosted on S3.
 * Allow to the set the upload time (aka `created_at`) from the timestamp of the filesystem as an _independent_ option, not as a fall-back if the takedate cannot be set. These two timestamps are completely independent.

ildyria: > Drop the touch upon uploading. This way the timestamp of the file always consistently match the upload time. In particular, touch won't work anyway for Flysystem disks hosted on S3.

:+1: 

> Allow to the set the upload time (aka created_at) from the timestamp of the filesystem as an independent option, not as a fall-back if the takedate cannot be set. These two timestamps are completely independent.

:+1:

nagmat84: Applied changes as discussed.

PR #1213, opened 2022-02-03
Fix some potential null references

nagmat84:
As a side result of the discussion on the annotated front-end, @kamil4 pointed out that a photo might not have a thumbnail, if the photo is a video and FFmpeg is not installed (probably there are some other cases, too, but this is the most prominent one).

Hence, I checked with the back-end if there is any code which wrongly assumes that thumbnails always exists. Luckily, the situation has not been as bad as I initially feared, but I found to issues nonetheless.


ildyria (approved):
LGTM.

kamil4 (approved):
Looks good!

PR #1239, opened 2022-03-03
Move filesize to size variants

Chostakovitch:
Hi,

As discussed on Gitter, this PR moves the filesize from `Photo` to `SizeVariant`. This was done because it was impossible to know the filesize of a variant from the API, which could be useful in some situations (in my case, to implement a FUSE-based Lychee filesystem).

This PR :
- Adds a migration which does **not** set real filesize values of variants, to avoid bottlenecks for large instances. Instead, a value of `0` indicates that the filesize has not been set yet.
- Adds a  `lychee:variant_filesize` artisan command to manually set all variant filesizes
- Updates model and tests to fit the new table layout.

I did my best to understand how things work, to respect the code style and to thoroughly test my modifications, but this is my first time with Laravel, so your attention is especially needed to avoid mistakes.

Thanks!


app/Console/Commands/VariantFilesize.php
```diff
@@ -0,0 +1,56 @@
...
+	 * @return mixed
+	 */
+	public function handle()
+	{
+		if ($this->confirm('This command can take tens of minutes for large instances. Do you really want to run it now ?')) {
+			$variants = SizeVariant::query()
+				->where('filesize', '=', 0)
+				->get();
+
+			if (count($variants) == 0) {
```
ildyria: Here is a better approach:
```php
$variants_query = SizeVariant::query()
				->where('filesize', '=', 0);
$count = $variants_query->count();
if ($count == 0) { 
...
} 
$variants = $variants_query->get();
...
```

The difference being that if you do count on `$variants` it will fetch all the variants, hydrate the models and potentially explode the memory usage.
While if you do count() directly, the count is in SQL an no longer in PHP. Saving a large amount of time.

Also worth considering is a from / limit. That way you allow to work on smaller chunk of data.

app/Contracts/SizeVariantFactory.php
```diff
@@ -50,12 +50,13 @@ abstract public function cleanup(): void;
...
-	 * @param int $width  the width of the original size variant
-	 * @param int $height the height of the original size variant
+	 * @param int $width    the width of the original size variant
+	 * @param int $height   the height of the original size variant
+	 * @param int $filesize the filesize of the original size variant
 	 *
 	 * @return SizeVariant the freshly created and persisted size variant
 	 */
-	abstract public function createOriginal(int $width, int $height): SizeVariant;
+	abstract public function createOriginal(int $width, int $height, int $filesize): SizeVariant;
```
ildyria, line 59:
```suggestion
	abstract public function createOriginal(int $width, int $height, int $filesize = 0): SizeVariant;
```

app/Models/Extensions/SizeVariants.php
```diff
@@ -178,10 +178,11 @@ public function getThumb(): ?SizeVariant
 	 *                                size variant shall point to
 	 * @param int    $width           the width of the size variant
 	 * @param int    $height          the height of the size variant
+	 * @param int    $filesize        the filesize of the size variant
 	 *
 	 * @return SizeVariant The newly created and persisted size variant
 	 */
-	public function create(int $sizeVariantType, string $shortPath, int $width, int $height): SizeVariant
+	public function create(int $sizeVariantType, string $shortPath, int $width, int $height, int $filesize): SizeVariant
```
ildyria, line 185:
```suggestion
	public function create(int $sizeVariantType, string $shortPath, int $width, int $height, int $filesize = 0): SizeVariant
```

app/Image/SizeVariantDefaultFactory.php
```diff
@@ -277,15 +277,23 @@ protected function createSizeVariantInternal(int $sizeVariant, int $maxWidth, in
 
 		$sv = $this->photo->size_variants->getSizeVariant($sizeVariant);
 		if (!$sv) {
-			$sv = $this->photo->size_variants->create($sizeVariant, $shortPath, $maxWidth, $maxHeight);
+			// Filesize is not known before creating the file, set to 0 meanwhile
+			$sv = $this->photo->size_variants->create($sizeVariant, $shortPath, $maxWidth, $maxHeight, 0);
```
ildyria, line 283: by using a default value, you no longer need to add the argument here.

ildyria (commented):
Still need to be tested, but it is already a nice step.

You need to fix the command. This will make it faster and allow a lower memory footprint.

database/migrations/2022_02_02_203008_filesize_size_variants.php
```diff
@@ -0,0 +1,108 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+/**
+ * Add a filesize colum for size variants to make
```
kamil4:
```suggestion
 * Add a filesize column for size variants to make it
```

app/Console/Commands/VariantFilesize.php
```diff
@@ -0,0 +1,56 @@
...
+	protected $description = 'Set filesize of size variants if missing';
+
+	/**
+	 * Execute the console command.
+	 *
+	 * @return mixed
+	 */
+	public function handle()
+	{
+		if ($this->confirm('This command can take tens of minutes for large instances. Do you really want to run it now ?')) {
```
kamil4:
```suggestion
		if ($this->confirm('This command can take tens a long time for large instances. Do you really want to run it now?')) {
```

kamil4 (approved):
Looks good overall!

app/Console/Commands/VariantFilesize.php
d7415: I think you meant

```suggestion
		if ($this->confirm('This command can take a long time for large instances. Do you really want to run it now?')) {
```

app/Image/SizeVariantDefaultFactory.php
Chostakovitch: Done.

app/Models/Extensions/SizeVariants.php
Chostakovitch: Done with -1 default value (see migration and command).

app/Console/Commands/VariantFilesize.php
Chostakovitch: Thanks a lot for your suggestions, indeed it would be way better.
I changed the default value from 0 to -1 to be sure that -1 means "no value yet", because a filesize could still be 0 (even if very unlikely) and mess up with the offset/limit logic. 

Tell me if you think the command is now satisfying. :)

Chostakovitch:
Thanks everyone for reviewing the PR. I tried to apply your suggestion, I tested again migrations/adding new photo/rotating etc and it seems to work.

nagmat84:
You will get my review tomorrow. I won't get back to the computer sooner.

Just a quick thought about preliminary creating size variants preliminary with a file size of zero and update them later. (In this context, @ildyria suggested to use a default parameter of zero.) I am wondering if this is really necessary or whether we can create the final DB entry with the correct file size right away.

However, I haven't checked with the code, if this is possible at all.

A more thorough review will come tomorrow.

app/Actions/Photo/Strategies/AddStandaloneStrategy.php
```diff
@@ -171,7 +172,7 @@ protected function normalizeOrientation(): void
 			// and the checksum may have changed.
 			/* @var  Extractor $metadataExtractor */
 			$metadataExtractor = resolve(Extractor::class);
-			$this->photo->filesize = $metadataExtractor->filesize($absolutePath);
+			$this->photo->size_variants->getOriginal()->filesize = $metadataExtractor->filesize($absolutePath);
```
nagmat84: I wonder why this does not trigger an error. Did you test the code with images that require auto-rotation upon import?

The problem is that `normalizeOrientation` is called very early during import on the temporarily uploaded file. At this point no size variant (not even the original size variant) has been created.

`normalizeOrientation` is called in https://github.com/LycheeOrg/Lychee/blob/a8c1354890fdd8d49b3a73931764df1b5ed566b6/app/Actions/Photo/Strategies/AddStandaloneStrategy.php#L47 and the original size variant is created some lines of code later in https://github.com/LycheeOrg/Lychee/blob/a8c1354890fdd8d49b3a73931764df1b5ed566b6/app/Actions/Photo/Strategies/AddStandaloneStrategy.php#L60

Hence, `photo->size_variants->getOriginal()` inside `normalizeOrientation` should return `null`.

app/Console/Commands/VariantFilesize.php
```diff
@@ -0,0 +1,72 @@
...
+
+			// Number of queries
+			$pages = ceil($count / $limit);
+
+			$bar = $this->output->createProgressBar($count);
+			$bar->start();
+
+			for ($i = 0; $i < $pages; $i++) {
+				// No need to offset, because previous variants have no -1 filesize anymore
+				$variants = $variants_query->limit($limit)->get();
```
nagmat84: This is overly complicated. There is no need to implement the paging algorithm oneself. Simply use `->lazyById` of Eloquent. This generates an iterable collection which only holds a single page in memory at a time. Like this

```php
$variants = $variant_query->lazyById($limit);

foreach($variants as $variant) {
  // do your stuff here
}
``` 

There is no need for an outer loop which iterates over the pages manually. By the way, you already use lazy iteration in the migration.

app/Console/Commands/VariantFilesize.php
```diff
@@ -0,0 +1,72 @@
...
+
+				/* @var SizeVariant $variant */
+				foreach ($variants as $variant) {
+					$fullPath = $variant->full_path;
```
nagmat84: Please don't use the attribute `full_path` of size variant in new code. I know it is still there for legacy reasons. Using `full_path` will make the upcoming work of this other guy which wants to implement support for other than local storage only more difficult.

At least use the following code `$fullPath = $variant->getFile()->getAbsolutePath()`. `getFile()` returns a `FlysystemFile` object. Although this code is slightly better, it is still not optimal, because `file_exist($fullPath)` and `filesize($fullPath)` won't work if `getAbsolutePath` might return a non-local path in the future (but maybe a AWS S3 bucket ID).

So, optimally you would use `$variant->getFile()->getSize()` and not use any path at all. However, I know that the method `getSize` does not exist yet and still has to be implemented in a way which uses the Flysystem API. So I won't ask you to do that. But at least do step 1.

app/Models/SizeVariant.php
```diff
@@ -42,6 +42,7 @@
  * @property string              full_path
  * @property int                 width
  * @property int                 height
+ * @property int          		 filesize
```
nagmat84:
```suggestion
 * @property int          		  filesize
```

app/Models/Extensions/SizeVariants.php
```diff
@@ -178,10 +178,11 @@ public function getThumb(): ?SizeVariant
 	 *                                size variant shall point to
 	 * @param int    $width           the width of the size variant
 	 * @param int    $height          the height of the size variant
+	 * @param int    $filesize        the filesize of the size variant
 	 *
 	 * @return SizeVariant The newly created and persisted size variant
 	 */
-	public function create(int $sizeVariantType, string $shortPath, int $width, int $height): SizeVariant
+	public function create(int $sizeVariantType, string $shortPath, int $width, int $height, int $filesize = -1): SizeVariant
```
nagmat84: I know that @ildyria asked you to add a default parameter. However, I haven't yet found the the place where we need to create a size variant without the correct, final size right away. I probably have to take a second review round.

database/migrations/2022_02_02_203008_filesize_size_variants.php
```diff
@@ -0,0 +1,108 @@
...
+	{
+		// To avoid doing I/O on every photo, which would be prohibitive
+		// on large instances, and because JPEG compression makes approximations
+		// from original filesize too approximative, just set filesize to -1
+		// for every variant except the original, indicating that the size
+		// has not been calculated yes.
+		// The real calculation can be done by calling `artisan lychee:variant_filesize` from the CLI.
+		// This will replace the default value by the actual filesize but takes time.
+		Schema::table(self::VAR_TAB, function (Blueprint $table) {
+			$table->bigInteger(self::SIZE_COL)->nullable(false)->default(-1);
```
nagmat84: I know that we need some kind of "sentinel" element which indicates a missing size, because we don't add the correct size during migration.

However, I wonder whether

1. we should use `null` instead of `-1`,
2. make the column unsigned instead, and
3. enhance the console command `lychee:variant_filesize` to make the column not-nullable after all file sizes have been added

This will give us the chance to have a correctly typed column in the future. This is meant as an open discussion, because I have no string opinion about that.

However, I have an opinion about not calculating file sizes right during migration. Have we actually tested how long it takes? I have the feeling that we only claim that it will be inefficient. I did some preliminary tests some time ago and reported that on Gitter. Even for a large DB like @ildyria's it should only be around 10minutes. Yes, this is longer than normally, but IMHO not prohibitive. I would really like to see @ildyria, @d7415 and @kamil4 to run the console command on their installation for a test and report some runtime figures.

If the results turn out to be not too bad, I would really recommend to calculate file sizes during migration. That would allow us to make the column a non-nullable, unsigned integer. Otherwise we will be hunted either by a signed or nullable column in the future.

nagmat84: Have you checked if the foreign key constraint from `size_variants.photo_id` to `photo.id` remains intact if you alter the table scheme (i.e. add a column here and drop a column during the reverse migration)?

I had some bad experiences with that in the past. Eloquent tends to drop foreign keys and column restrictions of the whole table, if you modify the table scheme even if the modified/added/dropped column is completely unrelated to the constraints. Even if it does not happen for MySQL or PostgreSQL it sometimes happens with SQLite.

d7415: > it should only be around 10minutes. Yes, this is longer than normally, but IMHO not prohibitive.

10 mins for a field that 99% of users won't care about? That sounds prohibitive.

app/Image/SizeVariantDefaultFactory.php
```diff
@@ -280,12 +280,19 @@ protected function createSizeVariantInternal(int $sizeVariant, int $maxWidth, in
 			$sv = $this->photo->size_variants->create($sizeVariant, $shortPath, $maxWidth, $maxHeight);
```
nagmat84: OK, I found the cause why you made the parameter `filesize` of `SizeVariants::create` optional.

As far as I see it, the only reason why we create the size variant first before we have the actual media file is that we need to get the absolute file path via `$sv->full_path`, because the image handler requires it. Otherwise all attributes (i.e. height, width, etc.) are already known, so we could create the media file first, obtain the file size and _then_ create the size variant. That would lead to a much more consistent approach.

Also note what I said about `->short_path` earlier. As soon as we support anything else than local storage, the image handler must use proper file streams anyway and not rely on absolute paths. From that moment on we could swap the order easily.

However, this is nothing which can be done now easily. So let us keep it that way for the moment.

app/Models/Extensions/SizeVariants.php
nagmat84: OK, found the cause here: https://github.com/LycheeOrg/Lychee/pull/1239#discussion_r820224978

However, as there is only a single place where we need to create a size variant with a dummy file size of `-1` first and even this specific line of code should better not exist (see linked comment), I would rather prefer not to add a default value here. This is only an invitation to create size variants without a proper file size. Optimally, size variants a are only created with a proper file size the same way we also require a proper with and height.

@ildyria What do you think? Can we revert this change?

app/Actions/Photo/Strategies/AddStandaloneStrategy.php
Chostakovitch: As I will probably say again later, I feel sorry to propose poor quality code. I thought that I sufficiently tested and understood the code, but I clearly did not take enough time to understand consequences of the code.

I did not test the auto-rotation feature (didn't know it) and obviously it throws an error.

Anyway, I will remove this line because the size variant filesize is set in the variant factory.

Thanks for your vigilance.

nagmat84: Au contraire, I honestly think that you delivered high-quality code. In particular, you used type annotations and you added comments to your code. Both aspects are not a matter of course. Moreover, your comments were helpful and provided the reader (and future programmer) with helpful insights. Unfortunately, I have read too many poor comments in my life which only repeated information which a reader could also deduce from the code itself, but completely missed to comment on the why and how the code relates to the rest of the program. Your comments don't fall into that category and this is a rather rare occasion. 👍

app/Models/Extensions/SizeVariants.php
ildyria: Sure. No strong opinion.

app/Actions/Photo/Strategies/AddStandaloneStrategy.php
Chostakovitch: Thanks for your kind words. 

See 0a54f0eb.

Chostakovitch: However @nagmat84  I have a really bizarre phenomenon with auto resize and filesizes. I cannot think to an obvious reason, do you have any hints ?

1. Take `mongolia.jpeg` example image and fake orientation :
```
exiftool -Orientation=8 mongolia.jpeg
```
2. Import it in Lychee : auto-rotation takes place.
3. Check values for filesize in DB :
```
sqlite> select * from size_variants;
44|LOXWvW3mP85ugGUVoaas_fuy|0|big/97c0d9b2b4e2b17c0a41e34eb1099167.jpeg|1280|850|201315
45|LOXWvW3mP85ugGUVoaas_fuy|6|thumb/97c0d9b2b4e2b17c0a41e34eb1099167.jpeg|200|200|14800
46|LOXWvW3mP85ugGUVoaas_fuy|5|thumb/97c0d9b2b4e2b17c0a41e34eb1099167@2x.jpeg|400|400|45564
47|LOXWvW3mP85ugGUVoaas_fuy|4|small/97c0d9b2b4e2b17c0a41e34eb1099167.jpeg|542|360|51115
48|LOXWvW3mP85ugGUVoaas_fuy|3|small/97c0d9b2b4e2b17c0a41e34eb1099167@2x.jpeg|1084|720|160063
```
4. Drop all values :
```
sqlite> update size_variants set filesize=-1;
```
5. Run the command to set filesizes :
```
php artisan lychee:variant_filesize
```
6. Re-check filesizes values :
```
sqlite> select * from size_variants;
44|LOXWvW3mP85ugGUVoaas_fuy|0|big/97c0d9b2b4e2b17c0a41e34eb1099167.jpeg|1280|850|203237
45|LOXWvW3mP85ugGUVoaas_fuy|6|thumb/97c0d9b2b4e2b17c0a41e34eb1099167.jpeg|200|200|14800
46|LOXWvW3mP85ugGUVoaas_fuy|5|thumb/97c0d9b2b4e2b17c0a41e34eb1099167@2x.jpeg|400|400|45564
47|LOXWvW3mP85ugGUVoaas_fuy|4|small/97c0d9b2b4e2b17c0a41e34eb1099167.jpeg|542|360|51115
48|LOXWvW3mP85ugGUVoaas_fuy|3|small/97c0d9b2b4e2b17c0a41e34eb1099167@2x.jpeg|1084|720|160063
```

Notice how the original variant filesize is a little bit different.

app/Console/Commands/VariantFilesize.php
Chostakovitch: Tried to do it better there : d20d6e24

app/Console/Commands/VariantFilesize.php
Chostakovitch: See 1128a51b.

Chostakovitch: plus, didn't know Flysystem, seems cool :)

app/Models/SizeVariant.php
Chostakovitch: See 1128a51b2

app/Models/Extensions/SizeVariants.php
Chostakovitch: Indeed I agree that using a default value only for that case, that will probably change, is not useful. See d02ea7e68.

app/Actions/Photo/Strategies/AddStandaloneStrategy.php
nagmat84: No, not by heart, but I have an educated guess. PHP caches access to files and their meta-data internally (i.e. the low-level commands `filesize`, `is_file`, `file_exist`, etc.) don't access the underlying filesystem, if there was a previous call for the same file and PHP finds the result in its cache.

As the metadata extractor calls `filesize` before we auto-rotate the file and then calls `filesize` again, we might get the (wrong, outdated) result from before the rotation.

However, if you nuke the entries in the DB and then call the console command, this is another PHP process with an empty cache.

nagmat84: For example, see here: https://stackoverflow.com/questions/3747982/php-filesize-reporting-old-size

database/migrations/2022_02_02_203008_filesize_size_variants.php
Chostakovitch: @nagmat84 I cannot be really helpful on this one. `-1` seems fine to me and modifying the schema in a command seems odd, intuitively.   At first I used `0` as a default value and I changed it because if a filesize was really `0` (should not, but technically can be), I was not sure that the paging query would correctly work, but `lazyById` seems to do the job.

What do you think ? Could `0` do the job as a default value, avoiding nullable or signed column ?

Chostakovitch: Hmmm, I remember that during my test, one time, SQLite complained about a foreign key drop, but it never happened again.

How can I make sure, in the code, not to drop restrictions of the whole table ?

app/Actions/Photo/Strategies/AddStandaloneStrategy.php
Chostakovitch: Nice guess. However, running `clearstatcache()`, even globally (i.e. without a file path) after the image has been rotated (and before variant creations, according to your first comment) does not change the behaviour. 

I noticed that the same thing occurs, but only with 1 byte difference, on some image without auto-rotation. 

This may be worth investigating, as I guess it was already behaving that way before my modifications, and could mislead programs depending on it (not much, for now...) ?

database/migrations/2022_02_02_203008_filesize_size_variants.php
nagmat84: IMHO, in this case `0` should do as a sentinel for "no value" as well, because there should be no real file with size `0`.

nagmat84: I don't know any better way than testing with every DB engine and manually comparing the table definitions before and after the migration.

For PostgreSQL it is `\d size_variants`, for SQLite it is `.schema size_variants` and for MySQL (I assume) `SHOW CREATE TABLE size_variants`. In the case of MySQL, I am not sure if this command also shows the foreign key constraints or if you need another command.

Chostakovitch: Done again in 790bf82124.

Chostakovitch: I managed to reproduce the problem I've had.

When trying to migrate existing data with SQLite, I got :

```
SQLSTATE[23000]: Integrity constraint violation: 19 FOREIGN KEY constraint failed (SQL: DROP TABLE photos)
```

with error pointing [from there](https://github.com/LycheeOrg/Lychee/pull/1239/files#diff-9a22abf7cfc5b5a793e236cd8d4a040d80f66d308a0457ca3617a5b4fa0dcafcR66) - i.e. the `dropColumn`. How could a  `dropColumn` ending up trying to delete the `photos` table ??

nagmat84: So my gut feeling was right.

Dropping a column in SQLite is implemented by renaming the existing table to some temporary name (e.g. `size_variants_old`), re-creating the table with the one column removed, copying all data from the renamed, old table to the freshly created one and finally deleting the whole old table.

Sadly, this implementation does not take foreign constraints into account.

There are only two options here:

a) Do the same as above manually, but also properly handle foreign constraints. Unfortunately, this will quickly evolve into a mess, because eventually you will end up re-creating the whole database (size variants will lead to photos, photos to albums, photos and albums to users, ...)

b) Do not drop the column on the downward migration.

Chostakovitch: @nagmat84 The problem happens in the upward migration, when trying to drop the `filesize` column of the `photos` table.

Chostakovitch: > Also as indicated [there](https://www.sqlite.org/lang_altertable.html) the `filesize` column seems to check all conditions to that SQLite manages to drop the column.
>
> Even doing it manually :
> ```
> alter table size_variants add column filesize integer not null default '0';
alter table photos drop column filesize;
> ```
> 
> throws no error. I'll try to investigate further...

This doesn't surprise me. This strange implementation is part of Laravel. So it is likely that SQLite might be able to drop the column anyway. That's why I recommend trying a `DB::raw` statement.

nagmat84: I didn't see that the upward migration requires a `DROP COLUMN`, too. The explanation remains valid though, i.e. dropping a column from `photos` requires to rename the table, etc.

That's bad, because we obviously need to upward migration.

I checked with the SQLite documentation. It seems that SQLite should support `DROP COLUMN`, so you could try a `DB::raw` statement.

Chostakovitch: Oh, so you mean that Eloquent is implementing column drop with SQLite this way ? Seems a bit odd. I should probably drop the column with a raw statement, thanks for your time and advice.

Chostakovitch: I used a raw statement to delete the column in af1cfe59ae650d11258385070f7a2435301568e9.

app/Actions/Photo/Strategies/AddStandaloneStrategy.php
Chostakovitch: Again you were right, clearing the cache was needed as the path do not changes.
Also, simply removing the line was not enough as the variant creation code was using the old size.

Could you please review 3c0a33bae61deba0ef51d9e863b2290a53528c1c ?

database/migrations/2022_02_02_203008_filesize_size_variants.php
```diff
@@ -0,0 +1,123 @@
...
+		 * However, in this scenario, the command fails because of :
+		 * `FOREIGN KEY constraint failed (SQL: DROP TABLE photos)`.
+		 * This is a really strange bug, because redoing by hand all migration commands just work.
+		 * To avoid corrupting user databases, and because SQLite now support column deletion,
+		 * just run the command manually.
+		 */
+		// Schema::table(self::PHOTOS_TAB, function (Blueprint $table) {
+		// 	$table->dropColumn(self::SIZE_COL);
+		// });
+		DB::statement('ALTER TABLE ' . self::PHOTOS_TAB . ' DROP COLUMN ' . self::SIZE_COL);
```
ildyria, line 80: This is not valid SQLite query. That is why it is breaking.

app/Actions/Photo/Strategies/AddStandaloneStrategy.php
nagmat84: Looks good, although a little bit hack-ish, but again this is not your fault but due to Lychee's code which has already been there.

IHMO, in an optimal world, we would use `FlysystemFile` everywhere and the image handlers (i.e. `GdHandler` and `ImagickHandler`) would take care of flushing the case when they modify an image file. It should be the concern of the code which modifies data to invalidate cached data, not the concern of the caller.

But we should leave it like this for now. From my point of view, this conversation can be resolved.

app/Console/Commands/VariantFilesize.php
nagmat84: Looks good. The conversation can be resolved! :+1:

database/migrations/2022_02_02_203008_filesize_size_variants.php
nagmat84: Looks good. :+1: Conversation can be resolved!

nagmat84: Looks good from inspection. I cannot test it myself right now, because I am on a business trip and I don't have a development environment at hand. But I assume you tested it.

However, our automatic test suite [fails with `General error: 1 near "DROP": syntax error (SQL: ALTER TABLE photos DROP COLUMN filesize)` for SQLite](https://github.com/LycheeOrg/Lychee/runs/5440206812?check_suite_focus=true). I don't know why. Maybe the test environment uses an ancient version of SQLite?

app/Console/Commands/ExifLens.php
```diff
@@ -80,8 +80,8 @@ public function handle()
 			if (file_exists($fullPath)) {
 				$info = $this->metadataExtractor->extract($fullPath, $photo->type);
 				$updated = false;
-				if ($photo->filesize == '' && $info['filesize'] != '') {
-					$photo->filesize = $info['filesize'];
+				if ($photo->size_variants->getOriginal()->filesize == 0 && $info['filesize'] != '') {
```
nagmat84:
```suggestion
				if ($photo->size_variants->getOriginal()->filesize === 0 && $info['filesize'] != '') {
```

app/Console/Commands/VariantFilesize.php
```diff
@@ -0,0 +1,61 @@
...
+	 * @var string
+	 */
+	protected $description = 'Set filesize of size variants if missing';
+
+	/**
+	 * Execute the console command.
+	 *
+	 * @return mixed
+	 */
+	public function handle()
```
nagmat84: The return values are inconsistent. You return `false`, if the command does not find anything to do, and you return nothing (or `void`) in case of success.

Actually, the return type should be `int` and the method should return `0` on success and a non-zero POSIX error code in case of failure. For the sake of simplicity return `-1` in case of doubt.

```suggestion
	 * @return int
	 */
	public function handle(): int
```

app/Console/Commands/VariantFilesize.php
```diff
@@ -0,0 +1,61 @@
...
+
+		if ($this->confirm('This command can take a long time for large instances. Do you really want to run it now ?')) {
+			$variants_query = SizeVariant::query()
+				->where('filesize', '=', 0)->orderBy('id');
+
+			$count = $variants_query->count();
+			if ($count == 0) {
+				$this->line('All filesize variants already set in database.');
+
+				return false;
```
nagmat84:
```suggestion
				return -1;
```

See above.

app/Console/Commands/VariantFilesize.php
```diff
@@ -0,0 +1,61 @@
...
+					$variant->filesize = filesize($fullPath);
+					if (!$variant->save()) {
+						$this->line('Failed to update filesize for ' . $fullPath . '.');
+					}
+				} else {
+					$this->line('File does not exist for ' . $fullPath . '.');
+				}
+			});
+		}
+	}
```
nagmat84, line 65:
```suggestion
	
	return 0;
	}
```

See comment above about return codes.

database/migrations/2022_02_02_203008_filesize_size_variants.php
Chostakovitch: @nagmat84 Indeed, the ubuntu image uses [SQLite v3.31.1](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md) whereas `DROP COLUMN` is supported from [v3.35.0](https://www.sqlite.org/releaselog/3_35_0.html) (dec. 2021)

This version is quite new and not included yet in LTS distributions repositories... thus is seems hard to ask users to have that minimum version, right ? :/

d7415: > This version is quite new and not included yet in LTS distributions repositories... thus is seems hard to ask users to have that minimum version, right ? :/

I see you haven't met our PHP requirements :laughing:

app/Console/Commands/VariantFilesize.php
Chostakovitch: To be honest I took inspiration from `ExifLens` command as I didn't know the "Laravel-way" to do stuff. Didn't know that PHP had type hints either. Happy to learn!

See c0c52ae6e44bd36f4f0419945b2713ed367bf48f.

database/migrations/2022_02_02_203008_filesize_size_variants.php
Chostakovitch: Haha, indeed PHP 8 is not in bullseye either but in bookworm, as SQLite 3.35. But I feel that requiring this version only for a feature that nobody will use is maybe a bit harsh. What would you do ? Try to migrate the table with copy/drop "by hand" ?

nagmat84: A migration "by hand" is not as easy as you might think, because dropping the tables `photos` and re-creating it is like opening a can of worms. You will end up re-creating nearly the whole DB due to the foreign constraints. If you want to see what happens, I recommend you to look at https://github.com/LycheeOrg/Lychee/blob/a8c1354890fdd8d49b3a73931764df1b5ed566b6/database/migrations/2021_12_04_181200_refactor_models.php It is a tiny, small, innocent migration of only 2000 LOC. 😂

I am not sure how many users actually use SQLite. Unfortunately, we don't have any usage statistics about that. However, it is not the recommended solution. Maybe we could require a higher minimal version for SQLite without annoying too many users.

I am unsure. @ildyria should decide.

d7415: I guess the other option is to leave the column there and null and try to remember to drop it later, when it's more widely available? That might not help with the `down` though. And I'm not sure how many other things that would mess up.

nagmat84: I propose the following compromise and like all compromises it is really ugly, horrible and hack-ish. I am ashamed of myself for even suggesting it.

If the migration detects MySQL or PostgreSQL as the DB engine, it could execute the raw DB statement to drop the column `filesize`. If the migration detects SQLite, it simply does nothing but leaves the column as it. The column won't be used anymore, but simply remains as it is. I don't like the idea of having an obsolete, unmaintained, abandoned column in the DB, because everybody will wonder what it is for in the future, but it is the least worst solution I can think of.

Note, the column has a set default value, so it does not interfere with future `INSERT` statements which don't provide a value for that column. On top, we could also update all existing values to `0` to indicate that the column is meaningless.

nagmat84: > I guess the other option is to leave the column there and null and try to remember to drop it later, when it's more widely available? That might not help with the `down` though. And I'm not sure how many other things that would mess up.

You were faster than me with the same ugly idea. :laughing:

d7415: Yeah, I'd leave it consistent across engines though. Then maybe open an issue against a future milestone (5?) so that we're somewhat less likely to forget.

app/Console/Commands/VariantFilesize.php
nagmat84: Most of the console commands are really ugly, better don't take them for inspiration.

May I ask why you introduced a separate variable `exit_code` if you constantly return `0` in all cases? Besides this minor negligibility, it looks good. I don't mind if the variable remains there, but personally it strikes me as unnecessary.

IMHO, this issue is resolved.

database/migrations/2022_02_02_203008_filesize_size_variants.php
nagmat84: What is the matter of state for this aspect? I assume that @d7415 and I have already agreed to not drop the column now, but leave it untouched and open an issue such that we remember to drop the column with the next big re-factoring (although I don't know if and when this will happen) or after SQLite 3.35 has become more widely spread.

What is the opinion of the other developers, esp. @ildyria? Is this solution acceptable?

I would really like to see this PR being merged soon.

app/Console/Commands/VariantFilesize.php
Chostakovitch: `exit_code` is `-1` if any update fails (_e.g._ [there](https://github.com/LycheeOrg/Lychee/commit/c0c52ae6e44bd36f4f0419945b2713ed367bf48f#diff-f9ded72faf81271d63a161dffa3455545a42249acf10efcd0bab9b5ceb483b7eR55)). 

To do that I needed to "bind" `exit_code` to the closure; this is because I don't want to exit on failure but try to do the maximum.

Is it ok ?

database/migrations/2022_02_02_203008_filesize_size_variants.php
Chostakovitch: > I would really like to see this PR being merged soon.

Right, I'll send a commit tonight or tomorrow!

Chostakovitch: See cc9d88f905ec5a13157a19b6fb875d09f6ab2803.

I don't think it is worth trying to set the column nullable and I just changed the "old" column values so they are all zeros.

I also added a check in the upward migrations to avoid creating the column which already would exist after a cycle of migrate/rollback.

I tested uploading an image before the migration, migrate, launch command, rollback, migrate again (while switching code) and it seems good to me. Hope it is... :p

Chostakovitch: I don't get why tests are failing with bad JSON (as if no size variants are created) while they do work in my environment. What could I do?

app/Console/Commands/VariantFilesize.php
nagmat84: Sure it is. My fault that I did not spot that.

database/migrations/2022_02_02_203008_filesize_size_variants.php
```diff
@@ -107,6 +111,7 @@ public function down()
 
 			$original_filesize = $original_variant->filesize;
 
+			// See comment if the upward migration : column still there
```
nagmat84:
```suggestion
			// See comment in the upward migration: the column is still there
```

nagmat84 (commented):
New changes look good to me with one minor spelling correction.

database/migrations/2022_02_02_203008_filesize_size_variants.php
nagmat84: > I don't think it is worth trying to set the column nullable and I just changed the "old" column values so they are all zeros.

Totally agreed. You must not even try to make the column nullable, because you would run into the same problem you have with dropping a column. Making a non-nullable column nullable in SQLite means to create a new column (with the same definition minus nullability), copying over all values, dropping the old column and renaming the new column to the old one. Hence, setting everything to `0` is just fine.

> I also added a check in the upward migrations to avoid creating the column which already would exist after a cycle of migrate/rollback.

:+1:

> I tested uploading an image before the migration, migrate, launch command, rollback, migrate again (while switching code) and it seems good to me. Hope it is... :p

Seems so.

nagmat84: > I don't get why tests are failing with bad JSON (as if no size variants are created) while they do work in my environment. What could I do?

That is really weird, in particular as the tests succeed with SQLite as well as PostgreSQL and the choice of the DB engine should not change the result of this test at all.

Did you run the test with MySQL locally, too?

I could try to checkout your branch to my machine and see if I find something in the step debugger. But this might take some time.

Chostakovitch: Well, again, my test environment was bad because I only tested SQLite. Now, I know I should test PostgreSQL, MariaDB & SQLite.

Problem was I forgot to remove the temporary `-1` filesize when creating variants. SQLite does not seems to enforce that filesize is signed, but MariaDB does. See 6c241babc7f8d6915cf95c73b2d72236ce4a03ae.

Maybe now I'm ready for a future PR with less bugs. :smile:

nagmat84: I’m glad you found the solution. :tada: :confetti_balls: Congratulations!

I would say this aspect can be regarded as resolved now.

nagmat84:
Looks good to me now. IMHO, we are ready for merge. @ildyria You have been surprisingly silent the last weeks. Do you want to run a final review?

nagmat84 (approved):

ildyria (approved):
LGTM.

ildyria:
Worth to be considered #1242 once this one is merged.

PR #1245, opened 2022-03-24
Belated sync of frontend for PR #1239

nagmat84:
I merged PR #1239 by @chostakovitch some days ago, but did not notice that is it was not in sync with the associated PR to the frontend. So the current master is in an unusable state because the backend code and the frontend are out-of-sync. This PR fixes the problem.

Shame on me that I did not check that back then and did not follow our process, merge frontend first, merge new master of frontend into PR of backend, merge backend. :-(


ildyria (commented):
LGTM.

PR #1252, opened 2022-03-31
Added a console command to fix the tree

nagmat84:
A simple CLI command to check and fix the nested set model of the tree.


kamil4 (approved):
LGTM! (found no issues on my devel instance)

PR #1258, opened 2022-04-09
Fix geoposition types

nagmat84:
The front-end expects the attributes `latitude` and `longitude` to be JS numbers not strings. Otherwise the calculation of the wen frontend for the bounding box of the map view fails (the JS operator `+` erroneously concatenates strings, but does not add numbers)

If the DB engine is MySQL, Laraval converts the SQL type `decimal` to a PHP string by default. This causes the backend to eventually send strings to the frontend. This PR rectifies the issue.

Also see https://github.com/LycheeOrg/Lychee-front/pull/286 for the frontend.


ildyria (approved):
LGTM.

PR #1264, opened 2022-04-10
Center search clear `X`

qwerty287:
Center the `X` in the search box.
Before:
Screenshot_20220410_194850

After:
Screenshot_20220410_194731


nagmat84:
I need to test that first locally. I have never noticed that problem before and I assume that the necessary offset depends on the specific font, browser and operating system.

qwerty287:
I pushed https://github.com/LycheeOrg/Lychee/pull/1264/commits/db856a0942cfe24cbb4375aeed79f8ce0c166b8c which follows [the instructions here](https://www.w3schools.com/howto/howto_css_center-vertical.asp). It uses "relative" values (like `50%`), which is probably better.

ildyria:
The change to app.css will only impact the Livewire (future frontend) part.
To change the dist.css you need to update Lychee-front and recompile the css/js files.

BTW you are more than welcome to join us on https://gitter.im/LycheeOrg/Lobby :)

qwerty287:
> The change to app.css will only impact the Livewire (future frontend) part.
> To change the dist.css you need to update Lychee-front and recompile the css/js files.

Ah, that's how it works. I thought this is a minfied variant of app.css because I searched through all the files for this `header__clear` key and only found these two, will update it.

> BTW you are more than welcome to join us on https://gitter.im/LycheeOrg/Lobby :)

Joined :)

kamil4:
I remember that I tried to fix this at least once before but at some point it "broke" again (either due to seemingly unrelated changes or browser updates). It would be nice to get a correct fix in! :smiley:

Issue #923, opened 2021-02-14
Polish translation


elvisef:
I am working on it, and it's almost done.
But some words made me think I cannot udnderstand it : ]

![2021-02-14_16-36](https://user-images.githubusercontent.com/25976749/107881158-5a818000-6ee3-11eb-845f-af7ef5c9b5a7.png)

Alright, what is a "sensitive" album ?
Maybe I did not read FAQ thoroughly but just in short, please.

Some words or sentences need to be longer, as for my language.
I tried to make it understandable and not so general in meaning some functions, in settings generally.

And if I'm done (I am working with GIT, but I am new to github pushing), do I need to use the master branch or how this should be properly done ?

ildyria:
>Alright, what is a "sensitive" album ?


See here : https://github.com/LycheeOrg/Lychee/pull/822 :laughing: 

> And if I'm done (I am working with GIT, but I am new to github pushing), do I need to use the master branch or how this should be properly done ?

https://lycheeorg.github.io/docs/contributions.html#how-to-properly-submit-a-pull-request-to-lychee

elvisef:
Alright, thanks, piece of cake ; )

One question, shouldn't "Owner" be in the locale files as well ?
![2021-02-14_19-32](https://user-images.githubusercontent.com/25976749/107885595-8c9edc00-6efb-11eb-8748-5c92da7b1370.png)

I could not find it there.

And one more, I could not find "Tags" in locales:
![2021-02-14_20-32](https://user-images.githubusercontent.com/25976749/107887061-cd9aee80-6f03-11eb-956c-d83a69d7f3aa.png)

It's when I wanted to edit tags for an album.

Ooops, one more : ) "About" is missing as well. While translating I think this should be easily changeable.
![2021-02-14_20-34](https://user-images.githubusercontent.com/25976749/107887125-1c488880-6f04-11eb-949c-0a28b5016e2d.png)

I'll try to find it in the code maybe. I will try to finish translating today or tomorrow max.
Will see how it looks :P

Anyway what exactly means "Take Date" or "Latest Take Date" ?
I suppose it could mean a date picture was taken, right ? Or not ?
It is confusing for me, sorry. If You could write what it means in other words, I will be happy.

ildyria:
There might indeed be missing some parts as you noticed for Owner, Tags, and other.

Now to answer your simpler questions:
- "Take Date" is the date the picture is taken (obviously)
[...]

PR #1270, opened 2022-04-11
Fix untranslatable strings

qwerty287:
Add some strings that were hard-coded into the JS files. Also addresses https://github.com/LycheeOrg/Lychee/issues/923#issuecomment-778820674=


ildyria:
LGTM you are just missing the strings in the other languages (it is currently complaining about Chinese)

qwerty287:
OK, so it is necessary to add them (in English) to every other locale? Didn't know that, thanks!

ildyria:
> OK, so it is necessary to add them (in English) to every other locale? Didn't know that, thanks!

yes, the idea being that if you want to translate to another language/improve a translation, you do no need to search for the missing entries, the untranslated ones are directly visible.

qwerty287:
Done.

ildyria (approved):

PR #1217, opened 2022-02-06
Consistent JSON API incl. Proper Handling of Errors (2nd half of big refactoring)

nagmat84:
# Summary

## Custom Exceptions with Optional HTTP Status Code

A lot of custom exception classes have been created. They are organized as follows:

Exceptions

All exceptions realize the interface `LycheeException`. This interface does not declare any methods. It is only used for grouping all Lychee exceptions. Lychee exceptions are categorized into external and internal exceptions via the interfaces `ExternalLycheeException` and `InternalLycheeException`.

External exceptions are meant to be returned to the frontend. To foster this, external Lychee exceptions must also realize
`HttpExceptionInterface` to provide meaningful HTTP status codes to the frontend. In particular, the message text of external Lychee exceptions should

 1. be understandable and provide a meaningful information for an end user as the message text is displayed to the user inside the status bar of the front-end
 2. not disclose any sensitive, security-relevant information

Internal exceptions are used for internal error propagation. While internal exceptions bubble up the call graph they should become wrapped into an external exception at some point.

Ideally, the methods of the HTTP controllers eventually only throw external exceptions. However, if an internal exception slips through the exception handler will take care of that and return a generic `500 Internal Server Error` to the frontend as a last line of defense to avoid unintended leakage of sensitive information.

## Revised Exception Handler

The exception handler `App\Exceptions\Handler` has heavily been revised.

### Correct content type

The old method `render` is gone. This method was bad, because it did not honor the content type which was expected by the client but unconditionally returned a HTML representation of the exception even if the frontend requested JSON.

The method has been replaced by the two distinct methods `renderHttpException` and `convertExceptionToArray`.
The former is called by the Laravel framework for HTML responses, the latter for JSON responses. Note, that `renderHttpException` is actually a misnomer, it should be named `renderHTMLException` (because JSON also uses HTTP as a transport layer), but heh, this is Laravel. :cry:

### Logging

Exceptions are now properly logged to our own home-brewed logging mechanism. The method `report` of the Laravel default exception handler has been overridden. The Larvavel default method tries to use a service which realizes the `\Psr\Log\LoggerInterface` which is not implemented by our class `Logs` and hence it has not worked until now.

This gives us two benefits:

 - We can avoid the "double" error handling in our code, i.e. we do not need to throw an exception and separately create a log entry with the same error message. Creating and throwing the exception now suffices.
[...]

PR #1272, opened 2022-04-12
Added shortcuts via `RewriteRule` to .htaccess

nagmat84:
While I was looking for a solution to https://github.com/LycheeOrg/Lychee/pull/1217#discussion_r836552854, I discovered that our `.htaccess` could be improved.

As there is a lot (I mean really A LOT) requests for the `upload` directory, we can take a shortcut there. The new rules also avoid some unnecessary exceptions and logging, if a browser requests a non-existing file.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -14,6 +14,15 @@
     RewriteCond %{REQUEST_URI} (.+)/$
     RewriteRule ^ %1 [L,R=301]
 
+	# Any request to the public folders can be served directly.
+	# There is no need to direct them to the front-controller.
+	# Hence, we take a short-cut here, [L] means last rule.
+	# This also avoids some unneccessary execptions and exception logging
+	# inside Lychee, if a non-existing file is requested.
+	# Also disable compression for already highly compressed media files.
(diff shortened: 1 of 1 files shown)


nagmat84:
While I was working on the `.htaccess` file, I also wondered if these lines https://github.com/LycheeOrg/Lychee/blob/4d1b5f0df078bd6160f7815f61d1e9f0447ef6b6/public/.htaccess#L12-L15 are actually correct. Here, `%1` in the `RewriteRule` captures everything of the URL but the trailing `/`, then the `RewriteRule` replaces the beginning of the URL `^` by `%1`. The value of `%1` is put it front of the URL which is already there, i.e. if `my-url/` is not an existing directory, then the browser is redirected to the URL `my-urlmy-url/`. I don't think this is intended, and the rule should be `RewriteRule .* %1 [L,R=301]`.

Why do we need the rules at all? What purpose do they serve? @ildyria added them to `.htaccess`.

ildyria:
I never touched that part of the file.
C.f. original here: https://github.com/laravel/laravel/blob/9.x/public/.htaccess

ildyria (approved):
Tested. LGTM.

PR #1271, opened 2022-04-11
Localize errors displayed in UI

qwerty287:
Followup of https://github.com/LycheeOrg/Lychee/pull/1270: also localize errors. This is dependent from https://github.com/LycheeOrg/Lychee-front/pull/289 to work, please merge this PR as well.

PR #1274, opened 2022-04-13
Add album tracks

qwerty287:
Adds a new feature that allows to upload GPX tracks/files and link them to an album (only regular, no smart/tag albums). This track is displayed together with the photos on the map, means this is only usable if the map is enabled.
To display the track, the Leaflet plugin https://github.com/mpetazzoni/leaflet-gpx is used, it is very easy and only needs one file.

An example usecase for this is, if you have an album of photos of a hike, you can display them on a map and add a track/GPX if you recorded one that is displayed on the map, this makes it easier to locate them. I took a screenshot of a simple example: 
Screenshot 2022-04-13 at 14-01-58 Lychee - 2309

This is the first time I'm contributing something else than just localization and simple CSS updates, I hope most things are in a usable state.


nagmat84:
I really love this feature. 😍😍

But two aspects:

 - We are in the middle of a big re-factoring which which changes a lot, especially the way what the controller do, how requests are look like and how error are handled. I would kindly ask you to base your PR on the branch `consistent_json_api`

 - I know that the previous code style of Lychee was not optimal. In particular, many parameters were not typed, anonymous arrays were used as dictionaries to pass arguments around and parameters were not annotated. So I know this is not your fault, you only adopted the code style which has already been there. I tried to change most of that in the branch `consistent_json_api`. If you base your work on that branch, I would also like to ask you to adopt to the new style. For example, I saw that one of your methods takes a `mixed` parameter. Mixed types shouldn't be necessary 99% of the time.

qwerty287:
I'll try to refactor it so that it works with `consistent_json_api` - didn't look at it yet, but this can take some time :grinning:

qwerty287:
@nagmat84 done - was easier as expected. :wink: (please note that this contains 2 commits from `master` that are not yet in `consistent_json_api`) The only missing thing might be the code style.

app/Actions/Album/Tracks.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace App\Actions\Album;
+
+use App\Models\Album;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+use Illuminate\Http\UploadedFile;
+use Illuminate\Support\Facades\Storage;
+
+class Tracks extends Action
```
nagmat84: I am not sure if we really need this action anymore. The new branch removed most of the actions, because they passed IDs in form of strings around. (See comments below.)

app/Actions/Album/Tracks.php
```diff
@@ -0,0 +1,53 @@
...
+{
+	/**
+	 * @param string $albumID the ID of the album
+	 * @param mixed  $value   the value to be set
+	 */
+	public function set(string $albumID, UploadedFile $value): void
+	{
+		$album = $this->albumFactory->findAlbumOrFail($albumID, false);
```
nagmat84: Please don't pass around the ID of the album as string. The new approach is to fetch the required album as part of the request, so we can properly fail with an HTTP error 422, 404, 401 or 403 in the correct order (see classes below `\App\Http\Requests`). This way the instance of the album model becomes part of the request and we only need to query the DB once.

app/Actions/Album/Tracks.php
```diff
@@ -0,0 +1,53 @@
...
+		if ($album->track_id != null) {
+			Storage::delete("tracks/$album->track_id.xml");
```
nagmat84: Please store the whole relative path in the `album` and not only the id. This would a) be in line with how the size variants work, and b) it will allow to use other paths in the future if the `tracks/` part is not hard-coded, but part of the DB field.

The attribute is called `short_path` in the model `SizeVariant`, because this is the terminology used by Flysystem.

Moreover, I personally, don't like the PHP syntax which allows you to put variables into strings. Here it is somewhat cleat that the variable is `track_id` because the dot `.` works as a visual separator, but personally I would prefer something like `'tracks/' . $album->track_id . '.xml'`. It is more verbose, but IMHO is clearer.

But anyway it should rather be

```suggestion
	public function set(string $albumID, UploadedFile $value): void
	{
		$album = $this->albumFactory->findAlbumOrFail($albumID, false);
		if ($album->track_short_path != null) {
			Storage::delete($album->track_short_path);
```

app/Actions/Album/Tracks.php
```diff
@@ -0,0 +1,53 @@
...
+		}
+
+		$new_track_id = uniqid();
```
nagmat84: `uniqid` is time-based and hence not really secure. If you have uploaded an album at a certain point of time, an attacker might guess the ID of the track and download it subsequently, because an attacker only needs to scan a small range of values. This could infringe privacy.

We just got rid of time-based IDs and use 24 alpha-numeric characters for such things. Also a (cryptographically secure hash) over the track would be fine.

app/Actions/Album/Tracks.php
```diff
@@ -0,0 +1,53 @@
...
+use Illuminate\Http\UploadedFile;
+use Illuminate\Support\Facades\Storage;
+
+class Tracks extends Action
+{
+	/**
+	 * @param string $albumID the ID of the album
+	 * @param mixed  $value   the value to be set
+	 */
+	public function set(string $albumID, UploadedFile $value): void
```
nagmat84: Please don't call the uploaded file `$value`.

app/Actions/Album/Tracks.php
```diff
@@ -0,0 +1,53 @@
...
+		if ($album->track_id != null) {
+			Storage::delete("tracks/$album->track_id.xml");
+		}
+
+		$new_track_id = uniqid();
+		Storage::putFileAs('tracks/', $value, "$new_track_id.xml");
+
+		if (Album::query()
+				->where('id', '=', $albumID)
+				->update(['track_id' => $new_track_id]) !== 1
```
nagmat84: Please don't do that.

`UPDATE` queries are evil, because the don't update the `updated_at` attribute of the album properly. However, this will be necessary, when we will have other clients than the web frontend, because other clients might use caching of objects and need a proper `created_at` and `updated_at` value to perform a 3-way merge during syncronization.

Moreover, you already have the album object above, so you can use it. Do something like

```php
$album->track_short_path = $new_track_path;
$album->save()
```

app/Actions/Album/Tracks.php
```diff
@@ -0,0 +1,53 @@
...
+		) {
+			throw new ModelNotFoundException();
```
nagmat84: If you use `$album->save` as recommended as above, explicitly throwing an exception won't be necessary, because `$album->save` already do so.

Moreover, just as a pointer: If you ever need to throw an exception, try to throw one of the Lychee specific exceptions (see `\App\Exceptions\` and `\App\Exceptions\Internal`). If you don't find any which seems to fit, then create a new one.

app/Actions/Album/Tracks.php
```diff
@@ -0,0 +1,53 @@
...
+		}
+	}
+
+	/**
+	 * Delete the track of an album.
+	 *
+	 * @param string $albumID the ID of the album
+	 */
+	public function delete(string $albumID): void
```
nagmat84: I skip comments on this method. Basically, everything from above applies. ;-) And as I said, I would like to get rid of the entire action anyway.

app/Factories/AlbumFactory.php
```diff
@@ -84,6 +84,34 @@ public function findBaseAlbumOrFail(string $albumId, bool $withRelations = true)
...
+	 * @param string $albumId       the ID of the requested album
+	 * @param bool   $withRelations indicates if the relations of an
+	 *                              album (i.e. photos and sub-albums,
+	 *                              if applicable) shall be loaded, too.
+	 *
+	 * @return Album the album for the ID
+	 *
+	 * @noinspection PhpIncompatibleReturnTypeInspection
+	 */
+	public function findAlbumOrFail(string $albumId, bool $withRelations = true): Album
```
nagmat84: Why has this method been added?

app/Http/Controllers/AlbumController.php
```diff
@@ -208,6 +211,32 @@ public function setLicense(SetAlbumLicenseRequest $request): void
...
+	 * Upload a track for the Album.
+	 *
+	 * @param SetAlbumTrackRequest $request
+	 * @param Tracks               $tracks
+	 *
+	 * @return void
+	 */
+	public function setTrack(SetAlbumTrackRequest $request, Tracks $tracks): void
+	{
+		$tracks->set($request->album()->id, $request->file);
```
nagmat84: I just realize that you already have implemented a specific request class :+1: :tada: 

But if you already have the album as part of the request, it is not necessary to get the ID of the album, just to fetch the same album from the DB inside the action again. This is an unnessary DB roundtrip.

You can simply do something like that

```php
$request->album()->track_short path = ...
```

app/Http/Controllers/AlbumController.php
```diff
@@ -208,6 +211,32 @@ public function setLicense(SetAlbumLicenseRequest $request): void
...
+	 * Upload a track for the Album.
+	 *
+	 * @param DeleteTrackRequest $request
+	 * @param Tracks             $tracks
+	 *
+	 * @return void
+	 */
+	public function deleteTrack(DeleteTrackRequest $request, Tracks $tracks): void
+	{
+		$tracks->delete($request->album()->id);
```
nagmat84: Same here. If you already have the album, don't use the ID to fetch the album again.

app/Http/Requests/Album/DeleteTrackRequest.php
```diff
@@ -0,0 +1,45 @@
...
+		return $this->authorizeAlbumWrite($this->album);
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function rules(): array
+	{
+		return [
+			HasAbstractAlbum::ALBUM_ID_ATTRIBUTE => ['present', new AlbumIDRule(false)],
```
nagmat84:
```suggestion
			HasAbstractAlbum::ALBUM_ID_ATTRIBUTE => ['required', new AlbumIDRule(false)],
```

app/Http/Requests/Album/DeleteTrackRequest.php
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	protected function processValidatedValues(array $values, array $files): void
+	{
+		$this->album = $this->albumFactory->findAlbumOrFail(
+			$values[HasAbstractAlbum::ALBUM_ID_ATTRIBUTE], false
+		);
```
nagmat84:
```suggestion
		$this->album = Album::query()->findOrFail($values[HasAbstractAlbum::ALBUM_ID_ATTRIBUTE]);
```

suffices. No need for the added factory method. The only reason why I had to added a factory method for `BaseAlbum` (the one which you copied) was that a `BaseAlbum` can either be an `Album` or a `TagAlbum`. But as tracks are only supported for real albums, a simply Eloquent query suffices.

app/Http/Requests/Album/SetAlbumTrackRequest.php
```diff
@@ -0,0 +1,52 @@
...
+		return $this->authorizeAlbumWrite($this->album);
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function rules(): array
+	{
+		return [
+			HasAbstractAlbum::ALBUM_ID_ATTRIBUTE => ['present', new AlbumIDRule(false)],
```
nagmat84:
```suggestion
			HasAbstractAlbum::ALBUM_ID_ATTRIBUTE => ['required', new AlbumIDRule(false)],
```

app/Http/Requests/Album/SetAlbumTrackRequest.php
```diff
@@ -0,0 +1,52 @@
...
+		];
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	protected function processValidatedValues(array $values, array $files): void
+	{
+		$albumID = $values[HasAbstractAlbum::ALBUM_ID_ATTRIBUTE];
+		$this->album = $this->albumFactory->findAlbumOrFail($albumID);
```
nagmat84: See above.

app/Locale/German.php
```diff
@@ -111,6 +124,8 @@ public function get_locale(): array
 			'IMPORT_SERVER' => 'Von Server importieren',
 			'NEW_ALBUM' => 'Neues Album',
 			'NEW_TAG_ALBUM' => 'Neues Tag-Album',
+			'UPLOAD_TRACK' => 'Upload track',
+			'DELETE_TRACK' => 'Delete track',
```
nagmat84:
```suggestion
			'UPLOAD_TRACK' => 'Track hochladen',
			'DELETE_TRACK' => 'Track löschen',
```

I am German ;-)

app/Models/Album.php
```diff
@@ -27,6 +28,7 @@
  * @property string            $license
  * @property string|null       $cover_id
  * @property Photo|null        $cover
+ * @property string|null       $track_id
```
nagmat84:
```suggestion
 * @property string|null       $track_short_path
```

app/Models/Album.php
```diff
@@ -183,6 +185,9 @@ public function delete(): bool
 		try {
 			$this->refreshNode();
 
+			// delete track
+			(new Tracks())->delete($this->id);
```
nagmat84: This does only delete the tracks of the current album, but does not delete the tracks of sub-albums.

app/Models/Extensions/BaseAlbum.php
```diff
@@ -50,6 +50,8 @@ abstract class BaseAlbum extends Model implements AbstractAlbum, HasRandomID
 	 */
 	protected $keyType = HasRandomID::ID_TYPE;
 
+	protected ?string $trackId = null;
```
nagmat84: You said, that only `Album` support tracks. So this attribute should not exist on `BaseAlbum`.

Also not that none of the Eloquent DB attributes exist as proper PHP class attributes (something I really hate about Laravel). They are added through PHP magic methods.

app/SmartAlbums/BaseSmartAlbum.php
```diff
@@ -43,6 +43,7 @@ abstract class BaseSmartAlbum implements AbstractAlbum
 	protected ?Thumb $thumb;
 	protected Collection $photos;
 	protected \Closure $smartPhotoCondition;
+	protected ?string $trackId = null;
```
nagmat84: Same here. Why do you add it?

database/migrations/2022_04_13_094611_add_track_id_to_album_table.php
```diff
@@ -0,0 +1,32 @@
...
+{
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		Schema::table('albums', function (Blueprint $table) {
+			$table->char('track_id', 20)->after('cover_id')->nullable();
```
nagmat84:
```suggestion
			$table->string('track_short_path')->after('cover_id')->nullable()->default(null);
```

database/migrations/2022_04_13_094611_add_track_id_to_album_table.php
```diff
@@ -0,0 +1,32 @@
...
+
+	/**
+	 * Reverse the migrations.
+	 *
+	 * @return void
+	 */
+	public function down()
+	{
+		Schema::table('albums', function (Blueprint $table) {
+			$table->dropColumn('track_id');
```
nagmat84:
```suggestion
			$table->dropColumn('track_short_path');
```

nagmat84 (changes requested):
Sorry for the myriad of comments. I hope they are not too harsh and you are disappointed. I really love this feature.

app/Actions/Album/Tracks.php
ildyria: > `UPDATE` queries are evil, because the don't update the `updated_at`

On the other hand, such queries do not require the hydration of a model. :) 
I would be fine with update query, if they also include the `update_at` update (but that is nittpicking). :p

app/Locale/French.php
```diff
@@ -209,6 +225,7 @@ public function get_locale(): array
 			'PHOTO_NEW_LICENSE' => 'Ajouter une License',
 			'PHOTO_SET_LICENSE' => 'Sélectionner License',
 			'PHOTO_LICENSE' => 'License',
+			'PHOTO_LICENSE_HELP' => 'Need help choosing?',
```
ildyria:
```suggestion
			'PHOTO_LICENSE_HELP' => 'Besoin d\'aide pour choisir ?',
```

app/Locale/French.php
```diff
@@ -295,6 +312,7 @@ public function get_locale(): array
 			'SETTINGS_SUCCESS_MAP_DISPLAY' => 'Parametres de la carte mis à jour.',
 			'SETTINGS_SUCCESS_MAP_DISPLAY_PUBLIC' => 'Parametres de la carte pour les albums publics mis à jour.',
 			'SETTINGS_SUCCESS_MAP_PROVIDER' => 'Fournisseur de la Carte mis à jour.',
+			'SETTINGS_DROPBOX_KEY' => 'Dropbox API Key',
```
ildyria:
```suggestion
			'SETTINGS_DROPBOX_KEY' => 'Clé de l\'API Dropbox',
```

app/Locale/French.php
```diff
@@ -303,6 +321,7 @@ public function get_locale(): array
 			'U2F_AUTHENTIFICATION_SUCCESS' => 'Authentication réussie!',
 			'U2F_CREDENTIALS' => 'Clés',
 			'U2F_CREDENTIALS_DELETED' => 'Clé supprimée!',
+			'U2F_NO_CREDENTIALS' => 'Credentials list is empty!',
```
ildyria:
```suggestion
			'U2F_NO_CREDENTIALS' => 'La liste de clé est vide !',
```

app/Locale/French.php
```diff
@@ -465,11 +484,15 @@ public function get_locale(): array
 			'PHOTO_SMALL_HIDPI' => 'Petite taille HiDPI',
 			'PHOTO_THUMB' => 'Mignature carrée',
 			'PHOTO_THUMB_HIDPI' => 'Mignature carrée HiDPI',
+			'PHOTO_THUMBNAIL' => 'Photo thumbnail',
```
ildyria:
```suggestion
			'PHOTO_THUMBNAIL' => 'Mignature photo',
```

app/Locale/French.php
```diff
@@ -465,11 +484,15 @@ public function get_locale(): array
...
 			'PHOTO_LIVE_VIDEO' => 'Partie vidéo d’une live-photo',
 			'PHOTO_VIEW' => 'Vue photo de Lychee :',
+			'VIDEO_NOT_SUPPORTED' => 'Your browser does not support the video tag.',
```
ildyria:
```suggestion
			'VIDEO_NOT_SUPPORTED' => 'Votre navigateur ne supporte pas la balise vidéo.',
```

app/Locale/French.php
```diff
@@ -465,11 +484,15 @@ public function get_locale(): array
...
 
 			'PHOTO_EDIT_ROTATECWISE' => 'Pivoter dans le sens des aiguilles d’une montre.',
 			'PHOTO_EDIT_ROTATECCWISE' => 'Pivoter dans le sens contraire des aiguilles d’une montre.',
+
+			'ERROR_GPX' => 'Error loading GPX file: ',
```
ildyria:
```suggestion
			'ERROR_GPX' => 'Erreur lors du chargement du fichier GPX : ',
```

app/Locale/French.php
```diff
@@ -111,6 +125,8 @@ public function get_locale(): array
 			'IMPORT_SERVER' => 'Importer depuis le serveur',
 			'NEW_ALBUM' => 'Nouvel album',
 			'NEW_TAG_ALBUM' => 'Nouvel album d’étiquette',
+			'UPLOAD_TRACK' => 'Upload track',
```
ildyria:
```suggestion
			'UPLOAD_TRACK' => 'Ajouter une trace',
```

app/Locale/French.php
```diff
@@ -111,6 +125,8 @@ public function get_locale(): array
...
+			'DELETE_TRACK' => 'Delete track',
```
ildyria:
```suggestion
			'DELETE_TRACK' => 'Supprimer la trace',
```

app/Locale/French.php
```diff
@@ -102,6 +115,7 @@ public function get_locale(): array
 			'COPY_TO' => 'Copier vers…',
 			'COPY_ALL_TO' => 'Copier la sélection vers…',
 			'DELETE' => 'Supprimer',
+			'SAVE' => 'Save',
```
ildyria:
```suggestion
			'SAVE' => 'Sauvegarder',
```

app/Locale/French.php
```diff
@@ -26,25 +26,38 @@ public function get_locale(): array
...
+			'CLEAN_LOGS' => 'Clean Noise',
 			'SIGN_OUT' => 'Se déconnecter',
 			'UPDATE_AVAILABLE' => 'Une mise à jour est disponible !',
 			'MIGRATION_AVAILABLE' => 'Une migration disponible !',
 			'DEFAULT_LICENSE' => 'License par defaut pour les nouveaux ajouts:',
 			'SET_LICENSE' => 'Selectioner une license',
 			'SET_OVERLAY_TYPE' => 'Selectioner le type d’Overlay',
 			'SET_MAP_PROVIDER' => 'Selectioner le fournisseur de données cartographiques',
+			'FULL_SETTINGS' => 'Full Settings',
+			'UPDATE' => 'Update',
```
ildyria:
```suggestion
			'UPDATE' => 'Mettre-a-jour',
```

app/Locale/French.php
```diff
@@ -26,25 +26,38 @@ public function get_locale(): array
...
+			'SHARING_UNDO' => 'undo',
+			'SHARING_REDO' => 'redo',
 			'CHANGE_LOGIN' => 'Changer le nom d’utilisateur',
 			'CHANGE_SORTING' => 'Changer le tri',
 			'SET_DROPBOX' => 'Paramétrer Dropbox',
 			'ABOUT_LYCHEE' => 'À propos de Lychee',
 			'DIAGNOSTICS' => 'Diagnostiques',
 			'DIAGNOSTICS_GET_SIZE' => 'Calculer l’espace utilisé',
 			'LOGS' => 'Afficher les logs',
+			'CLEAN_LOGS' => 'Clean Noise',
```
ildyria:
```suggestion
			'CLEAN_LOGS' => 'Nettoyer le bruit.',
```

app/Locale/French.php
```diff
@@ -26,25 +26,38 @@ public function get_locale(): array
...
 			'SIGN_OUT' => 'Se déconnecter',
 			'UPDATE_AVAILABLE' => 'Une mise à jour est disponible !',
 			'MIGRATION_AVAILABLE' => 'Une migration disponible !',
 			'DEFAULT_LICENSE' => 'License par defaut pour les nouveaux ajouts:',
 			'SET_LICENSE' => 'Selectioner une license',
 			'SET_OVERLAY_TYPE' => 'Selectioner le type d’Overlay',
 			'SET_MAP_PROVIDER' => 'Selectioner le fournisseur de données cartographiques',
+			'FULL_SETTINGS' => 'Full Settings',
```
ildyria:
```suggestion
			'FULL_SETTINGS' => 'Tous les paramètres',
```

app/Locale/French.php
```diff
@@ -26,25 +26,38 @@ public function get_locale(): array
 			'DEFAULT' => 'Valeur par defaut',
 			'ALBUM_SET_ORDER' => 'Changer l\'ordre',
 			'ALBUM_ORDERING' => 'Trier par',
+			'ALBUM_OWNER' => 'Owner',
```
ildyria:
```suggestion
			'ALBUM_OWNER' => 'Propriétaire',
```

app/Locale/French.php
```diff
@@ -26,25 +26,38 @@ public function get_locale(): array
...
 
 			'USERS' => 'Utilisateurs',
+			'ALLOW_UPLOAD' => 'Allow uploads',
+			'RESTRICTED_ACCOUT' => 'Restricted account',
```
ildyria:
```suggestion
			'RESTRICTED_ACCOUT' => 'Compte limité',
```

app/Locale/French.php
```diff
@@ -26,25 +26,38 @@ public function get_locale(): array
 			'DEFAULT' => 'Valeur par defaut',
 			'ALBUM_SET_ORDER' => 'Changer l\'ordre',
 			'ALBUM_ORDERING' => 'Trier par',
+			'ALBUM_OWNER' => 'Owner',
 
 			'USERS' => 'Utilisateurs',
+			'ALLOW_UPLOAD' => 'Allow uploads',
```
ildyria, line 32:
```suggestion
			'ALLOW_UPLOAD' => 'Autoriser les ajouts de photos',
```

app/Locale/French.php
```diff
@@ -26,25 +26,38 @@ public function get_locale(): array
...
+			'RESTRICTED_ACCOUT' => 'Restricted account',
+			'CREATE' => 'Create',
```
ildyria:
```suggestion
			'CREATE' => 'Créer',
```

app/Locale/French.php
```diff
@@ -26,25 +26,38 @@ public function get_locale(): array
...
+			'USERS_EMPTY' => 'User list is empty!',
```
ildyria:
```suggestion
			'USERS_EMPTY' => 'La liste d\'utilisateurs est vide !',
```

app/Locale/French.php
```diff
@@ -26,25 +26,38 @@ public function get_locale(): array
...
 			'U2F' => 'U2F',
 			'NOTIFICATIONS' => 'Notifications',
 			'SHARING' => 'Partage',
+			'SHARING_EMPTY' => 'Sharing list is empty!',
```
ildyria:
```suggestion
			'SHARING_EMPTY' => 'La liste des partages est vide !',
```

app/Locale/French.php
```diff
@@ -26,25 +26,38 @@ public function get_locale(): array
...
+			'SHARING_WITH' => 'with',
```
ildyria:
```suggestion
			'SHARING_WITH' => 'avec',
```

app/Locale/French.php
```diff
@@ -26,25 +26,38 @@ public function get_locale(): array
...
+			'SHARING_SHARE' => 'Share',
```
ildyria:
```suggestion
			'SHARING_SHARE' => 'Partager',
```

app/Locale/French.php
```diff
@@ -26,25 +26,38 @@ public function get_locale(): array
...
+			'SHARING_UNDO' => 'undo',
```
ildyria:
```suggestion
			'SHARING_UNDO' => 'défaire',
```

app/Locale/French.php
```diff
@@ -26,25 +26,38 @@ public function get_locale(): array
...
+			'SHARING_REDO' => 'redo',
```
ildyria:
```suggestion
			'SHARING_REDO' => 'refaire',
```

app/Models/Extensions/BaseAlbum.php
qwerty287: This was necessary to prevent exceptions that this field (`track_id`) does norexist if requesting it. It will always be `null`. I'll try to look for another way.

app/Locale/German.php
qwerty287: Me too 🙂, but I wasn't sure if I should do it in this PR. I'll translate them in my future  PRs directly.

app/Locale/French.php
qwerty287: Thisis actually from a commit in master that is not yet in `consisten_json_api`. Should I apply this anyway?

app/Locale/French.php
qwerty287: See above.

app/Locale/French.php
qwerty287: The strings I didn't apply yet are not added in this PR, they are already in `master`. Should I add them anyway?

app/Factories/AlbumFactory.php
qwerty287: I need a method which only returns "real" albums - no tag albums. `findBaseAlbumOrFail` also returns tag albums.

nagmat84: Well, but you don't need to fetch the album at all a second time. You can simply use the album object which is part of the request. See my comments on the request class below. So this whole method is not needed.

app/Actions/Album/Tracks.php
qwerty287: To which other class the methods should belong then? Directly call their contents in `AlbumController` or add them to `Album` model or to the request classes?

qwerty287: I [removed it](https://github.com/LycheeOrg/Lychee/pull/1274/commits/9beebb3b0c408f4feb0b314667e7e293a705f2ec) and added both methods to the `Album` model. Please review this and action-related stuff again.

app/Locale/French.php
ildyria: There are multiple point of view on this.
- One could say it needs to be in a separate PR.
- One could aslo say, given that we are modifying those lines, let's also make sure they are correct. (I am of the later).

@nagmat84 opinion ?

app/Models/Album.php
```diff
@@ -268,4 +274,38 @@ public function newEloquentBuilder($query): AlbumBuilder
...
+	 *
+	 * @return void
+	 */
+	public function setTrack(UploadedFile $file): void
+	{
+		if ($this->track_short_path != null) {
+			Storage::delete($this->track_short_path);
+		}
+
+		$new_track_id = sha1($file);
```
ildyria:
```suggestion
		$new_track_id = hash('sha256',$file->get());
```
I am not entirely sure of this but this should avoid using sha1 and this should make sure we are hashing the content of the file.

qwerty287: Fixed (but my commit message was wrong, should be SHA256)

app/Locale/French.php
qwerty287: Actually, we do not modify the lines. In general, I agree with the later option (and I applied the changes you suggested that really belong to this PR), but the other lines were modified in https://github.com/LycheeOrg/Lychee/commit/4d1b5f0df078bd6160f7815f61d1e9f0447ef6b6, the `consistent_json_api` branch does not contain them and is not up-to-date - if you update this, these lines won't be modified by this PR anymore.

app/Models/Album.php
qwerty287: How are they deleted? It seems that this `delete` method is never called for subalbums, do they use another method or do I have to fetch all the subalbums and run `deleteTrack` separately on them?

app/Models/Album.php
```diff
@@ -268,4 +274,38 @@ public function newEloquentBuilder($query): AlbumBuilder
...
+	/**
+	 * Set the GPX track for the album.
+	 *
+	 * @param UploadedFile $file the GPX track file to be set
+	 *
+	 * @return void
+	 */
+	public function setTrack(UploadedFile $file): void
+	{
+		if ($this->track_short_path != null) {
```
kamil4:
```suggestion
		if ($this->track_short_path !== null) {
```
We're trying to use stronger `!==`/`===` wherever possible.

app/Models/Album.php
kamil4: What if somebody uploads the same track file to two different albums? That's not even that unlikely, to be honest. We need to make sure that we handle it well...

app/Models/Album.php
```diff
@@ -268,4 +274,38 @@ public function newEloquentBuilder($query): AlbumBuilder
...
+	}
+
+	/**
+	 * Delete the track of the album.
+	 *
+	 * @return void
+	 */
+	public function deleteTrack(): void
+	{
+		if ($this->track_short_path == null) {
```
kamil4:
```suggestion
		if ($this->track_short_path === null) {
```

kamil4 (approved):

kamil4:
I looked through @nagmat84's initial comments and based on my reading of the (nearly) current version I must say that this PR has gone through quite a transformation! Good job! :smiley:

There's one question I piggybacked on one of @ildyria's comments that may have been lost in the noise so let me ask it here again:

_What if somebody uploads the same track file to two different albums? That's not even that unlikely, to be honest. We need to make sure that we handle it well..._

qwerty287:
> _What if somebody uploads the same track file to two different albums? That's not even that unlikely, to be honest. We need to make sure that we handle it well..._

Tested this and fixed the issue that it is deleted (on the disk) if it's used on other albums. The track will be stored only once (same SHA -> same filename).

app/Models/Album.php
nagmat84: > or do I have to fetch all the subalbums and run deleteTrack separately on them?

Basically this it is. Unfortunately, it is completely inefficient due to Laravels insance architecture. We have the same problem when we need to delete the photos and symbolic links, when an album is deleted. That is why we fetch all photos  via `->all_photos` some lines above.

However, I have another suggestion. The existing code (not your,s our code some lines below) is architecturally correct, but causes timeouts when you delete large sub-trees with many photos in it. (Because it hydrates all the photos within.) I already planned to fix that. I will fix that part and then you put your code to delete tracks on top of it.

Just give me some days to do it.

app/Locale/French.php
nagmat84: @ildyria: Please create another PR for it. I already have been wanting to ask for that, but I did not get to a keyboard fast enough. I am fine with the french translations for `UPLOAD_TRACK` and `DELETE_TRACK` because these strings are added by this PR.

But everything else should have been fixed in the scope of https://github.com/LycheeOrg/Lychee/pull/1270 (which has already been merged) and https://github.com/LycheeOrg/Lychee/pull/1271. As soon as the latter PR and the PR for `consistent_json_api` are merged into master and this PR is re-based these lines vanish anyway.

> One could aslo say, given that we are modifying those lines, let's also make sure they are correct. (I am of the later).

This argument is wrong. We are **not** modifying these lines here.

These lines only show up in this PR as a diff, because this PR is based on `consistent_json_api` and `consistent_json_api` is two or three commits behind current `master`. If the current PRs to `master` are also merged into `consistent_json_api` such that `consistent_json_api` is on par with `master` again, then these lines vanish.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: No, please not. This should be in another PR. All these should be added to https://github.com/LycheeOrg/Lychee/pull/1271.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Locale/French.php
nagmat84: Another PR.

app/Actions/Album/PositionData.php
```diff
@@ -45,6 +45,9 @@ public function get(AbstractAlbum $album, bool $includeSubAlbums = false): array
 			->whereNotNull('longitude')
 			->get()
 			->toArray();
+		if ($album instanceof Album) {
+			$result['track_short_path'] = $album->track_short_path;
+		}
```
nagmat84: Much better :+1:

```suggestion
		$result['track_url'] = $album instanceof Album ? $album->track_url : null;
```

But I have to excuse, that I didn't told you the whole truth earlier. My previous explanation was incomplete. 

In the DB we store the _short path_ which is relative to a certain disk of the Flysystem File adapter. But we need to send a _URL_ to the browser. The Flysystem storage provides the method `url(string $shortPath)` for that. That will become crucial, when some day in the future the disk isn't the same (local) disk on which Lychee is installed, but for example a S3 bucket.

The whole aspect should be in the same spirit as already done for `SizeVariant` or `Symlink`. For example the model `SizeVariant` provides to attributes: `short_path` and `url` (Please ignore the `full_path`. This is evil and will break with other than local storage. It is only there for legacy reasons, because our image manipulation code is based on absolute paths and does not work with proper file streams.)

https://github.com/LycheeOrg/Lychee/blob/019921260b4d8e1bac0e61ec99fd4d8782c68710/app/Models/SizeVariant.php#L40-L48

The `Album` model should have two attributes as well `track_short_path` and `track_url`.

The attribute `track_url` shall be a read-only attribute. To enforce that, the `Album` model shall use the `MustNotSet` cast. As a template see again `SizeVariant`:

https://github.com/LycheeOrg/Lychee/blob/019921260b4d8e1bac0e61ec99fd4d8782c68710/app/Models/SizeVariant.php#L70-L78

Moreover, the `Album` model needs a getter which provides the attribute `track_url` based on `track_short_path`. The role model is

https://github.com/LycheeOrg/Lychee/blob/019921260b4d8e1bac0e61ec99fd4d8782c68710/app/Models/SizeVariant.php#L123-L136

Please forget all the complexity of the method above. In your case the additional getter method on the `Album` model will be rather simple:

```php
	/**
	 * Accessor for the "virtual" attribute {@link Album::$track_url}.
	 *
	 * This is a convenient method which wraps
	 * {@link Album::$track_short_path} into
	 * {@link \Illuminate\Support\Facades\Storage::url()}.
	 *
	 * @return string the url of the track
	 */
	public function getTrackUrlAttribute(): string
	{
		return Storage::url($this->track_short_path);
	}
```

A final word about the ternary operator. The returned JSON object should always be identical. (This is necessary for statically, strongly typed clients.) Hence, if we have a tag album, the attribute `track_url` should still be provided, but be `null`.

nagmat84 (changes requested):
This made a great progress :+1:

I only have a single (but quite) long change request. Sorry, I didn't tell you earlier.

nagmat84:
I merged the current head of master into `consistent_json_api` on which this branch is based. @qwerty287 so you can proceed with merging  `consistent_json_api` into this branch. But *only after* @ildyria deleted his suggestions for French translations which are links to code lines which will be removed from the diff due to the merge.

app/Actions/Album/PositionData.php
qwerty287: I don't work with PHP usually, so what would be the "correct" way to add the `track_url` attribute with value `null` to tag albums (and smart albums, I guess, should have this as well)?

nagmat84: See the last line of my suggestion above. Here again:

```php
$result['track_url'] = $album instanceof Album ? $album->track_url : null;
```

qwerty287: This is only used in `PosutionData` - what happens if directly requesting Albums (`Abum::get` and `Albuns::get`)?

nagmat84: Ah, I see. Then we had a misunderstanding here. I didn't want to say that the attribute must be everywhere, I only wanted to say that an object of  the same type (here: `PositionData`) should always have the same attributes. So it is fine if `TagAlbum` and `SmartAlbum` don't return the attribute.

However, speaking about the serialization of `Album`, I noted another aspect which needs attention. When converted to JSON a `Album` object should include the "virtual" attribute `track_url`, but skip the attribute `track_short_path`. The `Album` class contains two arrays for that: `append` and `hidden`. The name of the attributes must be added to these arrays accordingly.

ildyria:
>  But only after @ildyria deleted his suggestions for French translations which are links to code lines which will be removed from the diff due to the merge.

There is no need to delete the suggestions because they have been marked as _outdated_.

app/Models/Album.php
nagmat84: @qwerty287 This is just a heads up. The PR https://github.com/LycheeOrg/Lychee/pull/1277 cleans up the deletion of media files and symlinks for current master. After this has been reviewed, approved and merged into the master branch, I will port it to `consistent_json_api`. Then you can add the code for deleting tracks.

Just in case you already would like to have a look.

qwerty287: Thanks! I'll look into it once it is ported.

nagmat84:
I merged the current changes of the `master` branch into `consistent_json_api`. In turn, you can now merge `consistent_json_api` and add the code to all delete tracks if albums are deleted recursively.

The method you need to edit is this one:

https://github.com/LycheeOrg/Lychee/blob/consistent_json_api/app/Actions/Album/Delete.php#L59

By the way, I would recommend to not create the file name of the track based on the hash, but use a truly random file name.
You could do it like that:

```php
const NAME_LENGTH = 24;
$baseFileName = strtr(base64_encode(random_bytes(3 * NAME_LENGTH / 4)), '+/', '-_')
```

Otherwise you need to take care of tracks which are used by more than one album. You do it correctly here

https://github.com/LycheeOrg/Lychee/pull/1274/files#diff-611b1c6b31b277082e42583d35aea738841a58a007ca0cf2b6a19f67ca5a53ebR333-R335

and this is fine, if you only need to delete a single album. But if you need to delete a set of albums (and the sets may get huge!), then this will become a PITA, because you need to take care of not deleting a track which is used by another album unless this other album is going to be deleted, too.

There are similar queries for that in `Actions\Photo\Delete`, for example

https://github.com/LycheeOrg/Lychee/blob/31f93f7949f5d3014cffc6fad6a7de6166867200/app/Actions/Photo/Delete.php#L100

which may serve as a role model. But these queries are complicated and inefficient. (The DB needs to perform a join on strings.) While I see the storage benefit for media files, track files are usually rather small. So I don't think it is worth the effort.

qwerty287:
@nagmat84 
I updated the branch, changed the filename generation and tried a recursive deletion, but it doesn't use a nice way. The core issue is that I need to have the `Album` models to run `deleteTrack` on it - or should I refactor this and move the method to an own action?
The ugly commit is https://github.com/LycheeOrg/Lychee/pull/1274/commits/34072b5d87c41b1387f954c4443e47aa73e25b2d.

nagmat84:
> @nagmat84 Should I refactor this and move the method to an own action? The ugly commit is [34072b5](https://github.com/LycheeOrg/Lychee/commit/34072b5d87c41b1387f954c4443e47aa73e25b2d).

Ah, no, the idea is to add all the track files to `$fileDeleter`, too, via `$fileDeleter->addRegularFile`. This way the track files will also be deleted in a deferred way after the actual response has already been sent to the client.

To do so, you have to move your added code down. You obtain the object `$fileDeleter` some lines below from the `PhotoDelete` action. Then you re-use this object and add the track files, too.

qwerty287:
I changed it in https://github.com/LycheeOrg/Lychee/pull/1274/commits/ec3971a733bce0d232bd36da3489285b04e8140d.

nagmat84:
> I changed it in [ec3971a](https://github.com/LycheeOrg/Lychee/commit/ec3971a733bce0d232bd36da3489285b04e8140d).

Great! That is what I meant. :+1:

Now, I am really nit-picking here. I am wondering, if we can improve the efficiency by avoiding this additional DB query. (But this is really gold-plating). 

**1st change**

Some lines above, we already have this code

```php
$albums = Album::query()
  ->without(['cover', 'thumb'])
  ->select(['id', 'parent_id', '_lft', '_rgt'])
  ->findMany($albumIDs);
```

You could add `'track_short_path'` to the set of queried columns there.

**2nd change**

Some lines below, we have

```php
$recursiveAlbumIDs = array_merge($recursiveAlbumIDs, $album->descendants()->pluck('id')->all());
```

This could be rewritten to something like

```php
$recursiveAlbums = $album->descendants()->select(['id', 'track_short_path'])->all();
$recursiveAlbumIDs = array_merge($recursiveAlbumIDs, $recursiveAlbums->pluck('id')->all());
```

and later you can use `$recursiveAlbums->pluck('track_short_path')` for your own code.

**Further remark**

I don't know why automatic tests fail for some jobs with the error message `'Could not determine the branch'` but not for others. I guess it might be related to the fact that your branch is hosted in another repository and not the Lychee repository. But I don't know for sure. If you don't mind, I would try to move this branch into the Lychee repository and give you write access to that. But I must check first, if I can do so without loosing all discussion.

qwerty287:
You can take a look at https://github.com/LycheeOrg/Lychee/pull/1274/commits/cb92363ff150ac1805334065f276e17493bb0fe0, I think that's (almost) what you mean.

> I don't know why automatic tests fail for some jobs with the error message 'Could not determine the branch' but not for others. I guess it might be related to the fact that your branch is hosted in another repository and not the Lychee repository. But I don't know for sure. If you don't mind, I would try to move this branch into the Lychee repository and give you write access to that. But I must check first, if I can do so without loosing all discussion.

That's fine for me, but not possible with GitHub as far as I know. GitHub doesn't allow to change the head branch, so this only works with loosing the discussion and opening a new PR from the new branch.

app/Actions/Album/Delete.php
```diff
@@ -78,18 +78,23 @@ public function do(array $albumIDs): FileDeleter
 			// loading expensive `min_taken_at` and `max_taken_at`.
 			$albums = Album::query()
 				->without(['cover', 'thumb'])
-				->select(['id', 'parent_id', '_lft', '_rgt'])
+				->select(['id', 'parent_id', '_lft', '_rgt', 'track_short_path'])
 				->findMany($albumIDs);
 
+			$recursiveAlbums = $albums;
```
nagmat84:
```suggestion
```

app/Actions/Album/Delete.php
```diff
@@ -78,18 +78,23 @@ public function do(array $albumIDs): FileDeleter
...
+
 			/** @var Album $album */
 			foreach ($albums as $album) {
-				// Collect the IDs of all (aka recursive) sub-albums in each album
-				$recursiveAlbumIDs = array_merge($recursiveAlbumIDs, $album->descendants()->pluck('id')->all());
+				// Collect all (aka recursive) sub-albums in each album
+				$subAlbums = $album->descendants()->select(['id', 'track_short_path'])->all();
+				$recursiveAlbumIDs = array_merge($recursiveAlbumIDs, $subAlbums->pluck('id')->all());
+				$recursiveAlbums = array_merge($recursiveAlbums, $subAlbums);
```
nagmat84:
```suggestion
```

app/Actions/Album/Delete.php
```diff
@@ -78,18 +78,23 @@ public function do(array $albumIDs): FileDeleter
...
 			}
 
 			// Delete the photos from DB and obtain the list of files which need
 			// to be deleted later
 			$fileDeleter = (new PhotoDelete())->do($unsortedPhotoIDs, $recursiveAlbumIDs);
+			$fileDeleter->addRegularFiles($recursiveAlbums->pluck('track_short_path'));
```
nagmat84: Nearly there! My suggestions avoid some unnecessary memory consumption and copying.

```suggestion
			$fileDeleter->addRegularFiles($albums->pluck('track_short_path'));
			$fileDeleter->addRegularFiles($subAlbums->pluck('track_short_path'));
```

nagmat84 (commented):
Great job! Just some minor suggestions to reduce memory consumption. I won't have the time to test it today, but it looks good from the code.

nagmat84:
> That's fine for me, but not possible with GitHub as far as I know. GitHub doesn't allow to change the head branch, so this only works with loosing the discussion and opening a new PR from the new branch.

Too bad! :(

app/Actions/Album/Delete.php
qwerty287: But `subAlbums` is a variable which is set every time in the `foreach` statement in line 89 and only contains sub albums of one top-level album. That's the reason for this new `recursiveAlbums` variable.

nagmat84: My fault! :-(

> That's the reason for this new `recursiveAlbums` variable.

Understood. But maybe we could just change it into a `recursiveAlbumTracks` in analogy to `recursiveAlbumIDs`. I would like to avoid duplicating entire `Album` models.

qwerty287: https://github.com/LycheeOrg/Lychee/pull/1274/commits/37f013e88c3b7d62b7d912cb21336b201052a85f

qwerty287:
> > That's fine for me, but not possible with GitHub as far as I know. GitHub doesn't allow to change the head branch, so this only works with loosing the discussion and opening a new PR from the new branch.
> 
> Too bad! :(

Just looked at the code, and the issue is [this line](https://github.com/LycheeOrg/Lychee/blob/consistent_json_api/app/Metadata/GitHubFunctions.php#L63=). The array does not have a key `2`. I tried this with some changes to the GH actions workflow: the `.git/HEAD` file looks like `abc4732b0347d357df6b7ea9f679e7dc1097fa5b` and this aren't enough `/` chars for the array - an exception is thrown.

d7415:
Should we be adding a `public/uploads/tracks/index.html` to block file listings? Only because we have them elsewhere.

d7415:
> > > That's fine for me, but not possible with GitHub as far as I know. GitHub doesn't allow to change the head branch, so this only works with loosing the discussion and opening a new PR from the new branch.
> > 
> > 
> > Too bad! :(
> 
> Just looked at the code, and the issue is [this line](https://github.com/LycheeOrg/Lychee/blob/consistent_json_api/app/Metadata/GitHubFunctions.php#L63=). The array does not have a key `2`. I tried this with some changes to the GH actions workflow: the `.git/HEAD` file looks like `abc4732b0347d357df6b7ea9f679e7dc1097fa5b` and this aren't enough `/` chars for the array - an exception is thrown.

As someone who's run into issues with this stuff before, it doesn't surprise me. It also does sort of make sense that GH is only taking the relevant commit and not branch info. We should probably allow a placeholder (e.g. "unknown", "undetermined") or otherwise allow testing to continue. I'd have to double check where this is used though. Or, more accurately, whoever fixes it will have to.

qwerty287:
I suggested a change in the refactor PR which's the source for this. https://github.com/LycheeOrg/Lychee/pull/1217#pullrequestreview-951224283=

nagmat84 (commented):
Just reviewed your recent changes. I am happy with the code. No further change requests from my side. :tada: :confetti_ball:

I will approve it after I have successfully tested it.

nagmat84:
> I suggested a change in the refactor PR which's the source for this. [#1217 (review)](https://github.com/LycheeOrg/Lychee/pull/1217#pullrequestreview-951224283)

I believe I fixed the problem. See here https://github.com/LycheeOrg/Lychee/pull/1217#discussion_r857164326.

Please try to merge. I already tried it locally and run into another test error, but this time in your code, because of a non-existing method which indeed does not exist.

qwerty287:
> Please try to merge. I already tried it locally and run into another test error, but this time in your code, because of a non-existing method which indeed does not exist.

Sorry, but if this is an issue I have to fix, could you please tell me which method is missing? I don't really understand the test logs of GH actions and have completely different results locally.

qwerty287:
Ah, nevermind, found one which doesn't work, probably that is it.

qwerty287:
The CI fail still happens during an github update test.

nagmat84:
OK, let me see what I can do

nagmat84:
This should do it: https://github.com/LycheeOrg/Lychee/pull/1217/commits/6e380596d491c028fd9ceee9fe1195009074700f (OK, I said that before :see_no_evil:)

nagmat84:
@qwerty287 As I don't know if you have already joined our developers chat on Gitter, I get in touch here. This evening today (my timezone, CEST, i.e. in approx. 9 hours from now) I am going to merge my big re-factoring `consistent_json_api` into `master` :tada:

This will have some effect on your branches. I asked you some time ago to base your feature branches on top of `consistent_json_api`. What I meant and what I assumed was that your branch should have been re-based on `consistent_json_api`, i.e. be strictly ahead of that. If that was the case you could have simply run

    git rebase --onto master consistent_json_api track

on your branch after `consistent_json_api` had been merged into `master`. The GIT command above simply means take all the changes of `track` from where it has branched from `consistent_json_api` and re-apply them to the (new) `master`. As the head of `master` after the merge will be exactly identical to `consistent_json_api`, this would have worked smoothly.

However, as I see from the GIT history your branch `tracks` is based on some ancient version of `master` and at some point you merged `consistent_json_api` into it. As we use squash-and-merge for the master branch all the ~500 commits of `consistent_json_api` will be gone after the merge. However, they will still be included in the history of your branch. This will make your branch being diverted from `master` by that. Re-merging the new `master` back into `tracks` will probably lead you right into merge hell, because GIT won't recognize that the ~500 commits in your branch are already included in `master`.

Unfortunately, I am no GIT expert and can't give you an easy solution. This is what I would probably do, but it is likely not the most elegant way.

 - Rename your local working copy of your branch to something like `track_backup`.
- Create a new local branch `track` on top of the new master (after the merge)
- Cherry-pick all of your relevant commit from `track_backup` manually thereby building a clean history for `track` leaving out all the commits from `consistent_json_api`
- Double-check that at the end of it `track` and `track_backup` are identical (via `git diff track_backup track`)!!!
- If everything is right force-push the newly built branch `track` to the upstream branch `track`. This will preserve the discussion inside the PSR.

And: Check out a local working-copy of `consistent_json_api` in case you need it. After the merge, the upstream branch will be gone.

d7415:
If they're actual merges (not squashes), and if you get in while the `consistent_json_api` branch still exists (or so long as you have an up-to-date-enough copy) you might still be able to get away with a rebase onto `consistent_json_api` as I *think* it should still recognise the duplicated work and discount it.

I'd still recommend taking a copy first though.

Edit to add: They look like "proper" merges at a glance, so this may work.

Edit again: Looks like this successfully strips out the duplicates, but there are still conflicts that will need looking at. Not many though.

qwerty287:
Thanks for the ideas, but I don't know if it's easier if I just re-apply the diff to a new branch created from `master`? Something like `git diff consistent_json_api..master | git apply` on a newly added branch? Since the changes of `consistent_json_api` are in `master `, there shouldn't be any conflicts. It might be good if you could wait with deleting the `consistent_json_api` until I rebased (or which other strategy I use) this to target `master`.

ildyria:
> Thanks for the ideas, but I don't know if it's easier if I just re-apply the diff to a new branch created from `master`? Something like `git diff consistent_json_api..master | git apply` on a newly added branch? Since the changes of `consistent_json_api` are in `master `, there shouldn't be any conflicts. It might be good if you could wait with deleting the `consistent_json_api` until I rebased (or which other strategy I use) this to target `master`.

I would also wait and see. In worse case, the diff would also be the approach I would take. :)

nagmat84:
> > Thanks for the ideas, but I don't know if it's easier if I just re-apply the diff to a new branch created from `master`? Something like `git diff consistent_json_api..master | git apply` on a newly added branch? Since the changes of `consistent_json_api` are in `master `, there shouldn't be any conflicts. It might be good if you could wait with deleting the `consistent_json_api` until I rebased (or which other strategy I use) this to target `master`.
> 
> I would also wait and see. In worse case, the diff would also be the approach I would take. :)

I am on it. I already did it for the Lychee-front branches:

 - https://github.com/LycheeOrg/Lychee-front/tree/drag-drop_rebased
 - https://github.com/LycheeOrg/Lychee-front/tree/qrcode-35_rebased
 - https://github.com/LycheeOrg/Lychee-front/tree/sass_rebased
 - https://github.com/LycheeOrg/Lychee-front/tree/tracks_rebased

But I only did it for those branches which have an open PR. At the moment they are all based on the branch `consistent_json_api`, because I haven't yet merge that branch into `master`. This step will follow soon. After I am ready, I have to do the following. Check them out locally to your development machine to a local branch without the `_rebased` suffix, then force push them to your branches of your Github repository. This will also update the PRs without loosing the discussion.

But wait until I am ready with the master branch.

qwerty287:
Nice and thanks for the effort!

nagmat84:
Ane here is your re-based branch: https://github.com/LycheeOrg/Lychee/tree/tracks_rebased

nagmat84:
I re-open the pull request as I restored the already merged branch `consistent_json_api`. However, actually the re-based branch above should be taken and merged into master. I am not sure if the targeted branch of this PR can be changed.

nagmat84:
Ah, it is indeed possible. See here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

So there are two things to do:

- Change the target of this PR from `consistent_json_api` to `master`. This will probably result in a lot "this brach has conflicts" warnings.
- Use the linked, rebased branch `LycheeOrg/Lychee/tracks_rebased` and push it to `querty287/Lychee/tracks`. Then you should be fine again.

app/Http/Controllers/AlbumController.php
```diff
@@ -211,6 +213,30 @@ public function setLicense(SetAlbumLicenseRequest $request): void
...
+	 *
+	 * @return void
+	 */
+	public function setTrack(SetAlbumTrackRequest $request): void
+	{
+		$request->album()->setTrack($request->file);
+	}
+
+	/**
+	 * Upload a track for the Album.
```
kamil4:
```suggestion
	 * Delete a track from the Album.
```

public/dist/leaflet.markercluster.js.map
```diff
@@ -1 +1 @@
...
...B7a,KAAKkC,eAGNA,YAAa,SAAUkW,GAClBpY,KAAK6M,aACR7M,KAAK6M,YAAYsL,WAAWC,IAI9B3S,uBAAwB,WACnBzF,KAAK6M,aACR7M,KAAK6M,YAAYpH,0BAKnBxC,iBAAkB,SAAUvB,GACvBA,EAAM0X,aACTpZ,KAAKE,cAAc6C,YAAYrB,GAE3BA,EAAM0B,aACT1B,EAAM0B,cAGH1B,EAAMyX,iBACTzX,EAAMyX,gBAAgB,GAGvBnZ,KAAK+B,KAAKgB,YAAYrB,EAAM0X,mBACrB1X,EAAM0X,eCjdhBrb,EAAED,mBAAmB+R,QAAQ,CAS5BuL,gBAAiB,SAAU5U,GAoB1B,OAnBKA,EAEMA,aAAkBzI,EAAED,mBAC9B0I,EAASA,EAAOnE,iBAAiByC,qBACvB0B,aAAkBzI,EAAE4D,WAC9B6E,EAASA,EAAO6U,QACN7U,aAAkBzI,EAAEwD,cAC9BiF,EAASA,EAAO1B,qBACN0B,aAAkBzI,EAAEO,SAC9BkI,EAAS,CAACA,IARVA,EAASxG,KAAKqC,iBAAiByC,qBAUhC9E,KAAKsb,4BAA4B9U,GACjCxG,KAAKuC,wBAGDvC,KAAK9B,QAAQW,kBAChBmB,KAAKub,gCAAgC/U,GAG/BxG,MAQRsb,4BAA6B,SAAU9U,GACtC,IAAIE,EAAIoH,EAGR,IAAKpH,KAAMF,EAOV,IADAsH,EAAStH,EAAOE,GAAI/D,SACbmL,GACNA,EAAOzC,kBAAmB,EAC1ByC,EAASA,EAAOnL,UAWnB4Y,gCAAiC,SAAU/U,GAC1C,IAAIE,EAAIhF,EAER,IAAKgF,KAAMF,EACV9E,EAAQ8E,EAAOE,GAGX1G,KAAKiC,SAASP,IAEjBA,EAAMyQ,QAAQnS,KAAK0N,oBAAoBhM,OAM3C3D,EAAEO,OAAOuR,QAAQ,CAQhB2L,mBAAoB,SAAUtd,EAASud,GACtC,IAAI7L,EAAO5P,KAAK9B,QAAQ0R,KAcxB,OAZA7R,EAAEgC,WAAW6P,EAAM1R,GAEnB8B,KAAKmS,QAAQvC,GAMT6L,GAA2Bzb,KAAK2C,UACnC3C,KAAK2C,SAASoE,OAAOqU,gBAAgBpb,MAG/BA","file":"dist/leaflet.markercluster.js.map"}
```
kamil4: General question: why have the permission bits for this file changed from `644` to `755`? @qwerty287, are you developing on Windows?

public/dist/_frame--javascript.js
```diff
@@ -0,0 +1,851 @@
+"use strict";
```
kamil4: This file should not be part of the commit at all.

public/dist/_main--javascript.js
```diff
@@ -0,0 +1,13112 @@
+"use strict";
```
kamil4: This file should not be part of the commit at all.

public/dist/_view--javascript.js
```diff
@@ -0,0 +1,3962 @@
+"use strict";
```
kamil4: This file should not be part of the commit at all.

kamil4 (changes requested):
I looked through this PR one more time. I spotted one trivial comment issue and a couple of weird issues with files (temporary files being part of the PR, one file changing permissions for unclear reasons). Otherwise it looks good.

public/dist/leaflet.markercluster.js.map
qwerty287: No, I'm actually on Linux/Manjaro, and I didn't change that manually. Will rollback.

public/dist/_frame--javascript.js
qwerty287: They were deleted after I applied the rebased branch.

public/dist/leaflet.markercluster.js.map
qwerty287: Seems that moving this to the rebased branch fixed it.

kamil4 (approved):

app/Models/Album.php
```diff
@@ -259,4 +271,58 @@ public function newEloquentBuilder($query): AlbumBuilder
...
+		if ($this->track_short_path === null) {
+			return;
+		}
+		// make sure track is not used for other albums
+		if (self::query()->where('track_short_path', '=', $this->track_short_path)->count() <= 1) {
+			Storage::delete($this->track_short_path);
+		}
+		$this->track_short_path = null;
+		$this->save();
+	}
```
nagmat84: Same thing here. BTW. The query is not necessary anymore as the file names are truly random.

```suggestion
	/**
	 * Delete the track of the album.
	 *
	 * @return void
	 *
	 * @throws ModelDBException
	 */
	public function deleteTrack(): void
	{
		if ($this->track_short_path === null) {
			return;
		}
		Storage::delete($this->track_short_path);
		$this->track_short_path = null;
		$this->save();
	}
```

app/Models/Album.php
```diff
@@ -259,4 +271,58 @@ public function newEloquentBuilder($query): AlbumBuilder
...
+	{
+		if ($this->track_short_path !== null) {
+			Storage::delete($this->track_short_path);
+		}
+
+		$new_track_id = strtr(base64_encode(random_bytes(18)), '+/', '-_');
+		Storage::putFileAs('tracks/', $file, "$new_track_id.xml");
+		$this->track_short_path = "tracks/$new_track_id.xml";
+		$this->save();
+	}
```
nagmat84: My  IDE complains about several unhandled/uncaught exceptions here. `random_bytes` may throw `\Exception`, `save()` may throw `ModelDBException`. The latter is fine, because it already inherits the `HttpException` interface, but it should be listed in the PHPdoc. The former is unfortunate, because legacy PHP throws such kind of generic exceptions. I would probably do something like this

```suggestion
	/**
	 * Set the GPX track for the album.
	 *
	 * @param UploadedFile $file the GPX track file to be set
	 *
	 * @return void
	 *
	 * @throws ModelDBException
	 * @throws MediaFileOperationException
	 */
	public function setTrack(UploadedFile $file): void
	{
		try {
			if ($this->track_short_path !== null) {
				Storage::delete($this->track_short_path);
			}

			$new_track_id = strtr(base64_encode(random_bytes(18)), '+/', '-_');
			Storage::putFileAs('tracks/', $file, "$new_track_id.xml");
			$this->track_short_path = "tracks/$new_track_id.xml";
			$this->save();
		} catch (ModelDBException $e) {
			throw $e;
		} catch (\Exception $e) {
			throw new MediaFileOperationException('Could not save track file', $e);
		}
	}
```

app/Http/Controllers/AlbumController.php
```diff
@@ -211,6 +213,30 @@ public function setLicense(SetAlbumLicenseRequest $request): void
 		$request->album()->save();
 	}
 
+	/**
+	 * Upload a track for the Album.
+	 *
+	 * @param SetAlbumTrackRequest $request
+	 *
+	 * @return void
+	 */
```
nagmat84:
```suggestion
	 * @return void
	 *
	 * @throws MediaFileOperationException
	 * @throws ModelDBException
	 */
```

Make my IDE happy.

app/Http/Controllers/AlbumController.php
```diff
@@ -211,6 +213,30 @@ public function setLicense(SetAlbumLicenseRequest $request): void
...
+		$request->album()->setTrack($request->file);
+	}
+
+	/**
+	 * Delete a track from the Album.
+	 *
+	 * @param DeleteTrackRequest $request
+	 *
+	 * @return void
+	 */
```
nagmat84:
```suggestion
	 * @return void
	 *
	 * @throws ModelDBException
	 */
```

dito.

nagmat84 (changes requested):
Some more changes :smirk:

nagmat84 (approved):
Ttesting finished. I created several folders, added and deleted tracks individually and also recursively deleted a sub-tree of albums. Tracks were deleted as expected. :+1:

PR #1275, opened 2022-04-14
Added file extension to temporary files

### LINKED ISSUE #1269: No thumbnails on videos with ffmpeg
No previews are generated on videos, even with ffmpeg installed and configured in Lychee. Thumbnails were generated file in older revisions. Up to date with master branch.

Upload a video in album, no thumbnail. The video can't be played too.

I think composer dependencies are broken for ffmpeg-php. Here is what happens when checking with composer.

```
sudo -u www-data composer require php-ffmpeg/php-ffmpeg
Info from https://repo.packagist.org: #StandWithUkraine
Using version ^1.0 for php-ffmpeg/php-ffmpeg
./composer.json has been updated
Running composer update php-ffmpeg/php-ffmpeg
Loading composer repositories with package information                                      Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires php-ffmpeg/php-ffmpeg ^1.0, found php-ffmpeg/php-ffmpeg[v1.0.0, v1.0.1, v1.x-dev] but these were not loaded, likely because it conflicts with another require.
  Problem 2
    - lychee-org/php-exif is locked to version dev-master and an update of this package was not requested.
    - lychee-org/php-exif dev-master requires php-ffmpeg/php-ffmpeg ^0.17.0 -> found php-ffmpeg/php-ffmpeg[v0.17.0] but it conflicts with your root composer.json require (^1.0).

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.
You can also try re-running composer require with an explicit version constraint, e.g. "composer require php-ffmpeg/php-ffmpeg:*" to figure out if any version is installable, or "composer require php-ffmpeg/php-ffmpeg:^2.1" if you know which you need.

Installation failed, reverting ./composer.json and ./composer.lock to their original content.
```

After some tests, logs from Lychee

```
 2022-04-11 21:57:22 -- error   -- App\Actions\Photo\Strategies\AddStandaloneStrategy::do -- 88 -- Failed to generate size variants, error was Media file is reported to be a video, but aperture (aka duration) has not been extracted
 2022-04-11 21:57:22 -- error   -- App\Image\SizeVariantDefaultFactory::extractReferenceImage -- 60 -- Media file is reported to be a video, but aperture (aka duration) has not been extracted
 2022-04-11 21:57:20 -- notice  -- App\Metadata\Extractor::extract -- 160 -- Falling back to native adapter.
```

```
    Diagnostics
    -------
    Warning: lossless_optimization set to 1 but cwebp not found!
    Warning: lossless_optimization set to 1 but svgo not found!
    Warning: You may experience problems when uploading a photo of large size or handling many/large albums. Take a look in the FAQ for details.




    System Information
    --------------
    Lychee Version (git):            master (d8279c8) - Data not in Cache
    DB Version:                      4.5.0
    
    composer install:                dev
    APP_ENV:                         local
    
    PHP Version:                     8.1
    Max uploaded file size:          100M
    Max post size:                   100M
    Max execution time:              90
    MySQL Version:                   10.3.34-MariaDB-0+deb10u1-log
    
    Imagick Version:                 1690




    Config Information
    --------------
    version:                         040500
    sorting_Photos_col:              title
    sorting_Albums_order:            ASC
    layout:                          0
    image_overlay_type:              none
    landing_owner:                   
    landing_title:                   
    landing_subtitle:                
    landing_facebook:                
    landing_flickr:                  
    landing_twitter:                 
    landing_instagram:               
    landing_youtube:                 
    landing_background:              
    site_title:                      BakaGamer - Screenshots
    site_copyright_enable:           0
    site_copyright_begin:            2010
    additional_footer_text:          
    photos_wraparound:               0
    has_ffmpeg:                      1
    location_show:                   0
    lossless_optimization:           1
    nsfw_warning:                    1
```

nagmat84:
I assume this should fix #1269, if @kamil4's idea is correct.

Adding file extensions to the temporary files was quite straight forward. However, I didn't test it, because I don't know how to trigger the error. @kamil4 could you do it?


nagmat84:
Sorry, my fault. I did a silly mistake.

kamil4:
This code may _also_ be needed (if we try to import from a URL, which I haven't tried yet) but it doesn't solve  #1269, where the issue is with _uploads_. I believe that for uploads the file is already in `/tmp` (put in there by logic below Lychee) with a file name without extension, and we need to rename it to give it the original extension.

FYI, a sample file that triggers the problem is https://screenshots.baka-gamer.net/uploads/big/6b6370b37add1991cd72af52b660b9dc.mp4

nagmat84:
I added some work-around and a lot of comments which solves the problem. I tested it with the given sample.

However, I don't really like the work-around. It introduces an unnecessary copy operation. Moreover, I fear that this all will break again when the storage adapter will be replaced by something other than a local storage in the future.

In the light of the newly found bug and fact that my comments on that are already some months old, I guess this part has obtained a higher priority than cleaning up the web-based installer/updater. As long as our MIME type handling code is cluttered like this

https://github.com/LycheeOrg/Lychee/blob/019921260b4d8e1bac0e61ec99fd4d8782c68710/app/Metadata/Extractor.php#L81-L109

https://github.com/LycheeOrg/Lychee/blob/019921260b4d8e1bac0e61ec99fd4d8782c68710/app/Actions/Import/Exec.php#L255-L258

and inconsistent, the whole process will always be very fragile.

kamil4:
Totally agreed. We are extremely inconsistent about dealing with MIME types and such in Lychee, and on top of that php-exif does its own thing. We should be establishing MIME type and file kind in one place and passing that info around as needed, not depend on file extensions and such, which will bite us with non-local file storage.

So do I understand it correctly that your latest commit uses the client-supplied MIME type in at least one place?

nagmat84:
> So do I understand it correctly that your latest commit uses the client-supplied MIME type in at least one place?

Yes, you did.

I first changed the code to use the client-supplied MIME type which alone did not solve the problem. As the temporary file created by PHP had no file extension, PHPexif still determined the file as `application/octet-stream`. Then I added the code which copies the uploaded file into a new file with the client-supplied original file extension. Note that the original file extension is also client-supplied. This made PHPexif finally happy. However, I did not try, if it was sufficient to only use the client-supplied file extension.

app/Http/Controllers/PhotoController.php
```diff
@@ -93,6 +94,37 @@ public function add(AlbumIDRequest $request): Photo
...
+		// re-determine the MIME type of the file based on the file path.
+		// This is not only inefficient, but the original MIME type (of the
+		// uploaded file) gets lost on the way.
+		// As a work-around we store the uploaded file with a file extension.
+		// Unfortunately, we cannot simply re-name the file, because this
+		// might break due to permission problems for certain installation
+		// if the temporarily uploaded file is stored in the system-global
+		// temporary directory below another mount point or another Docker
+		// image than the Lychee installation.
+		// Hence, we must make a deep copy.
```
kamil4, line 113: I feel like ideally we should try to rename and only if that fails, do a deep copy. But since this is a temporary fix, I'm OK with the way it is.

app/Image/TemporaryLocalFile.php
```diff
@@ -10,12 +10,23 @@
...
+		// We must not use the usual PHP method `tempnam`, because that
+		// method does not handle file extensions well, but our temporary
+		// files need a proper (and correct) extension for the MIME extractor
+		// to work.
+		$tempFilePath = sys_get_temp_dir() .
+			DIRECTORY_SEPARATOR .
+			'lychee-' .
+			strtr(base64_encode(random_bytes(12)), '+/', '-_') .
+			$fileExtension;
+		touch($tempFilePath);
```
kamil4: I'm confused by this code. Shouldn't it be in a loop so that if the generated random file name turns out to already exist, we generate a new one and try again? Instead we just a have `touch` here, which I don't think has the semantics of `O_CREAT | O_EXCL` that is typically required to make such temp file generation safe?

kamil4 (commented):
Sorry; this fell through the cracks...

I verified that the PR works as expected both when importing from URL and uploading. So that's good. However, I do have a question or two based on the code review.

app/Image/TemporaryLocalFile.php
nagmat84: You know more about that stuff than I do. I just thought, we have 96bit or randomness in the filename so collisions are negligible. But I agree, we should do it the right way. Need to read about `O_CREAT | O_EXCL` first.

nagmat84: Fixed it in: https://github.com/LycheeOrg/Lychee/pull/1275/commits/74b9c45137908ab5918ee5c29259b14db5f79e45

kamil4 (approved):

PR #1277, opened 2022-04-15
More efficient deletion (but ugly)

nagmat84:
When the backend receives a request to delete (regular) albums the backend needs to do the following steps:
- determine all photos in all sub-albums of the album at hand
- determine all size variants of these photos
- determine all symbolic links which point to these size variants
- delete all the entities from the DB (in the correct order)
- delete the associated files unless they are not shared by a duplicate which is not going to be deleted

In the current code each of the model classes has a overloaded method `delete()` which deletes the model from the DB and also ensures proper deletion of any associated files and/or child models. This is fine, because it follows the principle of separation of concerns. Every model class knows how to correctly delete itself and we (the programmers) cannot forget to delete the media files, because `delete()` will take care of it.

However, this approach is rather inefficient due to Laravel's (bad) architecture. In particular, if we want the `delete` method to be executed, we must iterate over each model and call `delete` explicitly. We must no use "mass deletion" via SQL queries like `Album::query()->...->delete()`, because these queries won't call `delete` on the model. :unamused: (See warning box in https://laravel.com/docs/9.x/eloquent#events.)

This inefficiency was mostly introduced by my refactoring, because I am a big fan of proper encapsulation and defensive programming, i.e. deleting a model should not leave you in an inconsistent state, because the code allows you to accidentally forget to delete the media files, too. However, the current situation is prohibitive. It takes 45s to delete an album with a deeply nested sub-tree of 150 sub-albums and 5000 photos.

This PR solves this problem by partially reverting some of my changes. In particular, it introduces three `Action` classes again to delete albums, photos and size variants. These action classes violates the principle of separation of concerns by using `Album`, `Photo` and `SizeVariant` in a white-box fashion. They pull out the code for (physical) deletion of files and use SQL mass deletion commands to delete the models from the DB. As a bonus point deletion of files is deferred after sending the response to the client.


app/Models/Album.php
```diff
@@ -150,44 +151,32 @@ public function toArray(): array
...
-		} catch (\Exception $e) {
-			try {
-				// if anything goes wrong, don't leave the tree in an inconsistent state
-				$this->newModelQuery()->fixTree();
-			} catch (\Throwable) {
-				// Sic! We cannot do anything about the inner exception
-			}
-			throw $e;
-		}
+		$fileDeleter = (new Delete())->do([$this->id]);
```
ildyria, line 159:
```suggestion
		$albumDeleter = (new Delete())->do([$this->id]);
```

app/Models/Album.php
```diff
@@ -150,44 +151,32 @@ public function toArray(): array
...
+		$this->exists = false;
+		$fileDeleter->do();
```
ildyria, line 161:
```suggestion
		$albumDeleter->do();
```

app/Models/Photo.php
```diff
@@ -427,25 +428,13 @@ public function replicate(array $except = null): Photo
...
-		if (!$keepFiles && !empty($livePhotoShortPath) && Storage::exists($livePhotoShortPath)) {
-			$success &= Storage::delete($livePhotoShortPath);
-		}
-
-		if (!$success) {
-			return false;
-		}
-
-		return parent::delete() !== false;
+		$fileDeleter = (new Delete())->do([$this->id]);
```
ildyria, line 436:
```suggestion
		$photoDeleter = (new Delete())->do([$this->id]);
```

app/Models/Photo.php
```diff
@@ -427,25 +428,13 @@ public function replicate(array $except = null): Photo
...
+		$this->exists = false;
+		$fileDeleter->do();
```
ildyria, line 438:
```suggestion
		$photoDeleter->do();
```

app/Models/SizeVariant.php
```diff
@@ -212,39 +213,12 @@ public function setSizeVariantAttribute(int $sizeVariantType): void
...
-			$shortPath = $this->short_path;
-			if (!empty($shortPath) && $disk->exists($shortPath)) {
-				if ($disk->delete($shortPath) === false) {
-					return false;
-				}
-			}
-		}
-
-		return parent::delete() !== false;
+		$fileDeleter = (new Delete())->do([$this->id]);
```
ildyria, line 220:
```suggestion
		$sizeVariantDeleter = (new Delete())->do([$this->id]);
```

app/Models/SizeVariant.php
```diff
@@ -212,39 +213,12 @@ public function setSizeVariantAttribute(int $sizeVariantType): void
...
+		$this->exists = false;
+		$fileDeleter->do();
```
ildyria, line 222:
```suggestion
		$sizeVariantDeleter->do();
```

app/Models/Album.php
nagmat84: No. `FileDeleter` is correct, because `->do` returns an object which delete files. The albums have already been deleted from the DB.

app/Models/Album.php
nagmat84: See above.

app/Models/Photo.php
nagmat84: See above.

app/Models/Photo.php
nagmat84: See above.

app/Models/SizeVariant.php
nagmat84: See above.

app/Models/SizeVariant.php
nagmat84: See above.

app/Actions/Album/Delete.php
```diff
@@ -2,29 +2,130 @@
...
+			// because it provides deletion of photos
+			if (in_array(UnsortedAlbum::ID, $albumIDs, true)) {
+				$query = UnsortedAlbum::getInstance()->photos();
+				if (!AccessControl::is_admin()) {
+					$query->where('owner_id', '=', AccessControl::id());
+				}
+				$photoIDs = UnsortedAlbum::getInstance()->photos()->pluck('id')->all();
+			}
+
+			// Only regular albums are owner of photos, so we only need to
```
kamil4:
```suggestion
			// Only regular albums are owners of photos, so we only need to
```

app/Actions/Album/Delete.php
```diff
@@ -2,29 +2,130 @@
...
-			if ($album instanceof BaseAlbum || $album instanceof UnsortedAlbum) {
-				$success &= $album->delete();
+			// Among the smart albums, the unsorted album is special,
+			// because it provides deletion of photos
+			if (in_array(UnsortedAlbum::ID, $albumIDs, true)) {
+				$query = UnsortedAlbum::getInstance()->photos();
+				if (!AccessControl::is_admin()) {
+					$query->where('owner_id', '=', AccessControl::id());
+				}
+				$photoIDs = UnsortedAlbum::getInstance()->photos()->pluck('id')->all();
```
kamil4:
```suggestion
				$photoIDs = $query->pluck('id')->all();
```

Untested, but we currently aren't using `$query` anywhere?!

app/Actions/Album/Delete.php
```diff
@@ -2,29 +2,130 @@
...
+				->delete();
+				$height = $rgt - $lft + 1;
+				$album->newNestedSetQuery()->makeGap($rgt + 1, -$height);
+				Album::$actionsPerformed++;
+			}
+			Album::query()->whereIn('id', $albumIDs)->delete();
+			TagAlbum::query()->whereIn('id', $albumIDs)->delete();
+
+			// Note, we may need to delete more base albums than those whose
+			// ID is in `$albumsIDs`.
```
kamil4:
```suggestion
			// ID is in `$albumIDs`.
```

app/Actions/Album/Merge.php
```diff
@@ -41,12 +41,11 @@ public function do(string $albumID, array $sourceAlbumIDs): void
...
 		// Now we delete the source albums
-		// ! we have to do it via Model::delete() in order to not break the tree
-		$albums = Album::query()->whereIn('id', $sourceAlbumIDs)->get();
-		/** @var Album $album */
-		foreach ($albums as $album) {
-			$album->delete();
-		}
+		// We must use the special `Delete` action in order to not break the
+		// tree.
+		// he returned `FileDeleter` can be ignored as all photos have been
```
kamil4:
```suggestion
		// The returned `FileDeleter` can be ignored as all photos have been
```

app/Http/Controllers/AlbumController.php
```diff
@@ -238,11 +239,10 @@ public function setLicense(AlbumModelIDRequest $request, SetLicense $setLicense)
 	 */
 	public function delete(AlbumIDsRequest $request, Delete $delete): IlluminateResponse
 	{
-		if ($delete->do(explode(',', $request['albumIDs']))) {
-			return response()->noContent();
-		} else {
-			return response('', 500);
-		}
+		$fileDeleter = $delete->do(explode(',', $request['albumIDs']));
+		App::terminating(fn () => $fileDeleter->do());
```
kamil4, line 243: Interesting. Are there any restrictions on what can be done in this callback / for how long /etc?

kamil4 (commented):
Untested. Looks good overall, though pretty complicated!

app/Http/Controllers/AlbumController.php
nagmat84: The only theoretic restriction is the maximum execution time for a script as configured in `php.ini`.

Of course, the script continues to run in its original executioncontext. For example, if you use Apache with `mod_php` and the prefork worker model, then this script occupies one Apache worker as long as it runs. The affected Apache process won't be available for serving new HTTP requests as long as it is busy executing the script of the old request. Of course, this is somehow obvious, but should be mentioned.

nagmat84:
> Untested. Looks good overall, though pretty complicated!

Actually, it is (more or less) the original approach from before I started the refactoring. The only thing which makes it more complicated than before is the additional indirection due to the size variants which are now independent models.

nagmat84:
Memo to my future me: Two aspects are missing:

 - Deletion of the video part of live photos
 - Files below `/upload/{big|raw}` may also be symbolic links, if the photo has been imported with the symlink option

nagmat84:
> Memo to my future me: Two aspects are missing:
> 
>  * Deletion of the video part of live photos
>  * Files below `/upload/{big|raw}` may also be symbolic links, if the photo has been imported with the symlink option

Done.

app/Image/FileDeleter.php
```diff
@@ -73,6 +89,31 @@ public function do(): bool
...
+				// Also note, the `file_exist` returns `false` for existing,
+				// but dead links.
+				// So the first part takes care of deleting links no matter
+				// if they are dead or alive.
+				// The latter part deletes (regular) files, but avoids errors
+				// in case the file doesn't exist.
+				$success &= ((is_link($absolutePath) && unlink($absolutePath)) || !file_exists($absolutePath) || unlink($absolutePath));
+			}
+		} else {
+			// If the disk is no local, we can assume that each file is a regular file
```
kamil4:
```suggestion
			// If the disk is not local, we can assume that each file is a regular file
```

kamil4 (approved):
Still not tested but LGTM.

nagmat84:
> Still not tested but LGTM.

Do you want to test it first before merge?

kamil4:
No, I'm so far behind the testing of #1217 that I don't want to hold other things up...

PR #1278, opened 2022-04-16
Add missing indices

nagmat84:
See discussion on Gitter


ildyria (approved):
Tested, LGTM.

@nagmat84 I added a bump version to 4.5.1. Merge or revert. :)

nagmat84:
I fear I haven't yet fully understood why we need these version bumps and when we add them. For example, we already have a couple of migrations after the last version bump and they just work well.

I don't mind whether it is there or not.

ildyria:
> I fear I haven't yet fully understood why we need these version dumps and when we add them. For example, we already have a couple of migrations after the last version bump and they just work well.
> 
> I don't mind if it is there or not.

The bump of minor versions are mostly here when there is one of the following condition applied:

- interesting minor changes added (e.g. support of the safe use of light sabers)
- a migration that really helps with the use of Lychee and really needs to be run.

We are in the later case with a bunch of migrations that might have been waiting to be run.
```
| Yes  | 2022_01_13_183131_bump_version040500                        | 59    |
| Yes  | 2022_01_16_181337_optimize_tables                           | 60    |
| Yes  | 2022_02_02_203008_filesize_size_variants                    | 61    |
| Yes  | 2022_04_06_091900_drop_objectionable_indices                | 63    |
| Yes  | 2022_04_16_170724_add_missing_indices                       | 64    |
| No   | 2022_04_16_174503_bump_version040501                        |       |
```

Though in my opinion, the file_size_variants should have required a minor bump already...

PR #1285, opened 2022-04-21
Update composer.lock

nagmat84:
@kamil4 has thankfully fixed the long-standing bug about invalid EXIF dates in https://github.com/LycheeOrg/php-exif. This PR updates our dependencies such that his bug fix also finds it way into Lychee.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1902,16 +1902,16 @@
         },
         {
             "name": "laravel/framework",
-            "version": "v8.83.7",
+            "version": "v8.83.9",
             "source": {
                 "type": "git",
                 "url": "https://github.com/laravel/framework.git",
-                "reference": "0706abf5c99d3b14149fcf79c95074dadd40b203"
+                "reference": "ac6e16bb56510eda2e373c6365d5c92da3fd559d"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/laravel/framework/zipball/0706abf5c99d3b14149fcf79c95074dadd40b203",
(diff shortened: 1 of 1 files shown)


kamil4 (approved):
I've always assumed that Lychee picks up the latest commit of php-exif, based on https://github.com/LycheeOrg/Lychee/blob/e75ea766b6d923cc1a562d3a1ad93bec62a950a7/composer.json#L31

So that's not happening without updating `composer.lock`? Bummer...

nagmat84:
> I've always assumed that Lychee picks up the latest commit of php-exif, based on
> 
> https://github.com/LycheeOrg/Lychee/blob/e75ea766b6d923cc1a562d3a1ad93bec62a950a7/composer.json#L31
> 
> So that's not happening without updating `composer.lock`? Bummer...

What do you mean by "happening". Nothing is happening there. The `composer.json` only describes what we would like to have. Then we (as the developers) need to the run `./composer update` which creates the file `composer.lock`.

Composer resolves all the dependency descriptions in all `composer.json` of all dependencies recursively and tries to find an optimal solution. For example, if we have a dependencies A and B in our dependency tree and both want another dependency C, but A wants C in version 1.1-1.3 and B wants C in version 1.5-1.7, then composer fill fail and complain. The found solution is written into `composer.lock` and persisted.

Our users only must run `composer install`. Composer install reads `composer.lock` and always installs the exact set of dependencies which has been committed by us.

TLTR: Whenever we change anything about our dependencies or just want a newer version, we must run `composer update` and commit the resulting `composer.lock`. `composer.json` is not used for installation.

PR #1288, opened 2022-04-25
Don't use locale

kamil4:
Undoes one change from #1270: the view mode didn't work because it does not support the localization of the blade.

### DIFF
--- a/resources/views/view.blade.php
+++ b/resources/views/view.blade.php
@@ -58,7 +58,7 @@
 <!-- Infobox -->
 <div class="sidebar">
     <div class="sidebar__header">
-        <h1>{{ $locale['PHOTO_ABOUT'] }}</h1>
+        <h1>About</h1>
     </div>
     <div class="sidebar__wrapper"></div>
 </div>


qwerty287:
That must be something I missed here... But couldn't you just add the `locale` variable to https://github.com/LycheeOrg/Lychee/blob/master/app/Http/Controllers/IndexController.php#L101=?
How it is done here: https://github.com/LycheeOrg/Lychee/blob/master/app/Http/Controllers/ViewController.php#L68=

ildyria:
> That must be something I missed here... But couldn't you just add the `locale` variable to [https://github.com/LycheeOrg/Lychee/blob/master/app/Http/Controllers/IndexController.php#L101=](https://github.com/LycheeOrg/Lychee/blob/master/app/Http/Controllers/IndexController.php?rgh-link-date=2022-04-25T05%3A56%3A36Z#L101=)? How it is done here: [https://github.com/LycheeOrg/Lychee/blob/master/app/Http/Controllers/ViewController.php#L68=](https://github.com/LycheeOrg/Lychee/blob/master/app/Http/Controllers/ViewController.php?rgh-link-date=2022-04-25T05%3A56%3A36Z#L68=)

That is the way to fix it indeed.

nagmat84:
> > That must be something I missed here... But couldn't you just add the `locale` variable to?
> 
> That is the way to fix it indeed.

OK. I am out of here. I never had a look at the Blade templates and I not going to start it, otherwise it might end with another refactoring. :grin:

If @qwerty287 fixed it, and @ildyria says it is the right way to do it, then I don't understand why @kamil4 want's to revert it. Did it throw an error? I will leave the discussion to you guys.

qwerty287:
Yes, it throw an error. @kamil4 reverted the change completely, while I proposed another fix with keeps support for localized content, but I only proposed it and didn't fix it myself.

kamil4:
I don't have any deep attachment to my fix; it was a minimal patch to get the view mode working again (it's completely broken right now). I'm perfectly fine with a better fix of course so if @qwerty287 wants to do it, I say go ahead and submit a new PR and we will close this one.

ildyria:
See https://github.com/LycheeOrg/Lychee/pull/1291

PR #1289, opened 2022-04-25
Fix rollback code for mysql

kamil4:
Fixes rollback code from #1278

The regular migration code special-cases mysql but the same is missing from the rollback code, resulting in one index not being deleted, so a subsequent forward migration fails.


nagmat84 (approved):
Looks good and tested on MySQL.

PR #1291, opened 2022-04-25
Add locale variable

qwerty287:
Replaces https://github.com/LycheeOrg/Lychee/pull/1288.

### DIFF
--- a/app/Http/Controllers/ViewController.php
+++ b/app/Http/Controllers/ViewController.php
@@ -1,9 +1,8 @@
 <?php
 
-/** @noinspection PhpUndefinedClassInspection */
-
 namespace App\Http\Controllers;
 
+use App\Facades\Lang;
 use App\Legacy\Legacy;
 use App\Models\Configs;
 use App\Models\Logs;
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/ViewController.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+/** @noinspection PhpUndefinedClassInspection */
```
nagmat84: I really don't like those lines. They suppress potential errors which are otherwise reported by a proper IDE. Please import the `Lang` facade via a proper `use`-statement.

nagmat84 (changes requested):
I haven't tested it. Personally, I don't use the view page of Lychee. So maybe @ildyria can test it. In only have a single CR.

kamil4:
> I haven't tested it. Personally, I don't use the view page of Lychee. So maybe @ildyria can test it. In only have a single CR.

`Share Photo` / `Direct Links` -- first link on top. Open preferably in a different browser session (one that is not authenticated with the server).

ildyria (approved):
LGTM.

nagmat84 (approved):
LGTM

qwerty287:
@ildyria thanks for fixing

PR #1292, opened 2022-04-27
Fix version of php-exif instead of using dev

ildyria:
Avoid using dev branches for php-exif and update composer.lock

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "laravel/framework": "^8.0",
         "livewire/livewire": "^2.7",
         "lychee-org/nestedset": "^6",
-        "lychee-org/php-exif": "dev-master",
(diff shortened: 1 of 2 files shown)


kamil4 (approved):
Yeah, I think it's cleaner this way.

I was actually confused about what `dev-master` would do; without ever verifying, I assumed that it would result in the latest commit always being pulled, but @nagmat84 set me straight that it's still locked to the commit specified in `composer.lock`. So specifying `dev-master` hardly simplified anything.

d7415 (approved):
Good idea. Untested.

As we fully control that repo, is it worth specifying the version exactly and changing it ourselves rather than the `^`, or do you want to leave that in case we miss one?

nagmat84 (approved):
LGTM

PR #1296, opened 2022-04-29
Update php-exif to 0.7.5

kamil4:
I updated just php-exif in the `composer.lock`, to keep it simple; do we actually have any policy on whether all the dependencies should be updated on every update?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "laravel/framework": "^8.0",
         "livewire/livewire": "^2.7",
         "lychee-org/nestedset": "^6",
-        "lychee-org/php-exif": "^0.7.4",
+        "lychee-org/php-exif": "^0.7.5",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -28,7 +28,7 @@
         "laravel/framework": "^8.0",
         "livewire/livewire": "^2.7",
         "lychee-org/nestedset": "^6",
-        "lychee-org/php-exif": "^0.7.4",
+        "lychee-org/php-exif": "^0.7.5",
```
nagmat84, line 31: I wonder if you should agree upon some kind of "dependency update strategy". For example, in this case I personally would not have updated `composer.json`, but only run `composer update` and generated a new `composer.lock`. But that is mostly matter of taste.

nagmat84 (approved):
LGTM

PR #1297, opened 2022-04-30
Use server's default timezone as default timezone

qwerty287:
Use the default timezone of the server instead of UTC as default timezone.


config/app.php
```diff
@@ -78,7 +78,7 @@
 	|
 	*/
 
-	'timezone' => env('TIMEZONE', 'UTC'),
+	'timezone' => env('TIMEZONE', date('T')),
```
nagmat84:
```suggestion
	'timezone' => env('TIMEZONE', date('e')),
```

We need the timezone identifier or a so-called level-3 timezone, e.g. `Europe/Paris`. A timezone abbreviation like "CEST" (level-2) or a simple time offset like "+02:00" does not suffice.

.env.example
```diff
@@ -29,7 +29,8 @@ DB_USERNAME=
 DB_PASSWORD=
 DB_LOG_SQL=false
 
-TIMEZONE=UTC
+# Application timezone. If not specified, the server's default timezone is used.
+# TIMEZONE=UTC
```
nagmat84:
```suggestion
# Application timezone. If not specified, the server's default timezone is used.
# Requires a named timezone identifier.
# See https://www.php.net/manual/en/timezones.php for the list of supported timezones,
# except for those listed below "Others".
# Don't use a timezone offset (like +01:00) or a timezone abbreviation (like CEST)
# TIMEZONE=Europe/Paris
```

nagmat84 (approved):
LGTM, but see proposed changes.

I don't see any drawback or problem in using the server's timezone as a default, but maybe @kamil4 or @ildyria. So I invited them for a review.

config/app.php
qwerty287: But wasn't the default before just `UTC` which is the level-2 variant?

.env.example
qwerty287: Should we use `Europe/Paris` as default timezone?

qwerty287: Actually that's just an example so it's not important

config/app.php
nagmat84: Yes and it produced a warning in the diagnostics to change that.

.env.example
```diff
@@ -29,7 +29,12 @@ DB_USERNAME=
 DB_PASSWORD=
 DB_LOG_SQL=false
 
-TIMEZONE=UTC
+# Application timezone. If not specified, the server's default timezone is used.
+# Requires a named timezone identifier.
+# See https://www.php.net/manual/en/timezones.php for the list of supported timezones,
+# except for those listed below "Others".
```
d7415:
```suggestion
# See https://www.php.net/manual/en/timezones.php for the list of supported timezones.
```

The list itself advises against these, so I don't see that we have to. Additionally, UTC remains a valid option.

kamil4 (approved):
I'm OK with that.

PR #1302, opened 2022-05-05
Remove Google Fonts

### LINKED ISSUE #1261: Remove Google Fonts usage
This is actually more a background question than a bug report or feature request. I don't really know any CSS and web styling, however, there are some lines importing Google Fonts APIs. I think, you should replace them with local files - take a look at your website:
```
Your server, your data, your rules. Never depend on someone else. Lychee is a self-hosted solution, so you are in the full control of your photos.
```
Sure, I'm in the full control of my photos, but connections to Google Fonts are still very bad and `Never depend on someone else` also isn't correct.

The lines are:
https://github.com/LycheeOrg/Lychee/blob/cb2a2262bfff56dee1a8468fbddc1bd143ea0ea0/simple_error_template.html#L11-L12
https://github.com/LycheeOrg/Lychee/blob/cb2a2262bfff56dee1a8468fbddc1bd143ea0ea0/public/dist/landing.css#L1
https://github.com/LycheeOrg/Lychee/blob/cb2a2262bfff56dee1a8468fbddc1bd143ea0ea0/config/secure-headers.php#L362 (more a side effect of others)
https://github.com/LycheeOrg/Lychee/blob/cb2a2262bfff56dee1a8468fbddc1bd143ea0ea0/config/secure-headers.php#L417 (more a side effect of others)
https://github.com/LycheeOrg/Lychee/blob/cb2a2262bfff56dee1a8468fbddc1bd143ea0ea0/public/dist/page.css#L1
https://github.com/LycheeOrg/Lychee/blob/cb2a2262bfff56dee1a8468fbddc1bd143ea0ea0/public/installer/assets/css/style.css#L8
https://github.com/LycheeOrg/Lychee/blob/cb2a2262bfff56dee1a8468fbddc1bd143ea0ea0/resources/views/error/error.blade.php#L10-L11
https://github.com/LycheeOrg/Lychee/blob/cb2a2262bfff56dee1a8468fbddc1bd143ea0ea0/resources/views/error/update.blade.php#L10-L11
https://github.com/LycheeOrg/Lychee-front/blob/feba79086476c7c522052b9b8d4c81b397a888c3/styles/landing/landing.scss#L2
https://github.com/LycheeOrg/Lychee-front/blob/feba79086476c7c522052b9b8d4c81b397a888c3/styles/page/page.scss#L2

qwerty287:
Serve fonts directly from local files.
Closes https://github.com/LycheeOrg/Lychee/issues/1261


ildyria (approved):

nagmat84:
Did you check the Google font license? Are we allowed to host a local copy? Or more specifically, distribute a copy which allows other people to host that copy locally.

qwerty287:
1. Google fonts has a built-in download option (but it uses a wrong format, I used a [third-party tool](https://github.com/majodev/google-webfonts-helper))
2. Roboto is under apache 2.0, they have a build step for web fonts [in their makefile](https://github.com/googlefonts/roboto/blob/main/Makefile)
3. Nunito is under open font license (https://github.com/googlefonts/nunito/blob/main/OFL.txt)

So I don't think there are any issues with it.

kamil4:
Agreed; it looks like we can.

I have a different questions though: _should we_? What I mean by that is: Is there a compelling reason to ship these particular fonts rather than relying on fonts that the users already have? Correct me if I'm wrong, but I believe they are used in only a couple of places (the landing page and the error pages?); in particular, they are not used in the main Lychee UI. So I think it makes sense for us to consider whether we should self-host these fonts (as this PR does) or whether we should instead replace their usage by standard fonts and not ship any custom fonts at all?

qwerty287:
I definitely agree on this, and I did the changes locally (but did not push yet). If we decide to remove them completely, I'll just push them.

nagmat84:
Completely agreed. Personally, I would prefer to not ship any fonts at all. But if removing a definite font and replacing it by some user-provided font breaked the layout and/or resulted in much work to fix that, I would be open to accept this solution. However, according to @kamil4 statement, it should not break anything. So I would even prefer to use the same font setting as we do for our main layout.

qwerty287:
The commit I pushed right now remove external fonts completely.

qwerty287:
@ildyria what's the state of this? Are you fine with merging it?

kamil4:
Under the circumstances, I think the best way forward for this PR is to revert some of the changes and go with local copies of Google Fonts, as the PR originally did. Nobody was opposed to that initial version; however, we decided to try to improve on it and then "better is the enemy of good" kicked in...

qwerty287:
@LycheeOrg/developers I added Roboto back (but only Roboto, not Nunito which was used on the error pages).

resources/views/landing.blade.php
```diff
@@ -13,7 +13,7 @@
 
     <div id="header" class="animate animate-down">
         <div id="logo">
-            <a href="{{ URL::to('/') }}"><h1>{{ $infos['title'] }}<span>{{ $infos['subtitle'] }}</span></h1></a>
+            <a href="{{ URL::to('/') }}"><h1>{{ $infos['title'] }}<span class="subtitle">{{ $infos['subtitle'] }}</span></h1></a>
```
kamil4: What is the reason for this change?

resources/views/landing.blade.php
```diff
@@ -23,7 +23,7 @@
     <div id="intro">
         <div id="intro_content">
             <h1 class="animate_slower pop-in">{{ $infos['title'] }}</h1>
-            <h2><span class="animate_slower pop-in">{{ $infos['subtitle'] }}</span></h2>
+            <h2 class="subtitle"><span class="animate_slower pop-in">{{ $infos['subtitle'] }}</span></h2>
```
kamil4: Same question...

kamil4 (approved):
Looks good overall; I'm just wondering what the changes in the landing blade do, since grepping our CSS style files for `subtitles` returns nothing relevant?

PR #1305, opened 2022-05-06
Fix coloring in scripts

qwerty287:
Must use `printf` instead of `echo`. Tested with bash and zsh.


scripts/install_files.sh
```diff
@@ -4,19 +4,19 @@ GREEN="\033[38;5;010m"
...
 echo "cp scripts/pre-commit .git/hooks/"
 cp scripts/pre-commit .git/hooks/
 echo "cp scripts/post-merge .git/hooks/"
 cp scripts/post-merge .git/hooks/
 
-echo "\n${ORANGE}To disable the call of composer and migration on pull add${NO_COLOR}"
-echo "${ORANGE}a file named '.NO_AUTO_COMPOSER_MIGRATE' in this directory.${NO_COLOR}\n"
\ No newline at end of file
+printf "\n${ORANGE}To disable the call of composer and migration on pull add${NO_COLOR}\n"
+printf "${ORANGE}a file named '.NO_AUTO_COMPOSER_MIGRATE' in this directory.${NO_COLOR}\n"
```
nagmat84:
```suggestion
printf "${ORANGE}a file named '.NO_AUTO_COMPOSER_MIGRATE' in this directory.${NO_COLOR}\n"

```

nagmat84 (approved):
Thanks! That makes a lot of sense. I was constantly bothered by seeing ANSI control sequences in my terminal, but I did not feel summoned to look into it.

PR #1308, opened 2022-05-07
Sync frontend for drag / drop

qwerty287:


nagmat84 (commented):
Nothing against it as it is just a sync with the frontend. You may merge it directly or wait until https://github.com/LycheeOrg/Lychee-front/pull/295 and https://github.com/LycheeOrg/Lychee-front/pull/294 also will have been merged into the frontend and then simply make a single commit to the backend which collectively syncs all recent changes of the frontend.

That's more a matter of taste and how fast you want to see the drag-and-drop feature being rolled out with Lychee. Personally, I would like to wait in order to keep the number of commits to the backend small which do not actually change the backend itself but only update the reference to the frontend. But as I said, that is a question of taste.

qwerty287:
The Google Fonts PR has it's own backend PR, so there is no need for it. I'm going to merge this to get some PRs done and will do this again after the mobile improvements has been discussed and merged.

PR #1309, opened 2022-05-07
Improve file handling

nagmat84:
### Intro and Final Goal

This is the first of a series of four planned PRs which ultimately clean up the way how the MIME type of media files is handled.

 1. Concentrate all file handling in a bunch of "file" classes
    **That is this PR.**
 2. Use data streams everywhere not file paths
    This PR is already in the pipeline, but not yet ready, see https://github.com/LycheeOrg/Lychee/tree/use_filestreams if curious
 3. Clean up of a lot of source code files and Lychee directory structure
    This does not change any functionality but it is/will be necessary.
 4. Make the live photo the eights size variant and store filesize in database
    This was discussed on Gitter recently.

All these PRs together will also eventually and "accidentally" enable the option to host media files on a remote location such as AWS S3. :tada: :confetti_ball: I decided to split the PRs into several ones to make reviews faster/easier.

### This PR

This first PR basically aims at two aspects

 1. Stop passing around file paths as strings
 2. Concentrate checks whether a file type is supported in one place

#### Aspect 1: Stop passing around file paths as strings

This is crucial in order to get rid of file paths eventually and make the transition to data streams possible. To this end, I introduced a couple of "file classes" (`FlysystemFile`, `NativeLocalFile`, etc.) which provide a unified interface for writing to and reading from files. The interface is inspired by Flysystem. Currently, the file paths sill remain to be used under the hood, because our image handlers, ffmpeg extractor, etc., need them, but this will change with the 2nd PR. This first step makes the rest of Lychee using proper file objects as method parameters.

#### Aspect 2: Concentrate checks whether a file type is supported in one place

Previously, there has been an unfortunate trait [`Actions\Photo\Extensions\Constants`](https://github.com/LycheeOrg/Lychee/blob/7d27f7da2dd72dc0258b69767d8d9a2ab4f272ae/app/Actions/Photo/Extensions/Constants.php) which collects some of the supported MIME types and file extensions. Some classes used this trait, but in different and inconsistent ways. These constants have been moved to a new central point [`Image\MediaFile.php`](https://github.com/LycheeOrg/Lychee/blob/f7c2c3ec1cac321c50de84fcb05113e20d50bb86/app/Image/MediaFile.php) which is the new parent class for all file classes. The trait `Constants` has been nuked.

Moreover, a long standing asymmetry has been resolved. Previously, there were constant arrays for supported EXIF image types (but not for videos) and for supported MIME types for videos (but not for images). Now there are arrays for both. This also changes the way how a file is recognized as raw or as a photo. Previously, we replaced the MIME type of files which had been identified as raw by the special keyword `'raw' ` upon upload. Then a file was recognized as a photo, if it was neither a video nor raw. Now, as we have a list of MIME types for videos _and_ for photos, videos and photos are identified as such, if there MIME type is included in the respective array. Everything else is raw. This means we don't use the special keyword `'raw'` anymore and don't loose the information about the real MIME type.

#### Addon: Type-safe EXIF extractor

The previous [`Extractor`](https://github.com/LycheeOrg/Lychee/blob/7d27f7da2dd72dc0258b69767d8d9a2ab4f272ae/app/Metadata/Extractor.php) used a [non-type-safe array](https://github.com/LycheeOrg/Lychee/blob/7d27f7da2dd72dc0258b69767d8d9a2ab4f272ae/app/Metadata/Extractor.php#L26) to transport EXIF data. This has been changed, too. The EXIF data are typed properties, now. Accidentally, this fixed a bug which seemingly nobody noticed so far. Matching the video and photo part of an Apple Live photos cannot have worked previously. The extractor used the attribute [`livePhotoContentID`](https://github.com/LycheeOrg/Lychee/blob/7d27f7da2dd72dc0258b69767d8d9a2ab4f272ae/app/Metadata/Extractor.php#L49), but we tried to save [`live_photo_content_id'`](https://github.com/LycheeOrg/Lychee/blob/7d27f7da2dd72dc0258b69767d8d9a2ab4f272ae/app/Actions/Photo/Strategies/AddBaseStrategy.php#L111) into the DB. This kind of bugs cannot happen anymore.


composer.json
```diff
@@ -19,6 +19,7 @@
 		"ext-fileinfo": "*",
         "ext-gd": "*",
         "ext-json": "*",
+		"ext-fileinfo": "*",
```
qwerty287: Isn't this already listed in line 19?
https://github.com/LycheeOrg/Lychee/blob/bd0fa1c00edd340c213302df09942f6f9a1fe699/composer.json#L19
https://github.com/LycheeOrg/Lychee/blob/bd0fa1c00edd340c213302df09942f6f9a1fe699/composer.json#L22

Also, it seems that indentation isn't the same in the lines above this (20 and 21).

app/Image/MediaFile.php
```diff
@@ -2,12 +2,13 @@
...
 
-use App\Exceptions\Internal\LycheeLogicException;
 use App\Exceptions\MediaFileOperationException;
+use App\Exceptions\MediaFileUnsupportedException;
+use App\Models\Configs;
 use Illuminate\Http\UploadedFile;
 
 /**
- * Interface MediaFile.
+ * Class MediaFile.
```
qwerty287: Is this line necessary?

qwerty287 (commented):
I'm relatively new to PHP, so don't expect this to be really helpful. It will take some time until I can really help with reviews.

composer.json
nagmat84: Yep, probably a failed merge/rebase.

app/Image/MediaFile.php
nagmat84: Yes and no. We need a short first line, because the first line of the PHPdoc comment is used as the title (during auto-completion by the IDE and if we will ever use it, in the generated PHP documentation). The following sentence is rather long:

```php
/**
 * Class MediaFile.
 *
 * This class abstracts from the differences of files which are provided
 * through a Flysystem adapter and files outside Flysystem.
 *
```

You definitely don't want the second sentence as a caption inside a table or similar.

But I agree that the caption should be a little bit more than just the repeated class name. I was simply too lazy to come up with another sentence which is longer than that but shorter than the following.

nagmat84: I rephrased it and put a little more information into the first line.

app/Actions/Photo/Strategies/AddStrategyParameters.php
```diff
@@ -2,23 +2,25 @@
...
+use App\Metadata\Extractor;
 use App\Models\Album;
 use App\Models\Configs;
 
 class AddStrategyParameters
 {
 	public ImportMode $importMode;
 
-	// Information about intended parent album
+	/** @var Album|null the intended parent album */
```
ildyria, line 13: Why the `/** @var Album|null` ?  Isn't the type in php enough ?

ildyria (commented):
Looks sane to me.
Need to test it.

app/Actions/Photo/Strategies/AddStrategyParameters.php
nagmat84: If the question is meant as "should it be theoretically sufficient to add the type hint to the variable in an optimal world?" Yes, I agree. :grinning: Giving the type as a type hint in the code and as an type annotation in the comment introduces unnecessary redundancy. 

Unfortunately, this is how PHPdoc expects it. I did not want to loose the original comment and attach it to the variable. So I changed to `//` into a `/**`. But in order to make the comment actually appear in the documentation, I had to add the `@var` keyword.

If one looks at https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/var.html#var how the syntax is defined, all three components of the `@var` tag - type, name, description - seem to be optional. But if you use the `@var` keyword and want to use the description you must add the type and the name, too. Otherwise, PHPdoc thinks that the first word of the description is the type and the second one the name.

Examples:

Does not work:

```php
/** intended parent album */
public ?Album parent
```

Makes PHpdoc falsely assume that "intended" is the actual type, "parent" the name and "album" the description. :roll_eyes:

```php
/** @var intended parent album */
public ?Album parent
```

Actually works, but occupies too many lines for my personal taste. This tiny, little variable gets more space in the source code than it deserves :grinning:

```php
/**
 * Intended parent album
 * 
  *@var 
  */
public ?Album parent
```

This works, too, is a succinct one-liner but at the costs of redundancy

```php
/** @var Album|null Intended parent album */
public ?Album parent
```

Hence, I decided to got with the last solution, especially as my IDE keeps the type hint and the type annotation in sync for me. :yum:

qwerty287 (approved):
I tested that some time ago already and it was working.

PR #1319, opened 2022-05-09
php artisan lychee:sync failing with Unresolvable dependency resolving bool $enableCLIFormatting

### LINKED ISSUE #1318: php artisan lychee:sync failing with Unresolvable dependency resolving bool $enableCLIFormatting
I am not probably the typical user that would do its album imports through the UI. I like to do it with the following command line:

```
php artisan lychee:sync path/to/folder
```

My installation is a direct git clone, which I regularly update. After the latest update I discovered this command stopped working with the following error message in the logs:

```
2022-05-09 18:59:07 -- error   -- Illuminate\Foundation\Console\Kernel::handle -- 37 -- Unresolvable dependency resolving [Parameter #1 [ <required> bool $enableCLIFormatting ]] in class App\Actions\Import\Exec; caused by
2022-05-09 18:59:07 -- error   -- Illuminate\Container\Container::unresolvablePrimitive -- 1104 -- Unresolvable dependency resolving [Parameter #1 [ <required> bool $enableCLIFormatting ]] in class App\Actions\Import\Exec
```

I have already probably found a fix that I will join with an upcoming PR, which is to define a default value to this parameter. But that is to be reviewed, as I do not know anything about Laravel.

mederel:
fixes #1318

Getting the following exception:

```
Unresolvable dependency resolving [Parameter #1 [ <required> bool $enableCLIFormatting ]] in class App\Actions\Import\Exec
```

when running:

```
php artisan lychee:sync myfolder
```

### DIFF
--- a/app/Console/Commands/Sync.php
+++ b/app/Console/Commands/Sync.php
@@ -55,13 +55,11 @@ class Sync extends Command
 	/**
 	 * Execute the console command.
 	 *
-	 * @param Exec $exec
-	 *
 	 * @return int
 	 *
 	 * @throws ExternalLycheeException
 	 */
-	public function handle(Exec $exec): int
+	public function handle(): int
 	{
 		try {
 			$directory = $this->argument('dir');


nagmat84:
I am not sure, if I like the change, because IMHO there is no sensible default whether CLI formatting should be enabled or not. If at all, then the default should be `false`, because the method is also used internally for HTTP request.

However, I am not sure, _why_ you get this error at all. I cannot reproduce it and there are only exactly two places places where the constructor is invoked.

 1. In `app/Console/Commands/Sync.php` at https://github.com/LycheeOrg/Lychee/blob/7a0c791d348e6f6b42123172d14e8cd354c43771/app/Console/Commands/Sync.php#L83-L92
 2. In `app/Actions/Import/FromServer.php` at https://github.com/LycheeOrg/Lychee/blob/7a0c791d348e6f6b42123172d14e8cd354c43771/app/Actions/Import/FromServer.php#L20

In both cases, the 2nd parameter is given. Are you sure that your working directory is clean and that you don't have a modified `app/Console/Commands/Sync.php`?

nagmat84:
Oh, sorry my fault. The true error is in this line

https://github.com/LycheeOrg/Lychee/blob/7a0c791d348e6f6b42123172d14e8cd354c43771/app/Console/Commands/Sync.php#L64

If you want, could you revert your previous commit and remove the parameter `Exec $exec` from the method `handle`? That is the actual reason why it fails.

The code

```php
 ​        ​/** 
 ​         * Execute the console command. 
 ​         * 
 ​         * @param Exec $exec 
 ​         * 
 ​         * @return int 
 ​         * 
 ​         * @throws ExternalLycheeException 
 ​         */ 
 ​        ​public​ ​function​ ​handle​(​Exec​ ​$​exec​): ​int
```

should be

```php
 ​        ​/** 
 ​         * Execute the console command. 
 ​         * 
 ​         * @return int 
 ​         * 
 ​         * @throws ExternalLycheeException 
 ​         */ 
 ​        ​public​ ​function​ ​handle​(​): ​int
```

mederel:
> Oh, sorry my fault. The true error is in this line
> 
> https://github.com/LycheeOrg/Lychee/blob/7a0c791d348e6f6b42123172d14e8cd354c43771/app/Console/Commands/Sync.php#L64
> 
> If you want, could you revert your previous commit and remove the parameter `Exec $exec` from the method `handle`? That is the actual reason why it fails.
> 
> The code
> 
> ```
>  ​        ​/** 
>  ​         * Execute the console command. 
>  ​         * 
>  ​         * @param Exec $exec 
>  ​         * 
>  ​         * @return int 
>  ​         * 
>  ​         * @throws ExternalLycheeException 
>  ​         */ 
>  ​        ​public​ ​function​ ​handle​(​Exec​ ​$​exec​): ​int
> ```
> 
> should be
> 
> ```
>  ​        ​/** 
>  ​         * Execute the console command. 
>  ​         * 
>  ​         * @return int 
>  ​         * 
>  ​         * @throws ExternalLycheeException 
>  ​         */ 
>  ​        ​public​ ​function​ ​handle​(​): ​int
> ```

Changes applied. Honestly thank you so much for your diligence!!!

app/Console/Commands/Sync.php
```diff
@@ -61,7 +61,7 @@ public function __construct()
 	 *
 	 * @throws ExternalLycheeException
 	 */
-	public function handle(Exec $exec): int
+	public function handle(): int
```
nagmat84, line 62: Please also remove the PHPdoc comment `@param ...` for the removed parameter.

nagmat84 (changes requested):

nagmat84:
No problem. Does it work now?

mederel:
> No problem. Does it work now?

I have to test tonight. Will update here.

app/Console/Commands/Sync.php
mederel: woops

nagmat84 (approved):

mederel:
> > No problem. Does it work now?
> 
> I have to test tonight. Will update here.

I could test it works well with that fix as well :-)

Issue #324, opened 2019-08-25
[Enhancement] LDAP authentication


jsaathof:
I have found Lychee just recently and migrated quickly to the new Laravel version which works just fine for me. Haven't found any issues.  I am really happy that I can use my PostgreSQL database instead of having to run an MySQL/MariaDB next to my PostgreSQL database. And I really like the feature of having multiple users.

However my users are in an LDAP database. I would really like to have LDAP authentication added to Lychee-Laravel to simplify user management and having just one password for users. I have found a project to add LDAP to a Laravel application called adldap2. See https://github.com/Adldap2/Adldap2-Laravel. I don't know if that can be used it Lychee-Laravel. I hope LDAP authentication makes it into the application.

Cheers.

Jurriaan

jsaathof:
I have found a successor of the project Adldap2 called ldaprecord. There is a lot of documentation available on the site. I can install the module but when I have to change the authentication I'm kind of lost.

https://ldaprecord.com/

ildyria:
Hummm.... 

I assume the following setting. You have:
- a LDAP server
- a Lychee install.

The question is more how do you determine the local ID for the used.
For the connection, you could basically add a line to check against a user in the LDAP server here:
https://github.com/LycheeOrg/Lychee/blob/376f8c7c9c463816252319dcbdab2975d221ed8a/app/ModelFunctions/SessionFunctions.php#L141

jsaathof:
Hi,

I do have an LDAP server, that is the main reason to request LDAP support. And I am running Lychee. But I’m not a PHP programmer. So I’m having a hard time implementing this myself. I will take a look at the code you mentioned.

--
Sincerely,

Jurriaan

julianfoad:
The [YunoHost Lychee integration](https://github.com/YunoHost-Apps/lychee_ynh) would become **most excellent** if it included LDAP integration. That Lychee-YunoHost integration is currently declared "YunoHost level 8" meaning "quite good" but its lack of integration with YunoHost's user account system is quite a drawback.

In fact, any self-hosting these days is hampered by having its own user account management. Single Sign-On (SSO) is the way to go, if we want to get ordinary people using our lovely self-hosted FOSS things. And LDAP support is a step towards SSO.

julianfoad:
Now the good news: apparently **Lychee LDAP integration already exists**: see [the Lychee-LDAP fork by 'hfr'](https://github.com/hfr/Lychee-LDAP).

[...]

PR #1322, opened 2022-05-12
Adding LDAP authentication

hfr:
LDAP authentication for Lychee

This implements the authentication for lychee using an LDAP server.

It is based on an expansion of the "SessionFunctions" class as suggested here: https://github.com/LycheeOrg/Lychee/issues/324#issuecomment-810106579

It simply automatically creates users in the lychee users table based on the data queried from the LDAP server when the user is logging in the first time in lychee. Obsolete users are not purged automatically. This needs to be done by some cron job accessing the database directly.

The LDAP 'uid' is mapped to the username and an additional column is created in the users table to store the common name (cn) of the user as 'fullname'.

The Admin Account is always stored in the lychee database.

After a normal installation one can switch to LDAP by setting up the following configuration in the settings:

ldap_enabled = 1 //to enable the LDAP login function
ldap_debug = 0 //enables LDAP log debug messages
ldap_server = <servername> //LDAP server name
ldap_port = 389  //LDAP server port
ldap_usertree = <spec>  //LDAP user tree specification e.g. "ou=People,dc=myorg,dc=com"
ldap_userfilter = <spec>  //LDAP user filter e.g. "(&(uid=%{user})(objectClass=posixAccount))"
ldap_version = 3  //LDAP protocol version
ldap_binddn = ''  //LDAP bind dn
ldap_bindpw = ''  //LDAP bind password
ldap_userkey = uid  //LDAP user key
ldap_userscope = sub  //LDAP user scope
ldap_starttls = 0  //LDAP use STARTTLS protocol
ldap_referrals = -1  //LDAP option referrals
ldap_deref = 0  //LDAP option deref

When LDAP authentication is enabled normal users cannot change their passwords in lychee anymore.

The implementation was developed and tested using an LDAP server running without STARTTLS on the local network, servicing about 20 users.


haivala:
would it be wise to add Oauth2 client too while you do this?

database/migrations/2022_05_11_154808_config_ldap_params.php
```diff
@@ -0,0 +1,187 @@
...
+	 * /**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		DB::table('configs')->insert([
+			[
+				'key' => 'ldap_enabled',
```
qwerty287, line 25: I'm not  sure if this fits better into the `.env` file. This is a pretty essential configuration, but I'm not really familiar with LDAP and don't know if this is fine as well :)

.gitignore
```diff
@@ -35,6 +35,7 @@ npm-debug.log
 yarn-error.log
 .env
 .env.*
+tmp/*
```
qwerty287: Is this related to LDAP?

hfr: Sorry, the tmp-dir is for my development stuff to keep it invisible from git. I can remove it.

database/migrations/2022_05_11_154808_config_ldap_params.php
hfr: It really does not matter where the configuration stuff is stored.
Since everything else like logs etc. is available from the web interface it might be convenient to have the LDAP-stuff there as well.

I would envision that in future one simply extracts the archives, sets up the web sever and everything else can be configured using the web interface.

hfr:
> would it be wise to add Oauth2 client too while you do this?

I personally see no need to secure my pictures using Oauth2. If I need more security I would add a client certificate to the web server.

I might getting Oauth2 up and running, but I cannot do the changes in the user interface. So if someone else can rewrite the login-dialog we can Oauth2 as well.

qwerty287 (commented):
> Also, I'd use a new PR for OAuth.

I completely agree with @querty287. If someone wants to add OAuth, then this should be in a different PR.

Personally, I would be even more happy, if this could wait until the unfortunate, home-brewed `SessionHandler` is gone. This is on my to-do list of task to clean up Lychee. This task will be a lot more easier if there is not too much other code depending on it. It might even be that adding OAuth becomes easier if the whole session handling logic is compatible with how Laravel does it normally.

But of course, if anybody want's to implement it, feel free to go. I don't want to block new features.

hfr:
> 1. The UI must be changed. Please update the `Lychee-Front` repo to hide the password change options for LDAP users.
> 
>     2. You updated a lot of dependencies. Did you test all the changes they introduced (if they contain any breaking)? Or was this an unintended change?
> 
> 
> Also, I'd use a new PR for OAuth.

LDAP itself does not introduce these changes. The problem is the different concept of username.

In an LDAP environment you would use the uid (user id) for login and the cn (common name) for display. I tried to use this concept for Lychee. So I left the numerical id of the user alone. This is how Lychee is managing everything by itself. I have add a new field 'fullname' to the user table which is filled by the cn delivered by the LDAP server. I just made sure that if there is no fullname present the backend will deliver the username instead.

The username is used for two purposes in the user interface: 1. displaying the owner of an album and when one sets up the sharing. I changed both places to deliver the fullname instead of the username if available.

The password change option can be easily suppressed by forcing LDAP users to be "locked". It is easy to do if we load the userdata.

It would be nice if the admin user interface could be changed to disable the feature to create new users in lychee which has at present no effect. A user which has no account on the LDAP server cannot login if the ldap_enabled option is set.

qwerty287:
Sorry, seems that I wasn't clear enough. You didn't update the [`Lychee-Front`](https://github.com/LycheeOrg/Lychee-Front) repository which holds the web frontend, this means that the options to change the user's password will still be there. You wrote that in the PR's description.

> It would be nice if the admin user interface could be changed to disable the feature to create new users in lychee which has at present no effect. A user which has no account on the LDAP server cannot login if the ldap_enabled option is set.

That's another reason why you have to update the frontend.

app/Actions/Sharing/ListShare.php
```diff
@@ -20,14 +21,16 @@ public function do(int $userId): Shares
 	{
 		try {
 			// prepare query
+			if (Configs::get_value('ldap_enabled', '0') != 0) {
+				$USERNAME = DB::raw("COALESCE(NULLIF(fullname,''), username) as username");
```
nagmat84: Why is this variable name written in capitals?

app/Http/Controllers/SessionController.php
```diff
@@ -83,7 +83,11 @@ public function init(): array
 					$user = User::query()->findorFail($user_id);
 					$return['status'] = Config::get('defines.status.LYCHEE_STATUS_LOGGEDIN');
 					$return['config'] = $this->configFunctions->public();
-					$return['is_locked'] = $user->is_locked;   // may user change their password?
+					if (Configs::get_value('ldap_enabled', '0')) {
+						$return['is_locked'] = true;
+					} else {
+						$return['is_locked'] = $user->is_locked;   // may user change their password?
+					}
```
nagmat84: Please don't do that here. Instead write a accessor function for the `is_locked` property of the `User` class. Otherwise code like this needs to be used everywhere in the code and this probably will be forgotten somewhere.

app/ModelFunctions/LDAPFunctions.php
```diff
@@ -0,0 +1,377 @@
...
+use App\Models\Configs;
+use App\Models\Logs;
+
+class LDAPFunctions
+{
+	/* @var resource $con holds the LDAP connection */
+	protected $con = null;
+
+	/* @var int $bound What type of connection does already exist? */
+	protected $bound = 0; // 0: anonymous, 1: user, 2: superuser
```
nagmat84:
```suggestion
	protected int $bound = 0; // 0: anonymous, 1: user, 2: superuser
```

Please use attributes with proper types. This remark is just an example and holds for all similar code lines.

app/ModelFunctions/LDAPFunctions.php
```diff
@@ -0,0 +1,377 @@
...
+	 * @param string     $base_dn
+	 * @param string     $filter
+	 * @param string     $scope           can be 'base', 'one' or 'sub'
+	 * @param array|null $attributes
+	 * @param int        $attrsonly
+	 * @param int        $sizelimit
+	 *
+	 * @return resource
+	 */
+	protected function _ldapsearch($link_identifier, $base_dn, $filter, $scope = 'sub', $attributes = null,
```
nagmat84: I haven't yet checked with my IDE and Linter, but a doubt that this method doesn't throw any exceptions. Please annotate all methods with proper `@throws` directives. This comment is an example and holds for all similar code lines.

app/ModelFunctions/LDAPFunctions.php
```diff
@@ -0,0 +1,377 @@
...
+						 $attrsonly = 0, $sizelimit = 0)
+	{
+		if (is_null($attributes)) {
+			$attributes = [];
+		}
+
+		if ($scope == 'base') {
+			return @ldap_read(
```
nagmat84: Please don't use the `@`-operator to silence engine errors. The `@`-operator is legacy and evil. For many functions the `@`-operator has already lost its function (e.g. see https://wiki.php.net/rfc/engine_warnings and https://php.watch/versions/8.0/fatal-error-suppression). As PHP is slowly phasing out this operator, its usage is not future proof. It might be that it won't work for ldap-functions is the next PHP release. Moreover, silencing errors makes tracking down errors only more difficult, because eventually the code will fail somewhere else and then it is only more difficult to find the real cause.

Lychee has only recently moved to use proper exceptions nearly everywhere and all remaining occurrences of the `@`-operator have been nuked. Please use exceptions.

This remark is an example and holds for all similar code lines.

app/ModelFunctions/LDAPFunctions.php
nagmat84: Please add proper type hints to the parameters and also to the return type (if possible). E.g.

```suggestion
	protected function _ldapsearch($link_identifier, string $base_dn, $filter, string $scope = 'sub', array $attributes = [],
```

Again, this is only an example.

app/ModelFunctions/LDAPFunctions.php
```diff
@@ -0,0 +1,377 @@
...
+	protected $user_info = [];
+
+	/**
+	 * Log debug message.
+	 *
+	 * @param string $method
+	 * @param string $line
+	 * @param string $text
+	 */
+	protected function _debug(string $method, string $line, string $text)
```
nagmat84: IMHO, this method should not be necessary. It should not be the responsibility of this class to decide if it logs certain messages or not. This means, the logic around `getConf('debug')` should be inside the logger.

app/ModelFunctions/LDAPFunctions.php
```diff
@@ -0,0 +1,377 @@
...
+
+			return $this->user_info[$user];
+		}
+		// force superuser bind if wanted and not bound as superuser yet
+		if ($this->getConf('binddn') && $this->getConf('bindpw') && $this->bound < 2) {
+			// use superuser credentials
+			if (!@ldap_bind($this->con, $this->getConf('binddn'), $this->getConf('bindpw'))) {
+				Logs::notice(__METHOD__, __LINE__, 'LDAP bind as superuser failed.');
+
+				return false;
```
nagmat84: Don't use `false` as a return type, if the claimed return type of the method is `array`. Use proper exceptions instead.

Moreover, if you throw a proper exception, the exception will become logged by our exception handler. There is no need to log it manually.

app/ModelFunctions/LDAPFunctions.php
```diff
@@ -0,0 +1,377 @@
...
+			}
+			$this->bound = 2;
+		}
+
+		$info = [];
+		$info['user'] = $user;
+		$info['server'] = $this->getConf('server');
```
nagmat84: Don't use associative array to pass around information or - in this case - as a return type. These kind of arrays are not type-safe and a real PITA, because it is difficult to find out which array keys exist and they are easily mistyped.

With a few remaining exceptions this kind of information transfer has been nuked from Lychee recently.

app/ModelFunctions/LDAPFunctions.php
```diff
@@ -0,0 +1,377 @@
...
+			);
+		} elseif (strpos($this->getConf('usertree'), '%{user}')) {
+			// direct user bind
+			$dn = $this->_makeFilter(
+				$this->getConf('usertree'),
+				['user' => $user, 'server' => $this->getConf('server')]
+			);
+		} else {
+			// Anonymous bind
+			if (!@ldap_bind($this->con, $this->getConf('binddn'), $this->getConf('bindpw'))) {
```
nagmat84: Please don't use magic string constants. Here, I am specifically referring to `'binddn'`  and `'bindpw'`. Introduce proper class constants, e.g. `const CONF_BIND_DN = 'ldap_binddn';` at the top of the class and then use those constants.

app/ModelFunctions/LDAPFunctions.php
```diff
@@ -0,0 +1,377 @@
...
+
+				return false;
+			}
+			$this->bound = 1;
+
+			return true;
+		}
+	}
+
+	protected function getConf(string $key, int|bool|string|null $default = null): int|bool|string|null
```
nagmat84: Is this method really necessary? I assume the method can be nuked, after your have introduced proper class constants for the configuration options.

app/Models/Configs.php
```diff
@@ -35,6 +35,7 @@ class Configs extends Model
 	use UseFixedQueryBuilder;
 
 	protected const INT = 'int';
+	protected const INTM = 'intm';
```
nagmat84: What is an "intm"?

app/Models/User.php
```diff
@@ -47,6 +47,7 @@ class User extends Authenticatable implements WebAuthnAuthenticatable
 		'username',
 		'password',
 		'email',
+		'fullname',
```
nagmat84: Please add the new property to the PHPdoc of the class, too.

Moreover, personally, I would like the denomination `display_name` better. This is also the name of an LDAP attribute and reflects more precisely the purpose of the attribute.

app/Models/User.php
```diff
@@ -108,4 +109,13 @@ public function name(): string
 	{
 		return ($this->id == 0) ? 'Admin' : $this->username;
 	}
+
+	public function fullname(): string
```
nagmat84: Accessor functions in Laravel use the somewhat acquired syntay `getFullnameAttribute`, see https://laravel.com/docs/8.x/eloquent-mutators#defining-an-accessor.

database/migrations/2022_05_11_201130_add_full_name_col.php
```diff
@@ -0,0 +1,35 @@
...
+
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		Schema::table(self::USERS, function (Blueprint $table) {
+			$table->string(self::NAME, 128)->default('')->after('password')->nullable();
```
nagmat84: If the column is nullable - which is fine - then the default should not be the empty string `''`. As we use `null` for "unknown value" or "not set", we should do that here, too.

database/migrations/2022_05_11_201130_add_full_name_col.php
```diff
@@ -0,0 +1,35 @@
...
+
+	/**
+	 * Reverse the migrations.
+	 *
+	 * @return void
+	 */
+	public function down()
+	{
+		Schema::table(self::USERS, function (Blueprint $table) {
+			$table->dropColumn(self::NAME);
```
nagmat84: Don't drop the column on downward migration. This breaks the foreign key constraints of the DB. This is due to a mix of a bug in Laravel and older versions of SQLite.

nagmat84 (changes requested):
Thank you for your interest in Lychee and this PR! LDAP authentication is a great feature.

I did a quick and sketchy review, so expect more to come. I haven't checked out the code and hence have neither tested it nor analyzed it with my IDE/Linter. I only commented on what immediate caught my eye during a quick scan.

Beside my initial comments on specific lines of code below, there are two more points I would like to address.

I would be really grateful, if you could add some automated tests for LDAP to our test suite. I am aware of the fact that we only have around 60% test coverage and hence it might appear somewhat odd to ask for tests from a first-time contributor. However, I have a particular reason. I am planning on replacing the whole, home-brewed `SessionHandler`, `SessionFunction`, etc., by their proper Laravel counterparts. Hopefully, I will do so within the next 6 month. This is point three on my to-do list. (Point one is fixing the EXIF/MIME type handling, point two is replacing the unfortunate web installer.) Our home-brewed session methods have some drawbacks which I want to get rid off. When I am going to do so, I naturally have to touch the whole LDAP logic, too. Hence, it would be wonderful to have some regression tests such that I won't accidentally break anything by then.

Actually, I would like to see the LDAP feature being merged after that, because merging it now will only make my job harder, but I don't want to block new feature, so this will be my problem.

Also, I noticed that you introduced a new styles for identifiers. I know that Lychee is inconsistent, too, and we use a wild mix of camel case (e.g. `thisIsAnExample`) and snake case (e.g. `this_is_an_example`). I tried to harmonize that once, but gave up, because Laravel actually forces us to use such a mix. (They mix it themselves and we must follow when we overwrite inherited methods, etc.) However, you added a new styles to the mix: a beginning underscore `_` to indicate non-public methods. That is not really a big deal and nothing which would prevent this PR from getting merged, nonetheless I am curious what the other maintainer think about it. In particular, as you are inconsistent within your own code. There are some non-public methods which don't start with an underscore.

haivala:
Of course there should be another PR for Oauth but the similarities with the LDAP are the thing I thought it would be easy to do as you have familiarized now with the login logic on Lychee side.

Side note: I think that all of the Oauth and LDAP settings should be ENV variables and default to be off.

Thank you for everything you have done already! and sorry for the suggestions in the middle of your work.

app/Models/Configs.php
hfr: I found out the hard way that someone before me came to the idea that an 'int' is by lychee definition is a positive integer number. Since I need a negative value in my configuration, I needed a definition of an integer which supports  negative numbers. I think that "intm" fits perfectly in relation to the definition of 'int' as a positive number. You might come up with something better. The best would be to define 'int' as full range integer and 'positive_int' as the positive only version.

ildyria: > I found out the hard way that someone before me came to the idea that an 'int' is by lychee definition is a positive integer number. Since I need a negative value in my configuration, I needed a definition of an integer which supports negative numbers. I think that "intm" fits perfectly in relation to the definition of 'int' as a positive number. You might come up with something better. The best would be to define 'int' as full range integer and 'positive_int' as the positive only version.

Sorryyyyyyy

app/Actions/Sharing/ListShare.php
hfr: There is no special reason. I frighted a lot to get the DB::raw statement working and that was the last version.

ildyria:
@hfr I added you as external to the project so that your PullRequest don't require approval all the time. :)

app/Models/Configs.php
nagmat84: > I found out the hard way that someone before me came to the idea that an 'int' is by lychee definition is a positive integer number.

Well, it seems that back then only positive integers were required. I believe, @ildyria just apologized for it. However, I believe that this has been OK in the sense that those things must always be evaluated on the background at that time.

> I think that "intm" fits perfectly in relation to the definition of 'int' as a positive number. You might come up with something better.

Sorry, but I disagree. "intm" ist just bad. I have just realized that it is supposed to mean "int minus", but I don't want to guess when I am reading code. If you need signed integers, it is now the time to fix that. So, please rename the first constant into `UNSIGNED_INTEGER` (or for my sake `UNSIGNED_INT` if you want it short) and `SIGNED_INTEGER`. That makes things clear and future developers won't need to guess.

And no, it is not up to us, to come up with something better. It is your PR.

app/Http/Controllers/SessionController.php
hfr: Done.

app/ModelFunctions/LDAPFunctions.php
hfr: The type of $con will be dependent of the definition of ldap_connect. So I think it is more robust to not fix the type.
But when the type is an internal choice then I agree that it should be fixed whenever feasable.

app/ModelFunctions/LDAPFunctions.php
hfr: I decided to use LDAPException only once if the connection to the LDAP server cannot be established.

The rest of the code would not fire Exceptions if the PHP LDAP functions do not fire them.
If the only consequence of omitting the @ operator now that you ask me to document  all the Exceptions which might be fired by the library functions in your code then I consider it a wast of time to remove the @ operator.

app/ModelFunctions/LDAPFunctions.php
hfr: I removed all @ operators.

app/Models/User.php
hfr: You can spend the rest of your life improving all the names. I left that game long ago.
fullname does not belong to the domain of LADP. There it is called cn (common name).

Lychee uses an numerical id, to internally separate users and as a 'display name', the user name is used.
This is a somewhat limited concept. 

I am fine with the numerical id for handling things internally. Then we need a login name which is called uid in the context of LDAP (at least most of the time). On top we need a nice display name,

So, now you can define names for this like: (id, username, fullname) or (id,uid,cn) or (user_id, user_name, display_name).

I leave it up to you...

database/migrations/2022_05_11_201130_add_full_name_col.php
hfr: It is the most robust definition for a column. I changed it.

app/ModelFunctions/LDAPFunctions.php
nagmat84: Another solution would be to wrap the whole function definition into a `try{ ... } catch(\ExceptionThrownByLibarary $e) {...}` block and then only throw your custom `LDAPException` _and_ attach the original exception as the predecessor exception.

database/migrations/2022_05_11_201130_add_full_name_col.php
hfr: I saw it but since dropping the column is in the migration of the email column as well, I thought it is the proper way to go.

So what should I put in instead?

hfr:
> Sorry, seems that I wasn't clear enough. You didn't update the [`Lychee-Front`](https://github.com/LycheeOrg/Lychee-Front) repository which holds the web frontend, this means that the options to change the user's password will still be there. You wrote that in the PR's description.
> 
> > It would be nice if the admin user interface could be changed to disable the feature to create new users in lychee which has at present no effect. A user which has no account on the LDAP server cannot login if the ldap_enabled option is set.
> 
> That's another reason why you have to update the frontend.

I do not care for the front-end development. Someone else need to do this. I think there is enough to do on the backend side

> Thank you for your interest in Lychee and this PR! LDAP authentication is a great feature.
> 
> I did a quick and sketchy review, so expect more to come. I haven't checked out the code and hence have neither tested it nor analyzed it with my IDE/Linter. I only commented on what immediate caught my eye during a quick scan.
> 
> Beside my initial comments on specific lines of code below, there are two more points I would like to address.
> 
> I would be really grateful, if you could add some automated tests for LDAP to our test suite. I am aware of the fact that we only have around 60% test coverage and hence it might appear somewhat odd to ask for tests from a first-time contributor. However, I have a particular reason. I am planning on replacing the whole, home-brewed `SessionHandler`, `SessionFunction`, etc., by their proper Laravel counterparts. Hopefully, I will do so within the next 6 month. This is point three on my to-do list. (Point one is fixing the EXIF/MIME type handling, point two is replacing the unfortunate web installer.) Our home-brewed session methods have some drawbacks which I want to get rid off. When I am going to do so, I naturally have to touch the whole LDAP logic, too. Hence, it would be wonderful to have some regression tests such that I won't accidentally break anything by then.
> 
> Actually, I would like to see the LDAP feature being merged after that, because merging it now will only make my job harder, but I don't want to block new feature, so this will be my problem.
> 
> Also, I noticed that you introduced a new styles for identifiers. I know that Lychee is inconsistent, too, and we use a wild mix of camel case (e.g. `thisIsAnExample`) and snake case (e.g. `this_is_an_example`). I tried to harmonize that once, but gave up, because Laravel actually forces us to use such a mix. (They mix it themselves and we must follow when we overwrite inherited methods, etc.) However, you added a new styles to the mix: a beginning underscore `_` to indicate non-public methods. That is not really a big deal and nothing which would prevent this PR from getting merged, nonetheless I am curious what the other maintainer think about it. In particular, as you are inconsistent within your own code. There are some non-public methods which don't start with an underscore.

I pushed a simple version of a unit test.

nagmat84:
> > Sorry, seems that I wasn't clear enough. You didn't update the [`Lychee-Front`](https://github.com/LycheeOrg/Lychee-Front) repository which holds the web frontend, 
> >
> > [...]
> > 
> > That's another reason why you have to update the frontend.
> 
> I do not care for the front-end development. Someone else need to do this. I think there is enough to do on the backend side

In defense of @qwerty287, you should watch you attitude. If you change things on the backend which effect the frontend, you will have to update the frontend as well. Period.

This is not a buffet where you can pick what you like, change some things and hope or even demand (!) that other people clean up the rest after you. If you don't care, we don't care for your PR.

For example, the only reason why I personally joined Lychee was that I wanted to develop a native Android client as a replacement for Google Photos. Unfortunately, it turned out that the whole JSON API was not well suited to be used with anything else than the current web frontend. So I had to refactor the API and thereby re-wrote a lot on the backend. Of course, I had to update the frontend as well, even though I don't like frontend development at all. But I did it.

The same is excepted of you.

database/migrations/2022_05_11_201130_add_full_name_col.php
nagmat84: Although it is not great, the only feasible option is to do nothing on the downward path and simply keep the column. In exchange, you have to add a test in the upward path whether the column already exists.

app/ModelFunctions/LDAPFunctions.php
hfr: This is all no solution, since the Exceptions fired by ldap_bind cannot be caught. So the only solution for not breaking the application is the use of the @-operator.

So the request to eliminate the @-operator in my code is simply nonsence. The @-operator is needed if you want a stable application. That is simply a fact. I figured this out long ago and I spend a lot of time verifying this fact with php 8.1.5 again only because you were not knowing what you are asking for.

So it was a waste of time trying to remove the @-operator. It is simply required until someone changes the PHP ldap-functions.

I would appreciate it if you can avoid such nonsense.requests in future.

app/ModelFunctions/LDAPFunctions.php
hfr: I put the @-operators in again, because they are required (see above).

app/Models/Configs.php
hfr: I changed the configuration entry to 'signed_int', someone else might change 'int' to 'unsigned_int' if he likes to do it.

app/Models/User.php
hfr: luchee is using a different syntax. (See app/Models/User.php)
I think the lychee syntax is better.

database/migrations/2022_05_11_201130_add_full_name_col.php
hfr: What you suggest is no solution at all, because it breaks the very concept of "migration".
What you suggest is simply "suppress the down migration" because your tools cannot handle them.

I am sorry, but this is nonsense. I tested the migrations with sqlite and they work. So we need to fix the workflow instead of breaking the concept of migration.

ildyria: > I am sorry, but this is nonsense. I tested the migrations with sqlite and they work. So we need to fix the workflow instead of breaking the concept of migration.

Yes they work, however they are silently dropping Foreign Key constraints in the case of SQL.
We discovered it at our own cost.

hfr:
> > > Sorry, seems that I wasn't clear enough. You didn't update the [`Lychee-Front`](https://github.com/LycheeOrg/Lychee-Front) repository which holds the web frontend,
> > > [...]
> > > That's another reason why you have to update the frontend.
> > 
> > 
> > I do not care for the front-end development. Someone else need to do this. I think there is enough to do on the backend side
> 
> In defense of @qwerty287, you should watch you attitude. If you change things on the backend which effect the frontend, you will have to update the frontend as well. Period.
> 
> This is not a buffet where you can pick what you like, change some things and hope or even demand (!) that other people clean up the rest after you. If you don't care, we don't care for your PR.
> 
> For example, the only reason why I personally joined Lychee was that I wanted to develop a native Android client as a replacement for Google Photos. Unfortunately, it turned out that the whole JSON API was not well suited to be used with anything else than the current web frontend. So I had to refactor the API and thereby re-wrote a lot on the backend. Of course, I had to update the frontend as well, even though I don't like frontend development at all. But I did it.
> 
> The same is excepted of you.

I am sorry, but I only do what I can do. Front-end development is not my cup of tea. fullstop.
I do not care if things look nice or not. I only care if they are working.

I am fine if you are not care about my PR. I needed to write most of this because I need to use LDAP for my sites. So I needed patches to get the LDAP-support in any time lychee will be updated. Most of your suggestions made sense, but me doing frontend programming does not.

So for me the line is, if lychee is usable by non-admin users. If it is I am fine. And this is fulfilled by my patches.

hfr:
Of course you need to drop Foreign Key constraints when you modify the
table structure of a sqlite database.
Sqlite cannot drop a column. At least the version installed on my machine
cannot do it.

So if you want to drop a column you need to recreate the table an drop the
old version.

So if you want to drop the fullname column from the users table, you need
to do the following in sqlite:

PRAGMA foreign_keys = OFF;
BEGIN TRANSACTION;
CREATE TABLE "users_backup" ("id" integer not null primary key
autoincrement, "created_at" datetime not null, "updated_at" datetime not
null, "username" varchar not null, "password" varchar, "email" varchar,
"may_upload" tinyint(1) not null default '0', "is_locked" tinyint(1) not
null default '0', "remember_token" varchar);
INSERT INTO "users_backup" SELECT
id,created_at,updated_at,username,password,email,may_upload,is_locked,remember_token
FROM "users";
DROP TABLE "users";
ALTER TABLE "users_backup" RENAME TO "users";
CREATE UNIQUE INDEX "users_username_unique" on "users" ("username");
CREATE UNIQUE INDEX "users_email_unique" on "users" ("email");
COMMIT;
PRAGMA foreign_keys = ON;

Since this is specific for sqlite it cannot be implemented in the
migration. It needs to be implemented in the workflow.
One solution ist to run "PRAGMA foreign_keys = OFF;" at the beginning of
the complete migration, check it with "PRAGMA foreign_key_check;"  and
switch it on with "PRAGMA foreign_keys = ON;" at the end.

The drop column handling for sqlite in the migration handling does all this
except switching off and on the "PRAGMA foreign_keys".
So just add it and everything will be fine.

hfr:
> Of course you need to drop Foreign Key constraints when you modify the table structure of a sqlite database. Sqlite cannot drop a column. At least the version installed on my machine cannot do it. So if you want to drop a column you need to recreate the table an drop the old version. So if you want to drop the fullname column from the users table, you need to do the following in sqlite: PRAGMA foreign_keys = OFF; BEGIN TRANSACTION; CREATE TABLE "users_backup" ("id" integer not null primary key autoincrement, "created_at" datetime not null, "updated_at" datetime not null, "username" varchar not null, "password" varchar, "email" varchar, "may_upload" tinyint(1) not null default '0', "is_locked" tinyint(1) not null default '0', "remember_token" varchar); INSERT INTO "users_backup" SELECT id,created_at,updated_at,username,password,email,may_upload,is_locked,remember_token FROM "users"; DROP TABLE "users"; ALTER TABLE "users_backup" RENAME TO "users"; CREATE UNIQUE INDEX "users_username_unique" on "users" ("username"); CREATE UNIQUE INDEX "users_email_unique" on "users" ("email"); COMMIT; PRAGMA foreign_keys = ON; Since this is specific for sqlite it cannot be implemented in the migration. It needs to be implemented in the workflow. One solution ist to run "PRAGMA foreign_keys = OFF;" at the beginning of the complete migration, check it with "PRAGMA foreign_key_check;" and switch it on with "PRAGMA foreign_keys = ON;" at the end. The drop column handling for sqlite in the migration handling does all this except switching off and on the "PRAGMA foreign_keys". So just add it and everything will be fine.

This can actually be done in the migration as a short fix:

                $dbc='database.connections.'.Config::get('database.default');
                $database = Config::get($dbc);
                if ($database['driver']=='sqlite') {
                    $fc=$database['foreign_key_constraints'];
                    DB::statement('PRAGMA foreign_keys = OFF') 
                }
		Schema::table(self::USERS, function (Blueprint $table) {
			$table->dropColumn(self::NAME);
		});
                if (($database['driver']=='sqlite') && ($fc!=0)) {
                    DB::statement('PRAGMA foreign_keys = ON')
                }

nagmat84:
> > Of course you need to drop Foreign Key constraints when you modify the table structure of a sqlite database.

That is not the point. The problem are not foreign key constraints from the table `users` to somewhere else, but foreign key constraints from somewhere else to the table `users`. We have them a lot. For example `base_albums(owner_id)` points to `users(id)`. Basically everything which has an owner.

The problem is the following. SQLite does not support dropping a column. Hence, the whole table is dropped and then re-created without the column. Thereby all _other_ foreign constraints from somewhere else to the table `users` are dropped, too, because the target table - in this case `users` - vanishes temporarily.

These _other_ foreign constraints are not re-created, because SQLite cannot add a foreign key constraint to an existing table - let's say `base_albums` - after its creation. SQLite only supports adding foreign key constraints as part of the `CREATE TABLE`-statement.

> So if you want to drop the fullname column from the users table, you need to do the following in sqlite:
> This can actually be done in the migration as a short fix:
>
> ```
> PRAGMA foreign_keys = OFF;
> ```

That is not a solution. It only disables the enforcement of a foreign key constraint. This is fine, if you are filling your tables with data and don't wand to take care of correct order. But it does not prevent that a key constraint is deleted. It is not that the foreign key constraint is being broken in the sense that it is temporarily not fulfilled, the whole constraint simply vanishes from the DB.

> What you suggest is no solution at all, because it breaks the very concept of "migration". What you suggest is simply "suppress the down migration" because your tools cannot handle them.

Unfortunately, it is not a good solution. We are aware of that fact. But it is not our tools which cannot handle it. It is just a limitation of SQLite and actually a bug in Laravel. It is even documented on the official SQLite homepage, see here: https://www.sqlite.org/lang_altertable.html#otheralter They even give a work-around for the problem in 12 steps which involves renaming the table and the give a warning (below "Caution") about the correct order of operations. So, if you want to implement a downward migration, then do it like that.

The bug in Laravel is not having implemented that official work-around, but drop the column and silently ignore any errors.

> I am sorry, but this is nonsense. I tested the migrations with sqlite and they work.

It does so in the sense that there is no obvious error message, because Laravel is great in suppressing error messages and just carry on silently (see above). However, it leaves the DB in a dangerous and broken state.

app/ModelFunctions/LDAPFunctions.php
nagmat84: > This is all no solution, since the Exceptions fired by ldap_bind cannot be caught. So the only solution for not breaking the application is the use of the @-operator.

Why should it be impossible to catch exceptions thrown by `ldap_bind` and related functions? As far as I know the PHP language does not support a feature like "uncaughtable" exceptions. If you really have encountered such an exception, this would imply a bug in the PHP engine and I highly doubt that.

A MWE which illustrates this problem would be highly welcome.

> So the request to eliminate the @-operator in my code is simply nonsence. The @-operator is needed if you want a stable application. That is simply a fact. I figured this out long ago and I spend a lot of time verifying this fact with php 8.1.5 again only because you were not knowing what you are asking for.

The links to the RFC for PHP and the release notes about the `@`-operator are included in the first comment. The `@`-operator is legacy and it is just a question of time when it will be removed entirely from PHP. We have removed the `@`-operator from Lychee and it works perfectly. Unless you can illustrate your problem, I still need to ask you to change it.

hfr: Look, if you think there are no "uncaughtable" exceptions in php then that is fine with me. Everybody can have its believes. For us who want stable code, do the testing and put in the @-operator.

I do not care if it a "bug" in PHP or a bug in the PHP ldap library. I put in the @-operator and try to solve other problems.

It is actually not a bug. It is a feature. If you cannot get your ldap_bind, the script does not run. It is very secure. It is just not the right way for modern web applications.

Its very simple to demonstrate. Just remove all the @-operators from my code and run the unit test I have written.
Now you will see ErrorExceptions. But how can that happen? There is a catch statement around all the library calls.

The next thing you should do is using mother google and there you will find comments from years ago reporting this feature. And the solution is always: use the @-operator.

Just because someone writes "the @-operator is legacy" you should not believe it. It is legacy if it is not needed any more. ldap_bind needs a @-operator if you do not want you application to crash if someone types in a wrong password.

There is no point in proving facts.

nagmat84: > Its very simple to demonstrate. Just remove all the @-operators from my code and run the unit test I have written. Now you will see ErrorExceptions. But how can that happen? There is a catch statement around all the library calls.

If you refer to this kind of `catch`-statement

https://github.com/hfr/Lychee/blob/6d8fff95bc4a88be7bd422c6e971e77f21e8d55a/app/ModelFunctions/LDAPFunctions.php#L78

it simply does not work, because you try to catch the wrong type of exception. `\ErrorException` does not inherit `\Exception`, because `\Exception` is not the parent class of all exceptions in PHP. This is `\Throwable`.

The top-level inheritance chain in PHP looks like this

                                        +--- \RuntimeException
                   +---  \Exception  ---+
                   |                    +--- \LogicException
    \Throwable  ---+
                   +---  \ErrorException

So in order to catch `\ErrorException` and `\Exception` within the same `catch`-statement you must use `\Throwable`

> Look, if you think there are no "uncaughtable" exceptions in php then that is fine with me. Everybody can have its believes.

"My believes" are substantiated by original, first-hand documentation which I posted you at the very beginning. 

> The next thing you should do is using mother google and there you will find comments from years ago reporting this feature. And the solution is always: use the @-operator.

The problem with mother Google and (I assume) StackOverflow is that they are usually a very bad source compared to original literature. They are just not authoritative. You will find a "proof" for everything on Google. Especially, in the field of programming language, if something has been posted years ago, i.e. in the "dark ages" of PHP, it probably won't be correct today.
 
> Just because someone writes "the @-operator is legacy" you should not believe it.

Except that this someone is a maintainer of the PHP core development team. I think this is amazing: You are willing to accept answers from "mother Google" without questioning, but a first-hand reference from the php.net site written by a PHP maintainer is "someone one should not believe".

> There is no point in proving facts.

Of course it is, because a fact isn't a fact until you have proven it or provided evidence for it. The reason why I asked you for a MWE was that I expected you would use `catch (\Exception $e)`.

So once again and for the last time: change your code.

hfr: > > Its very simple to demonstrate. Just remove all the @-operators from my code and run the unit test I have written. Now you will see ErrorExceptions. But how can that happen? There is a catch statement around all the library calls.
> 
> If you refer to this kind of `catch`-statement
> 
> https://github.com/hfr/Lychee/blob/6d8fff95bc4a88be7bd422c6e971e77f21e8d55a/app/ModelFunctions/LDAPFunctions.php#L78
> 
> it simply does not work, because you try to catch the wrong type of exception. `\ErrorException` does not inherit `\Exception`, because `\Exception` is not the parent class of all exceptions in PHP. This is `\Throwable`.
> 
> The top-level inheritance chain in PHP looks like this
> 
> ```
>                                     +--- \RuntimeException
>                +---  \Exception  ---+
>                |                    +--- \LogicException
> \Throwable  ---+
>                +---  \ErrorException
> ```
> 
> So in order to catch `\ErrorException` and `\Exception` within the same `catch`-statement you must use `\Throwable`
> 
> > Look, if you think there are no "uncaughtable" exceptions in php then that is fine with me. Everybody can have its believes.
> 
> "My believes" are substantiated by original, first-hand documentation which I posted you at the very beginning.
> 
> > The next thing you should do is using mother google and there you will find comments from years ago reporting this feature. And the solution is always: use the @-operator.
> 
> The problem with mother Google and (I assume) StackOverflow is that they are usually a very bad source compared to original literature. They are just not authoritative. You will find a "proof" for everything on Google. Especially, in the field of programming language, if something has been posted years ago, i.e. in the "dark ages" of PHP, it probably won't be correct today.
> 
> > Just because someone writes "the @-operator is legacy" you should not believe it.
> 
> Except that this someone is a maintainer of the PHP core development team. I think this is amazing: You are willing to accept answers from "mother Google" without questioning, but a first-hand reference from the php.net site written by a PHP maintainer is "someone one should not believe".
> 
> > There is no point in proving facts.
> 
> Of course it is, because a fact isn't a fact until you have proven it or provided evidence for it. The reason why I asked you for a MWE was that I expected you would use `catch (\Exception $e)`.
> 
> So once again and for the last time: change your code.

Of course I am not referring to that catch-statement.

You are obviously very stubben.

There is no point in trying to prove facts to stubben people. It simply does not work.

You have the strange idea that the ErrorExeptions created by ldap_bind can be caught. The fact is that they can't. Do not ask me why not. I simply do not care.

The information about the class inheritance was simply wrong is you can read here: https://www.php.net/manual/en/class.exception.php

ErrorException is an Exception from PHP 7.2 on.

The only thing which works is to set up the good old error handler. But for me this is even more ugly as the @-operator.

But I have implemented it to prove my assumption that it might work. It does, its a fact.

So my suggestion is that we stay with the @-operator as long as it is around and see how error handling might have evolved when it is not supported any more.

By the way, I am not willing to accept any answer without testing them. For me only facts count and it is a fact, if it runs on my machine...

nagmat84: > > If you refer to this kind of catch-statement [...]
>
> Of course I am not referring to that catch-statement. You are obviously very stubben. There is no point in trying to prove facts to stubben people. It simply does not work.

Until know you haven't specifically referred to any catch clause. Hence I wrote "if you refer ...". Also I did a typo, I didn't meant `ErrorException`, but simply `Error` this was my fault. But the rest is correct. See here https://www.php.net/manual/de/class.throwable.php and here https://www.php.net/manual/de/class.error.php.

Anyway, this is a last fair chance:

You are constantly insulting people and lecturing them. This is nothing, I would like to see any further. Either you watch your tone and do as you have been asked or this PR is not going to be merged. Any further misstep and this PR is going to be closed. As an ultima ratio we can also block you from this site.

hfr: You simply do not understand. I do not care if you accept my PR or not.

If have my installations up and running and if you do not want this feature then that is fine with me. Since it is in GIT I can merge it in any time, lychee gets updated.

Trying to get the @-operator out of the code was completely waste of time, just because you did not do the tests in practice.

This is something I cannot change. I can write reliable code and I can find problems, but manage you attitude to facts is simply out of my scope.

Since the problem is completely on your side, I do not have any understanding for the following sentence: 

"Either you watch your tone and do as you have been asked or this PR is not going to be merged."

Isn't it about getting useful and reliable features implemented?

I cannot change that you got it wrong. Since it now gets all obvious you get nasty. Why?

It would be simply better, if you try to avoid nonsense requirements.

By the way if you really a maintainer of the PHP core development team as you have suggested then you should take this as a chance for an improvement. The good old error handler can handle the ldap_bind Exceptions, but you cannot catch them. There must be a reason for this somewhere. If you find it, you can improve the world at least a bit...

hfr:
> > > Of course you need to drop Foreign Key constraints when you modify the table structure of a sqlite database.
> 
> That is not the point. The problem are not foreign key constraints from the table `users` to somewhere else, but foreign key constraints from somewhere else to the table `users`. We have them a lot. For example `base_albums(owner_id)` points to `users(id)`. Basically everything which has an owner.
> 
> The problem is the following. SQLite does not support dropping a column. Hence, the whole table is dropped and then re-created without the column. Thereby all _other_ foreign constraints from somewhere else to the table `users` are dropped, too, because the target table - in this case `users` - vanishes temporarily.
> 
> These _other_ foreign constraints are not re-created, because SQLite cannot add a foreign key constraint to an existing table - let's say `base_albums` - after its creation. SQLite only supports adding foreign key constraints as part of the `CREATE TABLE`-statement.
> 
> > So if you want to drop the fullname column from the users table, you need to do the following in sqlite:
> > This can actually be done in the migration as a short fix:
> > ```
> > PRAGMA foreign_keys = OFF;
> > ```
> 
> That is not a solution. It only disables the enforcement of a foreign key constraint. This is fine, if you are filling your tables with data and don't wand to take care of correct order. But it does not prevent that a key constraint is deleted. It is not that the foreign key constraint is being broken in the sense that it is temporarily not fulfilled, the whole constraint simply vanishes from the DB.
> 
> > What you suggest is no solution at all, because it breaks the very concept of "migration". What you suggest is simply "suppress the down migration" because your tools cannot handle them.
> 
> Unfortunately, it is not a good solution. We are aware of that fact. But it is not our tools which cannot handle it. It is just a limitation of SQLite and actually a bug in Laravel. It is even documented on the official SQLite homepage, see here: https://www.sqlite.org/lang_altertable.html#otheralter They even give a work-around for the problem in 12 steps which involves renaming the table and the give a warning (below "Caution") about the correct order of operations. So, if you want to implement a downward migration, then do it like that.
> 
> The bug in Laravel is not having implemented that official work-around, but drop the column and silently ignore any errors.
> 
> > I am sorry, but this is nonsense. I tested the migrations with sqlite and they work.
> 
> It does so in the sense that there is no obvious error message, because Laravel is great in suppressing error messages and just carry on silently (see above). However, it leaves the DB in a dangerous and broken state.

I do not care about sqlite, since I do not use it. People using sqlite should know these kind of limitations and should use migrations with care. One should not go backward on a production server or one should use a different database.

I found in the docs that sqlite has implemented ALTER TABLE DROP COLUMN. So this should be a temporary problem. There is no point in bothering about it. Just write a warning that migrating backwards with sqlite will erase the foreign key constrains.

ildyria:
@hfr
I did some refactoring, can you please double check & see what is going wrong in the test with php8.0 which is working fine with 8.1

hfr:
> @hfr I did some refactoring, can you please double check & see what is going wrong in the test with php8.0 which is working fine with 8.1

I only run PHP 8.1.5

nagmat84:
@hfr Please revert your last commit. If I am changing things to meet the code quality of Lychee it is not up to you to undo that changes. I am still working on it and I have absolute no patience to do things twice.

hfr:
[MODERATED]

app/LDAP/LDAPFunctions.php
```diff
@@ -0,0 +1,464 @@
...
+		// Either we intent to always call `check_pass` and `get_user_data`
+		// in that order.
+		// In that case `get_user_data` should not be a public function but
+		// the result of check user data should be returned as part of
+		// `check_pass`.
+		//
+		// Or, if we intent to use `get_user_data` stand-alone, then this
+		// must be fixed in a way such that it works independent of a
+		// previous call to `check_pass`.
+		$result = $this->LDAP_search($base, $filter, Configs::get_value(self::CONFIG_KEY_USER_SCOPE));
```
nagmat84: While I have been working on it, I had to understand the code and found this error in the logic of the program flow. This is your to-do. Please take care of it.

nagmat84 (changes requested):
I have fixed most of the code and brought the Linter errors down. In this context I stumbled about an problem which must still be solved (see below).

**Before:**

Screenshot_20220515_130336
 
**After:**

Screenshot_20220515_164734

hfr:
> I have fixed most of the code and brought the Linter errors down. In this context I stumbled about an problem which must still be solved (see below).
> 
> **Before:**
> 
> Screenshot_20220515_130336
> 
> **After:**
> 
> Screenshot_20220515_164734

I hope that you agree that the code is ready to be merged even by you standard.

Let me comment a moment about the way things got handled. 

Obviously we are very different characters. You want the code nice and beautiful and I do not care how it looks. I only care how it works. So for me design and coding decision are not a matter of style. It is good, when it flies and it isn't if it doesn't.

So my goal was to get it running and when it is running, keep it that way. You wanted all the nice and beautiful language constructs be used. This makes backporting a pain in the but, but you probably do not care about a wide user base.

One of the key design decisions of the original approach was to combine ldap_open with ldap_bind. In this way there is either a binding or there is not and therefore one only needs to worry if the binding level is ot high enough. This was "bound" for. Since you never need an open without bind, it is a very practical decision minimizing the number of ldap requests. This is important if you mange a large user base with different services.

I do not know, why you thought combining ldap_open with ldap_bind and honestly I really don't care. You did and in this way you created the problem which you could not easily solve. I think, the solution was easy, but I am obviously the expert here, when it comes to LDAP. But it is not difficult to be "an expert of ldap". That is no rocket science. But it is a pain of the but if it doesn't work.

By the way you added another difficult to find bug by keep the setting of bound to zero in open_LADP. This is ok, if open_LADP is already the binding as it was the case in my original approach, but it is wrong, when open_LADP is not call bind.

Unfortunately you kicked my test code out, which actually was able to expose all the problem. One needs to instrument open_LDAP, LDAP_seach and LDAP_bind to properly test the LDAP unit. I added the instrumentation functions clean and untouched. I also added some grep search for the code as a test module to detect if anybody uses the test_LDAP_* functions outside LDAPTest. I think that is enough protection and I hope you leave the stuff in place.

I actually tested the binding logic on a LDAP server with anonymous access and it passed all tests.

Just one last comment. The configuration entry ldap_server allows actually for a list of servers. This is nice to have in an environment where you have multiple servers for redundancy. open_LDAP walks to the list of servers and takes the first one which is available. The downside is that you cannot rely on the config entry to be the "server". open_LDAP needs therefore to set ldap_server. In most cases this will be simply the configuration entry.

One last comment:

I do not think that the \safe\ldap_bind() version of the code looks any nicer than @ldap_bind() but this is a matter of taste. As long as one puts enough backslashes in the code it will actually work, which is nice. If you have time you can explain the purpose of the backslashes to me. I simply do not understand the concept. But that is my limitation not yours.

I can only hope that the code is making it to the master branch so that I can update my running sites with it.

ildyria:
> I do not think that the \safe\ldap_bind() version of the code looks any nicer than @ldap_bind() but this is a matter of taste. As long as one puts enough backslashes in the code it will actually work, which is nice. If you have time you can explain the purpose of the backslashes to me. I simply do not understand the concept. But that is my limitation not yours.

It makes use of https://github.com/thecodingmachine/safe instead of the native php functions.

ildyria:
> You wanted all the nice and beautiful language constructs be used. This makes backporting a pain in the but, but you probably do not care about a wide user base.

Backporting is not an option we consider. Because if a user want to use Lychee with and outdated version of Php (e.g. 7) it is not our problem. Our users have to update their system to keep up to date with security, there is no point in staying behind.
We do not have a strong requirement in that regard: just stay in the green of that page ( https://www.php.net/supported-versions.php ) and you are fine.

There are other company (e.g. Spatie) which literally required their user base to switch to PHP 8 the day it was released. We are not that extreme.

The other reason why we have such strict requirement in term of code quality is because we want to be sure that the project does not becomes a huge mess of Spaghetti code and anti-patterns. By providing coding uniformity it also makes it easier for external user to interact with our code base.

Php 7 and 8 introduced types, return types, and union types. Those ensure stronger data path and less errors. It also allows linting to make sure that there are no sneaky bug inserted. Lychee has a powerful exception handler which makes it easier to track down when something fails, better make use of it.

ildyria:
I will be very Dutch for once.

1. **It is bad practice to have test code in production code. It has nothing to do and should not be placed there.  
    THIS IS NON-NEGOTIABLE.**
2. It seems to me that you do not grasp clearly how inheritance works. For this reason you consider it as "increased complexity". If you are not happy with the modern and sane architecture of Lychee, then please be my guest and use the [version 3](https://github.com/LycheeOrg/Lychee-v3). Don't expect any support from us and good luck with it.
3. > This code has become buggy as hell and it will never get to work, if you try to make things elegant instead of operational.

    If this is the case, I am sorry but then it means that the code was no-good from the beginning.
4. > At present the code is NOT validating the user correctly.

    If that is the case, then it is not a problem of using a wrapper test class to access `protected` methods (which is literally what you were doing by adding `public function test_...`)
5. Unit tests are passing, which means that the code using inheritance is correct and there are no change in the actual behavior.

hfr:
Yes, when you merge the LDAP code in the master branch you can move the
test code out.

But the LDAP driver was buggy like hell because of your "code quality"
improvements. Until I have sorted all bugs out and get it full functional
you need to keep my instrumentation in place. It is my fork. This is not
negotiable, too.

For me it is for operational quality first.

Look, do not try to teach me about concepts when it is about getting the
code reliable.

Yes, you can instrument protected methods be inheritance, but then they are
in a different file again and that makes my life more difficult.
In the class you can even instrument private methods using my way. At the
end you shift everything out, make the methods protected you really need
and you are done with it.
But don't do it in the middle of the game.

Look, a driver which does what it should is good, regardless how it is
coded. A driver where the code looks nice, but lets people login which have
no account is bad.

The key issue of the changes is the following:

The changes introduced a lot of text, which looks dramatically complex, but
simply uses the wrong concepts to get the logic clear.

The key function is ldap_bind(). It is the login function to the ldap
server. The return value of this function is super important. It tells you
if someone is allowed to login to lychee or not.

My tests showed that the nice \Safe\ in front of the ldap_bind() functions
changes the return value. So if you use \Safe\ the code can never work.

So ldap_bind() has a boolean return value and you need to keep it and you
need to return it.

There is no point in converting the boolean return value in an Exception
and in the calling function convert the exception back to a boolean
information. That makes the logic very difficult to grasp and leads to the
following code:

try {
    if (!ldap_bind(...)) { throw new Exception(); }
    return;
} catch (Exception $e) {
    throw new Exception($e);
}

Looks like super nonsense and it is super nonsense.

Obviously you do not have sufficient knowledge about the problem of test
coverage. That the code is passing the tests does not mean anything, if the
coverage is not good enough. Which it was not.

It had the following problem. I am using a publicly available LDAP server
which I have limited knowledge about. I needed a user which did not have an
account. Since the accounts where about scientists I chose the name test.
Everything was fine. I could not login with the user name test.

BUT, then I wrote the new function get_user_list() and when I listed the
accounts returned by the test server the user name "test" jumped in my eye.
This user was known, just the password was different and my tests did not
cover an unknown user at all. An unknown user could simply log in.

My original code did not had this problem.

I think it was necessary to fix this. In the middle of my work you started
to polish my test code.

Why do you not invest time in thinking about better tests? We need 100 %
path coverage for the test. How can we instrument the code to achieve this?

I did not write that using inheritance for the test instrumentation is
wrong or incorrect. It was just an unlucky point in time. Just keep the
patches and apply them when we get the stuff operational. That is all what
is needed.

I have a lot more test scenarios running, but there is no point in pushing
them because they rely on my servers. So if you start changing these things
in the middle of the game I have a lot of nonsense to do.

By the way, do you know what this \Safe\ is actually doing? I did not find
much docs about it.


Am Di., 17. Mai 2022 um 14:26 Uhr schrieb Benoît Viguier <
***@***.***>:

ildyria:
I did not know you were not done with the testing. You clearly said:
> **I hope that you agree that the code is ready to be merged** even by you[r] standard.

Which I understood as "_I am done, things are fixed and running properly_".

>  By the way, do you know what this \Safe\ is actually doing? I did not find
much docs about it.

For me those are pretty self explanatory. Upon error, throw an exception instead of returning false.
https://github.com/thecodingmachine/safe#the-problem
https://thecodingmachine.io/introducing-safe-php

hfr:
I am sorry, but I discovered the major problem when I implemented the
get_user_list() function.

Actually I did not expect this, but then I figured that the code quality
completely destroyed the logic.

Am Di., 17. Mai 2022 um 23:32 Uhr schrieb Benoît Viguier <
***@***.***>:

app/LDAP/LDAPActions.php
```diff
@@ -0,0 +1,80 @@
...
+	public function __construct()
+	{
+	}
+
+	/**
+	 * Create a user in the database if it does not exist already.
+	 */
+	public static function create_user_not_exist(string $username, LDAPUserData $userData): void
+	{
+		$user = User::query()->where('username', '=', $username)->first();
```
ildyria, line 22: use `->count()` instead.

`->first()` hydrates a `User` model which adds slow downs and increase the memory usage. This is especially true given that there is no use of `$user` after.

app/LDAP/LDAPActions.php
```diff
@@ -0,0 +1,80 @@
...
+				}
+			}
+		}
+	}
+
+	public static function update_all_users(?bool $purge = null): void
+	{
+		$ldap = new LDAPFunctions();
+		if (is_null($purge)) {
+			$purge = Configs::get_Value('ldap_purge', '0') == '1';
```
ildyria: Instead of having a is_null check, you can directly use:
```suggestion
			$purge = $purge ?? (Configs::get_Value('ldap_purge', '0') == '1');
```

database/migrations/2021_12_04_181200_refactor_models.php
```diff
@@ -475,7 +489,7 @@ private function createUsersTableUp(): void
 			$table->dateTime('updated_at', 6)->nullable(false);
 			$table->string('username', 100)->nullable(false)->unique();
 			$table->string('password', 100)->nullable(true);
-			$table->string('email', 100)->nullable()->unique();
+			$table->string('email', 100)->nullable();
```
qwerty287: You can't change an old migration without adding a new one. It won't run again, so this won't work with DBs that already applied the migration. (Correct me if I'm wrong, but I don't think Laravel will check if any migration file has changed and run this migration again.) Just move this to a different migration, if it's necessary.

qwerty287: There's another line in this file and a comment. Remove both.

hfr:
After days of fighting with the testing in the workflow I finally found a way to pass it.

The problem was obviously some kind of race condition with some processes in the test environment. The LDAP tests were started before the environment was setup completely. As a result the ldap test servers were not accessible.

The solution was a connection retry loop at the beginning of the tests. As a consequence the tests are delayed until the test server are reachable and then all tests could be performed.

Although in most cases it won't be necessary for a production system, the connection retry code is part of the LDAPFunctions class, with a retry count set to zero which means no retry by default.

From my point of view the code is ready for production after someone could review the line escapes for the test coverage which were necessary for the lines which cannot be tested. 

I changed the Configs class to integrate a function which returns a proper cron specification based on the integer value in the configs. While I was doing it I corrected the weakness that get_value was not setting the type of the return value correctly even if the type is known by the specification of the configs entry. Now the return type of an integer is int and of a boolean is bool. As a consequence some testing modules need minor changes. I could have separated this in its own PR but I consider it a minor thing and my rewriting of the kernel module relies on it.

If the public ldap test servers which are used in the tests (db.debian.org and ldap.forumsys.com) are available then the tests cover all bits and pieces of ldap. Thanks to db.debian.org which allows access with ldaps and starttls even secure connections are tested.

In case one wants to test lychee with a huge number of users, one could simply configure ldap to use db.debian.org and execute "php artisan lychee:LDAP_update_all_users". This will pull the complete debian user base to the users table.

Since the test cases are the guarantee that the LDAPFunctions are working, one should leave them untouched until one finds a bug. 

DO NOT CHANGE ANY NAMES IN THE TEST CODE!

But be very careful. Not everything which looks like a bug, is a bug. The tests try to provoke all kinds of situations to get a full test coverage.

If you want to change the names in LDAPFunctions then you can do so as long you do not change the Interface in LDAPTestFunctions.

The most stupid idea was to use LDAPUserData for the communication between the LDAPFunctions and the SessionFunctions. As one can see from the implementation of LDAPUserData, what is needed in practice is the array interface of this class. So basically the application has the need for an array and because you do not like arrays and consider it bad practice to used them in interfaces, a class was created to hold the data. The consequence is that we now have a dedicated class for the communication between the LDAPFunctions and the SessionFunctions when a simply array would have done a perfect job.

Also very stupid is the fashion of limiting the types of arguments as much as possible. Of course after years of experience with Delphi and Modula I completely understand the rational for it. It is just stupid when you write code in PHP. You should do the exact opposite: try to loosen the type requirement as much as possible. If the code can handle integers and strings as an input then it is just more versatile compared with code which cannot.

In addition I would like to offer some rules which I consider useful:

First, try to agree on some common naming rules. Write regex statements for them and put them as project rules in sonarcloud. Then they will show up in the code-smells. Everything which is not compatible with the rules should be changed. Leave all other names untouched even if you don't like them as long as you are not the author of the code.

It will be a constant battle if one of you wants the_underscore_naming and the other one prefers CamelCase or camelCase. So I recommend to write a naming guideline together with the regex statements.

Second, never change a running system. Do not improve the structure of code which perfectly does its job. The problem is that there is a chance that you do not know all the requirements and the idea that things should be done differently might end up in difficult situation.

An example of this can be found in the development of the LDAP code. The ldap_connect function will as long as you do not use STARTTLS not contact the ldap server. So you get back a connection, but you do not know if the server is usable or not. You gonna find out when you try to bind to that server. If you have a redundant configuration you need to go back and connect to a different server if ldap_bind fails. So the principle design decision of this implementation is to combine ldap_connect with ldap_bind so that after a successful open_LDAP call the ldap server is connected, bond and ready for a search or another user binding. open_LDAP knows how to bind to the server because we configure a binding dn and password or not and if it is not configured then the server is open for anonymous binding.

The LDAPFunctions class knows how to handle the binding by keeping a binding status which reflects the current level of binding.

nagmat84 came to the idea that the binding should not happen during open_LDAP and that it should be only a wrapper around ldap_connect(). Moving out the ldap_bind() from open_LDAP() had two consequences. A multi server configuration would not work and the LDAPFunctions class bound status got mixed up.

The code was passing the rudimentary test cases, but a user with a new username and an arbitrary password could have logged in any time. The reason is that the LDAPFunctions class need to be able to create new users in the users table which have been approved by the ldap server. If code in LDAPFunctions fails then the system will be open.

My solution was to correct the problems, move ldap_bind back into open_LDAP and write test code with 100 % coverage. As long as nobody messes with the test code one can be sure that the LDAPFunctions are reliable if they pass the test. Of course I invite people to review the test code. They just should not change it. Report all finding back to me and I will correct all bugs. Improving the names in the test code is nonsense and therefore I do not support it.

In addition to the test code which I have submitted I have test code which relies on my private LDAP servers and therefore can not be included in the project. If someone starts changing names in the test code I need to change these files. I do not accept this until there is a very good reason for it.

So I suggest the following:

DO NOT CHANGE ANYTHING ON MY FORK!

Decide if you want to use LDAP_auth or not. I you want to use it, then integrate the branch in the official project. Then I am out and it will be all yours. If you mess with it, you need to fix it.

If you do not want to do it, then I will create my own distribution "Lychee with LDAP" where I merge your stuff instead you merging mine.

I leave it up to you.

ildyria:
In light of your past interaction, passive-aggressive messages and non understanding of the core values of Lychee, I have the regret to inform you that this PR is not longer at a stage where it will be merged. We appreciate your work and efforts to bring your code quality up to our standards and good practices, unfortunately this still fails to meet our quality review.

Furtermore, it is a sad day today as this is the first time that we have to go to that point. LycheeOrg has been extremely kind with regard to your past behavior, such attitude will no longer be tolerated. Please refer to our [Code of Conduct](https://lycheeorg.github.io/docs/contributions.html#code-of-conduct), in particular the following two points: 

- Participants will be **tolerant of opposing views**.
- **Participants must ensure that their language and actions are free of personal attacks and disparaging personal remarks.**

Respectfully Yours.

PR #1325, opened 2022-05-14
Additional login screen

hfr:
This PR adds an additional login screen to Lychee to hide the public folder from the general public.

This is useful if the picture galleries should only be accessible to a closed community.

Everybody who wants to see the pictures needs to login first.

In combination with the LDAP authentication it is possible to manage larger groups and use lychee as an archive for pictures for the community.


nagmat84:
Honestly, I do not understand the purpose of that PR. This PR adds a complexity to Lychee and interfere with parts of Lychee which I doubt is actually necessary.

If you don't want unauthenticated users to not see your images, then simply don't make the photos public. If the only thing which bothers you is an empty public smart album on the start page, then there are easier ways to solve that blemish. I agree that an empty smart folder which one does not intend to use might be considered a cosmetic defect. However, the easiest and most appropriate solution for that is to simply exclude empty smart folders from being visible. This can either be done on the backend, i.e. a check inside `getAlbums` should be sufficient, or the frontend simply checks if a smart album is empty and does not render it in case. I would prefer the latter option as it keeps the JSON API consistent. Optimally, this behavior would be made configurable with two new options like `hide_empty_smart_albums` and `hide_empty_smart_albums_public`. The first controls this feature for all user of Lychee, the latter only for anonymous users on the front page.

The sketched solution above neither needs to interfere with the authentication system nor the routing system. Actually, your proposed solution breaks thing. You must not unconditionally redirect any unauthenticated request to `/`. In particular, it break public but password-protected albums and public but hidden albums. These albums needs to be accessed by anonymous users.

ildyria:
> Honestly, I do not understand the purpose of that PR. This PR adds a complexity to Lychee and interfere with parts of Lychee which I doubt is actually necessary.

Actually I am on the opposite team, I do understand the point of this PR.

> If the only thing which bothers you is an empty public smart album on the start page.

That is one of the possible idea behind this PR.

But I also consider the case that some may just want to have a login/password for a set of users. And that way your _public_ becomes the logged in users (and you don't have to manage manually who can see what.

Concrete example: you have a gallery of albums with a subscription model. You want to be able to manage your users but not have to manage the user access to each and every albums... Something that is very much used in the Adult industry for example.

hfr:
> > Honestly, I do not understand the purpose of that PR. This PR adds a complexity to Lychee and interfere with parts of Lychee which I doubt is actually necessary.
> 
> Actually I am on the opposite team, I do understand the point of this PR.
> 
> > If the only thing which bothers you is an empty public smart album on the start page.
> 
> That is one of the possible idea behind this PR.
> 
> But I also consider the case that some may just want to have a login/password for a set of users. And that way your _public_ becomes the logged in users (and you don't have to manage manually who can see what.
> 
> Concrete example: you have a gallery of albums with a subscription model. You want to be able to manage your users but not have to manage the user access to each and every albums... Something that is very much used in the Adult industry for example.

I am happy that someone understands the purpose of a login-screen.

But the real reason goes deeper. Lychee is missing a group model to manage a large number of users. If it would be there, one simply can mange the users using groups.

Now there is one group implemented in Lychee: the public. By adding a login screen one can use this to manage the users.

So the login screen is a way with a limited effort to overcome the limitation that there is no group model at all.

ildyria:
I don't think this should be implemented is such fashion.
It would be much better to have:
- a proper `/login` page & controller,
- a middleware which checks whether user is logged in, and if not, check if login page setting is enabled and if so redirects to said login page.

The current proposed solution may work as is but is not future oriented and increases the cognitive complexity of the code.

hfr:
> I don't think this should be implemented is such fashion. It would be much better to have:
> 
>     * a proper `/login` page & controller,
> 
>     * a middleware which checks whether user is logged in, and if not, check if login page setting is enabled and if so redirects to said login page.
> 
> 
> The current proposed solution may work as is but is not future oriented and increases the cognitive complexity of the code.

I just copied the design pattern of the landing page and added the middleware to secure the rest. I think it is fit for purpuse for a single page application. The application gets difficult to manage if you invent a new design-pattern for every new feature.

I believe that design has to follow function. Here the function is protecting the content. You cannot do that any better by creating a login url.

The login dialog is anyhow implemented in the front-end and therefore it does not matter which url we use for the login.

Actually the implementation is pretty manageable this way. You have a single index-class which operates like a switch board delivering different templates depending on the state while the url  does not change.

I have no idea what kind of future you are envision. I would not envision a multiple page application. From my point the front-end is by far to developed to switch to a different application model. I think it would be over the top to start with it, because of a simple login screen.

You know my bottom line. I do not care if you like my PR or not. I need the feature for my use cases. I have now the feature and I can merge it in any time when lynchee is updated. This is all I care about. If someone wants a different design, then he should go for it.

ildyria:
> I believe that design has to follow function. Here the function is protecting the content. You cannot do that any better by creating a login url.
> 
> The login dialog is anyhow implemented in the front-end and therefore it does not matter which url we use for the login.
> 
> Actually the implementation is pretty manageable this way. You have a single index-class which operates like a switch board delivering different templates depending on the state while the url does not change.

You make interesting points. I will need to think over it.

nagmat84 (changes requested):
I am still strongly against merging this PR into the mainline `master` branch. This PR mitigates a problem for a very specific use case by exploiting a work-around, but it does not solve the actual problem.

Lychee already has a login dialog and system. Adding a second one with a different approach will only make further development more difficult, because we always need to take care of this work-around, too. If this workaround is sufficient for @hfr, then I recommend that he maintains it locally for his installation.

Alternatively, I am always open for a proper solution.

**1. Empty smart album "Public" (and other smart albums)**

I agree that it may be considered a cosmetic defect, to render smart albums for an anonymous user on the frontpage if the smart albums are empty and do not serve any purpose. However, the proper solution is to add some additional logic to the frontend to skip empty smart albums and not render them.

Adding another login screen before that is not the solution.

**2. Direct guidance of the user to the login dialog**

If there is nothing what anonymous users can do, because all smart albums are empty (or not rendered at all, see previous section), usability demands that the login dialog is immediately shown. Otherwise, users may confused whether the site is not working. Moreover, it spares users to click on the cog symbol in the upper-left corner and select "login". I agree that usability can be improved here.

Again, this should be implemented  by the frontend. We already have a similar logic in the frontend to show the password dialog if no admin user is configured yet. This logic can be used as a basis for inspiration to show the login dialog, if no user is logged in and no public objects (i.e. photos, smart albums, regular albums) exist.

Adding another login screen is not the solution.

**3. Group management**

Currently, Lychee only knows individual users and the implicit group of anonymous users. Lychee does not support user groups. I completely understand that user groups are a desirable feature. This way you could manage shares for groups of users more easily.

But the solution is not to misuse the one implicit group of anonymous users as a replacement for the group "any authenticated user". This is what this additional login screen basically achieves. Any user needs to authenticate via the additional login screen and - in case of success - is allowed to see everything which is marked as public.

This will not only confuse admin users of Lychee, but also end in maintainability hell soon.

First of all, an admin (or a Lychee user with upload rights) who marks a photo or album as public, does not really make the photo or album public, but makes it visible to any authenticated users. This is simply confusing. Moreover, one looses the option to mark objects as really public.

Secondly, it will also confuse developers. All methods like `isPublic` (and similar) suddenly get a second, non-obvious meaning which depends on whether a user has used the additional login screen or not. This is simply bad.

If group management is required, then do it right.

If the group "any authenticated user" is the only group which is currently required, then a feature to share an object with any authenticated user can be added in the same spirit like the feature to mark an object as public. But don't misuse the latter for the former.

Alternatively, go the whole way and add proper groups to Lychee, i.e. add a table `groups`, the necessary relations, etc.

hfr:
> I am still strongly against merging this PR into the mainline `master` branch. This PR mitigates a problem for a very specific use case by exploiting a work-around, but it does not solve the actual problem.
> 
> Lychee already has a login dialog and system. Adding a second one with a different approach will only make further development more difficult, because we always need to take care of this work-around, too. If this workaround is sufficient for @hfr, then I recommend that he maintains it locally for his installation.
> 
> Alternatively, I am always open for a proper solution.
> 
> **1. Empty smart album "Public" (and other smart albums)**
> 
> I agree that it may be considered a cosmetic defect, to render smart albums for an anonymous user on the frontpage if the smart albums are empty and do not serve any purpose. However, the proper solution is to add some additional logic to the frontend to skip empty smart albums and not render them.
> 
> Adding another login screen before that is not the solution.
> 
> **2. Direct guidance of the user to the login dialog**
> 
> If there is nothing what anonymous users can do, because all smart albums are empty (or not rendered at all, see previous section), usability demands that the login dialog is immediately shown. Otherwise, users may confused whether the site is not working. Moreover, it spares users to click on the cog symbol in the upper-left corner and select "login". I agree that usability can be improved here.
> 
> Again, this should be implemented by the frontend. We already have a similar logic in the frontend to show the password dialog if no admin user is configured yet. This logic can be used as a basis for inspiration to show the login dialog, if no user is logged in and no public objects (i.e. photos, smart albums, regular albums) exist.
> 
> Adding another login screen is not the solution.
> 
> **3. Group management**
> 
> Currently, Lychee only knows individual users and the implicit group of anonymous users. Lychee does not support user groups. I completely understand that user groups are a desirable feature. This way you could manage shares for groups of users more easily.
> 
> But the solution is not to misuse the one implicit group of anonymous users as a replacement for the group "any authenticated user". This is what this additional login screen basically achieves. Any user needs to authenticate via the additional login screen and - in case of success - is allowed to see everything which is marked as public.
> 
> This will not only confuse admin users of Lychee, but also end in maintainability hell soon.
> 
> First of all, an admin (or a Lychee user with upload rights) who marks a photo or album as public, does not really make the photo or album public, but makes it visible to any authenticated users. This is simply confusing. Moreover, one looses the option to mark objects as really public.
> 
> Secondly, it will also confuse developers. All methods like `isPublic` (and similar) suddenly get a second, non-obvious meaning which depends on whether a user has used the additional login screen or not. This is simply bad.
> 
> If group management is required, then do it right.
> 
> If the group "any authenticated user" is the only group which is currently required, then a feature to share an object with any authenticated user can be added in the same spirit like the feature to mark an object as public. But don't misuse the latter for the former.
> 
> Alternatively, go the whole way and add proper groups to Lychee, i.e. add a table `groups`, the necessary relations, etc.

You are against this PR because you do not like my functional view on things.

I use lychee for quite some time, but not native but as an sandstorm application. In this context sandstorm manages the access to lychee. Unfortunately some of my lychee applications with more than 10000 pictures crashed and I could not bring it up again. So I had two options: Either I go into the business of getting version 4 of lychee ported to be compatible with sandstorm or mimic the sandstorm user experience of having non public lychee appliances.

If have chosen to modify sandstorm by adding LDAP support and a login screen.

This was a bad choice. Not because the technical aspects, because they were rather easy so solve.
It was bad because of the attitude of the people here. Basically the problem is the personal attitude of Matthias Nagel, which is in our country quite typical. I know all to well because I am Germans two.

So this is about two Germans, which cannot get along with each other. And they cannot settle it, because Mr. Nagel does not accept that he cannot boss everybody around.

So he uses his "power" to block my PRs. This one is easy to block, because most people use lychee for public display of the work and then they might use the landing screen but have no use for a login screen.

But even for those people the functionality might be interesting. They can offer a registration for free and then they know how has looked at their work.

My original approach was a complete solution using the existing pattern of the landing screen. Of course the maintainer did like this solution, because it was simply and straight forward. It did not even needed a single change in the front end. This was too functional. So they started to change things which just blows a simple solution up.

On the way Mr. Nagel learned because I wrote it, that what was really missing is some kind of group management.

But with his attitude he will never get it. I know it, because we earn our money with such things.

If you do not want my PR then I am fine with it. I will just create my own lychee distribution and maintain it. Then I can add the thing I find practical without having people around which deliver nothing functional but telling me how I should do things.

This will be a great improvement and I might find other people how do not want bossed around, too, to share the work.

The alternative is that Mr.  Nagel gets his acts together, accepts that my code is working, well tested and adds some interesting functionality and then I do not need to walk away.

I think it is his choice more or less and whatever he decides I am fine with it.

I think such a decision can be made within 14 days.

qwerty287:
I'm not sure if I understand how this works completely, but I assume, you, @hfr, would like to make your photos that are set as public in Lychee are only available to users logged in, but to every user that is logged in. I like this idea. However, I'm not sure if an additional login screen is a good idea. Isn't it possible to re-use the already existing login system and just returning empty responses if trying to access a public album without logged in? So, if the user is logged out and requests any related endpoint, return a 404 response or an empty list. Wouldn't this fix your issue as well?

PR #1328, opened 2022-05-14
PHP based basic install script

hfr:
Lychee requires some command line installation steps.

This is not only ugly but nor error proof.

So idea of this PR is to simply run these steps from PHP to make the process more reliable.

The script install.php can be simply moved to the public folder after cloning the repository or unpacking the zip-archive.

If the browser is pointed to the lychee url the install script will be executed exactly once. At the end it will be moved back to the project root.

If an installation does not have the public/index.php modified, the script can still be used. After moving install.php has been moved in place one needs to point the browser to <lychee-domain>/install.php once.

The idea is not to have a beauty queen but something which works out of the box.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -81,12 +81,24 @@
             "@php artisan key:generate"
         ],
         "post-autoload-dump": [
-            "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
-            "@php artisan package:discover",
+            "@autoload_dump",
+            "@package_discover",
             "@install_files"
         ],
+        "autoload_dump": [
+            "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump"
+        ],
+        "package_discover": [
+            "@php artisan package:discover"
+        ],
         "install_files": [
-            "sh scripts/install_files.sh"
+            "@php -r \"touch('public/dist/user.css');\"",
+            "@php -r \"touch('database/database.sqlite');\"",
+            "@php -r \"file_exists('.git/hooks') || copy('scripts/pre-commit', '.git/hooks/pre-commit');\"",
+            "@php -r \"file_exists('.git/hooks') || copy('scripts/pre-merge', '.git/hooks/pre-merge');\""
+        ],
+        "disable_auto_composer_upgrade": [
+            "@php -r \"touch('.NO_AUTO_COMPOSER_MIGRATE');\""
         ]
     },
     "config": {
--- a/install.php
+++ b/install.php
@@ -1,5 +1,5 @@
 <?php
-COMPOSER_INSTALLER='https://getcomposer.org/installer';
+define('COMPOSER_INSTALLER','https://getcomposer.org/installer');
 
 chdir('..');
 $OK=true;
--- a/install.php
+++ b/install.php
@@ -1,17 +1,30 @@
 <?php
+/**
+ *  Simple straight forward installation script to run the following installtion steps via the web browser:
+ *
+ *  1. Installation of composer.phar
+ *  2. copy .env.example to .env
+ *  3.
+ */
 define('COMPOSER_INSTALLER','https://getcomposer.org/installer');
+define('COMPOSER_HASH','55ce33d7678c5a611085589f1f3ddf8b3c52d662cd01d4ba75c0ee0459970c2200a51f492d557530c71c15d8dba01eae');
+define('ENV','.env');
+define('APP_KEY','APP_KEY=');
 
-chdir('..');
-$OK=true;
-if (!file_exists('.env')) {
-    exec("cp .env.example .env");
-    H('Copy .env.example to .env');
+function getappkey() {
+    $A=file(ENV);
+    foreach($A as $a) {
+        if (str_starts_with($a,APP_KEY)) {
+            return trim(substr($a,strlen(APP_KEY)));
+        }
+    }
 }
 
 function print_o($A) {
     foreach ($A as $a) {
       echo $a."<br/>\n";
     }
+    echo "<br/>\n";
     return true;
 }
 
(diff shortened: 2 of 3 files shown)


public/index.php
```diff
@@ -1,4 +1,6 @@
 <?php
+/* Run install.php, if present*/
+if (file_exists('install.php')) {header("Location: /install.php"); die();}
```
qwerty287, line 3: This shouldn't happen itself I'd say. Just running any command, downloading a Composer version etc.  is a really bad idea. The script should only run of the user requests it. Also, this will run **every** single time you open the gallery. Sorry, but that's completely unacceptable. Even if the file is moved, this should only run if the installation page is opened (if it should run itself).

At least, the user should confirm that the scripts runs, because just running a random (even official) script downloaded from the internet isn't good idea in terms of security.

install.php
```diff
@@ -0,0 +1,120 @@
...
+chdir('..');
+$OK=true;
+if (!file_exists('.env')) {
+    H('Copy .env.example to .env');
+    copy('.env.example', '.env');
+}
+
+H('Installing Composer.phar');
+copy(COMPOSER_INSTALLER, 'composer-setup.php');
+if (hash_file('sha384', 'composer-setup.php') === COMPOSER_HASH) {
```
qwerty287, line 66: What happens if they update their installer? Nobody is able to use it anymore because there's a bad hash sum.

install.php
```diff
@@ -0,0 +1,120 @@
...
+    touch('database/database.sqlite');
+}
+
+if ($OK) {
+    if (getappkey()=='') {
+        H('Generate key');
+        print_o(RUN('php artisan key:generate --force'));
+    }
+    H('Migrate database');
+    print_o(RUN('php artisan migrate --force'));
```
qwerty287, line 96: We actually have a web based installer for this task.

install.php
```diff
@@ -0,0 +1,120 @@
...
+    CL('php -r "'."if (hash_file('sha384', 'composer-setup.php') === '55ce33d7678c5a611085589f1f3ddf8b3c52d662cd01d4ba75c0ee0459970c2200a51f492d557530c71c15d8dba01eae') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;".'"');
+    CL('php composer-setup.php');
+    CL('php -r "'."unlink('composer-setup.php');".'"');
+    CL("cp .env.example .env");
+    CL("./composer.phar update");
+    CL("./composer.phar install --no-dev --prefer-dist");
+    CL("php artisan key:generate");
+    CL("php artisan migrate");
+    T("<li><code>mv public/install.php .</code></li>");
+    T('</ol>');
```
qwerty287, line 119: I would put this into an own function and call this if it fails (if you set `$OK=false`). After printing, just exit the script. Then you can also remove all the `if ($OK) {` checks.

install.php
```diff
@@ -0,0 +1,120 @@
...
+/**
+ *  Simple straight forward installation script to run the following installtion steps via the web browser:
+ *
+ *  1. Installation of composer.phar
+ *  2. copy .env.example to .env
+ *  3.
+ */
+define('COMPOSER_INSTALLER','https://getcomposer.org/installer');
+define('COMPOSER_HASH','55ce33d7678c5a611085589f1f3ddf8b3c52d662cd01d4ba75c0ee0459970c2200a51f492d557530c71c15d8dba01eae');
+define('ENV','.env');
```
qwerty287, line 11: What if you are using a different environment (e.g. testing)?

composer.json
```diff
@@ -81,12 +81,24 @@
...
             "@install_files"
         ],
+        "autoload_dump": [
+            "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump"
+        ],
+        "package_discover": [
+            "@php artisan package:discover"
+        ],
         "install_files": [
-            "sh scripts/install_files.sh"
```
qwerty287, line 89: Delete this script as well.

qwerty287 (changes requested):
Thanks! However, I don't think this should get merged because it is kind of a "bad" variant of the web installer and contains various issues, I added some notes and reviews.

d7415:
As @qwerty287 said, we have an installer script already. As such, I'm going to close this.

public/index.php
hfr: First, the user needs to move install.php in place, if he wants to use this install script.

Second, you can put some nice "Do you really want it? dialog" in front of it. I do not care.

It is only executed if install.php is present. If its not, nothing is executed. We can have a long debate, if this is a security issue. Actually, it works without this line in index.php. The user has to use /install.php as the url.

The problem is that you cannot rely on the installation of composer at least not on Ubuntu. So if you want this to be more reliable then you need to install composer automatically during installation.

Actually you did get the point. Installation is about reliability and not about beauty. It does not look nice. It does not need to follow coding guide lines. It needs to work for every single user. Nothing is more frustrating as having trouble with the installation. Since I had this trouble, I decided to do something about it.

So, as usual I do not care if you like my PR or not and I do not care about acceptance. You can take it or leave it. I need it to do my installation since the existing stuff is buggy.

At least you should copy the idea to have a PHP only install.

You have basically two options to do it. You can do either code them in composer.json and use the scripts section. But you should split the tasks in subtasks there. Your install-mechanic can then use composer run-script to execute them. The other option is to code them as php-scripts and run these.

Actually my problem was, that the shell tasks where no executed correctly, but I figured this out when my script was almost complete.

By the way, you should not rely on any shell commands or shell extension for the installation, if you want life easier for Windows users. At least I tried to do that.

install.php
hfr: You can come up with a better solution if you have an idea.

install.php
hfr: The Web based installer did not work for me.

install.php
hfr: Do what you want to do... You are free to improve things.

install.php
hfr: Go for it... You can do it if you try.

PR #1332, opened 2022-05-15
Make Linter happy about tests

nagmat84:
This is a really trivial and annoying PR, because it changes nothing on the functional level.

For my up-comping PR on [use_filestreams](https://github.com/LycheeOrg/Lychee/tree/use_filestreams) after https://github.com/LycheeOrg/Lychee/pull/1309 has been merged, I promised 85% test coverage for the stuff in [/app/Image/](https://github.com/LycheeOrg/Lychee/tree/master/app/Image) and [/app/Actions/Photo/](https://github.com/LycheeOrg/Lychee/tree/master/app/Actions/Photo). So I am busy adding tests to [use_filestreams](https://github.com/LycheeOrg/Lychee/tree/use_filestreams).

However, I want to get some "trivial" stuff in [/tests](https://github.com/LycheeOrg/Lychee/tree/master/tests) out of the way so that the eventual PR is not littered.

This PR changes three aspects:

 - Make the Linter happy; adding return types everywhere, etc. pp.
 - Added helper methods in `TestCase` to create `UploadedFiles` objects. I will need that a lot in the upcoming PR and I don't want to repeat boiler-plate code. This has also reduced some existing boiler plate code already.
 - Re-organized the files around photo tests. Now there are three instead of two:
    - `PhotosAddTests`: contains tests from the previous `PhotosTests`
    - `PhotosOperationTests`: contains tests from the previous `PhotosTests`
    - `PhotosRotateTest`: same as before
   
   The reason why I split the former `PhotosTest` is again the upcoming PR. There will be a lot of new tests in `PhotosAddTests` for the various combinations we support. So I don't wanted to get the old `PhotosTests` too crowded. It is also called `PhotosAddTests` (instead of `PhotosUploadTests`) on purpose: a) we also have imports; b) this way it comes before `PhotoOperations` in the alphabetical order and we should test photo uploads first before we operate on them.


tests/Feature/DemoTest.php
```diff
@@ -11,14 +21,14 @@ class DemoTest extends TestCase
...
-	public function testDemo0()
+	public function testDemo0(): void
 	{
 		// save initial value
 		$init_config_value = Configs::get_value('gen_demo_js');
 
 		// set to 0
 		Configs::set('gen_demo_js', '0');
-		$this->assertEquals(Configs::get_value('gen_demo_js'), '0');
+		static::assertEquals('0', Configs::get_value('gen_demo_js'));
```
d7415, line 31: Is there a particular reason for the flip? (`0==i` instead of `i==0`)

nagmat84: Yes, making the Linter happy and getting correct error messages.

The order of arguments is `expected`, `actual`. Hence, the desired argument must come first. This will also give you a reasonable error message in case the test fails, i.e. something like that "\<foo\> is expected to equal \<expected\>, but \<value of foo\> received". If the arguments are swapped, the error message becomes nonsense.

My Linter knows that. Basically it only triggers for comparison with constant values like here and if the constant value is in the 2nd argument, then the Linter complains.

d7415: Ok. Unintuitive, but this is PHP.

nagmat84: Totally. I would prefer the other way round, too. I can only guess that it is this way, because there are specialized methods like `assertEuqalsNull`, `assertEqualsZero`, etc., where the expected value is encoded into the name of the method and thus also comes first.

ildyria (approved):
LGTM

PR #1337, opened 2022-05-21
Update dependencies

qwerty287:
Update all dependencies in composer.lock by running `composer update` and add required PHP extensions.

Also I removed a Markdown parser which seems to be unused. Am I right that it can be removed?

Also update some dependencies in composer.json. The only dependencies here are Laravel itself (Larapass blocks an upgrade to Laravel 9) and php-ffmpeg which is blocked by our php-exif fork.


.github/workflows/Build.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         sql-versions: ['mariadb', 'postgresql', 'sqlite']
         mode: ['', '--no-dev']
 
-    name: PHP${{ matrix.php-versions }} - ${{ matrix.sql-versions }} - ${{ matrix.mode }}
+    name: PHP${{ matrix.php-versions }} - ${{ matrix.sql-versions }} ${{ matrix.mode }}
```
nagmat84, line 49: Github workflow files are out of my wheelhouse. Could you explain what this change does and why did it not work before? I am too lazy to read the Workflow Docs.

composer.json
```diff
@@ -15,17 +15,23 @@
     "require": {
         "php": "^8.0",
         "ext-bcmath": "*",
+        "ext-ctype": "*",
         "ext-exif": "*",
 		"ext-fileinfo": "*",
```
nagmat84:
```suggestion
        "ext-fileinfo": "*",
```

Replacing two tabs by eight spaces.

composer.json
```diff
@@ -15,17 +15,23 @@
...
+        "ext-tokenizer": "*",
+        "ext-xml": "*",
+        "ext-zip": "*",
         "bepsvpt/secure-headers": "^7.1",
         "darkghosthunter/larapass": "dev-LycheeSpecial",
         "doctrine/dbal": "^3.1",
         "fideloper/proxy": "^4.3",
         "geocoder-php/cache-provider": "^4.3",
         "geocoder-php/nominatim-provider": "^5.5",
-        "graham-campbell/markdown": "^13.1",
```
nagmat84, line 28: Why is this not needed anymore? We still use it, don't we?

nagmat84 (commented):
I am generally happy with this PR. Nothing against it. Just two questions about changes which I do not understand.

.github/workflows/Build.yml
qwerty287: It's just the name. Currently, for `--no-dev` it always looks like `PHP8.0 - mariadb - --no-dev (push)` which is a bit weird with the hyphen before `--no-dev`. After this change it looks like `PHP8.0 - mariadb --no-dev (push)` (doesn't affect dev builds because they are using an empty string as matrix.mode).

nagmat84: OK, agreed. It looks better.

composer.json
qwerty287: That's what I asked as well, but it doesn't look to be used. For emails, we use Laravel's built-in MD parser (which uses another library as backend).

d7415: It was added [here](https://github.com/LycheeOrg/Lychee/commit/72c3d8b3ea1f574ea34b5611accd2f80c79ed88f)

qwerty287: I looked that up as well and couldn't find any usages. There is this (unused?) PageContent model, but I changed this to use Laravel's MD parser.

nagmat84: My fault. I did not see the line `use Illuminate\Mail\Markdown`. Then I am fine with that.

And yes, the `PageContent` is unused. Ildyria once wanted to add some "blog" functionality which would allow to create photo-blog pages with photos and text. But the feature has never been completed, i.e. you cannot add, edit nor delete pages. IMHO, this should have been move into some development branch until this feature does something actual useful, but ildyria wanted to keep it in the master branch.

nagmat84:
> Also update some dependencies in composer.json. The only dependencies here are Laravel itself (Larapass blocks an upgrade to Laravel 9) and php-ffmpeg which is blocked by our php-exif fork.

There is another more subtle blocker for an upgrade to Laravel 9. Laravel 8 ships with Flysystem version 1, while Laravel 9 took a big step to Flysystem version 3. Even if Larapass would not be a blocker, we had to rewrite a lot of code related to file handling. That's another reason why I should finish my work on EXIF/MIME/etc.

nagmat84 (approved):
LGTM.

ildyria:
Markdown is used on the pages which don't have interfaces... Anyway, it **will** be used by the Livewire front-end to replace the PR of Kamil on using markdown on album descriptions.

For Larapass, I need to have a look. I already started to work towards Laravel 9 at some point in the past, especially when php 8 was released. Didn't really had time to dig further... Priorities...

qwerty287:
You can use the MD parser which is integrated anyway due to Laravel dependencies. This way we can avoid duplicated libraries.

PR #1338, opened 2022-05-22
Proper REST API

qwerty287:
Turn the API into a proper REST API. Mainly has two changes:

- use route instead of query params for some params
- use proper HTTP method (`POST`, `GET`, `PATCH`, `DELETE`)

Also addresses the content type issues I reported in https://github.com/LycheeOrg/Lychee/issues/1356.

TODO

- [x] migrate endpoints to route instead of query params
- [x] merge some endpoints (e.g. editing endpoints for albums like `setTitle`, `setDescription` etc.)
- [x] migrate editing endpoints to `PATCH`
- [x] rewrite frontend API logic to improve code
- [ ] improve ways how requests with lists are handled (see https://github.com/LycheeOrg/Lychee/pull/1338#issuecomment-1133957205 for an example)
- [x] fix tests


nagmat84:
HTTP `PUT` means to put an entire asset at the given URL. Subsequent `GET` requests for the same URL are expected to return the object which has previously been "put" there. HTTP `PATCH` allows only sent partial objects and only the given partial attributes are modified. The general idea of `GET`, `PUT`, `PATCH` is that you do not have a bunch of URLs for individual API calls (like `api/setTitle`), but that you have a URLs which represents the objects.

This is completely different approach then Lychee is using today. At the moment our URLs are action-centered. Proper REST URLs are entity-centered (or model-centered).

Don't get me wrong. I would **love** to see such an approach. But is demands a lot more than just changing the HTTP methods.

Here is an example.

**``PUT`` example**

First we `PUT` a new, complete asset at the denoted URL.

```http
PUT album/17 HTTP/1.1
Content-Type: application/json

{
  id: 17
  title: "Some title"
  parent_id: 42
}
```

The main issue here is that the object must be complete and contain all attributes. (Note: I left out most of the attributes of a proper album object for the sake of keeping the example small.)

Then `GET album/17` is supposed to return

```http
HTTP/1.1 200 OK
Content-Type: application/json

{
  id: 17
  title: "Some title"
  parent_id: 42
}
```

But we don't do it like that. We have a dedicated API endpoint `api/addAlbum`. But one cannot "put" an object into/onto  `api/addAlbum`.

**``PATCH`` example**

Let's assume that you called `GET album/17` and you got

```http
HTTP/1.1 200 OK
Content-Type: application/json

{
  id: 17
  title: "Some title"
  parent_id: 42
}
```

as above. Then it is fine to "patch" the object by providing a partial object like this:

```http
PATCH album/17
Content-Type: application/json

{
  title: "Some other title"
}
```

and the subsequent GET request is supposed to return

```http
HTTP/1.1 200 OK
Content-Type: application/json

{
  id: 17
  title: "Some other title"
  parent_id: 42
}
```

Again, we don't do it like this. We have a dedicated API endpoint `api/setTitle`. But once cannot "patch" `api/setTitle`, one can only patch `album/17`.

**Summary**

The main idea is that you directly work with URLs which identify the entitiy, e.g. `album/17` and then you put/get/patch that. But on must not/cannot put/patch a API endpoint.

We must also keep in mind that proxies and CDN are aware of the semantics of these methods and are allowed to make "clever" decisions. I fear that won't work with our API. As CDN know how these methods are supposed to work using them in a different way may result in a lot of funny and unexpected resutls.

The only HTTP method which has no intrinsic semantic is `POST`. So I would keep that unless we also change the entire routes. Personally, I never felt like doing all this work.

Don't get me wrong. I would love to see something like above for Lychee. This would be a proper REST API. But it would mean a completely different approach. In particular, we would not have a dozen different API endpoints for the individual actions which you can do with a model (such as api/setTitle, api/setDescription, etc.), but one URL for each instance of a model and then the actions to be performed depend on the attributes which are part of the "patch".

qwerty287:
@nagmat84 thanks for the explanation, so we should stay with POST requests? (Or we refactor the edit endpoints to be properly usable?)

nagmat84:
It's up to you. I don't have a strong opinion. But if we do it, we should do it properly. As I said, I was considering myself once, but then did not feel like doing it. If you want to do it, go for it.

Maybe @ildyria and @kamil4 have some opinion.

One other aspect which needs to be considered. If we go for a "pure" `PUT`/`GET`/`PATCH`/`DELETE` approach, we move more logic into the frontend. For example, at the moment we can delete a set of photos with a single `POST` to `deletePhotos` and transmitting a sequence of photo IDs. If you have one URL per entity you need to send one `DELETE` for every entity-URL.

Instead of a single request

```http
POST api/deletePhotos
Content-Type: application/json

{
  photo_ids: [0815, 42, 17]
}
```

we would get three requests

```http
DELETE photo/0815
```
```http
DELETE photo/42
```
```http
DELETE photo/17
```

I am not sure, if I like that. Of course we could do both. I.e. have a proper REST API with entity-based URLs and proper HTTP verbs `PUT`/`GET`/`PATCH`/`DELETE` **and** keep the API. But this sounds like over-kill.

qwerty287:
@nagmat84 I marked this as draft, but you can take a look at https://github.com/LycheeOrg/Lychee/pull/1338/commits/1525308d0aa79f6f4557f01d167cab582b07d5b9. (Tests will probably fail.) Please note that this is still WIP, but it uses for the single album endpoint the `album/{albumID}` notation for parameters.

qwerty287:
I believe this is now in a state where you can start reviewing. It changes a lot of internal functionalities and handlings, but I think it's fine (for me) how it is now.

To the question with the endpoints that consume lists as input: I am using a small workaround. In @nagmat84 's [example above](https://github.com/LycheeOrg/Lychee/pull/1338#issuecomment-1133957205=), you could do the following request:
```
DELETE api/photos/0815,42,17
```
IDs can be comma-separated in the route/URL params and are interpreted as list then. This is probably not REST-conform, but still better and easier than doing multiple requests :)

nagmat84:
I had quick look at this PR and basically skimmed the routes, because I was curious how you solved some of the problems which  I saw when I were considering to create a proper REST API and eventually led me to the conclusion to not do it and which I tried to warn you about.

I appreciate your enthusiasm. I also cannot overstate that I would really love to see a proper REST API which follows a consistent paradigm, but and this is a huge but, it must be done _consistently_. For me, a clean and consistent API is most important. Everything else will drive one crazy, if one tries to work with it. The previous (current) API might not be following the present spirit how web API should be written, but it is internally consistent, i.e. API endpoints represent methods `Photo::setTitle`, parameters are passed as JSON and everything uses `POST`. The naming pattern follows the scheme `<class>::<method>`. This kind of consistency is much more important when it comes to writing other clients than trying to follow a modern paradigm.

The `PUSH`/`PUT`, `GET`, `PATCH` and `DELETE` approach is basically nothing else than the renowned CRUD (Create, Read, Update, Delete) principle of databases. From my point of view, the biggest drawback of this approach is that it basically forces the developer to squeeze everything into this scheme and we hit this barrier with Lychee all the time. For some applications this might be easy and I agree that there are a lot of aspects of Lychee which fit into that nicely. You want to set the title of a album? `PATCH` the album. You want to set the license of photo? `PATCH` the photo. You want to delete a tag album? `DELETE` it. And so on.

However, we also have about a dozen high-level operations which cannot be squeezed into that scheme easily or at least require some thoughts and a lot of effort. These operations were the deciding factor for me not to pursue that path. My quick inspection of the PR unveiled that you basically have left these operations as it. (You renamed them and removed the previous `<class>::` prefix, but basically they are still there.) Hence, we now have proper entities as API endpoints such as `photo`, `album`, etc., which can be patched, fetched and deleted, but we also have operations such like `move`, `merge`, `import`, etc. This is the kind of result which I hoped to avoid.

Let me give you some tips how some of these things can be repaired. I am not promising that this solves every problem. These are only the thoughts which I have already had when I was considering this approach.

First of all, be consistent about singular and plural forms (ok, that probably just an oversight), but at the moment we have `album/{id}` but `photos/{id}`. Second, you need to expose the collections and relations as API endpoints as well, e.g. `album/{id}/photos` gives you all the photos in that album. This basically leads to this preliminary set of API enpoints

```
/albums                   # the collection of all regular albums on the root level
/tag-albums               # the collection of all tag albums on the root level
/smart-albums             # the collection of all tag albums on the root level
/photos                   # the collection of all photos on the root level

/album/{id}               # the album with the given ID
/album/{id}/albums        # the collection of all sub-albums in the album with the given ID
/album/{id}/photos        # the collection of photos in the album with the given ID
/album/{id}/cover         # the cover photo of the album

/tag-album/{id}           # the tag album with the given ID
/tag-album/{id}/photos    # the collection of photos in the tag album with the given ID
/tag-album/{id}/tags      # the collection of tags of the tag album

/smart-album/{id}         # the smart album with the given ID, i.e. public, recent, starred
/smart-album/{id}/photos  # the collection of photos in the smart album with the given ID

/photo/{id}               # the photo with the given ID
/photo/{id}/album         # the parent album of the photo with the given ID
                          # (not sure, if this is really ever required, but in pure theory it should exist)

...
```

I think you should get the idea. The crucial point for consistency is that URLs represent objects: either individual entities or collections thereof. Relations of objects are represented as sub-URLs.

Now, lets consider some of the more problematic operations:

 - _Creation, moving, copying and merging of albums_:
    - To create an album, you `POST` an album to the appropriate collection, i.e. to `/albums` or `/album/{id}/albums`. The former creates a new album on the root level, the latter creates a sub-album.
    - To move an album, there are two options. You could even support both: you `PATCH` its `parent_id` or you `PUT` it into the appropriate collection, e.g. `/albums` or `/album/{id}/albums`. Note that, to put something means to keep its ID, so you do not create a new version of the album.
    - To copy an album, you `POST` the existing album to the appropriate collection (see creation).
    - For merging, there is no real good standard strategy. My idea for a work-around would be to `POST` the existing album onto an album, i.e. `/album/{id}`. This is sufficiently distinct from moving the album which targets the collection of sub-albums, i.e. `/album/{id}/albums`.
 - _Uploading, importing from server, importing via URL of photos:_
    - To upload a photo, you `POST` the image file to the collection of photos, i.e. to `/photos` or `/album/{id}/photos` or `/smart-albums/{id}/photos`.
    - Importing from server and via URL is a little bit more difficult. I do not have a brilliant idea here, but I could imagine to simply `POST` a JSON object with the list of URLs and or paths to the collection of photos. This means, it is identical to uploading, but instead of a binary image file, you sent a JSON. Then the media type (image/* or application/json) decides wether it is a real upload or an import

I still don't have a good idea how to express "photo rotation". That was the point where I gave up thinking about it. In pure theory, you cannot. You have to `GET` the photo, rotate it client side and `PUT` it again. I think we all agree, that this is not what we want. I also do not claim that my ideas for the operations above are the best solution. But at least, the stay within the CRUD principle and they only use URLs that represent objects.

Moreover, this is nothing which is done quickly. There is a lot of work involved. This is what I wanted you to warn about. if you still feel like doing it, go ahead!

Two last remarks:

 - You said that deletion of photos and albums is now possible via `DELETE /(photo|album)/<list-of-ids>`. I see the advantage that this only needs a single request, but I would not do it. First of all, if we want to do it properly, we should do it properly and stay within the paradigm. This means to bite the bullet and use multiple `DELETE` request. Secondly, please keep in mind the length limitation of URLs. A JSON body of a POST request can become as large as needed, but URLs are length restricted to 2048 characters. We hit this barrier for 100 photos which I deem in the range of possibilities. (There are people with 1000 photos in a single album.)
 -  Backward compatibility: I know at least of two persons who already use our JSON API with there own frontends (not counting me). There was the one guy how wrote a FUSE adapter in python to represent Lychee as a file system and there was another person who opened an issue, because `checksum` had been renamed to `original_checksum` which broke his client and the person did not noticed the renaming but feared that the attribute was gone. Changing the API will break those 3rd party clients.

qwerty287:
I'm not sure if it's a good idea to go on with this. As @nagmat84 already pointed out, this would require some more work to be a proper REST API and has multiple, partially major issues.

The biggest issue I see is the backward compatibility. I was aware of this when I started working on it, but I didn't know that there are other clients out there. I thought there's only our web frontend (which is definitely a bit stupid from me because if there's a usable API there will be 3rd-party scripts of course). Changing the API would require big changes for the developers of these tools, depending on what their clients are doing and how many APIs they use. Just renaming a field is something you can change in a second, but changing the concept of the whole API isn't easy.
The other issue I might see is that if we don't change it now, we won't be able to change it later. Especially, if we write more clients that are not integrated directly (e.g. a mobile app), changing the API is something we should or even have to avoid.

Originally, my plan for this only was to migrate some `POST` endpoints to `GET` and `DELETE` (which is also the reason why the branch's name is `http_meth` and not `rest_api` or something like this). In general, this worked well (even if it's still a breaking API change), but it has some of the issues as well like the char limitation of URLs for e.g. `DELETE` statements with lists. Also, the consistency is somewhat lower. Then, you asked me to extend it to a proper REST API - including the warnings I ignored - which I tried to do. I don't have enough experience in web dev to notice myself how a proper REST API should look like, but you're right that it isn't correct and well done in it's current state.
In the future, I should remember that I have to pay more attention to your warnings :)

There are some changes unrelated to a REST API like https://github.com/LycheeOrg/Lychee/pull/1338/commits/c1ce38254ba89fd2fb26eba6d9a0bb9eb96013a2 (?), some parts of https://github.com/LycheeOrg/Lychee/pull/1338/commits/cdd549e8192cf4f17c8111bf97e23eb8e1af79f0 and https://github.com/LycheeOrg/Lychee/pull/1338/commits/34302068739ec5efb7174dd4b17cfc8d5a116d20, https://github.com/LycheeOrg/Lychee/pull/1338/commits/48a645df9f03ad5e1de52e9da055fd64c1325f1e (which is only a frontend sync, so the actual commit is https://github.com/LycheeOrg/Lychee-front/commit/095451ba1d9b1683c8191748c3ad6abc24c5d44e). If we decide to close this PR (which I would recommend currently), I'd cherry-pick these commits and open a new PR. I could also go on with working on it, but this will take some time. Also, we will have to deal with issues by 3rd-party tools.

What do you think on how we should continue with this? I could go on with it, and to speed development up you are free to commit to the branch, but this doesn't solve the problems mentioned above and by @nagmat84.

qwerty287:
I'm going to close this, for reasons see my comment above.

PR #1348, opened 2022-05-26
Remove executable bit from non-executable files

nagmat84:
While I have been working on more tests for my other PR, I noticed that some files on the master branch have set the executable bit which shouldn't. I don't know why, how or when it happened and it doesn't matter. (Probably a very old commit from an Windows machine.)


d7415 (approved):
Good spot!

PR #1351, opened 2022-05-26
Use filestreams

nagmat84:
fixes https://github.com/LycheeOrg/Lychee/issues/60

This is the follow up of #1309. The file objects for images use file streams internally and therewith save a lot of read-write cycles to disk. Previously, every image operation followed the pattern: read file, do image operation, write file. Now, the image file is loaded once, all image operations are done in memory and then the various target files (i.e. the original and the size variants) are written once.

In the long run, this approach will also allow to store the media files on a non-local storage (such as AWS 3). The old approach would have triggered a complete file transfer for every elementary image operation.

I promised 85% test coverage for the source code in [`/app/Actions/Photo`](https://github.com/LycheeOrg/Lychee/tree/use_filestreams/app/Actions/Photo) and [`/app/Image`](https://github.com/LycheeOrg/Lychee/tree/use_filestreams/app/Image). I already have pushed `/app/Actions/Photo` [above 90%](https://codecov.io/gh/LycheeOrg/Lychee/tree/use_filestreams/app/Actions/Photo). :tada: The coverage of `/app/Image` is still as low as it always has been. But I am going to push more tests to the PR. Nonetheless the PR is ready for review.

**A change being worth of discussion**

This PR includes a change in the naming strategy which needs discussion. We already had a discussion about that [on Gitter](https://gitter.im/LycheeOrg/Developers?at=6277ac8261f5ca3feeefe01c). Previously, the name of the image files was based on the checksum of the final image (i.e. after optional auto-rotation, etc.) With the new approach the final checksum is unknown until the image stream has been written onto disk which means that the file cannot be named after the checksum. (Well, it could, but at the cost of higher code complexity.)

Now, the image files are named randomly. This PR includes three different naming strategies. Eventually, I only want to include one naming strategy in order to reduce test and maintenance efforts:

 1. **`SizeVariantFullyRandomNamingStrategy`:** As the name suggests every image file (incl. the different size variants) is named completely random. There is no way to tell if two image files are size variants of the same photo. The directory structure of `/uploads/` contains two level of sub-directories from `/uploads/aa/aa/...` through `/uploads/ff/ff/...`. This avoids thousands of files in a single directory.
   The directories are created on-the-fly as necessary. The old directories `/uploads/thumb`, `/uploads/small`, ..., `/uploads/big` are not used anymore, but they remain in place for existing photos which already have been uploaded. There is no migration of existing photos to the new naming scheme.
[...]

PR #1352, opened 2022-05-29
Test samples for pending PR.

nagmat84:
This PR only adds the test samples to the `master` branch which are required by #1351. The test cases for import via URL download use URLs like these

```php
public const SAMPLE_DOWNLOAD_JPG = 'https://github.com/LycheeOrg/Lychee/raw/master/tests/Samples/mongolia.jpeg';
public const SAMPLE_DOWNLOAD_TIFF = 'https://github.com/LycheeOrg/Lychee/raw/use_filestreams/tests/Samples/tiff.tif';
```

To ensure that the test suite runs for the development branch `use_filestreams` before the merge and the `master` branch after the merge all URLs must already point to the master branch. However, this requires that the sample files are independently merged into master at first.

I added the EXIF tag `Location` to the header of the sample file `night.jpg`, because the related branch in the EXIF reader was untested. That's why the file is 266 bytes larger and appears as changed.


tests/TestCase.php
```diff
@@ -20,20 +20,59 @@ abstract class TestCase extends BaseTestCase
...
+	public const MIME_TYPE_IMG_JPEG = 'image/jpeg';
+	public const MIME_TYPE_IMG_PNG = 'image/png';
+	public const MIME_TYPE_IMG_TIFF = 'image/tiff';
+	public const MIME_TYPE_IMG_WEBP = 'image/webp';
+	public const MIME_TYPE_IMG_XCF = 'image/x-xcf';
+	public const MIME_TYPE_VID_MP4 = 'video/mp4';
+	public const MIME_TYPE_VID_QUICKTIME = 'video/quicktime';
+
+	public const SAMPLE_DOWNLOAD_JPG = 'https://github.com/LycheeOrg/Lychee/raw/master/tests/Samples/mongolia.jpeg';
+	public const SAMPLE_DOWNLOAD_TIFF = 'https://github.com/LycheeOrg/Lychee/raw/use_filestreams/tests/Samples/tiff.tif';
```
qwerty287: I'm not sure if I like this. I often develope offline which will break these, also we don't need the additional traffic if we have the files locally.

nagmat84: We need to download them to test the feature "import via URL". This is something completely different than import from a local file and also takes a different code path in Lychee. For example, the check whether the MIME type of a file is acceptable differs depending on whether the file is locally available are needs to be downloaded. I already found some bugs in the "import via URL code" this way. So we need both: local and download.

qwerty287: Ah okay, then I am completely wrong here. Sorry.

nagmat84: No problem. Questions are always welcome. Most tests are actually run locally as you can see from the path constants some lines below. That's also the reason why there are only two files to be tested with a proper download: a JPEG and a TIFF.

TIFF is treated as "raw" by Lychee (and thus has a different code path than JPEG) but is still supported my Imagick. Everything else which can be tested about images (EXIF, auto-rotation, broken binary encoding, etc.) is tested with import from local files.

ildyria (approved):

PR #1353, opened 2022-05-29
Fix of tests for console commands for pending PR.

nagmat84:
The pending PR #1351 needs to test some of the console commands to cover all methods in `\App\Images` as certain methods are only invoked by those commands. (For example, only the command `lychee:generate_thumbs` checks for the existence of size variants.)

When I tried to write test cases for the console commands, I noticed that

 - the existing tests were never executed, because the name of test class did not end with `...Test`. 
 - a  composer dependency to `Mockery` was missing which is required to mock and capture the console output of the commands
 - two commands were so fundamentally broken that they could not even be started (oh, oh)

This PR repairs all that in order to keep it out of PR #1351. Also note that I have already split the existing tests for console command into three files, because the tests for the command `lychee:generate_thumbs` and `lychee:takedate` will significantly be increased by PR #1351 and hence the deserve their own files already.


qwerty287 (approved):
Looks good, but untested.

Issue #1323, opened 2022-05-13
install errors


sunosv:
**Steps to reproduce the issue**

download zip
Run composer install --no-dev to install dependencies
Copy .env.example as .env and edit it to match your parameters
Generate your secret key with php artisan key:generate
Go to web http://server/Lychee/public/install
First screen shows Lychee-installer -> click on Next button
Second screen shows error: "Please fix the errors before going to the next step." But all the checks are 'green'
Clicking Re-check button does not help

**Diagnostics**

php artisan lychee:diagnostics

In Connection.php line 712:

  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'lychee.users' do
  esn't exist (SQL: select * from `users` where `users`.`id` = 0 limit 1)

In Connection.php line 368:

  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'lychee.users' do
  esn't exist

System Information
Lychee Version (git): master  Default

composer install: --no-dev
APP_ENV: production
APP_DEBUG: false

System: win
PHP Version: 8.1

Browser and system
Server: Fedora 35/Apache
Client: Windows 10
Browser: any

[![QQ截图20220514001949.png](https://sunos.top:1098/2022/05/14/c0c3ca1b41640.png)](https://sunos.top:1098/2022/05/14/c0c3ca1b41640.png)
Where did I go wrong?

I tried to php artisan migrate 
then I could not open the page 
[![QQ截图20220514015700.png](https://sunos.top:1098/2022/05/14/d5a3a4080bac6.png)](https://sunos.top:1098/2022/05/14/d5a3a4080bac6.png)

and nginx REWRITE :

    if (!-e $request_filename)
    {
        rewrite ^/(.*)$ /index.php?/$1 last;
        break;
    }

[![QQ截图20220514020406.png](https://sunos.top:1098/2022/05/14/04077a3ab1000.png)](https://sunos.top:1098/2022/05/14/04077a3ab1000.png)

Diagnostics
-----------
Warning: 'D:\phpEnv\www\Lychee\public\dist\user.css' does not exist or has insufficient read/write privileges.
Warning: Dropbox import not working. dropbox_key is empty.
Warning: zend.assertions is enabled although Lychee is not in debug mode. Outside of debugging, code generation for assertions is recommended to be disabled for efficiency reasons

System Information
------------------
Lychee Version (release):        4.5.1
DB Version:                      4.5.1

composer install:                --no-dev
APP_ENV:                         production
[...]

PR #1361, opened 2022-06-04
Migrate install_files scripts to PHP and add more config to `.env`

qwerty287:
Allow to change some config options via `.env` and migrate the `install_files` to PHP (see https://github.com/LycheeOrg/Lychee/issues/1323#issuecomment-1126657461=). Also fixes a deprecation.


config/app.php
```diff
@@ -91,7 +91,7 @@
 	|
 	*/
 
-	'locale' => 'en',
+	'locale' => env('APP_LOCALE', 'en'),
```
nagmat84: In generally, I agree to exposing this and the other configuration settings via the `.env` file. But two aspects:

 1. I am not sure if Lychee honors this setting at all. Lychee uses its own home-brewed localization whose setting is stored in the DB. I don't mind to expose these settings nonetheless, but they will probably only affect some error messages and Blade templates from the framework. I only wanted to make you aware of this.
 2. If you add new environment variables, you should also update the example `.env` file. Otherwise nobody will notice that these variables exist.

config/app.php
```diff
@@ -216,7 +216,7 @@
 		'Cookie' => Illuminate\Support\Facades\Cookie::class,
 		'Crypt' => Illuminate\Support\Facades\Crypt::class,
 		'DB' => Illuminate\Support\Facades\DB::class,
-		'DebugBar' => Barryvdh\Debugbar\Facade::class,
+		'DebugBar' => Barryvdh\Debugbar\Facades\Debugbar::class,
```
nagmat84, line 219: Nicely spotted. :+1:

scripts/install_files.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+$NO_COLOR = "\033[0m";
+$GREEN = "\033[38;5;010m";
+$YELLOW = "\033[38;5;011m";
+$ORANGE = "\033[38;5;214m";
+
+echo "\n${YELLOW}creating file for CSS personalization$NO_COLOR\n";
+touch('public/dist/user.css');
```
nagmat84:
```suggestion
\Safe\touch('public/dist/user.css');
```

scripts/install_files.php
```diff
@@ -0,0 +1,19 @@
...
+
+echo "\n${YELLOW}creating default SQLite database$NO_COLOR\n";
+touch('database/database.sqlite');
```
nagmat84:
```suggestion
\Safe\touch('database/database.sqlite');
```

scripts/install_files.php
```diff
@@ -0,0 +1,19 @@
...
+
+echo "\n${YELLOW}setting up hooks for git pull and git commits$NO_COLOR\n";
+copy('scripts/pre-commit', '.git/hooks/pre-commit');
+copy('scripts/post-merge', '.git/hooks/post-merge');
```
nagmat84:
```suggestion
\Safe\copy('scripts/pre-commit', '.git/hooks/pre-commit');
\Safe\copy('scripts/post-merge', '.git/hooks/post-merge');
```

scripts/install_files.php
```diff
@@ -0,0 +1,19 @@
...
+
+echo "\n${ORANGE}To disable the call of composer and migration on pull add$NO_COLOR\n";
+echo "${ORANGE}a file named '.NO_AUTO_COMPOSER_MIGRATE' in this directory.$NO_COLOR\n";
```
nagmat84: This comment is for the whole script not just this line. I am not sure, if this script works nicely on Windows. Obviously, nobody has ever being bothered by this question, because the old file was a `.sh`-script which obviously has never been working with Windows.

But now, after the script has become runnable for Windows we should maybe also ensure that it runs nicely. Here are the two things, I am not sure about:

 1. We use `\n` as the new line character. PHP offers the constant `PHP_EOL` which is `\n\r` on Windows, because Windows needs an additional carriage return. However, I do not know if this is only required for text files our for output to the command line, too. I would suspect the letter, because the programmer cannot know if the STDOUT targets the screen or is redirected to a file. So probably, we should use `PHP_EOL`.
 2. I am not sure, if Windows can handle the ANSI escape sequence and color codes.

nagmat84 (commented):
Overall I am in favor of the PR. I added some remarks and suggestions, but nothing severe.

scripts/install_files.php
qwerty287: Not tested, but Windows seems to support ANSI. See https://gist.github.com/mlocati/fdabcaeb8071d5c75a2d51712db24011 for example.

config/app.php
qwerty287: I don't think this is used because we have our own system here. Is this something might be used by Laravel internally? Then I would rollback this change, otherwise I'd remove it.

qwerty287: It is used by Laravel internally, so I'd go with `en` as value since it doesn't affect the web UI and we can ignore it.

config/session.php
```diff
@@ -71,20 +71,7 @@
...
-	|--------------------------------------------------------------------------
-	|
-	| When using the "database" or "redis" session drivers, you may specify a
-	| connection that should be used to manage these sessions. This should
-	| correspond to a connection in your database configuration options.
-	|
-	*/
-
-	'connection' => env('SESSION_CONNECTION', null),
+	'connection' => env('SESSION_CONNECTION'),
```
nagmat84, line 74: I totally agree that it was nonsense to set the same option twice. But why have you removed the comment (of the repeated option)?

config/hashing.php
```diff
@@ -30,21 +30,4 @@
...
-	| passwords are hashed using the Argon algorithm. These will allow you
-	| to control the amount of time it takes to hash the given password.
-	|
-	*/
-
-	'argon' => [
-		'memory' => 1024,
-		'threads' => 2,
-		'time' => 2,
-	],
```
nagmat84: At the moment, Lychee does not use Argon as a hashing algorithms anywhere. Hence, these options are probably not used currently. I agree on that, even though I am not that sure about any Laravel internal code which might even use Argon.

But what happens if someone starts to use Argon? Will it fail? (I would even call this a good case) Or will it silently fall back to insecure defaults if no settings are found? (This would be the an extremely bad case.) Have you checked that?

Personally, I feel unsure about the (possibly bad) consequences of removing these settings even though I am generally in favor of nuking everything which is unnecessary.

scripts/install_files.php
nagmat84: Great! Then we can keep the color codes. I feared that Windows would get confused in such a way that it would totally break console output. Glad to hear that it is the other way around!

But I still believe we need `PHP_EOL` instead of `\n`.

Moreover, another thought crossed my mind. I somehow remember that you figured out that the ANSI color codes are either only supported by `echo` or by `printf`. I don't remember what script you repaired and what the correct solution was, but have you checked that?

config/session.php
qwerty287: It was duplicated as well (above these lines).

config/app.php
```diff
@@ -133,7 +133,7 @@
 
 	'key' => env('APP_KEY'),
 
-	'cipher' => 'AES-256-CBC',
+	'cipher' => env('APP_CIPHER', 'AES-256-CBC'),
```
d7415, line 136: I'm not sure how much benefit there is to exposing this, but I guess there's no harm.

config/session.php
d7415: :+1:

config/trustedproxy.php
```diff
@@ -33,5 +33,6 @@
 	 *
 	 * @link https://symfony.com/doc/current/deployment/proxies.html
 	 */
-	'headers' => Request::HEADER_X_FORWARDED_ALL,
+	'headers' => Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_HOST | Request::HEADER_X_FORWARDED_PORT
+		| Request::HEADER_X_FORWARDED_PROTO | Request::HEADER_X_FORWARDED_PREFIX,
```
d7415, line 37: :+1:

config/urls.php
```diff
@@ -3,7 +3,7 @@
 return [
 	'update' => [
 		// we need this in case the URL of the project changes
-		'git' => 'http://api.github.com/repos/LycheeOrg/Lychee/commits',
+		'git' => 'https://api.github.com/repos/LycheeOrg/Lychee/commits',
```
d7415, line 6: :+1:

d7415 (commented):
Looks good. I think it's worth adding some of the missing variables to the example `.env`, so that users know what's available.

qwerty287:
I wouldn't add them to `.env`, instead, use the docs for this task (https://github.com/LycheeOrg/LycheeOrg.github.io/pull/56).

d7415:
https://github.com/LycheeOrg/LycheeOrg.github.io/pull/56#discussion_r889679493
> I'm very in favour of complete examples (ideally with the default values - I suggested changing those to match elsewhere).
> 
> As a user, having to copy every directive is a pain, and then you'd probably want to add context in case you need to edit later.

qwerty287:
I understand this and agree with you, but I wouldn't add the experimental/advanced settings to the example. And the non-experimental are all available there already.

d7415:
Looking again at the list, I think I mostly agree. At which point it doesn't need to be in this PR.

FWIW, the one that caught my eye was `SESSION_SECURE_COOKIE`, which seems like a reasonable thing to expose at some point.

kamil4 (approved):
Untested but looks good.

PR #1364, opened 2022-06-07
Override `key:generete` command to fix recreating key every time

qwerty287:
https://github.com/LycheeOrg/Lychee/commit/d9fa818ed811c0fd7ac4557a14a4bbc7696e3298 introduced a minor issue: every time you call `composer install` the `APP_KEY` will be recreated. This PR fixes this by overriding the default `key:generate` command (can be found in `vendor/laravel/framework/src/Illuminate/Foundation/Console/KeyGenerateCommand.php`) so that it will only create the key if there isn't already one.

### DIFF
--- a/app/Console/Commands/KeyGenerateCommand.php
+++ b/app/Console/Commands/KeyGenerateCommand.php
@@ -2,34 +2,34 @@
 
 namespace App\Console\Commands;
 
-use Illuminate\Encryption\Encrypter;
-use Safe\Exceptions\UrlException;
-
+/**
+ * Generate the `APP_KEY` config variable.
+ */
 class KeyGenerateCommand extends \Illuminate\Foundation\Console\KeyGenerateCommand
 {
+	/**
+	 * The name and signature of the console command.
+	 *
+	 * @var string
+	 */
+	protected $signature = 'key:generate
+	                {--no-override : Do not override an existing key}
+                    {--show : Display the key instead of modifying files}
+                    {--force : Force the operation to run when in production}';
+
 	/**
 	 * Set the application key in the environment file.
 	 *
 	 * @param string $key
 	 *
 	 * @return bool
-	 *
-	 * @throws UrlException
 	 */
 	protected function setKeyInEnvironmentFile($key): bool
 	{
-		$currentKey = $this->laravel['config']['app.key'];
-		if (str_starts_with($currentKey, 'base64:')) {
-			$currentKey = substr($currentKey, 7);
-		}
-		$supported = Encrypter::supported(\Safe\base64_decode($currentKey), $this->laravel['config']['app.cipher']);
-
-		if (strlen($currentKey) !== 0 && ($supported || ($this->getDefaultConfirmCallback()() && !$this->confirmToProceed()))) {
-			return false;
+		if (!$this->hasOption('no-override') || !$this->option('no-override')) {
+			return parent::setKeyInEnvironmentFile($key);
 		}
 
-		$this->writeNewEnvironmentFileWith($key);
-
-		return true;
+		return strlen($this->laravel['config']['app.key']) === 0 || parent::setKeyInEnvironmentFile($key);
 	}
 }
--- a/app/Console/Commands/KeyGenerateCommand.php
+++ b/app/Console/Commands/KeyGenerateCommand.php
@@ -2,68 +2,11 @@
 
 namespace App\Console\Commands;
 
-use Illuminate\Console\Command;
 use Illuminate\Console\ConfirmableTrait;
-use Illuminate\Encryption\Encrypter;
 
-class KeyGenerateCommand extends Command
+class KeyGenerateCommand extends \Illuminate\Foundation\Console\KeyGenerateCommand
 {
 	use ConfirmableTrait;
-
-	/**
-	 * The name and signature of the console command.
-	 *
-	 * @var string
-	 */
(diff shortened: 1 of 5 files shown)


app/Console/Commands/KeyGenerateCommand.php
```diff
@@ -0,0 +1,35 @@
...
+	 *
+	 * @throws UrlException
+	 */
+	protected function setKeyInEnvironmentFile($key): bool
+	{
+		$currentKey = $this->laravel['config']['app.key'];
+		if (str_starts_with($currentKey, 'base64:')) {
+			$currentKey = substr($currentKey, 7);
+		}
+		$supported = Encrypter::supported(\Safe\base64_decode($currentKey), $this->laravel['config']['app.cipher']);
```
nagmat84: Why is `\base64_decode` outside of the condition which checks whether the value starts with `base64:`? I haven't looked up where the key is used inside the framework, so it might be correct and the key is always encoded in Base64 even if it does not start with `base64`. But is looks suspicious, I would have expected `base64_encode` inside the condition.

app/Console/Commands/KeyGenerateCommand.php
```diff
@@ -0,0 +1,35 @@
...
+
+		if (strlen($currentKey) !== 0 && ($supported || ($this->getDefaultConfirmCallback()() && !$this->confirmToProceed()))) {
```
nagmat84: Why do you call `getDefaultConfirmCallback` explicitly`? It is called inside `confirmToProceed` anyway. The original condition from the parent class does not do it either. There it is only:

```php
if (strlen($currentKey) !== 0 && (! $this->confirmToProceed())) {
```

What do you want to achieve with that? If there is a non-obvious reason you should add the explanation to the PhpDoc comment above the method and explain what this method does differently than the original and why. Other developers may wonder the same in the future.

app/Console/Commands/KeyGenerateCommand.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Encryption\Encrypter;
+use Safe\Exceptions\UrlException;
+
+class KeyGenerateCommand extends \Illuminate\Foundation\Console\KeyGenerateCommand
```
nagmat84, line 13: Please add a PhpDoc comment to the class which explains why it is necessary to override the command and provide reference to the Composer install script. Other developer in the future may ask themselves the same question.

nagmat84 (changes requested):
The original variant of the command allowed to replace an existing key after confirmation. With the modification, the confirmation never appears when a valid encryption key is set, but the command is a silent no-op. This means the command cannot be used interactively anymore to replace an existing key with confirmation.

This is good for our install script, but it removes functionality. Maybe a better approach would be to introduce a new command line flag, e.g. `--batch-mode` or `--dont-ask` or `--non-interactive`, which explicitly enables the new behavior and we use that flag in our install script. This would leave the original behavior untouched.

qwerty287:
I changed the behavior:
1. if used without the new option `--no-override`, it is not changed
2. otherwise, it checks if the key is not empty and uses the `parent::` method if it is, otherwise if just returns

app/Console/Commands/KeyGenerateCommand.php
nagmat84: That is not exactly what I meant. The comment only explains what the original command does anyway. This does not explain _why_ we need to overwrite it. I meant background information, not the obvious. More something like this

```php
/**
 * Generates the `APP_KEY` config variable.
 *
 * This class extends the original command by the additional option `--no-override`
 * which - if enabled - let the command silently do nothing, if the variable has already been set.
 * This special mode is needed in the automatic install script which is called by Composer
 * upon package installation
 */
```

This is what I meant by `provide reference to the Composer install script`. In 6 months from now on, nobody will remember we we needed that.

app/Console/Commands/KeyGenerateCommand.php
```diff
@@ -2,34 +2,34 @@
...
-		if (strlen($currentKey) !== 0 && ($supported || ($this->getDefaultConfirmCallback()() && !$this->confirmToProceed()))) {
-			return false;
+		if (!$this->hasOption('no-override') || !$this->option('no-override')) {
+			return parent::setKeyInEnvironmentFile($key);
 		}
 
-		$this->writeNewEnvironmentFileWith($key);
-
-		return true;
+		return strlen($this->laravel['config']['app.key']) === 0 || parent::setKeyInEnvironmentFile($key);
```
nagmat84: This shortcut is wrong, isn't it? If the key has length 0, i.e. if it is not set, then the first part evaluates to `true` and nothing happens. I don't think that is what we want. Or am I completely confused?

Moreover, I think you can combine this short-cut condition with the `if`-clause above.

nagmat84 (changes requested):

app/Console/Commands/KeyGenerateCommand.php
qwerty287: I noticed that a few minutes before you commented as well, I'm sorry. I can't change it right now, but will fix this (and the other issue) tomorrow.

qwerty287: NVM, did it now.

nagmat84 (approved):
LGTM. :+1:

Issue #486, opened 2020-04-19
[Enhancement] Document migration from 4.0.0 to 4.0.1


LudovicRousseau:
Yesterday I installed and migrated to Lychee 4.0.0 using the .zip file.
The .zip file created the directory `Lychee-v4.0.0`.
On disk I have:
```
Lychee -> Lychee-v4.0.0
Lychee-v4.0.0
```
I use a symbolic link from `Lychee` to `Lychee-v4.0.0`.

The .zip for version 4.0.1 created a new folder and I now have:
```
Lychee -> Lychee-v4.0.0
Lychee-v4.0.0
Lychee-v4.0.1
```
I can't just modify the symbolic link since the images are in `Lychee-v4.0.0/public/uploads/big/`, etc.
I need to copy/move them from `Lychee-v4.0.0/` to `Lychee-v4.0.1/`.

Can you provide a script to run to migrate from `Lychee-v4.0.0` to `Lychee-v4.0.1`?

En even better solution would be to have Lychee in two distinct directories:
- (static) code in `Lychee-v4.0.0/`
- images, configuration, logs, etc. in `Lychee-data/` or something similar.

One benefit is that www-data only need read access for `Lychee-v4.0.0/` and read-write access only for `Lychee-data/`.

ildyria:
Interesting use case.
Also, there should be the possibility to change your upload folders, your can't you just use the `.env` ?

LudovicRousseau:
Good idea.

Now I have in my `.env`:
```
LYCHEE_DIST="/var/www/Lychee-data/public/dist/"
LYCHEE_UPLOADS="/var/www/Lychee-data/public/uploads/"
```
And in the Apache2 `/etc/apache2/sites-available/lychee.conf` file
```
  Alias "/dist/" "/var/www/Lychee-data/public/dist/"
  Alias "/uploads/" "/var/www/Lychee-data/public/uploads/"
```
It works fine.

I then recreated the `Lychee-v4.0.1` from the .zip file. But I get the error:
```
[...]

PR #1366, opened 2022-06-11
Improve config options

qwerty287:
**:warning: might be breaking**

- Reduce options by removing/determining automatically
    - removed logs config completely because it's only used for SQL logs
    - other removals are mostly unused options
    - removed options that might result in wrong config without real warnings
- add some new env options (like `HASHING_ALGORITHM`)


.env.example
```diff
@@ -17,7 +15,7 @@ LOG_CHANNEL=stack
 # Table prefix (e.g. lychee_) of a Lychee v3 instance for migration
 DB_OLD_LYCHEE_PREFIX=
 
-# DB_CONNECTION can be sqlite, mysql or pgsql. For sqlite the other entries are
+# DB_CONNECTION can be sqlite, mysql, pgsql or sqlsrv. For sqlite the other entries are
```
d7415: Do we want this here? We're not testing it, none of the developers use it, I'd leave it unsupported. Maybe mention it in the proper docs where there's space for a disclaimer - I already don't like how long this bit is, and I wrote it.

composer.json
```diff
@@ -58,7 +58,10 @@
         "nunomaduro/collision": "^5.0",
         "phpunit/phpunit": "^9"
     },
-    "autoload": {
+	"suggest": {
+		"etx-phpredis": "*"
```
d7415: ?

config/database.php
```diff
@@ -38,7 +38,6 @@
 			'url' => env('DATABASE_URL'),
 			'database' => env('DB_DATABASE', database_path('database.sqlite')),
 			'prefix' => '',
-			'foreign_key_constraints' => env('DB_FOREIGN_KEYS', true),
```
d7415, line 41: Does this definitely default to `true` if it's removed?

config/database.php
```diff
@@ -49,7 +48,6 @@
 			'database' => env('DB_DATABASE', 'forge'),
 			'username' => env('DB_USERNAME', 'forge'),
 			'password' => env('DB_PASSWORD', ''),
-			'unix_socket' => env('DB_SOCKET', ''),
```
d7415: I haven't tested using sockets here, but if they work I don't see why we should block them.

config/database.php
```diff
@@ -162,11 +160,11 @@
 	*/
 
 	'redis' => [
-		'client' => env('REDIS_CLIENT', 'phpredis'),
+		'client' => 'phpredis',
 
 		'options' => [
 			'cluster' => env('REDIS_CLUSTER', 'redis'),
-			'prefix' => env('REDIS_PREFIX', Str::slug(env('APP_NAME', 'laravel'), '_') . '_database_'),
+			'prefix' => env('REDIS_PREFIX', Str::slug(env('APP_NAME', 'Lychee'), '_') . '_database_'),
```
d7415, line 169: I don't know if the capitalisation is significant here, but it's worth considering.

config/cache.php
```diff
@@ -98,6 +95,6 @@
 
 	'prefix' => env(
 		'CACHE_PREFIX',
-		Str::slug(env('APP_NAME', 'laravel'), '_') . '_cache'
+		Str::slug(env('APP_NAME', 'Lychee'), '_') . '_cache'
```
d7415, line 98: I don't know if the capitalisation is significant here, but it's worth considering.

config/debugbar.php
```diff
@@ -135,32 +135,32 @@
...
 		],
 		'route' => [
-			'label' => true,  // show complete route on bar
+			'label' => true,
 		],
 		'logs' => [
 			'file' => null,
 		],
 		'cache' => [
-			'values' => true, // collect cache values
```
d7415: This looks potentially useful, and it doesn't look like it's been moved anywhere.

config/filesystems.php
```diff
@@ -57,7 +57,7 @@
 			'driver' => 's3',
 			'key' => env('AWS_ACCESS_KEY_ID'),
 			'secret' => env('AWS_SECRET_ACCESS_KEY'),
-			'region' => env('AWS_DEFAULT_REGION'),
+			'region' => env('AWS_DEFAULT_REGION', 'us-east-1'),
```
d7415, line 72: Is there any point adding this?

config/mail.php
```diff
@@ -35,7 +35,7 @@
 	'mailers' => [
 		'smtp' => [
 			'transport' => 'smtp',
-			'host' => env('MAIL_HOST', 'smtp.mailgun.org'),
```
d7415, line 38: Why?

config/session.php
```diff
@@ -127,7 +128,7 @@
 
 	'cookie' => env(
 		'SESSION_COOKIE',
-		Str::slug(env('APP_NAME', 'lychee'), '_') . '_session'
+		Str::slug(env('APP_NAME', 'Lychee'), '_') . '_session'
```
d7415, line 131: I guess this aligns with the defaults, but...

config/database.php
nagmat84: Please keep it! I am using sockets on my productive system and on my test system. I prefer sockets, because they cannot be accidentally exposed to the external world and they do not have the overhead of a TCP connection.

composer.json
qwerty287: This is necessary if you would like to use Redis.

d7415: I mean, we don't suggest the DB drivers, for example...

d7415:
Somewhat off-topic, but I've just seen that composer.json still has `"name": "lycheeorg/lychee-laravel"`...

config/database.php
qwerty287: Hmm I thought it is unused, that's why I removed it. I did a full text search in the code (including the `vendor` directory) for `unix_socket`, there aren't any.

config/database.php
qwerty287: Hmm I thought it is unused, that's why I removed it. I did a full text search in the code (including the `vendor` directory) for `foreign_key_constraints`, there aren't any.

config/filesystems.php
qwerty287: It's the default on other usages of this env var. Even if it's a comment, but so we can't forget it if we add S3 support at some point.

config/mail.php
qwerty287: 1. There's a separate Mailgun driver
2. Why should the server have a default value?

config/database.php
qwerty287: I added this mainly for consistency with other usages of `APP_NAME`. But usually, `APP_NAME` is set already to the string `Lychee` which means that this default will be used almost never.

config/mail.php
d7415: Sorry, my bad. I read this one backwards!

config/database.php
nagmat84: No, it is used. See here: https://github.com/laravel/framework/blob/2bc180295ebf8c4848a6b9c4ae8fa0d00ec75217/src/Illuminate/Database/SQLiteConnection.php#L111-L114

config/database.php
nagmat84: It is used here: https://github.com/doctrine/dbal/blob/e15d05d08172ceeb4f55066feed48d688fdb0250/src/Driver/Mysqli/Driver.php#L60-L68

.gitignore
```diff
@@ -3,10 +3,6 @@
 public/Lychee-front/node_modules/
 public/Lychee-front/package-lock.json
 
-public/Lychee-front/node_modules/
-public/Lychee-front/bower_components/
-public/Lychee-front/package-lock.json
-
```
nagmat84: What happens if we stop ignoring those files?  More over this makes me wonder if we should simply ignore the whole directory `Lychee-front`. It is a GIT submodule at last.  However, I was too lazy to look into the official GIT docs how `.gitignore` of the parent project and submodules interact with each other.

composer.json
```diff
@@ -1,15 +1,24 @@
...
+    "support": {
+        "source": "https://github.com/LycheeOrg/Lychee",
+        "issues": "https://github.com/LycheeOrg/Lychee/issues",
+        "docs": "https://lycheeorg.github.io/docs/",
+        "chat": "https://gitter.im/LycheeOrg/Lobby"
+    },
     "keywords": [
         "framework",
         "laravel"
     ],
```
nagmat84: If we are on it, I wonder whether these keywords make any sense. The seem to have been copied from a Laravel application template, but Lychee is not a framework. I would have expected keywords like

 - `"photo management"`,
 - `"gallery"`,
 - `"album software"`,
 - `"image organizer"`,
 - etc.

The latter came to my mind, because Wikipedia calls it like this: https://en.wikipedia.org/wiki/Image_organizer, https://en.wikipedia.org/wiki/Comparison_of_photo_gallery_software

config/app.php
```diff
@@ -65,8 +65,6 @@
 
 	'url' => env('APP_URL', 'http://localhost'),
 
-	'asset_url' => env('ASSET_URL'),
-
```
nagmat84: Are we sure that we do not need this config option? We use the method `asset()` in some Blade templates below `view/`.

Even if we do not use it, I assume that is is better to set this variable to `null`. I cannot assess what happens if the variable is unset. It is used here https://github.com/laravel/framework/blob/cd6e76fe864161af7307be7abb6d4d18cbe064ca/src/Illuminate/Routing/RoutingServiceProvider.php#L62-L66 I guess this will throw an exception, if the array key is not set.

config/mail.php
```diff
@@ -60,15 +60,6 @@
...
 		],
-
-		'log' => [
-			'transport' => 'log',
-			'channel' => env('MAIL_LOG_CHANNEL'),
-		],
-
-		'array' => [
-			'transport' => 'array',
-		],
```
nagmat84: What happens, if these array keys are not set? Are you sure that they are really not used anywhere? This time I was to lazy to look into the vendor source code myself.

config/session.php
```diff
@@ -99,7 +100,7 @@
 	|
 	*/
 
-	'store' => env('SESSION_STORE'),
+	'store' => env('SESSION_DRIVER'),
```
nagmat84: Are you sure that this is equivalent? I haven't look into the source code nor the documentation, but my gut feeling tells me that `SESSION_DRIVER` ans `SESSION_STORE` are different things. I would have assume, that the meaning of `SESSION_STORE` depends on the value of `SESSION_DRIVER`. For example, if `SESSION_DRIVER` equals `file` I would have expected `SESSION_STORE` to be the actual physical path, e.g. `storage/session`.

I don't know it, that is just me feeling. Have you confirmed with the Laravel source code that your change is correct?

nagmat84 (changes requested):
During the review I was wondering why Lychee provides two files `.env.example` and `.env.homestead`. Does anybody know why? What is the purpose of the latter?

.gitignore
d7415: 2 of the 3 are ignored just above, so only `bower_components`, whatever that is. I haven't seen it.

qwerty287: It's from bower, some deprecated npm alternative focused on browsers. Probably lychee used it in older versions.

config/app.php
qwerty287: This was never set before in the `.env` file, means it was always `null`. I don't think Laravel is that error-sensitive, but I can change it to `null`. Also, I didn't notice any issues while testing.

config/mail.php
qwerty287: Nothing. I'm sure. They are `driver`s or `connection`s that are only used if the corresponding option has this array key value. In this case, it's the `MAIL_DRIVER` config. As long nobody sets it to `array` (which I don't think somebody did), there's nothing that could break it.

config/session.php
qwerty287: This affects the cache stores defined in the cache config file. They have the same names as the driver name itself, so they are equivalent.

config/filesystems.php
```diff
@@ -68,8 +68,8 @@
 		// TODO: Maybe we should drop this Flysystem disk, because neither the driver nor the root must be changed and hence the whole point of using the Flysystem abstraction is gone.
 		'dist' => [
```
ildyria, line 81: I am unsure about this change. This remove the ability for user to configure external path to store their pictures.

See:
https://github.com/LycheeOrg/Lychee/issues/486
https://lycheeorg.github.io/docs/distributions.html#separating-data-from-the-code

qwerty287: You can still set `LYCHEE_UPLOADS` to geht this, or is this wrong?

qwerty287: Or do you only refer to the `dist` path? This doesn't affect the pictures, only the `user.css` file.

ildyria: True. My mistake.

d7415: That same concern does apply though, for separating the code from user data.

qwerty287: Should I undo this change?

config/session.php
nagmat84: > They have the same names as the driver name itself, so they are equivalent.

Yes and no. Only because they accidentally happen to have the same values, does not mean they are equivalent. Actually, I would call it a fail of the Laravel example configuration to use the same values for `SESSION_DRIVER` and `SESSION_STORE` in their example file because this suggests that there would be some relation which isn't.

That is the same mistake so many math teachers do on a blackboard if they - for example - need to draw to lines or triangles: most people tend to draw either parallel or orthogonal lines and tend to draw either right-angled triangles or equilateral triangle - only rarely people draw really skew lines and oblique triangles. Then, when they prove a statement, they falsely use an "accidental" property which does not hold in full generality.

Here we have the same thing. Let's assume we wanted to store the session in a database (Lychee does not). Further assume that we wanted to use two databases, for example one database for volatile, ephemeral data and one for data which needs to be persisted. Maybe the databases are even of completely different type: the first one might be a MySQL database with in-memory tables and the second one is a PostgeSQL database on disk. In this case the store in `cache.php` would not be called `'database' `, but maybe `'fast-in-mem-mysql'` and `'durable-psql'`, but the driver would still be `'database'` because this relates to the PHP class which implements the driver.

Hence, I would prefer one out of two options:

a) Either go the full way and make the whole thing completely not configurable via environment variables.
b) Revert the change to support configuration via environment variables in full generality without making assumptions.

Maybe a) is even the best option, because I doubt that we will ever support something else than we have now.

qwerty287:
@LycheeOrg/reviewers This PR should be reviewed because it contains some new config options that are already listed in the docs. Could you please re-review it?

kamil4 (approved):

config/session.php
qwerty287: @nagmat84 Can this be resolved or should I remove the config option? Removing will make it impossible to use apc, dynamodb, memcached or redis as session store, because they need a store config. This store config has *exactly* the same name as the driver itself, they are defined in https://github.com/LycheeOrg/Lychee/blob/5853686354d70401dd5d5a278104aa6f8147cfa2/config/cache.php#L33-L83.

nagmat84: Yes and no.

I would not recommend removing the option for the very reasons you pointed out yourself, but I wouldn't keep it the current way either. I would revert it to the old behavior, i.e. keeping `SESSION_STORE` and `SESSION_DRIVER` distinct.

As I wrote, they are semantically different things. That all examples accidentally happen to use the same values for them is a bad co-incident, which suggests a relationship between these options and their values which does simply not exist.

qwerty287: OK, I'll see when I get the time to do this.

qwerty287:
@nagmat84 I added the session store option back. You still need to approve because you requested changes :)

composer.json
```diff
@@ -1,10 +1,20 @@
 {
-    "name": "lycheeorg/lychee-laravel",
+    "name": "lycheeorg/lychee",
```
nagmat84: Just something which came to my mind again. The Github URL is `LycheeOrg/Lychee` with three capital letters. I don't know if this is relevant for anything at all or in case "something" interprets this entry, that "something" does so in a case-sensitive manner.

Anyway, I just re-remembered that we are astonishingly inconsistent about our capitalization: not only here, but also in the frontend, etc.

nagmat84 (approved):

composer.json
qwerty287: Can we ignore it or should I update it?

ildyria: I would go for `lychee-org/lychee` because our other packages on [packagist.org](https://packagist.org/?query=lychee) are under that common name

d7415: It's fairly common for package names etc to be lower case while the project itself has capitalisation. By no means universal, but common. Especially with package managers.

ildyria:
@qwerty287 There have been some changes with Laravel 9.
I would suggest to see if you could rebase this PR on top of https://github.com/LycheeOrg/Lychee/pull/1469 to avoid doing the work twice. :| 

Or we can merge this now and see what is conflicting with #1469 :)

qwerty287:
I'd merge this first and then look at #1469. I can see what I can do regarding conflicts there, but I'd merge this first.

composer.json
qwerty287:
```suggestion
    "name": "lychee-org/lychee",
```

nagmat84:
I guess we are good to merge this one?

qwerty287:
From my side yes :)

composer.json
d7415: I only just reread this. **Why** the "-"??? (@ildyria)

(I agree we should match what's there as we're effectively defining a package here, but...)

ildyria: No clue. Just matching.

d7415: Were you not the one to set it up?

ildyria: I honestly don't remember, the only packages that are uploaded on packagist anyway are dependencies.
I guess my intent was because Capital letters are not "allowed" in package names and thus I wanted to avoid lycheeorg (which is ugly). But don't quote me on that.

nagmat84: Yes, @ildyria was. That's another inconsistency we have and when I spotted it, it had been to late.

I guess we all agree that it would be awesome to use the same spelling of "Lychee Org" everywhere, i.e. here on Github and on [Packagist](https://packagist.org/). I don't know if it is possible to rename our packages and create some kind of "redirection" on Packagist from the old name to the new name such that we don't break any dependencies.

ildyria:
```json
    "require": {
        "php": "^8.0",
        "ext-bcmath": "*",
        "ext-ctype": "*",
        "ext-exif": "*",
        "ext-fileinfo": "*",
        "ext-gd": "*",
        "ext-json": "*",
        "ext-mbstring": "*",
        "ext-openssl": "*",
        "ext-pdo": "*",
        "ext-tokenizer": "*",
        "ext-xml": "*",
        "bepsvpt/secure-headers": "^7.1",
        "darkghosthunter/larapass": "dev-LycheeSpecial",
        "doctrine/dbal": "^3.1",
        "fideloper/proxy": "^4.3",
        "laravel/framework": "^8.83.14",
        "livewire/livewire": "^2.7",
        "lychee-org/nestedset": "^6",
        "lychee-org/php-exif": "^0.7.11",
        "maennchen/zipstream-php": "^2.1",
        "php-ffmpeg/php-ffmpeg": "^1.0",
        "php-http/guzzle7-adapter": "^1.0",
        "php-http/message": "^1.12",
        "spatie/guzzle-rate-limiter-middleware": "^2.0",
        "spatie/laravel-feed": "^4.0",
        "spatie/laravel-image-optimizer": "^1.6.2",
        "symfony/cache": "^v6.0.0",
        "whichbrowser/parser": "^2.0"
    },
    "require-dev": {
        "ext-imagick": "*",
        "ext-posix": "*",
        "ext-zip": "*",
        "barryvdh/laravel-debugbar": "^3.6",
        "barryvdh/laravel-ide-helper": "^2.10",
        "filp/whoops": "^2.5",
        "friendsofphp/php-cs-fixer": "^3.3",
        "itsgoingd/clockwork": "^5.0",
        "laravel/homestead": "^v13.2.1",
        "lychee-org/phpstan-lychee": "dev-master",
        "mockery/mockery": "^1.5",
        "nunomaduro/collision": "^5.0",
        "nunomaduro/larastan": "^1.0",
        "php-parallel-lint/php-parallel-lint": "^1.3",
        "phpunit/phpunit": "^9"
    },
```
camelCase does not seem standard there.

ildyria: If we take example on https://github.com/FriendsOfPHP it _should_ be "lycheeorg"

d7415: I was just curious :). I'm slightly in favour of `lycheeorg` but it's not a major issue for me.

FWIW, we're `lycheeorg` on Docker Hub. That one was probably me.

d7415: Found in a SO post: https://github.com/composer/packagist/issues/47

Which raises the question of why it was working in the first place.

Answer: Because the packages we publish are in `lychee-org` form https://github.com/LycheeOrg/laravel-nestedset/blob/v6/composer.json#L2

So I'd use `lychee-org` for now until/unless we want to change properly.

Issue #1321, opened 2022-05-10
Clean up schema definition of table `photos` (drop `filesize`, rename `type`)


nagmat84:
PR #1239 moved the filesize from the photos to the aize variants. However, the column `filesize` remained on table `photos` as it was (or still is) impossible to drop a column from a table definition without re-creating everything (see  https://github.com/LycheeOrg/Lychee/pull/1239#discussion_r820222170_). We should catch up on that as part of the next bigger migration.

nagmat84:
The column `type` should be renamed to `mimetype` (or `mime_type`) to better reflect what kind of type is stored.

There had been quite some code which wrongly used that column to store the kind/category of the photo (i.e. photo, video or raw).

PR #1368, opened 2022-06-12
Improve API client usability

### LINKED ISSUE #1356: [Enhancement] make API really usable with external clients
The API is currently not really usable for external clients. While there were/are refactorings like https://github.com/LycheeOrg/Lychee/pull/1217 and https://github.com/LycheeOrg/Lychee/pull/1338 that improve that, it is still not in a usable state. What I would like to have is an API that allows me to run `curl localhost:8000/api/albums` and I get a correct JSON answer including public albums. Main issue here is the content type middlewares which require a `Content-type` *and* an `Accept` header. I don't think this is a good idea. The `Accept` header can be ignored (it will be always JSON) and the content type of the request body should be recognized automatically. So that I can run `curl localhost:8000/api/albums` instead of `curl localhost:8000/api/albums -H "Content-type: application/json" -H "Accept: application/json"`.

Another issue is the authentication. Especially on a command line this is really hard. Expiring session keys are not good for API clients and just add extra complexity. The easier way would be to allow a HTTP Basic or Digest auth, the more complex, but better would be a token system which generates tokens for different API apps that can then be sent in query params or a header. In general, a way to authorize *without* using the `api/Session::` functions.

qwerty287:
- add an API token system. This assigns a token to every user which you can view, reset, copy etc. This token allows authorized access to API endpoints using the `Authorization` header.
- do not require Content-Type and Accept headers, if they allow any type or are missing, ignore them, only if they are wrong an exception is thrown.
- add an endpoint `User::getCurrent` which allows to get the user which is currently logged in

Closes https://github.com/LycheeOrg/Lychee/issues/1356


qwerty287:
Does anybody know why the SQLite rollback migration fails and how I can fix it?

qwerty287:
I think CI failure is not related. (Only pgsql fails, MariaDB and locally SQLite work)

app/Http/Middleware/AcceptContentType.php
```diff
@@ -37,7 +37,7 @@ class AcceptContentType
 	public function handle(Request $request, \Closure $next, string $contentType): mixed
 	{
 		if ($contentType === self::JSON) {
-			if (!$request->expectsJson()) {
+			if (!$request->acceptsJson()) {
```
kamil4: What does that change do?

app/Http/Middleware/ContentType.php
```diff
@@ -34,15 +34,17 @@ class ContentType
...
 			if (!$request->isJson()) {
 				throw new UnexpectedContentType(self::JSON);
 			}
-		} elseif ($contentType === self::MULTIPART) {
+		} elseif ($contentType === self::MULTIPART && !$any) {
 			if ($request->getContentType() !== 'form') {
 				throw new UnexpectedContentType(self::MULTIPART);
 			}
-		} else {
+		} elseif (!$any) {
```
kamil4: You've added `!$any` to all three cases. It feels like it should be handled up-front (start with `if (!$any)` and indent everything else by one level).

app/Http/Middleware/VerifyCsrfToken.php
```diff
@@ -36,26 +38,24 @@ class VerifyCsrfToken extends Middleware
...
-			 * We are currently checking for Authorization.
-			 * Do we also want to check if there is a POST value with the apiKey ?
-			 */
-			if ($apiKey && $request->header('Authorization') === $apiKey) {
-				return $next($request);
+			/** @var User $user */
+			$user = User::query()->where('token', '=', $token)->get();
+			if ($user instanceof Collection) {
+				$user = $user->get(0);
 			}
```
kamil4: Is that `if` meant to handle cases of token collision between users or is it something more fancy? If the former, wouldn't it be better to prevent such cases at token creation time (check for duplicates an if so, generate another random one)? Considering that you appear to simply pick the first user then, why not simply use something like `->first()` instead of `->get()`?

database/migrations/2022_06_12_075709_add_token_to_user_table.php
```diff
@@ -0,0 +1,64 @@
...
+	 * @return void
+	 */
+	public function down(): void
+	{
+		DB::table('configs')->insert([
+			[
+				'key' => 'api_key',
+				'value' => User::query()->findOrFail(0)->token,
+				'confidentiality' => 0,
+				'cat' => 'config',
```
kamil4: Isn't that config currently in `Admin` with confidentiality level `3`? Why are you changing that?

database/migrations/2022_06_12_075709_add_token_to_user_table.php
```diff
@@ -0,0 +1,64 @@
...
+	 * @throws Exception
+	 */
+	public function up(): void
+	{
+		$key_length = 16;
+
+		$old_key = Configs::get_value('api_key', '');
+		if (!$old_key) {
+			$old_key = strtr(base64_encode(random_bytes($key_length)), '+/', '-_');
+		}
```
kamil4: So the old code supported `api_key` as an _optional_ feature (if the default value of `''` was used, the feature was disabled). You seem to unconditionally enable it for all users, with no option of turning it off. Is there any reason for that?

kamil4 (commented):
I left some (hopefully useful) comments, but somebody with a greater understanding of the `api_key` feature should have a look (I've never used it or even bothered to fully understand it).

app/Http/Middleware/AcceptContentType.php
nagmat84: Yes, this is one of the changes I don't like at all. I had my reasons to use `expectsJson`, because otherwise we run into problems with the exception handler. I'm too busy to explain all the details right now, because I am still struggling with the PHPstan PR. I would really be glad, if we could wait with that PR until we have merged @ildyria's PR and maybe my PR about the file streams.

app/Http/Middleware/ContentType.php
nagmat84: Same problem here. We must be really careful about the content types because there are deeply hidden dependencies to the exception handler. It had cost me hours (if not even days) to read through the Laravel source code and figure that out.

If there is an urgent need to push this PR forward, I would be fine with it, if someone else did the review on the parts which are directly related to the API keys, but please leave the content types untouched in that case.

I also don't really see how this is related to API keys. For me these are two orthogonal things.

If these changes are really necessary (for whatever reason), then please, please wait with this PR until the other PRs are out of the door and I can have a look into this one.

database/migrations/2022_06_12_075709_add_token_to_user_table.php
qwerty287: Forgot to change the copied block.

app/Http/Middleware/AcceptContentType.php
qwerty287: I couldn't find any issues when testing it. The reason  why I changed it is mainly that this allows to set something like `*/*` as Accept header. The old variant *always* required an Accept header containing something like `json`. This is a bit annoying, but I can undo the change.

app/Http/Middleware/ContentType.php
qwerty287: @nagmat84 this is not related to API keys, but to the API usability. See my comment above for explanation.

database/migrations/2022_06_12_075709_add_token_to_user_table.php
qwerty287: I know that it was optional (actually, it doesn't have any effect before - the feature was broken completely), but what's a disadvantage of enabling it for all users? Just noticed that the biggest security concern is that the `api_key` setting doesn't have any permissions before, but now has admin permissions. It should be autogenerated for all users.

Adding a global option to disable it (not per user) is easy to add though.

app/Http/Middleware/ContentType.php
qwerty287: I removed this change. It's easy to add for clients.

app/Http/Middleware/AcceptContentType.php
qwerty287: I removed the change, clients can easily add the required headers.

database/migrations/2022_06_12_075709_add_token_to_user_table.php
ildyria: The point of the `api_key` was to bypass the CSRF verification. There is no authentication related to it.

qwerty287: I removed this as authentication key for the admin.

ildyria:
- Merged from master.
- fixed conflicts.
- tested.
- add possibility to disable the token feature by leaving it "empty".

ildyria (approved):
Tested - LGTM.

ildyria:
@qwerty287 can you check if you are happy with my changes? :)

nagmat84:
Please give me the chance to do a review, too, before merge.

qwerty287:
@ildyria in general yes, but adding the old api key for the admin means that an api key that had no permissions before now has admin permissions. I can fix it.

ildyria:
> api key that had no permissions before now has admin permissions. I can fix it.

Good point. I missed that one.

ildyria (approved):
IMHO ready for merge.

Would wait for #1403 and resolve conflicts, or merge and then fix conflicts in #1403

ildyria:
@nagmat84 kind reminder as you requested to have time to review 19 days ago. :)

database/migrations/2022_06_12_075709_add_token_to_user_table.php
```diff
@@ -0,0 +1,48 @@
...
+	 * @return void
+	 *
+	 * @throws Exception
+	 */
+	public function up(): void
+	{
+		Configs::where('key', '=', 'api_key')->delete();
+
+		Schema::table('users', function (Blueprint $table) {
+			$table->char('token', 100)->after('email')->default('');
```
nagmat84: Why `''` as default? The column is nullable hence we should use `null` as default or make the column non-nullable. 

But using `null` also yields another advantage. The tokens are unique with very high likelihood. Hence, we should create a `unique` index on this column which also gives higher efficiency when users are looked up by their token. However, a unique column may have multiple `null` entries, because `null` is not included in the index. But SQL does not allow multiple `''` in a unique index.

database/migrations/2022_06_12_075709_add_token_to_user_table.php
```diff
@@ -0,0 +1,48 @@
...
+				'key' => 'api_key',
+				'value' => '',
+				'confidentiality' => 3,
+				'cat' => 'Admin',
+			],
+		]);
+		Schema::disableForeignKeyConstraints();
+		Schema::table('users', function (Blueprint $table) {
+			$table->dropColumn('token');
+		});
```
nagmat84: The ongoing problem. SQLite does not support dropping columns yet. Dropping a column is mimic by re-creating a table without that column. But this throws away all foreign key constraints.

For the time being, we must abstain from dropping columns. But luckily this is only in the downwards path, not the upwards path. Otherwise we had to add it to the list in https://github.com/LycheeOrg/Lychee/issues/1321.

app/Models/User.php
```diff
@@ -30,6 +30,7 @@
  * @property string|null                                           $email
  * @property bool                                                  $may_upload
  * @property bool                                                  $is_locked
+ * @property string                                                $token
```
nagmat84: Even though the token is public information (is it?) I wonder, if we should it to the `hidden` array such that it is not sent to the client when the user object is converted into JSON.

routes/api-admin.php
```diff
@@ -49,3 +49,6 @@
 Route::post('/User::create', [UserController::class, 'create']);
 Route::post('/User::setEmail', [UserController::class, 'setEmail'])->withoutMiddleware([AdminCheck::class]);
 Route::post('/User::getEmail', [UserController::class, 'getEmail'])->withoutMiddleware([AdminCheck::class]);
+Route::post('/User::getCurrent', [UserController::class, 'getCurrent'])->withoutMiddleware([AdminCheck::class]);
```
nagmat84: Is the name of the route specified by some standard or can it be changed? I have not yet reviewed what `getCurrent` does but IMHO it should tell what it gets, e.g. `getCurrentSomething`. I assume token? Why not simply `getToken`?

tests/Feature/Lib/UsersUnitTest.php
```diff
@@ -205,4 +205,67 @@ public function get_email(
...
+	 *
+	 * @return TestResponse
+	 */
+	public function get_user(
+		int $expectedStatusCode = 200,
+		?string $assertSee = null
+	): TestResponse {
+		$response = $this->testCase->postJson('/api/User::getCurrent');
+		$response->assertStatus($expectedStatusCode);
+		if ($assertSee) {
```
nagmat84:
```suggestion
		if ($assertSee !== null) {
```

I know PHPStan is not (yet?) enabled for the tests, but maybe it will someday.

tests/Feature/Lib/UsersUnitTest.php
```diff
@@ -205,4 +205,67 @@ public function get_email(
...
+	 *
+	 * @return TestResponse
+	 */
+	public function reset_token(
+		int $expectedStatusCode = 200,
+		?string $assertSee = null
+	): TestResponse {
+		$response = $this->testCase->postJson('/api/User::resetToken');
+		$response->assertStatus($expectedStatusCode);
+		if ($assertSee) {
```
nagmat84:
```suggestion
		if ($assertSee !== null) {
```

tests/Feature/Lib/UsersUnitTest.php
```diff
@@ -205,4 +205,67 @@ public function get_email(
...
+	 *
+	 * @return TestResponse
+	 */
+	public function disable_token(
+		int $expectedStatusCode = 200,
+		?string $assertSee = null
+	): TestResponse {
+		$response = $this->testCase->postJson('/api/User::disableToken');
+		$response->assertStatus($expectedStatusCode);
+		if ($assertSee) {
```
nagmat84:
```suggestion
		if ($assertSee !== null) {
```

routes/api-admin.php
nagmat84: After I have been reading on, I assume `getAuthenticatedUser` is what has been meant?

routes/api-admin.php
```diff
@@ -49,3 +49,6 @@
...
+Route::post('/User::resetToken', [UserController::class, 'resetToken'])->withoutMiddleware([AdminCheck::class]);
+Route::post('/User::disableToken', [UserController::class, 'disableToken'])->withoutMiddleware([AdminCheck::class]);
```
nagmat84:
```suggestion
Route::post('/User::deleteToken', [UserController::class, 'deleteToken'])->withoutMiddleware([AdminCheck::class]);
```

maybe?

Or

```suggestion
Route::post('/User::unsetToken', [UserController::class, 'unsetToken'])->withoutMiddleware([AdminCheck::class]);
```

maybe? `unset` matches `reset`.

If there is a method `disableX` I would also expect to have a method `enabledX` which we don't.

app/Http/Middleware/VerifyCsrfToken.php
```diff
@@ -37,26 +38,25 @@ class VerifyCsrfToken extends Middleware
...
 
-			/*
-			 * if apiKey is the empty string we directly return the parent handle.
-			 */
-			if ($apiKey === '') {
+			/** @var User|null $user */
+			$user = User::query()
+				->where('token', '=', $token)
+				->where('token', '!=', '')
+				->first();
```
nagmat84: That's why we should have a unique index on the `token` column. (See comment wrt. the migration file.) Also inequality conditions are always less efficient in SQL then equality checks on an indexed column. However, if we store `null` for non-existing tokens, then the inequality condition can be dropped.

app/Http/Controllers/Administration/UserController.php
```diff
@@ -130,4 +130,48 @@ public function getEmail(): array
 			'email' => AccessControl::user()->email,
 		];
 	}
+
+	/**
+	 * Returns the currently authenticated user.
+	 *
+	 * @return User
+	 */
+	public function getCurrent(): User
```
nagmat84: I would prefer

```suggestion
	public function getAuthenticatedUser(): User
```

app/Http/Controllers/Administration/UserController.php
```diff
@@ -130,4 +130,48 @@ public function getEmail(): array
...
+	 */
+	public function resetToken(): array
+	{
+		AccessControl::user()->token = strtr(base64_encode(random_bytes(16)), '+/', '-_');
+		AccessControl::user()->save();
+
+		return [
+			'token' => AccessControl::user()->token,
+		];
+	}
```
nagmat84: I would prefer to return the entire user object. That is what we do in most cases. Yes I know there are still exceptions to this rule like the unfortunate `getEmail`.

```suggestion
	public function resetToken(): array
	{
		$user = AccessControl::user();
		$user->token = strtr(base64_encode(random_bytes(16)), '+/', '-_');
		$user->save();

		return $user;
	}
```

app/Http/Controllers/Administration/UserController.php
```diff
@@ -130,4 +130,48 @@ public function getEmail(): array
...
+	 */
+	public function disableToken(): array
+	{
+		AccessControl::user()->token = '';
+		AccessControl::user()->save();
+
+		return [
+			'token' => '',
+		];
+	}
```
nagmat84: Why do we need to return anything at all? Also I would prefer `unsetToken` because it matches `resetToken`.

```suggestion
	public function unsetToken(): void
	{
		$user = AccessControl::user();
		$user->token = null;
		$user->save();
	}
```

This will return `204 - No Content` in case of success and a proper error code in any other case.

If you wonder why I would like to use `null` instead of the empty string `'' ` see my other comments about unique constraints and DB query efficiency.

app/Actions/User/Create.php
```diff
@@ -24,6 +24,7 @@ public function do(string $username, string $password, bool $mayUpload, bool $is
 			$user->is_locked = $isLocked;
 			$user->username = $username;
 			$user->password = bcrypt($password);
+			$user->token = '';
```
nagmat84:
```suggestion
```

Not necessary if default was `null` (see other comments).

nagmat84 (changes requested):
Overall it looks really good. I like it. Even though this might not be reflected by the number of my comments. :see_no_evil:

nagmat84:
After I went through the whole PR and the frontend I have learned that the token is a symmetric, shared secret. This answers my question from above. But in this case, it should also be treated as one. In particular, it should not be sent back to the user except after initial creation. (This is also how Github does it.) If the user forget the token, then the user must reset the token and obtain a new one.

app/Actions/User/Create.php
qwerty287: ~~It can't be `null` right now, this would require changes to the migration. Is it better to set this to `null` instead of an empty string?~~ Yes we should.

qwerty287:
~~The tests don't fail locally for me~~ :thinking: Seems working now.

qwerty287:
The main difference to Github I see here is that Github allows as many keys as you need, while this PR only allows *one* per user. This makes it hard to reuse the key if you write it down anywhere (I wouldn't do this if it's not in some app's config) and every time I need a new key I have to recreate it.

nagmat84:
Yes, I saw that difference. And I am somehow inclined to say that users need to be authenticated first, before they can see their own token. But this argument would hold for passwords as well and we don't show the password to the user either. Moreover I cannot give you a good example of what might possibly go wrong. However, the little security guy inside me tells me that it is a bad idea.

As I have no strong feelings about that we can leave it like that for now. I just wanted to inform you about my bad gut feeling.

qwerty287:
I know. We can try to build a multi-token system in a later PR that makes this easier for us and the users.

tests/Feature/UsersTest.php
```diff
@@ -290,7 +290,8 @@ public function testDisableToken(): void
 		$oldToken = $users_test->reset_token()->offsetGet('token');
 		assertNotEquals('', $oldToken);
 
-		$newToken = $users_test->disable_token()->offsetGet('token');
+		$users_test->unset_token();
+		$newToken = $users_test->get_user()->offsetGet('token');
 		assertEquals('', $newToken);
```
nagmat84: I am surprised. Shouldn't be that `null`?

qwerty287: Actually yes. Maybe `offsetGet` returns an empty string if this key doesn't exist? (But it should fail...)

nagmat84:
I had to simplify the middleware `VerifyCrsfToken` because I could not stand the number of `return` in the method which really makes it hard to reason about what this method does.

tests/Feature/UsersTest.php
nagmat84: I don't know. But if it does I would prefer the following approach to have a stricter comparison.

```php
$userResponse = $users_test->get_user();
$userResponse->assertJson([
	'token' => null,
]);
```

ildyria:
>  However, the little security guy inside me tells me that it is a bad idea.

I am actually more concerned about something else... which is the reason why we hash password.
If the database is compromised, at least secrets are non recoverable.

I would actually like to suggest the following. Have the token encrypted in the database.
When we receive the token, we encrypt it, and search for the encrypted match.

That way, it is possible for a user to "read" is current token, but if the DB is compromised but not the `.env` then we are less in a pinch.
Relevant: https://laravel.com/docs/8.x/encryption

nagmat84:
@qwerty287 Given my remark about putting the user object properly into the initialization part. I will create a PR for that. This will close some the TODOs which I added to the source code and it will make your PR for the front-end cleaner.

nagmat84:
> @qwerty287 Given my remark about putting the user object properly into the initialization part. I will create a PR for that. This will close some the TODOs which I added to the source code and it will make your PR for the front-end cleaner.

With PR https://github.com/LycheeOrg/Lychee/pull/1443 the entire user object becomes part of `lychee.user` in the front-end. This frees this PR from fetching the current user before opening the dialog. Even though `User::getAuthenticatedUser` becomes unnecessary with that, I would keep that method. It does not harm to have it.

qwerty287:
@ildyria I tried to implement encryption of the tokens now, but it didn't work. I can't search for encrypted values in the DB, because they are always different.
They use some random string to kind of salt it (but the salt isn't necessary to decrypt it). This would only be possible if `encryptString` always returns the same output for the same input. (See https://github.com/laravel/framework/blob/8.x/src/Illuminate/Encryption/Encrypter.php#L100).

Also, encrypting the token isn't something where I'd say it's necessary. Hashing a password is something else, because they are often used multiple times, but the tokens are random. I don't really see a security risk here, because if your DB is compromised, they have access to every data anyways (except data that is added later of course).

ildyria:
> @ildyria I tried to implement encryption of the tokens now, but it didn't work. I can't search for encrypted values in the DB, because they are always different. They use some random string to kind of salt it (but the salt isn't necessary to decrypt it). This would only be possible if `encryptString` always returns the same output for the same input. (See https://github.com/laravel/framework/blob/8.x/src/Illuminate/Encryption/Encrypter.php#L100).
> 
> Also, encrypting the token isn't something where I'd say it's necessary. Hashing a password is something else, because they are often used multiple times, but the tokens are random. I don't really see a security risk here, because if your DB is compromised, they have access to every data anyways (except data that is added later of course).

true.

ildyria:
May I suggest to get out in 4.6.1, once @nagmat84 is happy with it ? :)

d7415:
We could always hash the token. You're back to "show once", but that could well be why "proper" services do it that way. I don't think the argument that these aren't passwords holds.

nagmat84:
> We could always hash the token. You're back to "show once", but that could well be why "proper" services do it that way. I don't think the argument that these aren't passwords holds.

This way my initial idea when I reviewed this PR. I wanted "show once" combined with "hash it". Who gave an argument based on "it is not a password hence we are good"?! :astonished: :open_mouth:

qwerty287:
That's not what I meant :) Of course it's best to encrypt (or hash) it, but the main difference to a password is that it's randomly generated. The main reason to hash the password is to prevent that you can log in to other accounts with the same password, and since this is a random string, you can't use it to take over other accounts with it.

Yes, I thought about hashing too, but show once is not that user-friendly if you only have one key. I can implement it though.

nagmat84:
> That's not what I meant :) Of course it's best to encrypt (or hash) it, but the main difference to a password is that it's randomly generated. The main reason to hash the password is to prevent that you can log in to other accounts with the same password

No that is not the reason for hashing. One hashes the DB such that an attacker cannot use the password after the attacker has corrupted the DB, because the attacker cannot calculate the pre-image of the hashes. It does not matter whether all passwords are distinct or if some user use the same password. 

> and since this is a random string, you can't use it to take over other accounts with it.

No, that is not true. If the DB has been corrupted and the attacker can read the tokens, it does not matter whether they have been randomly generated or if they have been chosen by a human (like most passwords).

qwerty287:
I pushed https://github.com/LycheeOrg/Lychee/pull/1368/commits/d8ccc6e75488edc283dcca95e9e4bd71432ed274 which hashes the tokens using SHA-512 (not bcrypt though) and only allows to see them once.

tests/Feature/UsersTest.php
```diff
@@ -246,4 +248,35 @@ public function testUsers(): void
...
+	}
+
+	public function testUnsetToken(): void
+	{
+		$users_test = new UsersUnitTest($this);
+
+		Auth::loginUsingId(0);
+
+		$oldToken = $users_test->reset_token()->offsetGet('token');
+		assertNotEquals('', $oldToken);
```
ildyria:
```suggestion
		assertNotNull($oldToken);
```

tests/Feature/UsersTest.php
```diff
@@ -14,10 +14,12 @@
 
 use App\Legacy\AdminAuthentication;
 use App\Models\Configs;
+use App\Models\User;
 use App\SmartAlbums\PublicAlbum;
 use App\SmartAlbums\StarredAlbum;
 use App\SmartAlbums\UnsortedAlbum;
 use Illuminate\Support\Facades\Auth;
+use function PHPUnit\Framework\assertNotEquals;
```
ildyria:
```suggestion
use function PHPUnit\Framework\assertNotEquals;
use function PHPUnit\Framework\assertNotNull;
```

ildyria (approved):
LGTM.

app/Http/Controllers/Administration/UserController.php
```diff
@@ -139,4 +139,48 @@ public function getEmail(): array
...
+	/**
+	 * Returns the currently authenticated user.
+	 *
+	 * @return User
+	 */
+	public function getAuthenticatedUser(): User
+	{
+		/** @var User */
+		return Auth::user();
+	}
```
nagmat84: Do we still need this method? (More a reminder for myself to check it later.)

app/Http/Middleware/VerifyCsrfToken.php
```diff
@@ -23,38 +25,28 @@ class VerifyCsrfToken extends Middleware
...
-				return parent::handle($request, $next);
-			}
+		$token = $request->header('Authorization');
+		if ($request->is('api/*') && is_string($token)) {
+			/** @var User|null $user */
+			$user = User::query()
+				->where('token', '=', hash('SHA512', $token))
+				->first();
+			if ($user instanceof User) {
+				Auth::loginUsingId($user->id);
```
nagmat84: Authentication as the corresponding user is something new. I understand the intention of the code, but I am unsure whether we are good here.

Normally, the authenticated user is stored as part of the session. The session ID is persisted as a cookie between individual requests. Some other middelware re-reads the cookie value and starts the session (`\Illuminate\Session\Middleware\StartSession`) and still another one authenticates the user stored in that session (`\Illuminate\Session\Middleware\AuthenticateSession`). Here we interfere with this authentication. I don't know (yet) in which order these middelwares are executed of if there is even a deterministic order. I guess the middleware which is executed last and calls `Auth::login` "wins".

My question is: What happens if there is a session with a stored user inside it and the user authenticated via the session and the user authenticated via the API key are out-of-sync?
This would lead to some kind of unintentional re-authentication as a different user. Normally, Laravel requires us to call `Session::flush` and therewith destroy the session for intentional changes in the authenticated user. Obviously, that is nothing what we can or should do here, because we don't want to destroy the session.

However, I am wondering whether we need to take additional steps of precaution here and in `\Illuminate\Session\Middleware\AuthenticateSession`. Something like checking if there is a stored user in the session and a user associated to the API key, both user objects are referring to the same user and if not bail out.

The following somewhat "freaky" scenario is on my mind:

- A client starts a session and authenticates a some user A
- A client uses the same session ID and sends the API key for another user B

Of course, this should not happen for a well-behaving client, but you never know. Such crazy things do happen.

database/migrations/2022_06_12_075709_add_token_to_user_table.php
```diff
@@ -0,0 +1,48 @@
...
```
nagmat84: See https://github.com/LycheeOrg/Lychee/pull/1368#discussion_r938104216

> The ongoing problem. SQLite does not support dropping columns yet. Dropping a column is mimic by re-creating a table without that column. But this throws away all foreign key constraints.
> 
> For the time being, we must abstain from dropping columns. But luckily this is only in the downwards path, not the upwards path. Otherwise we had to add it to the list in https://github.com/LycheeOrg/Lychee/issues/1321.

nagmat84 (changes requested):
My main concern is addressed in the very long comment on the middleware check. Maybe my concerns are unsubstantiated, but at least I am missing a good justification why they are.

Moreover, as far as I see we have no test which actually tries to login with an API key. My concerns could also be part of a test which actually mimics the described "bad" client.

app/Http/Controllers/Administration/UserController.php
qwerty287: I think it's good to keep it because it allows clients to determine the current user. (But otherwise, no.)

app/Locale/German.php
```diff
@@ -55,6 +55,11 @@ public function get_locale(): array
 			'SET_MAP_PROVIDER' => 'Speichere Provider für OpenStreetMap Karten',
 			'FULL_SETTINGS' => 'Alle Einstellungen',
 			'UPDATE' => 'Update',
+			'RESET_TOKEN' => 'API-Schlüssel zurücksetzen',
+			'RESET' => 'Zurücksetzen',
+			'DISABLE' => 'Deaktivieren',
+			'ENABLE_TOKEN' => 'API-Schlüssel aktivieren',
+			'DISABLED' => 'deaktiviert',
```
nagmat84: I just had a closer look here, because the German translation for the key `DISABLED` does not work in the frontend as expected. But this is a bug in the frontend. Nonetheless another thought came to my mind.

Maybe we should use more specific keys here in order to avoid a clash of keys or strange keys in the future. Something more like

```suggestion
			'DISABLE_TOKEN_TOOLTIP' => 'Deaktivieren',
			'ENABLE_TOKEN' => 'API-Schlüssel aktivieren',
			'DISABLED_TOKEN_STATUS_MSG' => 'Deaktiviert',
```

app/Http/Middleware/VerifyCsrfToken.php
qwerty287: @nagmat84 I added a test which also tries this in https://github.com/LycheeOrg/Lychee/pull/1368/commits/b79293fdbdbe1b5383e2fddf228c64011ecad850.

The behavior is that the user is switched then.

qwerty287:
@nagmat84 would be good if you also re-review https://github.com/LycheeOrg/Lychee-front/pull/300.

app/Http/Controllers/Administration/UserController.php
nagmat84: I don't mind whether we keep it or not. It's your call.

But if we keep it, we should have one or two tests for it, i.e. one test which checks for the unauthenticated user that the JSON response is `null`, another test which checks for an authenticated user that the expected user object is returned. Without tests this method is just dead code and will become an "unbeloved child" in future updates. I agree that this method is so trivial, that it hardly will ever contain a bug, but the devil is a squirrel. Moreover, me secret goal is to push test coverage over 85%. So we should test what we can test and here this method even implements an external interface.

app/Http/Middleware/VerifyCsrfToken.php
nagmat84: Thanks. At least nothing bad seems to happen.

Still I would like to run this through the step debugger and have a closer look at the `AuthenticateSession` middleware, if I can spot anything which might look suspicious like a lurking bug. I will do so this weekend.

app/Http/Controllers/Administration/UserController.php
qwerty287: https://github.com/LycheeOrg/Lychee/pull/1368/commits/daf923861124f33247def943d181ba8e3b7c852e

nagmat84 (commented):
No text (just a "review" such that GH remembers what I have already seen)

app/Http/Middleware/HasUserCheck.php
```diff
@@ -0,0 +1,44 @@
...
+		if (!$this->isInstalled->assert()) {
+			return $next($request);
+		}
+
+		if (!Auth::hasUser()) {
+			throw new UnauthorizedException('Requires authentication');
+		}
+
+		return $next($request);
+	}
```
nagmat84: What is the purpose of this middleware? What does this middleware ensure what is not already ensured by other middleware?

app/Models/User.php
```diff
@@ -32,6 +32,7 @@
  * @property string|null                                           $email
  * @property bool                                                  $may_upload
  * @property bool                                                  $is_locked
+ * @property string|null                                           $token
```
nagmat84, line 35: The `has_token` property is missing. Moreover `'token'` still needs to be added to the array `$hidden` to prevent sending it to the frontend.

nagmat84 (changes requested):

app/Models/User.php
nagmat84: Fixed this in https://github.com/LycheeOrg/Lychee/pull/1368/commits/871eb59816a065adc81200fec7ce1883de73d1c7 myself. IMHO this can be resolved.

app/Http/Middleware/HasUserCheck.php
qwerty287: That there's an authenticated user. Otherwise, on the `api-admin` routes with disabled `AdminCheck` middleware, the endpoints are accessible without authorization. This can lead to 500 errors because `Auth::user()` returns `null`.

nagmat84: > Otherwise, on the `api-admin` routes with disabled `AdminCheck` middleware, the endpoints are accessible without authorization.

This sounds like a bug somewhere else. In particular, all non-admin routes (e.g. `Photo::setTitle`) don't use the `AdminCheck` middleware, but still require authentication and these routes have done so for quite a while. I don't get why we suddenly need _another_ middleware to solve a problem which has already been solved many times for other routes without this new middleware.

> [Otherwise, ] This can lead to 500 errors because Auth::user() returns null.

Could you please give a specific example where and when this happens? This sounds like a bug in a particular route.

In particular, the only "old" routes which are inside the `api-admin` group for historic reasons, but explicitly exclude the `AdminCheck` middleware are `User::setEmail` and `User::getEmail`. If no user is authenticated (even no non-admin user) these routes throw an `UnauthenticatedException` and return a HTTP status code 401 which is just fine.

The only other routes which are inside the `api-admin` group, but exclude the `AdminCheck` middleware are the new routes `User::getAuthenticatedUser`, `User::resetToken` and `User::unsetToken` which are added bis this PR. If these routes generate a 500 Server Error then they are broken somehow and we should fix that instead of trying to work around it.

nagmat84: Never mind. Fixed it myself in https://github.com/LycheeOrg/Lychee/pull/1368/commits/8e9cde160a630c275ac528bbd062d1bfbe51b33d.

nagmat84 (commented):
The only thing which remains to be done now, is step-debugging through user authentication middleware.

app/Http/Middleware/VerifyCsrfToken.php
nagmat84: I fixed the concerns I raised above in https://github.com/LycheeOrg/Lychee/pull/1368/commits/c5d2ff62ccfa44319f06b7ad228a2e4f8046d0a1.

It took me hours to find out how Laravel Authentication is supposed to work. The new guard should also handle all "bad" cases correctly. I haven't tried it yet, just coded. (It took longer than expected.) So chances are that there are some lurking bugs.

@qwerty287 You should play around with it, if this PR still does what you want it to do after the change.

nagmat84:
While I believe that the code is correct, I am currently unable to write proper tests for it. Again, the problem is that the application object survives between tests. As we already have had this problem once or twice and posted this question https://github.com/laravel/framework/discussions/44086.

In the past we could always help ourselves with some workaround, but in this case I am lost. I hope some answers.

nagmat84:
And another one: https://github.com/laravel/framework/discussions/44088

nagmat84:
I fixed the tests. While `SessionOrTokenGuard` has become more complicated than I initially thought, I am rather convinced that it should be fine. We have [96% test coverage](https://app.codecov.io/gh/LycheeOrg/Lychee/compare/1368/tree/app/Services/Auth/SessionOrTokenGuard.php) and the only two lines which are not tested affect the "remember-me" token which we don't use.

Only two things remain:

@qwerty287 please check if you are fine with my changes and if the PR still does what you want it to do.

The issue https://github.com/LycheeOrg/Lychee/pull/1368#discussion_r966251712 needs to be addressed.

qwerty287:
How should I fix the sqlite issue? Don't remove the column if it's SQLite?

Changes are fine otherwise.

nagmat84:
> How should I fix the sqlite issue? Don't remove the column if it's SQLite?
> 

For the same issue in other cases, we decided to _not_ delete the column for _any_ DBMS in order to keep the tables consistent between SQLite, MySQL and PostgreSQL. Otherwise it will be a nightmare to clean it up later.

Simply add the column to https://github.com/LycheeOrg/Lychee/issues/1321 such that we do not forget about it. When I will be ready with an PR for https://github.com/LycheeOrg/Lychee/issues/1462, the tables need an extensive change such that we can clean up all the postponed changes, too.

nagmat84 (approved):

qwerty287:
Let's wait for the tests and merge then.

qwerty287:
Hmm GH Actions seems not to work :thinking:

PR #1371, opened 2022-06-13
Fix untranslatable strings

qwerty287:
- allow translation of untranslatable strings
- remove some duplications
- fix and improve some German strings
- Never concat strings: Use a template system (`%s`, `%d`...) instead of just writing multiple strings behind each other


ildyria:
> Never concat strings: Use a template system ( %0, %1...) instead of just writing multiple strings behind each other

I like the idea and intent, but I am not sure of the execution as we will also be using those strings in Php.
I would rather use sprintf format (`%s` etc) and make use of e.g. https://www.npmjs.com/package/sprintf-js

ildyria:
> Never concat strings: Use a template system ( %0, %1...) instead of just writing multiple strings behind each other

I like the idea and intent, but I am not sure of the execution as we will also be using those strings in Php.
I would rather use sprintf format (`%s` etc) and make use of e.g. https://www.npmjs.com/package/sprintf-js

qwerty287:
~~Does PHP has a native template system (or a library similar to the JS lib you linked)? Because the current system supports PHP as well using `str_replace`.~~ Found it, https://www.php.net/manual/de/function.sprintf.php.

qwerty287:
@ildyria Updated it in https://github.com/LycheeOrg/Lychee/pull/1371/commits/a12bf2aa5ea344285c83285524b2b8fec2de936f.

nagmat84:
I haven't had a look at it. I like the idea and it isa step in the right direction. It is an improvement in any case compared to what has been there before.

In the long run (not this PR) we should try to replace `sprintf` by a proper translation engine which also supports multiple plural and inflections. For example in the pattern `'Do you want to delete %d photo(s)'`, the object "photo" must be in singular for "1" but in plural form for "0" and any number greater than one. Different rules hold for other languages. (I believe Russian has different forms for one, two and anything else.) Symfony (my framework of choice when in comes to PHP) has a great engine which supports nearly all languages, provides replacement for named place holders (i.e. something like `'Do you want to delete %number_of_photos% photo(s)'`) and correctly handles plural forms and inflections. I hope/assume that Laravel provides something similar.

qwerty287:
Our core issue is that we always need to work with two platforms: PHP and JS. We need easy support for both. If we fully switch over to Livewire, this is easier I believe. With a quick internet search I found https://www.codeandweb.com/babeledit/tutorials/how-to-translate-your-laravel-project which looks promising. It supports pluralization etc. in any way (yes, the Russian way is supported as well). This issue probably needs to wait for a full-livewire frontend.

ildyria (approved):
PHP side LGTM.

ildyria:
I would love to merge this soon, but I think we should wait for https://github.com/LycheeOrg/Lychee/pull/1382

ildyria:
I see that you are using `%1$s` and `%2$s`. Unfortunately those are not compatible with php sprintf. :(

```php
echo sprintf("Weet je zeker dat je dit album wilt samenvoegen '%1$s' met het album '%2$s'?", 'test', 'test');
```
Leads to:
> NOTICE Undefined variable: s on line number 3
> NOTICE Undefined variable: s on line number 3
> Weet je zeker dat je dit album wilt samenvoegen ' met het album '?

The reason being that $s is being interpolated from php variables.
I strongly believe this should be :
```php
echo sprintf("Weet je zeker dat je dit album wilt samenvoegen '%s' met het album '%s'?", 'test', 'test');
```
> Weet je zeker dat je dit album wilt samenvoegen 'test' met het album 'test'?

qwerty287:
While you're right, this doesn't affect us, I believe. The `$s` is part of the string and not parsed as variables. The reason is that we use `'` instead of `"` which does not support variable interpolation.
If you run:
```php
echo sprintf('Weet je zeker dat je dit album wilt samenvoegen \'%1$s\' met het album \'%2$s\'?', 'test', 'test');
```
the output is:
```
Weet je zeker dat je dit album wilt samenvoegen 'test' met het album 'test'?
```

That's exactly how we are doing it.

ildyria:
Thanks, TIL. :)

ildyria (approved):
PHP & JS side LGTM & tested.

PR #1374, opened 2022-06-16
Add "Clear" button to logs

qwerty287:
https://github.com/LycheeOrg/Lychee-front/pull/302


qwerty287:
CI failure is unrelated.

app/Locale/Dutch.php
```diff
@@ -46,6 +46,7 @@ public function get_locale(): array
 			'DIAGNOSTICS_GET_SIZE' => 'Request space usage',
 			'LOGS' => 'Laat logs zien',
 			'CLEAN_LOGS' => 'Clean Noise',
+			'CLEAR' => 'Leeren',
```
kamil4: My Dutch is a little rusty but I don't believe that's the right word, and the official Dutch dictionary seems to agree (https://www.vandale.nl/gratis-woordenboek/nederlands/betekenis/leeren).

kamil4 (changes requested):

app/Locale/Dutch.php
qwerty287: Whoops looks like I changed Dutch instead of German ;) ("German" in German is "Deutsch", so I probably misread the filename)

ildyria (approved):
LGTM. fix conflict & can merge.

qwerty287:
@kamil4 you have to approve because you requested changes and I don't want to use the force-merge option as admin :)

kamil4 (approved):

PR #1375, opened 2022-06-21
Synchronize/update the list of PHP extensions

kamil4:
I went through the `composer.lock` and I actually extracted the list of PHP extensions that are required, plus what Laravel claims to need at https://laravel.com/docs/8.x/deployment#server-requirements.

Overall, the list in `composer.json` was close (other than the not actually required ZIP). The lists in the installer and the Diagnostics page were a lot further from complete, and I also added the requirements of our various dependencies (which don't need to be listed in `composer.json` since composer obviously takes care of them automatically). I duplicated those lists across two source files which is admittedly ugly but I know that @nagmat84 intends to rewrite the installer so I saw no point in refactoring this code at this point.

There will be a separate PR to our online docs, which contain yet another list...

### DIFF
--- a/app/Actions/Diagnostics/Checks/PHPVersionCheck.php
+++ b/app/Actions/Diagnostics/Checks/PHPVersionCheck.php
@@ -44,7 +44,26 @@ class PHPVersionCheck implements DiagnosticCheckInterface
 		}
 
 		// Extensions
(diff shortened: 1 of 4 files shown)


ildyria (approved):
LGTM.

PR #1386, opened 2022-06-29
Fix GD (broken by #1336)

kamil4:
The first victim of merging #1336 is here :wink:: the GD image handler didn't work anymore for images not requiring rotation...


nagmat84 (approved):
Luckily, this was a one-liner. But we should really get #1351merged (after I have gone through merging hell). #1351 contains a bunch of new tests for Imagick _and_ GD incl. all sorts of auto-rotation in every combination. That why I added `orientation-xxx.jpg` and `{h|v]flip.jpeg` to [tests/Samples](https://github.com/LycheeOrg/Lychee/tree/master/tests/Samples). This will prevent us from problems like this in the future.

PR #1387, opened 2022-06-29
Ensure a local timestamp for mov videos

kamil4:
I don't know how long this has been broken, but probably a good while. I'm sure it worked when I first merged #680; I suppose it may have been broken since timezone support got merged in #1022 (but then, why wouldn't I have tested it at the time?).

Anyway, the issue is that for MOV files ffprobe outputs their timestamp with "Z" at the end even though typically the timestamp is actually local. Since we now support timezones, we actually need to explicitly correct this.

BTW, I'm also considering using exiftool for video files with ffprobe as a _fallback_ (right now the order is the opposite) as it seems to be able to extract more metadata in some cases (like the make and model of some cameras). But I need to test it on a larger sample to see if there are any regressions...

### DIFF
--- a/app/Metadata/Extractor.php
+++ b/app/Metadata/Extractor.php
@@ -240,7 +240,9 @@ class Extractor
 				//
 				// Here, we rely here on a simple filetype-based heuristics and,
 				// for a timestamp we suspect to be in UTC, we convert it to the
-				// application's default timezone.
+				// application's default timezone.  For a timestamp we suspect to
+				// be local, we ensure that the extractor didn't erroneously mark
+				// it as UTC.
 				// All other timestamps are not altered, but used "as is":
 				//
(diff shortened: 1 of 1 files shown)


ildyria (approved):

qwerty287 (approved):

nagmat84:
I was still reviewing.

ildyria:
oups.

nagmat84:
I guess there is an oversight in it. @all: If we have an "urgent" bug fix like the one about the sharing, it is fine for me if you merged it fast, but for everything else it would be nice, if a PR could at least wait a couple of days until it gets merged such someone has even the chance to do a review. I was just writing up my findings.

nagmat84:
OK, I will create a new PR.

PR #1393, opened 2022-07-02
remove old workflow runs

ildyria:
Add a simple action to nuke old workflows run.


d7415:
I don't see that they're causing a problem, but I guess we probably don't need 208 pages of old runs that no-one's going to read.

PR #1394, opened 2022-07-02
Make migration of filesize more robust wrt. broken DB

ildyria:
Throw an error (but do not stop the migration) if an image is missing the original link for some reason

### DIFF
--- a/database/migrations/2022_02_02_203008_filesize_size_variants.php
+++ b/database/migrations/2022_02_02_203008_filesize_size_variants.php
@@ -4,8 +4,6 @@ use Illuminate\Database\Migrations\Migration;
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Schema;
-use Symfony\Component\Console\Output\ConsoleOutput;
-use Symfony\Component\Console\Output\ConsoleSectionOutput;
 
 /**
  * Add a filesize column for size variants to make it
@@ -16,20 +14,9 @@ use Symfony\Component\Console\Output\ConsoleSectionOutput;
  */
 class FilesizeSizeVariants extends Migration
 {
-	private ConsoleOutput $output;
-	private ConsoleSectionOutput $msgSection;
-
-	public function __construct()
-	{
-		$this->output = new ConsoleOutput();
-		$this->msgSection = $this->output->section();
-	}
-
 	private const VAR_TAB = 'size_variants';
 	private const PHOTO_FK = 'photo_id';
 	private const TYPE_COL = 'type';
-	private const WIDTH_COL = 'width';
-	private const HEIGHT_COL = 'height';
 	private const ID_COL = 'id';
 	private const TYPE_ORIGINAL = 0;
 	private const PHOTOS_TAB = 'photos';
@@ -40,7 +27,7 @@ class FilesizeSizeVariants extends Migration
 	 *
 	 * @return void
 	 */
-	public function up()
+	public function up(): void
 	{
 		// To avoid doing I/O on every photo, which would be prohibitive
 		// on large instances, and because JPEG compression makes approximations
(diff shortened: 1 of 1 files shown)


ildyria:
When running the migration, I get (which is the expected behaviour).

```
▶ ./artisan migrate       
**************************************
*     Application In Production!     *
**************************************

 Do you really wish to run this command? (yes/no) [no]:
 > yes

Migrating: 2022_02_02_203008_filesize_size_variants
Error: FHaFPwDs767DAiT9v1dzgAUk does not have an original attached. Please double check your Database.
Migrated:  2022_02_02_203008_filesize_size_variants (56,227.45ms)
Migrating: 2022_02_22_194700_fix_sorting_config
Migrated:  2022_02_22_194700_fix_sorting_config (275.10ms)
Migrating: 2022_04_06_091900_drop_objectionable_indices
Migrated:  2022_04_06_091900_drop_objectionable_indices (3,986.66ms)
Migrating: 2022_04_13_094611_add_track_short_path_to_album_table
Migrated:  2022_04_13_094611_add_track_short_path_to_album_table (709.10ms)
Migrating: 2022_04_16_170724_add_missing_indices
Migrated:  2022_04_16_170724_add_missing_indices (2,643.37ms)
Migrating: 2022_04_16_174503_bump_version040501
Migrated:  2022_04_16_174503_bump_version040501 (64.47ms)
Migrating: 2022_04_18_150400_add_index_for_delete
Migrated:  2022_04_18_150400_add_index_for_delete (2,259.94ms)
Migrating: 2022_04_18_174417_fix_live_photo_short_path
Migrated:  2022_04_18_174417_fix_live_photo_short_path (97.26ms)
```

database/migrations/2022_02_02_203008_filesize_size_variants.php
```diff
@@ -14,6 +16,15 @@
  */
 class FilesizeSizeVariants extends Migration
 {
+	private ConsoleOutput $output;
+	private ConsoleSectionOutput $msgSection;
+
+	public function __construct()
+	{
+		$this->output = new ConsoleOutput();
+		$this->msgSection = $this->output->section();
```
qwerty287: Would it be possible to use
```suggestion
		$this->msgSection = new ConsoleOutput()->section();
```
and remove the variable in line 19?

database/migrations/2022_02_02_203008_filesize_size_variants.php
```diff
@@ -57,9 +68,13 @@ public function up()
 				->where(self::TYPE_COL, '=', self::TYPE_ORIGINAL)
 				->first();
 
-			DB::table(self::VAR_TAB)
+			if ($original_variant === null) {
+				$this->msgSection->writeln('<error>Error:</error> ' . $photo->id . ' does not have an original attached. Please double check your Database.');
+			} else {
+				DB::table(self::VAR_TAB)
 				->where(self::ID_COL, '=', $original_variant->id)
 				->update([self::SIZE_COL => $photo->filesize]);
```
kamil4: Isn't the indentation wrong?

kamil4 (commented):
Overall I think this is reasonable; I do believe that we should be coding the migrations as defensively as practical so that they don't destroy users' databases on some odd errors.

I'm not of fan of the title of this PR (clearly, it's the DB that's broken, not the migration; BTW, it would be interesting to figure out how that brokeness came to be), but that's not a deal-breaker...

nagmat84:
I just realized that this migration is highly inefficient and makes two times more SQL queries than necessary.

In particular fetching the original size variant from the DB is completely unnecessary. Fixing that would also indirectly solve the problem  of a non-existing original size variant, because the migration wouldn't notice a missing original size variant.

Give me one or two days. I will have a look into it in the evenings. Obviously, I have not reviewed this migration when it had been merged initially.

nagmat84:
@ildyria Feel free to test this variant of the PR. I squashed the whole copying into a single SQL query.

I also could not stand the title of this PR so I changed it. :see_no_evil:

kamil4 (approved):
The new SQL statements exceed my meager skills so I guess OK?

nagmat84:
Personally, I believe that all these class constants and raw statements make the SQL unreadable. Actually, it is rather simple. This is the upward migration in pure SQL

```sql
UPDATE size_variants
SET size_variants.filesize = ( SELECT filesize FROM photos WHERE photos.id = size_variants.photo_id )
WHERE size_variants.type = 0;
```

Is this clearer? The outer update query sets the `filesize` of the original size variants. The new value is picked from `filesize` in the photos of the corresponding entry.

Sometimes I feel that the Laravel query builder makes things more complicated than writing the native SQL query directly.

PR #1395, opened 2022-07-03
Add some tests

qwerty287:
Try to increase test coverage by adding some (partially useless) tests.


tests/Feature/LangTest.php
```diff
@@ -35,5 +38,21 @@ public function testLang(): void
...
+		self::assertEquals('en', $lang->get_code());
+
+		DB::table('configs')->insert([
+			[
+				'key' => 'lang',
+				'value' => 'en',
+				'confidentiality' => 0,
+				'cat' => 'Gallery',
+			],
+		]);
```
nagmat84, line 59: Only after inspecting the code I got what you wanted to test here. Instead of adding this test code to the already existing but imprecisely named method `testLang` I would rather move this code into its own test method like `testEnglishAsFallbackIfLangConfigIsMissing`. I know this method name is rather long, but it really helps reader to understand what you want to test. `testLang` is simply so untelling.

Moreover, there are authors who advocate that there must be only a single assertion per test as otherwise the test is too large for being a single test. I wouldn't go that far, because sometime you simply need multiple assertions for a test to make sense. Hence, I only consider this only a rule of thumb, but the underlying idea is correct.

Lychee is a prime example which simply puts to many and unrelated tests into a single test method. While it is tedious to try to repair that for the existing tests, we should try to do better for newly written tests.

tests/Feature/SearchTest.php
```diff
@@ -0,0 +1,97 @@
...
+use Tests\TestCase;
+
+class SearchTest extends TestCase
+{
+	/**
+	 * Test searching is working.
+	 *
+	 * @return void
+	 */
+	public function testSearch(): void
```
nagmat84: Again, please split this test methods into multiple test methods, see previous comment. If I am not mistaken, I am able to identify at least three test cases

- `testSearchPhotoByTitle`
- `testSearchPhotoByTag`
- `testSearchAlbumByTag`

This should at least be three different test methods with proper names.

tests/Feature/SearchTest.php
```diff
@@ -0,0 +1,97 @@
...
+		);
+		$photos_tests->set_title($photoId, 'photo search');
+
+		$response = $this->postJson(
+			'/api/Search::run',
+			['term' => 'search']
+		);
+		$response->assertStatus(200);
+
+		$response->assertJson([
```
nagmat84, line 53: This only works accidentally, because it implicitly assumes that there are no other photos and albums which match the search. This is indeed the case, because the other tests and test classes properly clean up after themselves.

In PR https://github.com/LycheeOrg/Lychee/pull/1351 I have added a special test class [`PhotoTestBase`](https://github.com/LycheeOrg/Lychee/blob/use_filestreams/tests/Feature/Base/PhotoTestBase.php) for that. I would like to wait with this PR until https://github.com/LycheeOrg/Lychee/pull/1351 has been merged and then this test class should extend `PhotoTestBase`.

tests/Feature/SettingsTest.php
```diff
@@ -0,0 +1,74 @@
...
+namespace Tests\Feature;
+
+use App\DTO\SortingCriterion;
+use App\Facades\AccessControl;
+use App\Http\Requests\Settings\SetSortingRequest;
+use Tests\TestCase;
+
+class SettingsTest extends TestCase
+{
+	public function testSetSorting(): void
```
nagmat84, line 22: Again, please split this method. I see at least

- `testSetSorting` (the postiive case)
- `testSetSortingWithIllegalAlbumAttribute`
- `testSetSortingWithIllegalPhotoAttribute`
- `testSetSortingWithUnknownOrder`

nagmat84 (changes requested):

ildyria:
I would wait for #1403 as it changes Legacy quite a lot and some of those code may already have changed.

nagmat84 (commented):
I had a look at the changes since my last review. Everything fine so far except the one open comment which can only be closed after https://github.com/LycheeOrg/Lychee/pull/1351. Except from that I am fine with the PR.

However, it would be possible to even improve on this PR further (not I "must", only something which crossed my mind). The tests for the search feature only check whether the uploaded photo is found. There are no negative tests, i.e. that an uploaded photo which does not match the search criteria is not found. This means a silly implementation of the search feature which always simply returns the whole DB without actually searching anything would also pass the tests. Hence, what I would have done is to upload two photos in each of the tests, e.g. the Night photo and the Mongolia photo, search and check that the search result only contains one element and the correct element. But I also understand that this might be out of the scope of this PR. I am already happy that we have more tests.

tests/Feature/SearchTest.php
nagmat84: I fixed that in https://github.com/LycheeOrg/Lychee/pull/1395/commits/cde6d6326434225d5f8f1f091985d22a987cd5a2

nagmat84:
> However, it would be possible to even improve on this PR further (not I "must", only something which crossed my mind). The tests for the search feature only check whether the uploaded photo is found. There are no negative tests [...] 

Added negative tests myself in https://github.com/LycheeOrg/Lychee/pull/1395/commits/e5d042a7f06fc8a54d173efc15b3ddcd5b1ae51f.

nagmat84 (approved):

nagmat84:
> I would wait for #1403 as it changes Legacy quite a lot and some of those code may already have changed.

@ildyria I assume you are speaking about the changes introduces by this PR in https://github.com/LycheeOrg/Lychee/pull/1395/commits/f732f5dd4695524008a464a2817c23f849b9c208?

I am not sure why @qwerty287 made this part of this PR. To me these changes seem unrelated to the purpose of this PR. I assume if https://github.com/LycheeOrg/Lychee/pull/1395/commits/f732f5dd4695524008a464a2817c23f849b9c208 was reverted, then the only changes introduced by this PR are more tests and the PR is fine for merge, right?

ildyria:
> @ildyria I assume you are speaking about the changes introduces by this PR in https://github.com/LycheeOrg/Lychee/commit/f732f5dd4695524008a464a2817c23f849b9c208?

Yes.

nagmat84:
@qwerty287 Do you mind to revert https://github.com/LycheeOrg/Lychee/commit/f732f5dd4695524008a464a2817c23f849b9c208? I am not sure if this should be part of this PR at all and it will be mostly outdated after https://github.com/LycheeOrg/Lychee/pull/1403. Without this commit, we could merge this PR right away.

qwerty287:
This is not related to the tests. But I noticed that these methods will *always* return the same result. The removed methods `noLogin` and `SetPassword` return always `false` (due to the version check). Thus I removed them and their usages. I can revert it in a few days.

qwerty287:
@nagmat84 @ildyria Reverted.

PR #1405, opened 2022-07-11
Improve safety and reliability of lychee:ghostbuster

nagmat84:
While working on my other PR, I noticed that `lychee:ghostbuster` became a victim of the recent addition of PHPstan. The comparison with `=== 1` or `=== '1'` is not reliable, because sometimes the framework returns an `int` and sometimes a `string`.

In case of `dryrun` this is extremely dangerous, because `dryrun` could accidentally be set to `false`. This PR should fix it.

### DIFF
--- a/app/Console/Commands/Ghostbuster.php
+++ b/app/Console/Commands/Ghostbuster.php
@@ -59,9 +59,15 @@ class Ghostbuster extends Command
 	public function handle(): int
 	{
 		try {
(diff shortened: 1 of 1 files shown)


ildyria:
> because sometimes the framework returns an `int` and sometimes a `string`.

FFS...

ildyria (approved):
LGTM.

PR #1407, opened 2022-07-16
SQLite does not return a `int` on `getCode()` but a `string`.

ildyria:
To trigger the bug:

- git clone ...
- composer install
- config for SQLite
- artisan migrate
- `chmod/chown -R storage`
- `chmod/chown uploads`
- try to login as new admin => Error 500 because `database/database.sqlite` is not web-user writable and SQLite code is a `string`: `H00`.


app/Exceptions/LycheeBaseException.php
```diff
@@ -53,6 +53,11 @@ abstract class LycheeBaseException extends HttpException implements ExternalLych
...
-		parent::__construct($httpStatusCode, $message, $previous, [], $previous !== null ? $previous->getCode() : 0);
+		$code = null;
+		if ($previous !== null) {
+			// Some Throwable will throw strings intead of int: SQLite when failing on read only DB for example.
+			$code = is_int($previous->getCode()) ? $previous->getCode() : 0;
+		}
+		parent::__construct($httpStatusCode, $message, $previous, [], $code ?? 0);
 	}
-}
+}
```
nagmat84:
```suggestion
}

```

nagmat84 (approved):

d7415:
I haven't looked at where this is used, but I guess `0` is not success in that context?

ildyria:
When you hit an exception, you are pretty much in a "not success" state anyway... :laughing:

nagmat84:
Zero is the default, i.e. it is fine to set it to `0`. The meaning of the code actually depends on the type of the exception itself. E.g. error code `500` for an `SQLException` might mean something entirely different than for an `HttpException`. It is not used by code except for putting it into logs.

PR #1403, opened 2022-07-11
Use Laravel Auth facade.

ildyria:
~This is preliminary work for some refactoring later removing the need of `SessionFunctions` in favor of the `Auth` facade.~
Remove the `AccessControl.php` facade in favor of the `Auth::` and `Gate::` Laravel facades ~~`App/Auth/Authorization.php` which is an extensions of Laravel `Auth` facade`~~.

### Rationale of using Auth
`Auth` is unfortunately necessary for the use of https://github.com/Laragear/WebAuthn as it is used in the middleware provided.
I would rather avoid to have a clone version of WebAuthn adapted to our need as I previously did for [Larapass](https://github.com/LycheeOrg/Larapass) as it keeps us out of the loop for updates etc.

As for the reason to use Laragear/WebAuthn instead of Larapass: the later is abandoned. As a result in order to upgrade to Laravel 9.0, we need to support Laragear/WebAuthn. Do note that this will also enable us to use Safe functions 2.2 instead of locked at 1.3.

I know that @nagmat84 intends to rewrite the full Auth section.
I would also like to point out that https://laravel.com/docs/9.x/passport would be an interesting addition in order to support Oauth2 in the future (and make some of our users happy in that regards).

With regard to `Auth`: https://laravel.com/docs/9.x/authentication
With regard to `Gate`: https://laravel.com/docs/9.x/authorization

PR #1414, opened 2022-07-24
Authorization tests

nagmat84:
This PR adds a lot of tests to increase the test coverage of `AlbumAuthorizationProvider` and `PhotoAuthorizationProvider`. This PR does not only aim at maximizing line coverage, but also tries to test many scenarios which do not contribute to a higher line coverage, because the decisions are made on the SQL layer.

All but one change are in the `tests/` directory. The only change within productive code affects `AlbumAuthorizationProvider`. This PR removes the parameter `$origin` from `applyBrowsabilityFilter(AlbumBuilder $query, ?Album $origin = null)`, because the method is only called in one place without using this parameter. Hence, this has been dead code.

*Potential issues which may need clarification*

During writing the tests I encountered two potential issues where the test result is questionable. Moreover, I am not sure what the desired behavior should be. They do not constitute security issues. 

 1. Password-protected and locked albums do not appear in the album tree. They do appear after unlocking.
 2. Public (but not password-protected) albums below a public and password-protected, locked album are accessible by direct link, i.e. these albums behave like hidden albums.


ildyria (approved):
LGTM.

As this PR only adds tests, I would also consider directly merging it tomorrow and having the two TODO left as a stand alone PR. :)

This PR is blocking for multiple successors: 
#1403
#1454
#1357

so I would be in favor to put it on expedited tracks :)

tests/Feature/SharingSpecialTest.php
```diff
@@ -0,0 +1,493 @@
...
+				$albumInfo[0], $albumInfo[1], $albumInfo[2], $albumInfo[3]
+			));
+		}
+		foreach ([
+			$this->photoID1,
+			$this->photoID2,
+			$this->photoID3,
+			$this->photoID4,
+			$this->photoID5,
+			$this->photoID5,
```
kamil4, line 463: A typo: this should be `photoID6`

tests/Feature/SharingWithAnonUserAndPublicSearchTest.php
```diff
@@ -0,0 +1,157 @@
...
+			true,
+			$this->photoID1, [
+				$this->generateExpectedPhotoJson(static::SAMPLE_FILE_TRAIN_IMAGE, $this->photoID1, null),
+			],
+		));
+		$responseForStarred->assertJsonMissing(['id' => $this->photoID2]);
+
+		$responseForTree = $this->root_album_tests->getTree();
+		$responseForTree->assertJson($this->generateExpectedTreeJson());
+		$responseForTree->assertJsonMissing(['id' => $this->photoID2]);
```
kamil4, line 151: Shouldn't `albumID1` and `photoID1` be missing here as well?

tests/Feature/SharingSpecialTest.php
nagmat84: Well spotted.

I wonder how we include these findings after the PR has already been merged. I will probably create a bug-fix PR based on your findings after you have finished your review.

tests/Feature/SharingWithAnonUserAndPublicSearchTest.php
nagmat84: Good question. I need to double check. Photo 1 is public, starred and appears in the recent album. As the global option `public_search` is enabled, I assumed photo 1 would be shown as a thumbnail for the root (i.e. unsorted album) in the album tree. But maybe our album tree does never show a thumbnail for the root at all. I cannot remember by heart how it is implemented.

At least it would not be an error from the security perspective, _if_ photo 1 was shown as a thumbnail for the root.

But sure, if it isn't, then it should be added to the list in order to avoid confusion of reviewers.

tests/Feature/SharingSpecialTest.php
kamil4: Personally, I would include a fix in the next (possibly unrelated) PR, simply because this is not fixing any critical bug.

Anyway, I'm done with my quick walkthrough of this PR.

tests/Feature/SharingWithAnonUserAndPublicSearchTest.php
kamil4: Speaking of thumbs, I noticed that if we explicitly specify a thumb for an album, it is returned to the front end even if the album is password-protected _and still locked_. This does not appear to be a regression (I could reproduce it with 4.4.0); still, we should fix it.

ildyria: > Speaking of thumbs, I noticed that if we explicitly specify a thumb for an album, it is returned to the front end even if the album is password-protected _and still locked_. This does not appear to be a regression (I could reproduce it with 4.4.0); still, we should fix it.

It is not a bug, it is a _feature_. :D
I am totally fine with that behavior personally and wanted to make a PR for it, so if it is already working, then even better.
The rationale being that the user is voluntarily setting the image. That being said, I would make this behavior more explicitly known to the user (however this needs a revamp of the UI). :)

kamil4: I disagree. Yes, the user set an image as a preferred cover of an album, but they never agreed for that image being shown to every random Joe who happens to browse by.

I can set a cover image of a private album; does that mean that somehow that thumb should now be exposed to everybody, since I explicitly set it? Of course not. The access permissions take precedence.

I mean, sure, I can imagine a scenario where one might _want_ the current behavior (showing a "taster", if you will), but it must be possible to configure it; I also don't think that the current behavior should be the default.

nagmat84: Sorry, I didn't want to resolve the conversation. I just hit the wrong button. But it is unresolved again.

I totally agree with @kamil4. This is a security bug and should be fixed. Actually I am disappointed (of myself) that I did not consider this case on my tests.

PR #1416, opened 2022-07-24
Moved unit tests for root album into separate class

nagmat84:
While working on https://github.com/LycheeOrg/Lychee/pull/1414 I noticed that we do not have a test class which collects all API calls to the root album in the same way as we have for albums and photos. So far we only have had a method to call `Albums::get` as part of the (normal) albums test, but it also has been quite different in style from the typical methods we have. As I am going to need more API calls to the root album in https://github.com/LycheeOrg/Lychee/pull/1414, I first want to remedy this situation.

### DIFF
--- a/tests/Feature/Lib/RootAlbumUnitTest.php
+++ b/tests/Feature/Lib/RootAlbumUnitTest.php
@@ -1,0 +1,52 @@
+<?php
+
+/**
+ * We don't care for unhandled exceptions in tests.
+ * It is the nature of a test to throw an exception.
+ * Without this suppression we had 100+ Linter warning in this file which
+ * don't help anything.
+ *
+ * @noinspection PhpDocMissingThrowsInspection
+ * @noinspection PhpUnhandledExceptionInspection
+ */
+
+namespace Tests\Feature\Lib;
+
+use Illuminate\Testing\TestResponse;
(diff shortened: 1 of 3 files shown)


tests/Feature/Lib/RootAlbumUnitTest.php
```diff
@@ -0,0 +1,52 @@
...
+	 * @return TestResponse
+	 */
+	public function get(
+		int $expectedStatusCode = 200,
+		?string $assertSee = null,
+		?string $assertDontSee = null
+	): TestResponse {
+		$response = $this->testCase->postJson('/api/Albums::get');
+		$response->assertStatus($expectedStatusCode);
+		if ($assertSee) {
```
ildyria, line 43: Good thing that phpstan is not parsing those files. :laughing:

ildyria (approved):
LGTM.

tests/Feature/Lib/RootAlbumUnitTest.php
nagmat84: Yes, I had the exact same though. I also inclined to write `if ($asserSee !== null)`, but this is copy & paste from the other files and then I decided to keep it consistent. At some point in the future I would like to include the tests into PHPStan, but not now.

PR #1424, opened 2022-07-27
Improve cicd

ildyria:
- add a linter step to catch early php syntax error.
- Add more structure and unify the CI.
- Still run test even if code style fails.
- Add support for artifact upload on master branch (which aims to simulate a master release build).


.github/workflows/CICD.yml
```diff
@@ -0,0 +1,298 @@
...
+    paths-ignore:
+      - '**/*.md'
+      - 'public/dist/*.js'
+      - 'public/dist/**/*.js'
+  # Allow manually triggering the workflow.
+  workflow_dispatch:
+
+jobs:
+  kill_previous:
+    name: "0️⃣ Kill previous runs"
```
nagmat84: What is this strange character which appears between the ordinal number and the first letter of the caption? First I thought Github had some rendering problems, but I also see a placeholder for an unknown Unicode character in my IDE.

.github/workflows/CICD.yml
```diff
@@ -0,0 +1,298 @@
...
+          access_token: ${{ github.token }}
+
+  php_syntax_errors:
+    name: "1️⃣ PHP Syntax errors"
+    runs-on: "ubuntu-latest"
+    needs:
+      - "kill_previous"
+    steps:
+      - name: "Set up PHP"
+        uses: "shivammathur/setup-php@v2"
```
nagmat84: We use different syntax in this workflow file. Sometimes we enclose the value of a `uses`-dependency in quotation marks like here, sometimes we don't. (I believe those which have been copied over from the former `Build-yml`. Obviously, both seems to work and I am too unfamiliar with Github workflow files to tell what the recommended syntax is, but I would like to see consistency.

Same holds for other settings such as `name`

.github/workflows/CICD.yml
```diff
@@ -0,0 +1,298 @@
...
+    name: "2️⃣ PHP: 8.0 | PHPStan"
+    runs-on: "ubuntu-latest"
+    needs:
+      - "php_syntax_errors"
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@2.15.0
```
nagmat84: Here we have an example for not using quotation marks.

Moreover, we sometimes specify a particular version of a package like here `shivammathur/setup-php@2.15.0`, sometimes we only specify a lower major version even for the same package as in `shivammathur/setup-php@v2`. Again, i would like to see consistency here. Otherwise someone in the future may wonder if this was by intention or accidentally. I assume it is accidentally and should be unified.

(If it is on purpose, then at least add a comment to the workflow, why.)

.github/workflows/CICD.yml
```diff
@@ -0,0 +1,298 @@
...
+  tests:
+    name: "2️⃣ PHP ${{ matrix.php-version }} - ${{ matrix.sql-versions }}"
+    needs:
+      - "php_syntax_errors"
+    runs-on: "ubuntu-latest"
+    strategy:
+      matrix:
+        php-version:
+          - "8.0"
+          - "8.1"
```
nagmat84: Another example for inconsistency. Here the versions are enquoted, but they are not somewhere else.

.github/workflows/CICD.yml
```diff
@@ -0,0 +1,298 @@
...
+    needs:
+      - "php_syntax_errors"
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@2.15.0
+        with:
+          php-version: 8.0
```
nagmat84, line 82: Here the version number is written without quotes (cp. with comment below).

nagmat84 (commented):
I only have limited understanding of Github workflow files. I checked out the old `Build.yml` and did a manual diff with the new `CICD.yml` on my local machine as Github does not recognize that `CICD.yml` is basically a merger of `Build.yml` and `phpstan.yml` plus some additions. From what I have seen it looks sane.

As I cannot say much about the workflow on the semantic level, I only have some less important comments about the syntax. I simply noticed some inconsistencies.

ildyria:
I will standardize the strings with double quotes (even though they are not necessary per say).

.github/workflows/CICD.yml
ildyria: Emoji Unicode. :)
https://emojipedia.org/keycap-digit-zero/
https://emojipedia.org/keycap-digit-one/
https://emojipedia.org/keycap-digit-two/
https://emojipedia.org/keycap-digit-three/
...

ildyria:
The big difference is that there are dependencies between the different parts of the runs to potentially fail early.
https://github.com/LycheeOrg/Lychee/actions/runs/2768728603
image

I just hope that the build artifact step will work on master. This should allow us to have a "release-like" zip file built on each commit on master, potentially allowing us to give the "lastest" build from master without having to build a formal release. :)

.github/workflows/CICD.yml
nagmat84: Seems as if I have to install a better/new font.

nagmat84 (approved):
With my little understand it looks good. Maybe we should just give it a try. :see_no_evil:

nagmat84:
> I will standardize the strings with double quotes (even though they are not necessary per say).

I believe you decided for _without_ double quotes.

.github/workflows/CICD.yml
```diff
@@ -0,0 +1,298 @@
...
+      - name: Codecov
+        uses: codecov/codecov-action@v1
+
+      - name: Make sure we can go backward
+        run: php artisan migrate:rollback
+
+  dist:
+    name: 2️⃣ PHP ${{ matrix.php-version }} - ${{ matrix.sql-versions }} - Dist
+    needs:
+      - code_style_errors
```
nagmat84, line 176:
```suggestion
      - php_syntax_errors
```

nagmat84 (changes requested):
Sorry, I just spotted that after I had a look on the dependency graph. I believe we want to check for syntax errors here.

.github/workflows/CICD.yml
```diff
@@ -0,0 +1,298 @@
...
+        uses: ramsey/composer-install@v2
+
+      - name: Check source code for syntax errors
+        run: vendor/bin/parallel-lint --exclude .git --exclude vendor .
+
+  code_style_errors:
+    name: 1️⃣ PHP - Code Style errors
+    runs-on: ubuntu-latest
+    needs:
+      - kill_previous
```
nagmat84:
```suggestion
    needs:
      - php_syntax_errors
```

Shouldn't code style depend on syntax? I mean what sense does it make to check code style like correct indention if the syntax, e.g. closing and opening braces, are wrong?

.github/workflows/CICD.yml
```diff
@@ -0,0 +1,298 @@
...
+          php artisan migrate:rollback
+
+  createArtifact:
+    name: 3️⃣ Build Artifact
+    needs:
+      - phpstan
+      - dist
+      - tests
+    runs-on: ubuntu-latest
+    if: (github.event_name == 'push' && github.event.pull_request.head.repo.full_name != github.repository && startsWith(github.ref, 'refs/heads/master'))
```
nagmat84: Given your comment about hoping that it runs on `master`, why not running it for every branch? Probably, nobody will ever need the resulting ZIP file of another branch than `master`, but this would give us the guarantee that it works. BTW: Where is the ZIP file finally put? Inside the branch or in a different location? Also see comment below.

.github/workflows/CICD.yml
```diff
@@ -0,0 +1,298 @@
...
+
+      - name: Build Dist
+        run: |
+          make clean dist
+
+      - name: Upload a Build Artifact
+        uses: actions/upload-artifact@v3.1.0
+        with:
+          name: Lychee-Dist.zip
+          path: Lychee.zip
```
nagmat84: Where is the resulting ZIP archive finally put? Inside the branch, if yes, where or in a different location?

.github/workflows/CICD.yml
```diff
@@ -0,0 +1,298 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set Up Imagick, FFmpeg & Exiftools
+        run: |
+          sudo apt-get update
+          sudo apt-get --fix-broken install
+          sudo apt-get -y install ffmpeg libimage-exiftool-perl
```
nagmat84: Why is this necessary for creating a ZIP archive? We don't run Lychee, we only want to pack it.

.github/workflows/CICD.yml
nagmat84: Found it myself: https://github.com/actions/upload-artifact#where-does-the-upload-go

.github/workflows/CICD.yml
nagmat84: Found the answer myself: https://github.com/actions/upload-artifact#where-does-the-upload-go.

But given the fact that the artifact is stored as an attachment to the workflows (and not in the branch), why do we restrict it to `master`?

Instead we should ensure that outdated workflows are deleted and purged. Then it doesn't matter if we create unnecessary ZIP archive.

ildyria:
Addressed:
- reordering of the task and dependency
- removing superfluous apt
- zip no longer only on master
- zip the zip.

nagmat84 (approved):

d7415:
Apparently GitHub Actions diagrams are terrible at showing dependencies clearly

(I accept that's a non-trivial task, and the interactive nature helps a lot, but it's still annoying)

PR #1425, opened 2022-07-28
move the traits to the class as it is only used once

ildyria:
Simple refactoring.


nagmat84 (approved):
Thanks for doing this. :partying_face: :confetti_ball:

This has been on my list of things which I wanted to clean up. In my PR before the file streaming, I announced that I had four PR in my mind and that the third one would be about cleaning up files and directories.

This would have been part of that. We have more of these strange things.

PR #1427, opened 2022-07-30
Fix album unit tests

nagmat84:
This is another PR with cherry-picking from https://github.com/LycheeOrg/Lychee/pull/1414 in order to keep https://github.com/LycheeOrg/Lychee/pull/1414 lean. While working on https://github.com/LycheeOrg/Lychee/pull/1414 I discovered three unrelated issues:

- Wrong expected default return value (200 instead of 204) for `Album::unlock`. All our currently existing tests specify an explicit error code (401 or 403), hence this bug was never spotted.
- Missing `use \Safe\json_encode`
- `Album::setProtectionPoilcy` did not support setting a password


ildyria (approved):
LGTM.

PR #1428, opened 2022-07-30
Fix test interaction with smart albums

nagmat84:
This is another PR with cherry-picking from https://github.com/LycheeOrg/Lychee/pull/1414 in order to keep https://github.com/LycheeOrg/Lychee/pull/1414 lean.

The smart albums (`PublicAlbum`, `RecentAlbum`, `StarredAlbum`, `PublicAlbum`)  have been implemented as singletons in order to cache them incl. their photo relation etc. during the same HTTP request. This is fine for productive code.

However, during testing all "requests" are part of the same test process and hence caching is a problem if we query the smart albums in the same test and expect different results. To this end the productive code has always been containing code which was only necessary for tests. (Yes, this has always been a bad design, but is was an easy work-around back then.)

While I have been working on https://github.com/LycheeOrg/Lychee/pull/1414 I tumbled across more issues due this caching effect. Hence, it has been time to eventually fix this problem.

This PR does three things:

- Remove the "testing only" code from `PublicAlbum`, `RecentAlbum`, `StarredAlbum` and `PublicAlbum`
- Add a new trait `InteractWithSmartAlbums` to the test code
- Adopt the existing tests to use this trait where necessary


ildyria (approved):
LGTM.

PR #1435, opened 2022-08-03
Use traits for `authorize()` instead of defined write functions

ildyria:
A potential idea of refactoring.

Unsure.

### DIFF
--- a/app/Http/Requests/Photo/DeletePhotosRequest.php
+++ b/app/Http/Requests/Photo/DeletePhotosRequest.php
@@ -5,8 +5,10 @@ namespace App\Http\Requests\Photo;
 use App\Http\Requests\BaseApiRequest;
 use App\Http\Requests\Contracts\HasPhotoIDs;
 use App\Http\Requests\Traits\HasPhotoIDsTrait;
+use App\Models\Photo;
+use App\Policies\PhotoPolicy;
 use App\Rules\RandomIDRule;
-use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Gate;
 
 class DeletePhotosRequest extends BaseApiRequest implements HasPhotoIDs
 {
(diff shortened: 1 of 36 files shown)


app/Http/Requests/Photo/DeletePhotosRequest.php
```diff
@@ -16,7 +17,7 @@ class DeletePhotosRequest extends BaseApiRequest implements HasPhotoIDs
 	 */
 	public function authorize(): bool
 	{
-		return $this->authorizePhotosWriteByIDs($this->photoIDs);
+		return $this->photoPolicy->canEditByID(Auth::user(), $this->photoIDs);
```
nagmat84: If I see lines like this, I wonder what benefits do we have from using Laravel Gate magic in the remaining places. Obviously, our authorization logic goes beyond what Laravel Gates can handle. Don't worry, I don't want to revert that, it is rather a philosophical, theoretical question.

We need to call the actual policies anyway. So from my perspective the Laravel Gates only yield

- less type safety, incl. this strange passing of strings instead of proper `null`-values
- more inconsistency, because we cannot use them always
- less efficiency, because the OpCode compiler cannot optimize dynamic reflection, but it could optimize direct method calls

nagmat84 (approved):
I am fine with that. I scanned most of the files only quickly, because it was rather repetitive and dull. I hope I did not miss anything. I have one philosophical question.

app/Http/Requests/Photo/DeletePhotosRequest.php
ildyria: > less type safety, incl. this strange passing of strings instead of proper null-values

I really wished it was better in that regard.

> more inconsistency, because we cannot use them always

I think If found a way. Cf last commit.

> less efficiency, because the OpCode compiler cannot optimize dynamic reflection, but it could optimize direct method calls

True. :/

nagmat84: I hoped you would give me some arguments for using Laravel Gates instead of agreeing with each of my points. Otherwise I could be inclined to remove all gates again. :laughing:

ildyria: Arguments for gates ? 

- Seamless interaction with current User.
- keywords `@can` etc... in blade templates (useful later in e.g. Livewire)
- `before()` which allows to skip checks directly.
- `after()` if we need to add even more checks.

Issue #1434, opened 2022-08-03
zip-download of a gallery fails after approx. 65Mb


SvenMertin:
Hi there, 
I cannot figure out, why in my case the zip-download is not possible. I tested three different galleries and in each case the download fails after reaching 65Mb. No file is stored.

### Steps to reproduce the issue
1. Go to gallery
2. Click on Download
3. Wait Download size to reach approx. 65Mb

Any ideas? Thank you :-)

### Diagnose output
    Diagnostics
    -------
    Warning: Dropbox import not working. dropbox_key is empty.

    System Information
    --------------
    Lychee Version (git):            master (f39be7a) -- Up to date (5 minutes ago).
    DB Version:                      4.5.2
    
    composer install:                --no-dev
    APP_ENV:                         production
    APP_DEBUG:                       false
    
    System:                          Linux
    PHP Version:                     8.1.8
    PHP User agent:                  Lychee/4 (https://lycheeorg.github.io/)
    Timezone:                        Europe/Berlin
    Max uploaded file size:          128M
    Max post size:                   128M
    Max execution time:              200
    MySQL Version:                   10.3.35-MariaDB-1:10.3.35+maria~bionic-log
    
    exec() Available:                yes
    Imagick Available:               1
    Imagick Enabled:                 1
    Imagick Version:                 1687
    GD Version:                      2.3.3

    Config Information
    --------------
    version:                         040502
    check_for_updates:               0
    sorting_photos_col:              taken_at
    sorting_photos_order:            ASC
    sorting_albums_col:              max_taken_at
    sorting_albums_order:            ASC
    imagick:                         1
    skip_duplicates:                 0
    small_max_width:                 0
    small_max_height:                360
    medium_max_width:                1920
    medium_max_height:               1080
    lang:                            de
    layout:                          1
    image_overlay_type:              desc
    default_license:                 none
    compression_quality:             90
    full_photo:                      1
    delete_imported:                 0
    Mod_Frame:                       1
    Mod_Frame_refresh:               30
    thumb_2x:                        1
    small_2x:                        1
    medium_2x:                       1
    landing_page_enable:             0
    landing_owner:                   Caro
    landing_title:                   
    landing_subtitle:                
    landing_facebook:                
    landing_flickr:                  
    landing_twitter:                 
    landing_instagram:               
    landing_youtube:                 
    landing_background:              
    site_title:                      Online-Gallerie
    site_copyright_enable:           1
    site_copyright_begin:            2022
[...]

PR #1436, opened 2022-08-03
Added options for ZIP compression

nagmat84:
Adds option to control compression of ZIP downloads.

And before anybody asks what the difference is between -1 = disable compression and 0 = no compression. I have no clue :grinning:  i only took it over from the docs.

### DIFF
--- a/app/Actions/Album/Archive.php
+++ b/app/Actions/Album/Archive.php
@@ -3,6 +3,7 @@
 namespace App\Actions\Album;
 
 use App\Contracts\AbstractAlbum;
+use App\Exceptions\ConfigurationKeyMissingException;
 use App\Exceptions\Handler;
 use App\Exceptions\Internal\FrameworkException;
 use App\Facades\AccessControl;
@@ -20,6 +21,9 @@ use Symfony\Component\HttpFoundation\HeaderUtils;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 use ZipStream\Exception\FileNotFoundException;
 use ZipStream\Exception\FileNotReadableException;
+use ZipStream\Option\Archive as ZipArchiveOption;
+use ZipStream\Option\File as ZipFileOption;
+use ZipStream\Option\Method as ZipMethod;
 use ZipStream\ZipStream;
 
 class Archive extends Action
@@ -32,20 +36,23 @@ class Archive extends Action
 		'<', '>', ':', '"', '/', '\\', '|', '?', '*',
 	];
 
+	protected int $deflateLevel = -1;
+
 	/**
 	 * @param Collection<AbstractAlbum> $albums
 	 *
 	 * @return StreamedResponse
 	 *
 	 * @throws FrameworkException
+	 * @throws ConfigurationKeyMissingException
 	 */
 	public function do(Collection $albums): StreamedResponse
 	{
+		$this->deflateLevel = Configs::getValueAsInt('zip_deflate_level');
+
 		$responseGenerator = function () use ($albums) {
-			$options = new \ZipStream\Option\Archive();
+			$options = new ZipArchiveOption();
 			$options->setEnableZip64(Configs::getValueAsBool('zip64'));
-			$options->setLargeFileSize(Configs::getValueAsInt('zip_large_file_size'));
-			$options->setDeflateLevel(Configs::getValueAsInt('zip_deflate_level'));
 			$options->setZeroHeader(true);
 			$zip = new ZipStream(null, $options);
 
@@ -199,7 +206,12 @@ class Archive extends Action
 				} catch (InfoException) {
 					// Silently do nothing, if `set_time_limit` is denied.
 				}
-				$zip->addFileFromStream($fileName, $file->read());
+				$zipFileOption = new ZipFileOption();
+				$zipFileOption->setMethod($this->deflateLevel === -1 ? ZipMethod::STORE() : ZipMethod::DEFLATE());
+				$zipFileOption->setDeflateLevel($this->deflateLevel);
+				$zipFileOption->setComment($photo->title);
+				$zipFileOption->setTime($photo->taken_at);
+				$zip->addFileFromStream($fileName, $file->read(), $zipFileOption);
 				$file->close();
 			} catch (\Throwable $e) {
 				Handler::reportSafely($e);
--- a/app/Actions/Album/Archive.php
+++ b/app/Actions/Album/Archive.php
@@ -44,6 +44,9 @@ class Archive extends Action
 		$responseGenerator = function () use ($albums) {
 			$options = new \ZipStream\Option\Archive();
 			$options->setEnableZip64(Configs::getValueAsBool('zip64'));
+			$options->setLargeFileSize(Configs::getValueAsInt('zip_large_file_size'));
+			$options->setDeflateLevel(Configs::getValueAsInt('zip_deflate_level'));
+			$options->setZeroHeader(true);
 			$zip = new ZipStream(null, $options);
 
 			$usedDirNames = [];
--- a/app/Actions/Album/Archive.php
+++ b/app/Actions/Album/Archive.php
@@ -210,7 +210,9 @@ class Archive extends Action
 				$zipFileOption->setMethod($this->deflateLevel === -1 ? ZipMethod::STORE() : ZipMethod::DEFLATE());
 				$zipFileOption->setDeflateLevel($this->deflateLevel);
 				$zipFileOption->setComment($photo->title);
-				$zipFileOption->setTime($photo->taken_at);
+				if ($photo->taken_at !== null) {
+					$zipFileOption->setTime($photo->taken_at);
+				}
 				$zip->addFileFromStream($fileName, $file->read(), $zipFileOption);
 				$file->close();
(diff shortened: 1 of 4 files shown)


ildyria (approved):
Untested, LGTM.

qwerty287 (approved):
Untested as well.

kamil4:
Honestly, I don't like it on the user interface level. I mean, I know that you just exposed what the zipstream library provides, but that's an API meant for programmers, not for the end users.

I think we should have a simple, single boolean config flag, `zip_compress` or something along these lines. I can't imagine why our users would want more control in this case.

ildyria:
Well, it's on the "More Settings"

nagmat84:
It is still less than what the standard Linux `zip` CLI tool provides for command line options. But I am also happy to simplify it.

However, I run several tests on my machine with different compression levels and different settings for large file and different sets of sample photos. It appears as if it does not make any difference. This is what currently bothers me more.

kamil4:
Exactly. The data is compressed already. Even providing the on/off switch seems like much (nobody would mind if it was always off); anything more is a complete overkill.

nagmat84:
That is not what I meant. I also tried some uncompressed ZIP images. And I can compress them do different size when I use the Linux command from CLI. Only with Lychee they always turned out the same.

Maybe I am totally mistaken, but I believe this is a upstream bug. See here:

https://github.com/maennchen/ZipStream-PHP/blob/184587ccc97d0034cb2300396f9e5262a433a489/src/File.php#L120-L125

This is the only place which I have found that uses the setting again. But the variable `$method` is never used again. Or I am totally blind.

d7415:
> This is the only place which I have found that uses the setting again. But the variable `$method` is never used again. Or I am totally blind.

https://github.com/maennchen/ZipStream-PHP/blob/184587ccc97d0034cb2300396f9e5262a433a489/src/File.php#L473 ?

nagmat84:
Yes, but the whole method is never called.

nagmat84:
And even on my developing machine with local download zipping is ridiculously slow.

nagmat84:
I am more and more inclined to try to test the PHP native ZIP methods https://www.php.net/manual/de/book.zip.php.

ildyria:
> I am more and more inclined to try to test the PHP native ZIP methods https://www.php.net/manual/de/book.zip.php.

[We actually moved away from that the original zip and send](https://github.com/LycheeOrg/Lychee-v3/issues/17)
If it is possible to use stream with Zip then sure. :)

nagmat84:
> If it is possible to use stream with Zip then sure. :)

I  knew, there has been a problem. :laughing:

nagmat84:
Found the problem. Unfortunately, the options which are globally set on the archive, seem to have no effect :-( When I set some on a per-file level like this

```php
$zipFileOption = new ZipStreamFileOption();
$zipFileOption->setMethod(ZipStreamMethod::STORE());
$zip->addFileFromStream($fileName, $file->read(), $zipFileOption);
```

it suddenly makes a difference. Now, I have to check how to make that nice.

nagmat84:
Sorry, for the failed first attempt. I should have tried it first with different compression settings and check whether it makes any difference. Now it does:

Screenshot_20220803_210618

Level -1 is lightning fast and level 9 is slow as hell and "reduces" the size.

I changed the following parts:
- Compression method and level are now set on a per file level
- This also allows us to set some other nice things

I don't know why it does not work if done on the archive level, but I don't feel like debugging an upstream library right now.

qwerty287 (approved):
Looks good, but still untested.

app/Actions/Album/Archive.php
```diff
@@ -196,7 +206,14 @@ private function compressAlbum(AbstractAlbum $album, array &$usedDirNames, ?stri
...
 					// Silently do nothing, if `set_time_limit` is denied.
 				}
-				$zip->addFileFromStream($fileName, $file->read());
+				$zipFileOption = new ZipFileOption();
+				$zipFileOption->setMethod($this->deflateLevel === -1 ? ZipMethod::STORE() : ZipMethod::DEFLATE());
+				$zipFileOption->setDeflateLevel($this->deflateLevel);
+				$zipFileOption->setComment($photo->title);
+				if ($photo->taken_at !== null) {
+					$zipFileOption->setTime($photo->taken_at);
+				}
```
kamil4, line 215: I think I understand the purpose of `setTime` but what does the `setComment` do?

app/Actions/Album/Archive.php
```diff
@@ -32,18 +36,24 @@ class Archive extends Action
...
 	 */
 	public function do(Collection $albums): StreamedResponse
 	{
+		$this->deflateLevel = Configs::getValueAsInt('zip_deflate_level');
+
 		$responseGenerator = function () use ($albums) {
-			$options = new \ZipStream\Option\Archive();
+			$options = new ZipArchiveOption();
 			$options->setEnableZip64(Configs::getValueAsBool('zip64'));
+			$options->setZeroHeader(true);
```
kamil4, line 56: The documentation on this is cryptic. Is this a speed optimization?

app/Actions/Photo/Archive.php
```diff
@@ -266,7 +275,10 @@ protected function zip(Collection $photos, string $variant): StreamedResponse
 						);
 					} while (array_key_exists($filename, $uniqueFilenames));
 				}
-				$zip->addFileFromStream($filename, $archiveFileInfo->getFile()->read());
+				$zipFileOption = new ZipFileOption();
+				$zipFileOption->setMethod($this->deflateLevel === -1 ? ZipMethod::STORE() : ZipMethod::DEFLATE());
+				$zipFileOption->setDeflateLevel($this->deflateLevel);
+				$zip->addFileFromStream($filename, $archiveFileInfo->getFile()->read(), $zipFileOption);
```
kamil4, line 281: Why aren't we calling `setComment` and `setTime` here (like we do for albums)?

kamil4 (commented):
Looks good overall but I have a few questions...

app/Actions/Album/Archive.php
nagmat84: The ZIP standard allows to create archives which attach certain metadata to every file. With an appropriate ZIP viewer this comment is shown next to the compression ration of the file, the CRC of the file and well yes a "comment". As this is UTF-8 I thought it would be nice to use it for the title. It is just a gimmick. I am not sure if it helps anybody.

app/Actions/Album/Archive.php
nagmat84: I assume you are referring to this sentence

> Enable streaming files with single read where general purpose bit 3 indicates local file header contain zero values in crc and size fields, these appear only after file contents in data descriptor block. Default is false. Set to true if your input stream is remote (used with addFileFromStream()).

from [ZipStream-PHP: Available-options](https://github.com/maennchen/ZipStream-PHP/wiki/Available-options). I am not even sure if the first sentence is actually a sentence :D

I am not 100% sure what it does, but I did some step debugging through the code and https://en.wikipedia.org/wiki/Deflate gave me some clues. My assumption is also backed up by looking at some ZIP archives in the hex editor.

I believe the documentation should have been as follows

> The DEFLATE algorithm allows to use a Huffman encoding which is globally optimized and constructed for the entire archive or individual, local Huffman encoding for each processing block. On a technical level the former type of blocks inside an archive are indicated by a `x01`-header, the latter type is indicated by a `x10-header (cp. [Wikipedia: Deflate](https://en.wikipedia.org/wiki/Deflate)).
> In order to get the next to optimal compression ratio the DEFLATE algorithm needs to read the entire payload three times (across file boundaries). This allows to construct a near to optimal Huffman encoding for the entire archive on a global level. However a 3-pass scan is prohibitive for large archives and does not work for file streams which are non-seekable, like remote file streams.

But this contains a lot of guess work here. I have no clue what the documentation wants to say about "contain zero values in crc and size fields". But I can confirm that without `setZeroHeader(true)` the library tries to read every file three times.

app/Actions/Photo/Archive.php
nagmat84: Because we do not iterate over photos here. I have no access to `$photos->title` in this place.

I wanted to but then I noticed that `Photo/Archive` and `Album/Archive` use two totally different approaches. They also solve collisions of files names differently. :( The classes also contain duplicate code. :( In the long run we should clean up that, i.e. move the common code into a shared base class, ensure that collisions in filenames are solved consistently, etc.

But I did not feel like this should be part of this PR. This would probably imply to rewrite both classes completely :D

app/Actions/Album/Archive.php
kamil4: It sounds like they set the crc and size to `0` in the header but also set some flag to indicate that the actual values will be provided after the payload (the compressed file contents). Otherwise an extra pass over the file contents would be needed just to calculate the crc (and size, if unknown) before the compression can even start.

nagmat84:
Although the original issue #1434 has been a Heisenberg bug, I believe this PR has still its merits on its own and should be merged.

@kamil4 Are my answers to your question to your satisfaction? Can we merge?

kamil4:
Yes, go ahead and merge.

kamil4 (approved):

PR #1439, opened 2022-08-03
Make VideoHandler support optional

### LINKED ISSUE #1438: Video upload fails with error: FFmpeg is disabled by configuration
When I try to upload a video to an album, it fails because FFmpeg is not available.

I'm on a shared host, without FFmpeg available.
From my understanding of [FAQ](https://lycheeorg.github.io/docs/faq.html#why-dont-my-videos-have-thumbnails), video upload is possible without FFmpeg but there will be no thumbnails. 

I also tried to upload https://github.com/LycheeOrg/Lychee/blob/master/tests/Samples/gaming.mp4, same results

1. Go to any album
2. Click on upload
3. Select a video file to upload
4. At the end of the upload, see error

**Screenshots**
image

```


    Diagnostics
    -------
    Warning: /home/xxxxx/Lychee/public/uploads/ has permissions 0775, but should have 2775
    Warning: /home/xxxxx/Lychee/public/uploads/thumb2x has permissions 0775, but should have 2775
    Warning: /home/xxxxx/Lychee/public/uploads/thumb2x/e3 has permissions 0775, but should have 2775
    Warning: /home/xxxxx/Lychee/public/uploads/thumb2x/e3/7c has permissions 0775, but should have 2775
    Warning: /home/xxxxx/Lychee/public/uploads/thumb2x/e3/ea has permissions 0775, but should have 2775
    Warning: 7586 more directories with wrong permissions

    System Information
    --------------
    Lychee Version (git):            master (1a31f3c) -- 3 commits behind master cd394af (1 days ago)
    DB Version:                      4.5.2
    
    PHP Version:                     8.0.21
    Max uploaded file size:          500M
    Max post size:                   500M
    Max execution time:              200
    MySQL Version:                   10.5.16-MariaDB-cll-lve
    
    Imagick Available:               1
    Imagick Version:                 1808

    Config Information
    --------------
    version:                         040502
    sorting_albums_col:              title
    landing_owner:                   xxxx
    landing_title:                   
    landing_subtitle:                
    landing_facebook:                
    landing_flickr:                  
    landing_twitter:                 
    landing_instagram:               
    landing_youtube:                 
    landing_background:              
    site_title:                      xxxx
    site_copyright_begin:            2022
    additional_footer_text:          
    downloadable:                    1

```

### Browser and system
Firefox 103 on Windows 10

kamil4:
Fixes #1438 

This is a semi-draft PR:
* I fixed only the case of uploading a regular video file but I see that `VideoHandler` is also used for Google Motion Pictures. I left the latter as-is as I don't understand how that part works and what (if anything) we can do about it.
* By catching the exception, we prevent a message about missing FFmpeg from being generated in the logs. It seems like it would be useful to have it there though under these circumstances, but I wasn't sure what the "approved" way of doing it is these days. @nagmat84?

### DIFF
--- a/app/Actions/Photo/Strategies/AddStandaloneStrategy.php
+++ b/app/Actions/Photo/Strategies/AddStandaloneStrategy.php
@@ -89,31 +89,29 @@ class AddStandaloneStrategy extends AddBaseStrategy
 		$this->photo->original_checksum = StreamStat::createFromLocalFile($this->sourceFile)->checksum;
 
 		try {
-			// Load source image if it is a supported photo format
-			if ($this->photo->isPhoto()) {
-				$this->sourceImage = new ImageHandler();
-				$this->sourceImage->load($this->sourceFile);
-			} elseif ($this->photo->isVideo()) {
-				// We try to extract a video frame using FFmpeg; however,
-				// that one's optional, so we need to be able to recover.
-				try {
+			try {
+				if ($this->photo->isPhoto()) {
+					// Load source image if it is a supported photo format
+					$this->sourceImage = new ImageHandler();
+					$this->sourceImage->load($this->sourceFile);
+				} elseif ($this->photo->isVideo()) {
 					$videoHandler = new VideoHandler();
 					$videoHandler->load($this->sourceFile);
 					$position = is_numeric($this->photo->aperture) ? floatval($this->photo->aperture) / 2 : 0.0;
 					$this->sourceImage = $videoHandler->extractFrame($position);
-				} catch (\Throwable) {
-					$this->sourceImage = null;
-				}
-			} else {
-				// If we have a raw file, we try to treat it as an image, as
-				// Imagick supports a lot of other image-like file types
-				// But if it fails we don't mind.
-				try {
+				} else {
 					$this->sourceImage = new ImageHandler();
 					$this->sourceImage->load($this->sourceFile);
-				} catch (\Throwable) {
-					$this->sourceImage = null;
 				}
+			} catch (\Throwable $e) {
+				// This may happen for videos if FFmpeg is not available to
+				// extract a still frame, or for raw files (Imagick may be
+				// able to convert them to jpeg, but there are no
+				// guarantees, and Imagick may not be available).
+				$this->sourceImage = null;
+
+				// Log an error without failing.
+				Handler::reportSafely($e);
 			}
 
 			// Handle Google Motion Pictures
@@ -121,10 +119,15 @@ class AddStandaloneStrategy extends AddBaseStrategy
 			// file and stash it away, before the original file is moved into
 			// its (potentially remote) final position
 			if ($this->parameters->exifInfo->microVideoOffset !== 0) {
-				$tmpVideoFile = new TemporaryLocalFile(GoogleMotionPictureHandler::FINAL_VIDEO_FILE_EXTENSION, $this->sourceFile->getBasename());
-				$gmpHandler = new GoogleMotionPictureHandler();
-				$gmpHandler->load($this->sourceFile, $this->parameters->exifInfo->microVideoOffset);
-				$gmpHandler->saveVideoStream($tmpVideoFile);
+				try {
+					$tmpVideoFile = new TemporaryLocalFile(GoogleMotionPictureHandler::FINAL_VIDEO_FILE_EXTENSION, $this->sourceFile->getBasename());
+					$gmpHandler = new GoogleMotionPictureHandler();
+					$gmpHandler->load($this->sourceFile, $this->parameters->exifInfo->microVideoOffset);
+					$gmpHandler->saveVideoStream($tmpVideoFile);
+				} catch (\Throwable $e) {
+					Handler::reportSafely($e);
+					$tmpVideoFile = null;
+				}
 			} else {
 				$tmpVideoFile = null;
 			}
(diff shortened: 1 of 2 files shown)


app/Actions/Photo/Strategies/AddStandaloneStrategy.php
```diff
@@ -94,10 +94,16 @@ public function do(): Photo
...
-				$this->sourceImage = $videoHandler->extractFrame($position);
+				// We try to extract a video frame using FFmpeg; however,
+				// that one's optional, so we need to be able to recover.
+				try {
+					$videoHandler = new VideoHandler();
+					$videoHandler->load($this->sourceFile);
+					$position = is_numeric($this->photo->aperture) ? floatval($this->photo->aperture) / 2 : 0.0;
+					$this->sourceImage = $videoHandler->extractFrame($position);
+				} catch (\Throwable) {
+					$this->sourceImage = null;
```
nagmat84: You can use `Handler::reportSafely($e)` to log the exception as if it would bubble up the call graph without actually throwing it.

nagmat84 (approved):
Looks good to me with respect to the reported issue.

Maybe we should wrap the whole if-then-else-branch into the try-block? This would also allow to upload regular photos which are too broken to generate thumbs but maybe still viewable by a browser. But I don't know if that case really exist.

We should also have a test which tries to upload a video with FFMpeg disabled and which asserts that the upload still succeeds, that no thumbnail has been generated and a log entry has been made. But I can also write the test tomorrow independently of this PR.

nagmat84:
With respect to GMP. In

https://github.com/LycheeOrg/Lychee/blob/f39be7a803ea1e30230e30fae0dca246c5f3a8c1/app/Actions/Photo/Strategies/AddStandaloneStrategy.php#L113-L124

the lines 118-121, i.e. the body of the if-branch could be wrapped into a try-block as well and then the catch-block needs to set `$tmpVideoFile` to `null` exactly as it is done in the else-branch as if it was no GMP image at all. Then we would be fine for that, too.

I would do it myself, but here it is 00:30am and I am not in front of a PC.

kamil4:
@nagmat84: I made some changes along the lines you suggested. Logging works now. An existing testcase (`testBrokenGoogleMotionPhotoUpload`) triggers now though -- I decided to leave it up to you to figure out what to do about it.

I also added a testcase for uploading a video file without ffmpeg (or exiftool for that matter, to ensure no video metadata). It works, though maybe it's not as tight as you would like (I don't believe it actually checks if there are no thumbnails?).

tests/Feature/PhotosAddHandlerTestAbstract.php
```diff
@@ -344,4 +345,37 @@ public function testTrickyVideoUpload(): void
...
+			'album_id' => null,
+			'title' => 'gaming',
+			'type' => TestCase::MIME_TYPE_VID_MP4,
+			'size_variants' => [
+				'original' => [
+					'width' => 0,
+					'height' => 0,
+					'filesize' => 66781184,
+				],
+			],
```
nagmat84: Maybe we should also assert that the other size variants are really `null`?


```suggestion
			'size_variants' => [
				'original' => [
					'width' => 0,
					'height' => 0,
					'filesize' => 66781184,
				],
				'medium2x' => null,
				'medium' => null,
				'small2x' => null,
				'small' => null,
				'thumb2x' => null,
				'thumb' => null,
			],
```

`FFMpeg` is always available to the the test suite. We only mimick a missing FFMpeg by disabling it in the configuration. If that fails (for some unknonwn reason) we might accidentally not test what we want, because FFmeg will be called without complain.

tests/Feature/PhotosAddHandlerTestAbstract.php
```diff
@@ -344,4 +345,37 @@ public function testTrickyVideoUpload(): void
...
+		]);
+
+		Configs::set(self::CONFIG_HAS_FFMPEG, $hasFFMpeg);
+		Configs::set(self::CONFIG_HAS_EXIF_TOOL, $hasExifTool);
+	}
```
nagmat84, line 430: Maybe we should additionally assert that the expected error message, e.g. "FFMpeg is disabled by configuration", can be found in the logs?

To this end we could clear the log table at the beginning of the test with `DB::tables('logs)->delete()`. At the end of the test we do a count-query and assert that we found an entry, e.g. something like

```php
$numErrorMsg = DB::tables('logs)->where('message', 'like', '%FFMpeg is disabled by configuration%')->count();
self::assertEquals(1, $numErrorMsg);
```

But maybe this would make the test too strict? I am unsure.

nagmat84 (commented):
Just some ideas.

tests/Feature/PhotosAddHandlerTestAbstract.php
kamil4: Yeah, I agree that merely disabling it in the configuration is imperfect, but realistically that's probably all we can do in the testsuite?

tests/Feature/PhotosAddHandlerTestAbstract.php
kamil4: I wouldn't do it myself (testing for log messages seems a little overly prescriptive to me) but feel free to add a commit to that effect to this PR if you want.

tests/Feature/PhotosAddHandlerTestAbstract.php
nagmat84: That's is why I considered checking the logs. This way, we can be sure that the exception has been triggered inside the code.

PR #1440, opened 2022-08-05
Update dependencies and add NPM lockfile

qwerty287:
- update PHP dependencies (`composer update`)
- add an NPM lockfile to lock JS dependencies for Livewire.

### DIFF
--- a/app/Metadata/Extractor.php
+++ b/app/Metadata/Extractor.php
@@ -16,6 +16,7 @@ use PHPExif\Adapter\NoAdapterException;
 use PHPExif\Exif;
 use PHPExif\Reader\Reader;
 use Safe\Exceptions\StringsException;
+
 use function Safe\substr;
 
 /**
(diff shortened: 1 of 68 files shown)


app/Metadata/Extractor.php
```diff
@@ -341,14 +341,14 @@ public static function createFromFile(NativeLocalFile $file): self
 							$taken_at->setTimezone(new \DateTimeZone(date_default_timezone_get()));
 						}
 						// In the remaining cases the timezone information was
-						// extracted and the recording time is assumed exhibit
-						// to original timezone of the location where the video
-						// has been recorded.
-						// So we don't need to do anything.
+					// extracted and the recording time is assumed exhibit
```
ildyria: weird change.

qwerty287: This was php-cs-fixer, let me look at it.

qwerty287: Hmm yes that's weird. Actually, I downgraded php-cs-fixer again after I noticed that it doesn't work... I'll try to find out why it'd doing this.

qwerty287:
Everything's fixed again.

ildyria:
Honestly, I would not bother much with that PR. The reason being that #1403 should allow user to use [Laragear/WebAuthn](https://github.com/Laragear/WebAuthn) instead of [our Larapass fork](https://github.com/LycheeOrg/Larapass).
By doing so we will be migrating to Laravel v9 and this will apply a whole bunch of package update on its own.

qwerty287:
@ildyria  It's finished, I fixed everything :)

ildyria (approved):

nagmat84:
I did not know that we also use NPM for the Lychee backend; the the frontend of course, yes. Could anybody enlighten me why we need NPM and by what it is used?

qwerty287:
It will be used for Livewire.

ildyria:
NPM is used to compile dependencies from the `resources/asset` folder.

ildyria:
> It will be used for Livewire.

Exactly that. :laughing:

PR #1450, opened 2022-08-07
Update readme.md

ildyria:
Avoid showing the pre-releases in readme.


qwerty287 (approved):

d7415:
This just makes me wonder why we have been all this time! Presumably it didn't come up because previously pre-releases were only "pre" for hours at most.

PR #1453, opened 2022-08-08
Provide an ASCII fallback for multibyte filenames

### LINKED ISSUE #1452: Error 500 when downloading an image with non-ASCII characters
When downloading an image with non-ASCII characters in its original filename, server fails with an error 500 HttpException "Symfony's response component failed".

Upload of the file via the import button was perfectly fine. Same for display. Internally, the file was renamed to a UUID filename.

Note that downloading full album is ok, the downloaded zip contains images named with non-ASCII characters.

1. Upload an image with non-ASCII characters in its filename. Here it was "Connecté.png"
2. Go to this image in gallery
3. Click on three-dot menu > Download
4. Pick any size
5. Server error 500 HttpException "Symfony's response component failed".

### Log outputs
```
2022-08-08 19:12:27 UTC -- error   -- Symfony\Component\HttpFoundation\HeaderUtils::makeDisposition -- 173 -- The filename fallback must only contain ASCII characters.
2022-08-08 19:12:27 UTC -- error   -- App\Actions\Photo\Archive::file -- 143 -- The filename fallback must only contain ASCII characters.; caused by
2022-08-08 19:12:27 UTC -- error   -- App\Actions\Photo\Archive::file -- 166 -- Symfony's response component failed; caused by
```

### Output of the diagnostics 
```

    Diagnostics
    -------
    Warning: /home/xxx/Lychee/public/uploads/ has permissions 0775, but should have 2775
    Warning: /home/xxx/Lychee/public/uploads/thumb2x has permissions 0775, but should have 2775
    Warning: /home/xxx/Lychee/public/uploads/thumb2x/e3 has permissions 0775, but should have 2775
    Warning: /home/xxx/Lychee/public/uploads/thumb2x/e3/7c has permissions 0775, but should have 2775
    Warning: /home/xxx/Lychee/public/uploads/thumb2x/e3/ea has permissions 0775, but should have 2775
    Warning: 7591 more directories with wrong permissions


    System Information
    --------------
    Lychee Version (git):            master (fe173e4) -- Up to date (7 hours ago).
    DB Version:                      4.5.3
    
    PHP Version:                     8.0.21
    Max uploaded file size:          500M
    Max post size:                   500M
    Max execution time:              14400
    MySQL Version:                   10.5.16-MariaDB-cll-lve
    
    Imagick Available:               1
    Imagick Version:                 1808


    Config Information
    --------------
    version:                         040503
    sorting_albums_col:              title
    landing_owner:                   xxxx
    landing_title:                   
    landing_subtitle:                
    landing_facebook:                
    landing_flickr:                  
    landing_twitter:                 
    landing_instagram:               
    landing_youtube:                 
    landing_background:              
    site_title:                      xxx
    site_copyright_begin:            2022
    additional_footer_text:          
    downloadable:                    1
    zip_deflate_level:               -1


```
### Browser and system
Firefox 103 on Windows 11

kamil4:
Fixes #1452 

An equivalent of an old fix we had implemented for albums in #384


nagmat84 (commented):
The fix looks good. I also compared it with the analogous code for the albums to ensure consistency, i.e.

https://github.com/LycheeOrg/Lychee/blob/fe173e4b77ee81a204e0ca6064f37403ab03cec2/app/Actions/Album/Archive.php#L72-L76

Let's see, if I find the time to add a test case this evening so that we won't tumble across this regression a third time. Or you can add it. It shouldn't be too difficult. Just upload a photo, rename it to something with multi-byte characters in it and then download it. Optimally, we also add a test for  an album, too, while we are on it.

ildyria (approved):
LGTM.

kamil4:
Strictly speaking, we don't need a new image file with a multi-byte character; we could just upload one of the existing ones and then change the title. But I guess there may be value in testing multi-byte uploads as well...

I agree with @nagmat84 that we should also add a test for multi-byte album downloads.

kamil4:
Looking through the testsuite, I could easily add a relevant album test by adding just a couple of lines to `testManyFunctionsAtOnce`, around: https://github.com/LycheeOrg/Lychee/blob/ef102c4ed91389373772282ff0dfab5879260da6/tests/Feature/PhotosOperationsTest.php#L139

Is that considered acceptable? There's a comment at the start of that method that says: _Preferably, all the tested actions should be seperated into individual tests._

tests/TestCase.php
```diff
@@ -48,6 +48,7 @@ abstract class TestCase extends BaseTestCase
 	public const SAMPLE_FILE_GMP_IMAGE = 'tests/Samples/google_motion_photo.jpg';
 	public const SAMPLE_FILE_MONGOLIA_IMAGE = 'tests/Samples/mongolia.jpeg';
 	public const SAMPLE_FILE_NIGHT_IMAGE = 'tests/Samples/night.jpg';
+	public const SAMPLE_FILE_SUNSET_IMAGE = 'tests/Samples/fin de journée.jpg';
```
nagmat84: I would prefer to keep that list in alphabetic order. I will push a commit myself.

tests/Feature/PhotosMultiBytesTest.php
```diff
@@ -0,0 +1,62 @@
...
+ * @noinspection PhpDocMissingThrowsInspection
+ * @noinspection PhpUnhandledExceptionInspection
+ */
+
+namespace Tests\Feature;
+
+use Tests\Feature\Base\PhotoTestBase;
+use Tests\TestCase;
+
+class PhotosMultiBytesTest extends PhotoTestBase
```
nagmat84: Why has the test been put into its own class? I doubt that this class with this special name will be extended by any other test in the future. So why not adding the test to one of the existing classes.

tests/Feature/PhotosMultiBytesTest.php
```diff
@@ -0,0 +1,62 @@
...
+				],
+				'original' => [
+					'width' => 914,
+					'height' => 1625,
+					'filesize' => 270345,
+				],
+			],
+		]);
+
+		$this->photos_tests->download([$id]);
```
nagmat84: Why isn't the result of the download checked more thoroughly, i.e. file extension, etc.?

nagmat84 (commented):
I will push some commits myself which address my remarks.

nagmat84:
> Strictly speaking, we don't need a new image file with a multi-byte character;

That's what I thought of, too.

> we could just upload one of the existing ones and then change the title. But I guess there may be value in testing multi-byte uploads as well...

Agreed. And as @ildyria has already added a sample file, I'm fine with that.

nagmat84:
> Looking through the testsuite, I could easily add a relevant album test by adding just a couple of lines to `testManyFunctionsAtOnce`, around:
> 
> https://github.com/LycheeOrg/Lychee/blob/ef102c4ed91389373772282ff0dfab5879260da6/tests/Feature/PhotosOperationsTest.php#L139
> 
> Is that considered acceptable? There's a comment at the start of that method that says: _Preferably, all the tested actions should be seperated into individual tests._

I would like to avoid that. When I added a lot of test as part of https://github.com/LycheeOrg/Lychee/pull/1351, I needed to get this "god test" out of my way but did not want to lose what has already been there. In the long run this god test should be split, hence I would prefer not to add to that mess.

Let's see if I can create a test this evening.

tests/TestCase.php
nagmat84: Fixed it myself in https://github.com/LycheeOrg/Lychee/pull/1453/commits/a28747ea885803d6280052f3490283d7429dbccc.

tests/Feature/PhotosMultiBytesTest.php
nagmat84: Fixed it myself in https://github.com/LycheeOrg/Lychee/pull/1453/commits/650c2577394c6bdf980c842b472af9e8b08e63e0. Moved the upload into an independent test to run it for all image handlers and also without EXIF tool. The download is tested more thoroughly.

tests/Feature/PhotosMultiBytesTest.php
nagmat84: Fixed it myself in https://github.com/LycheeOrg/Lychee/commit/650c2577394c6bdf980c842b472af9e8b08e63e0.

nagmat84 (approved):
Fixed all my concerns myself. I also added a test to download an archive with a multi-byte title. Feel free to merge and resolve my remarks. I only left them open for you such that they are more easily to read.

PR #1454, opened 2022-08-08
Fix 1411 visibility propagation

ildyria:


ildyria:
Just back-end stuff for now.

ildyria:
@nagmat84 are you working on the separation of `is_public` from the album with regard to sharing ?
If so I would drop this PR.

nagmat84:
Yes, I do, but it will probably take about three weeks until I will have something ready.

ildyria:
I will be closing this one for now because #1511 is going to introduce a lot of conflicts and change which makes it more interesting to start working from there.

PR #1457, opened 2022-08-10
Cleanup download tests

nagmat84:
This is a follow-up to #1453. I already wanted to do that yesterday, but wanted to keep the re-factoring out of #1453.

 1. It moves all download tests from the general `PhotosOperationsTest` to the new and specialized test class `PhotosDownloadTest` in order to keep them together.
 2. We test different ZIP archives in several tests. I moved the boiler-plate code to handle ZIP archives and make assertions about them into a new class `AssertableZIPArchive`

### DIFF
--- a/tests/Feature/Lib/AssertableZipArchive.php
+++ b/tests/Feature/Lib/AssertableZipArchive.php
@@ -65,7 +65,7 @@ class AssertableZipArchive extends \ZipArchive
 	}
 
 	/**
-	 * Asserts that the ZIP archive exactly contains the given files and no more.
+	 * Asserts that the ZIP archive contains exactly the given files and no more.
 	 *
 	 * @param array<string, array{size?: ?int}> $expectedFiles a list of expected file names together with optional file attributes
 	 *
--- a/tests/Feature/Lib/AssertableZipArchive.php
+++ b/tests/Feature/Lib/AssertableZipArchive.php
@@ -1,0 +1,84 @@
+<?php
+
+/**
+ * We don't care for unhandled exceptions in tests.
+ * It is the nature of a test to throw an exception.
+ * Without this suppression we had 100+ Linter warning in this file which
+ * don't help anything.
+ *
+ * @noinspection PhpDocMissingThrowsInspection
+ * @noinspection PhpUnhandledExceptionInspection
+ */
+
+namespace Tests\Feature\Lib;
+
+use App\Image\InMemoryBuffer;
+use App\Image\TemporaryLocalFile;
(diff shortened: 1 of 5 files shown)


ildyria:
I am wondering if we should add a helper function which given a path `foo/../bar/image.ext` returns `image.ext`. 
This would avoid having magic constants in our tests.

nagmat84:
No need to add helper function for that. `fileinfo` (I believe, or `pathinfo`?) does the job.

But I am more curious where you want to _use_ it?

ildyria:
e.g. here:
```php
$zipArchive->assertContainsFilesExactly([
	'train.jpg' => ['size' => filesize(base_path(self::SAMPLE_FILE_TRAIN_IMAGE))],
	'mongolia-1.jpeg' => ['size' => filesize(base_path(self::SAMPLE_FILE_MONGOLIA_IMAGE))],
	'mongolia-2.jpeg' => ['size' => filesize(base_path(self::SAMPLE_FILE_MONGOLIA_IMAGE))],
```
Well `mongalia-1`, and `mongolia-2` are not helping... but here for example makes a bit more sense:
```php
$zipArchive->assertContainsFilesExactly([
	self::MULTI_BYTE_ALBUM_TITLE . '/fin de journée.jpg' => ['size' => filesize(base_path(TestCase::SAMPLE_FILE_SUNSET_IMAGE))],
	self::MULTI_BYTE_ALBUM_TITLE . '/mongolia.jpeg' => ['size' => filesize(base_path(TestCase::SAMPLE_FILE_MONGOLIA_IMAGE))],
]);
```

nagmat84:
At these places such a helper function does not help. The name of downloaded file is not based on the original name of the initially uploaded file, but on the title of the photo which has been assigned by Lychee.

ildyria (approved):
LGTM.

tests/Feature/Lib/AssertableZipArchive.php
```diff
@@ -0,0 +1,84 @@
...
+	{
+		$stat = $this->statName($fileName);
+		PHPUnit::assertNotFalse($stat, 'Could not assert that ZIP archive contains ' . $fileName);
+		if ($expectedFileSize !== null) {
+			PHPUnit::assertEquals($expectedFileSize, $stat[self::ZIP_STAT_SIZE]);
+		}
+	}
+
+	/**
+	 * Asserts that the ZIP archive exactly contains the given files and no more.
```
kamil4:
```suggestion
	 * Asserts that the ZIP archive contains exactly the given files and no more.
```

kamil4 (approved):

PR #1467, opened 2022-08-18
Move 'installation:complete' to web instead of per route

ildyria:
Minor refactoring to simplify middleware in routes in `web.php`.


qwerty287 (approved):
Are Installation routes excluded from this?

ildyria:
Yes, because the installation routes are in `web-install.php`.
This middleware is just applied to `web.php`.
See `app/Providers/RouteServiceProvider.php`

qwerty287 (approved):

qwerty287:
Why is this middleware not applied to the API? If there's no reason, you should apply it there too.

nagmat84:
Yes, there is a reason. The API client expects JSON responses. A redirection to a HTML page (with the installer) is nothing an API client can handle.

qwerty287:
Does it return something else instead or does it just fail if there's e.g. a DB request?

nagmat84:
If the request fails, it returns the error message of the code line which caused the request to fail as all failing requests do. If the request fails, it uses the normal error reporting mechanisms as always.

qwerty287:
Sure, but it would be better to report something like "not installed".

nagmat84:
Definitely yes.

And I have checked to code again. When I wrote the exception handler and created the exception `InstallationRequiredException`, I had been better than I thought I had been. :grinning: If the middleware attempts to assert that the installation is complete and finds the installation to be incomplete, then this middleware throws an `InstallationRequiredException` exception. The exception handler catches the exception but _only_ replies with a redirection, if the client requested a HTML page not `application/json`. Otherwise, i.e. if the client wants `application/json`, then the error reporting proceeds normally and the client receives a proper `InstallationRequiredException` through JSON.

So back to what you have initially suggested in https://github.com/LycheeOrg/Lychee/pull/1467#issuecomment-1227632272.

We actually can apply this middleware to the API and we are good.

PR #1469, opened 2022-08-19
Upgrade to Laravel 9 + switch to Laragear/WebAuthn

ildyria:


nagmat84:
Please keep also in mind that the stream handling with respect to the change from Flysystem v1 to v3 needs changes.

ildyria:
> Please keep also in mind that the stream handling with respect to the change from Flysystem v1 to v3 needs changes.

No worries, I have not forgotten. Do you mind if I let you fix that part while I focus on the WebAuthn problem? :)
I can also give a shot to the streams, but I have never worked with them before. But I can try. :)

ildyria:
```

There were 3 errors:

1) Tests\Feature\CommandVideoDataTest::testThumbRecreation
ErrorException: Attempt to read property "url" on null

~/Documents/Lychee/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php:259
~/Documents/Lychee/tests/Feature/CommandVideoDataTest.php:31

2) Tests\Feature\PhotosAddMethodsTest::testImportViaSymlink
TypeError: Tests\Feature\Lib\PhotosUnitTest::get(): Argument #1 ($photo_id) must be of type string, null given, called in ~/Documents/Lychee/tests/Feature/PhotosAddMethodsTest.php on line 61

~/Documents/Lychee/tests/Feature/Lib/PhotosUnitTest.php:109
~/Documents/Lychee/tests/Feature/PhotosAddMethodsTest.php:61

3) Tests\Feature\PhotosAddMethodsTest::testAppleLivePhotoImportViaSymlink
TypeError: Tests\Feature\Lib\PhotosUnitTest::get(): Argument #1 ($photo_id) must be of type string, null given, called in ~/Documents/Lychee/tests/Feature/PhotosAddMethodsTest.php on line 177

~/Documents/Lychee/tests/Feature/Lib/PhotosUnitTest.php:109
~/Documents/Lychee/tests/Feature/PhotosAddMethodsTest.php:177

--

There were 1 failures:

1) Tests\Feature\CommandFixPermissionsTest::testFixPermissions
~/Documents/Lychee/public/uploads/original/6b/a7/a7b075b2b80527e9ad9ece5ad041.jpeg
Failed asserting that 256 matches expected 436.

~/Documents/Lychee/tests/Feature/CommandFixPermissionsTest.php:53
```
Getting there. :)

ildyria:
Finally !

nagmat84:
I guess I need at least the upcoming weekend to step through this PR. First, I would like to put my attention on https://github.com/LycheeOrg/Lychee/pull/1368 and https://github.com/LycheeOrg/Lychee/pull/1366 which are a) much smaller and b) longer in the pipeline.

ildyria:
> I guess I need at least the upcoming weekend to step through this PR. First, I would like to put my attention on #1368 and #1366 which are a) much smaller and b) longer in the pipeline.

I would be happy if also #1480 would take priority over this one. :'D

app/Actions/Album/Delete.php
```diff
@@ -101,6 +100,8 @@ public function do(array $albumIDs): FileDeleter
 				$recursiveAlbumIDs = array_merge($recursiveAlbumIDs, $subAlbums->pluck('id')->all());
 				$recursiveAlbumTracks = $recursiveAlbumTracks->merge($subAlbums->pluck('track_short_path'));
 			}
+			// prune the null values
+			$recursiveAlbumTracks = $recursiveAlbumTracks->filter(fn ($val) => $val !== null);
```
nagmat84, line 104: Why? Is this a an independent bug fix?

app/Actions/Install/PermissionsChecker.php
```diff
@@ -3,7 +3,6 @@
 namespace App\Actions\Install;
 
 use function Safe\preg_match;
-use function Safe\substr;
```
nagmat84, line 6: I have seen this being removed a couple of times now. So I am finally going to ask: why? What has changed that his isn't required anymore?

app/Actions/User/Notify.php
```diff
@@ -19,7 +19,7 @@ public function do(Photo $photo): void
 		}
 
 		// The admin is always informed
-		$users = new Collection([User::query()->find(0)]);
+		$users = new Collection(User::query()->find(0));
```
nagmat84: Why that change?

I am not even sure that this change is correct. The constructor checks if the passed value is an instance of `Arrayable` and the calls `toArray()` on the passed value. `User` is a model and a such an instance of `Arrayable`. However `User::toArray()` returns the array representation of a user not an array with a single user object.

This looks really wrong.

app/Assets/SizeVariantGroupedWithRandomSuffixNamingStrategy.php
```diff
@@ -156,11 +156,11 @@ protected static function createRndMiddlePath(): string
 			$rndStr = bin2hex(random_bytes(self::NAME_LENGTH / 2));
 
 			return
-				\Safe\substr($rndStr, 0, 2) .
+				substr($rndStr, 0, 2) .
```
nagmat84, line 159: What happened to the safe variant? Here and below?

app/Console/Commands/FixPermissions.php
```diff
@@ -88,8 +89,9 @@ public function handle(): int
 	private function fixPermissionsRecursively(string $path): void
 	{
 		try {
-			$actualPerm = fileperms($path);
-			if ($actualPerm === false) {
+			try {
+				$actualPerm = fileperms($path);
+			} catch (FilesystemException) {
```
nagmat84: Why adding a separate `catch`-block here when we return anyway? Why not using the `try`-`catch`-structure we already have and catching this exception at the end of the whole block? This would avoid an unnecessary level of indentation.

app/Factories/AlbumFactory.php
```diff
@@ -153,6 +153,7 @@ public function findBaseAlbumsOrFail(array $albumIDs, bool $withRelations = true
 			$albumQuery->with(['photos', 'children', 'photos.size_variants']);
 		}
 
+		/** @var Collection<BaseAlbum> $result */
```
nagmat84: As far as I know, for Laravel 9 and Larastan 2, `Collection` needs two type parameters. The type of the key (either `string` or `int`) and the type of the value.

app/Http/Controllers/Administration/SettingsController.php
```diff
@@ -387,7 +387,7 @@ public function setCSS(Request $request): void
 		$request->validate(['css' => 'present|nullable|string']);
 		$css = $request->get('css') ?? '';
 
-		if (!Storage::disk('dist')->put('user.css', $css)) {
+		if (Storage::disk('dist')->put('user.css', $css) === false) {
```
nagmat84, line 390: Why that change?

app/Http/Controllers/WebAuthn/WebAuthnLoginController.php
```diff
@@ -0,0 +1,51 @@
...
+		->send(new AssertionValidation($request, User::findOrFail(0)))
+		->thenReturn()->credential;
+
+		if ($credential !== null) {
+			/** @var \Illuminate\Contracts\Auth\Authenticatable $authenticatable */
+			$authenticatable = $credential->authenticatable;
+			Auth::login($authenticatable);
+		}
+
+		return response()->noContent($credential !== null ? 204 : throw new UnauthenticatedException('Invalid credentials'));
```
nagmat84: Normally, I love the ternary operator, in particular if it helps to only assign a variable once and hence making it effectively a constant. But I deem this construction is really ugly, because it is not about assigning.

```suggestion
		if ($credential === null) {
			throw new UnauthenticatedException('Invalid credentials'));
		}
```

app/Http/Controllers/WebAuthn/WebAuthnLoginController.php
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	/**
+	 * Log the user in.
+	 *
+	 * @param AssertedRequest $request
+	 *
+	 * @return \Illuminate\Http\Response
+	 */
+	public function login(AssertedRequest $request, AssertionValidator $assertion): Response
```
nagmat84: If I am not mistaken, we either throw an exception or return an empty `204` response. In this case it also OK, to not return anything which is the same as returning `204` to the client.

```suggestion
	public function login(AssertedRequest $request, AssertionValidator $assertion): void
```

app/Actions/Install/PermissionsChecker.php
ildyria: With Laravel 9, I am also upgrading to nunomaduro/larastan 2 and thecodingmachine/safe 2.
As a result the rules of phpstan are also updated. phpstan now indicates that Safe\substr is no longer need if you are using php8. Before we were using thecodingmachine/safe 1 which was designed for php 7.2+.

app/Assets/SizeVariantGroupedWithRandomSuffixNamingStrategy.php
ildyria: https://github.com/LycheeOrg/Lychee/pull/1469#discussion_r983842184

app/Actions/User/Notify.php
ildyria: If you add the brackets back:

```
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   app/Actions/User/Notify.php                                                                                                                                                                                           
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  25     Parameter #1 ...$values of method Illuminate\Support\Collection<int,Illuminate\Database\Eloquent\Model|null>::push() expects Illuminate\Database\Eloquent\Model|null, Illuminate\Database\Eloquent\Collection given.  
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
 ```

File: app/Http/Controllers/Administration/SettingsController.php (hunk 7 above)
Comment (ildyria) [reply app/Http/Controllers/Administration/SettingsController.php:390]: Because put() does not return a `bool|string` :
```
 ------ -------------------------------------------------------------------- 
  Line   app/Http/Controllers/Administration/SettingsController.php          
 ------ -------------------------------------------------------------------- 
  390    Only booleans are allowed in a negated boolean, bool|string given.  
 ------ -------------------------------------------------------------------- 
```

File: app/Http/Controllers/WebAuthn/WebAuthnLoginController.php [hunk 10]
```diff
@@ -0,0 +1,51 @@
...
+	{
+		$credential = $assertion
+		->send(new AssertionValidation($request, User::findOrFail(0)))
+		->thenReturn()->credential;
```
nagmat84:
```suggestion
			->send(new AssertionValidation($request, User::findOrFail(0)))
			->thenReturn()
			->credential;
```

app/Http/Controllers/WebAuthn/WebAuthnManageController.php
```diff
@@ -0,0 +1,34 @@
...
+	 * @throws UnauthenticatedException
+	 */
+	public function delete(Request $request): void
+	{
+		$id = $request->validate(['id' => 'required|string']);
+
+		/** @var \App\Models\User $user */
+		$user = Auth::user() ?? throw new UnauthenticatedException();
+		$user->webAuthnCredentials()->where('id', $id)->delete();
+	}
```
nagmat84: In order to follow the behaviour authenticate first or throw 401, validate request or throw 422 and finally authorize or throw 403, the order of checks must be swapped. Or create a proper request class for this method as argument.

```suggestion
	public function delete(Request $request): void
	{
		/** @var \App\Models\User $user */
		$user = Auth::user() ?? throw new UnauthenticatedException();
		$id = $request->validate(['id' => 'required|string']);
		$user->webAuthnCredentials()->where('id', $id)->delete();
	}
```

app/Http/Controllers/WebAuthn/WebAuthnRegisterController.php
```diff
@@ -0,0 +1,48 @@
...
+	 */
+	public function options(AttestationRequest $request): Responsable
+	{
+		$request->user = Auth::user() ?? throw new UnauthenticatedException();
+
+		return $request
+			->fastRegistration()
+//            ->userless()
+//            ->allowDuplicates()
+			->toCreate();
```
nagmat84: Why is there commented code in a newly written class? If we don't need that code, delete it.

app/Http/Controllers/WebAuthn/WebAuthnRegisterController.php
```diff
@@ -0,0 +1,48 @@
...
+	 * @return \Illuminate\Http\Response
+	 */
+	public function register(AttestedRequest $request): Response
+	{
+		$request->user = Auth::user() ?? throw new UnauthenticatedException();
+
+		$request->save();
+
+		return response()->noContent();
+	}
```
nagmat84:
```suggestion
	public function register(AttestedRequest $request): void
	{
		$request->user = Auth::user() ?? throw new UnauthenticatedException();
		$request->save();
	}
```

app/Factories/AlbumFactory.php
ildyria: fixed [a066720](https://github.com/LycheeOrg/Lychee/pull/1469/commits/a0667203456e97901e8985e4236d633c788a3336)

app/Http/Controllers/WebAuthn/WebAuthnLoginController.php
ildyria: fixed [773f085](https://github.com/LycheeOrg/Lychee/pull/1469/commits/773f0855084f1c229cfe17229574fd747be83d02)

app/Http/Controllers/WebAuthn/WebAuthnLoginController.php
ildyria: fixed [773f085](https://github.com/LycheeOrg/Lychee/pull/1469/commits/773f0855084f1c229cfe17229574fd747be83d02)

app/Http/Middleware/LocalStorageOnly.php
```diff
@@ -21,8 +21,8 @@ class LocalStorageOnly
 	 */
 	public function handle(Request $request, \Closure $next)
 	{
-		$storageAdapter = SizeVariantNamingStrategy::getImageDisk()->getDriver()->getAdapter();
-		if (!($storageAdapter instanceof Local)) {
+		$storageAdapter = Storage::disk()->getAdapter();
```
nagmat84: No, there was a reason to query the `SizeVariantNamingStrategy` to get the disk, because that is guaranteed to return the disk which is really used by the naming strategy and might be a remote disk someday (e.g. AWS S3 or something similar). `Storage::disk` simply returns the default disk which currently happens to be the same, but this may not be guaranteed in the future.

app/Console/Commands/FixPermissions.php
ildyria: fixed [ae607b8](https://github.com/LycheeOrg/Lychee/pull/1469/commits/ae607b8418eb244707f42e45b970298328d2eb6e)

app/Http/Requests/Album/GetAlbumRequest.php
```diff
@@ -7,6 +7,7 @@
 use App\Http\Requests\BaseApiRequest;
 use App\Http\Requests\Contracts\HasAbstractAlbum;
 use App\Http\Requests\Traits\HasAbstractAlbumTrait;
+use App\Models\Album;
```
nagmat84: Why? If nothing else has changed in the file?

app/Image/BinaryBlob.php
```diff
@@ -104,11 +105,11 @@ public function close(): void
 	protected static function appendStatFilter($stream): StreamStat
 	{
 		$streamStat = new StreamStat();
-		\Safe\stream_filter_append(
+		stream_filter_append(
 			$stream,
 			StreamStatFilter::REGISTERED_NAME,
 			STREAM_FILTER_READ,
-			$streamStat
+			[$streamStat]
```
nagmat84: Why that change to an array? Also the code in `StreamStatFilter` which deals with this change is very inefficient.

app/Image/FileDeleter.php
```diff
@@ -104,11 +104,11 @@ public function do(): void
 
 		// If the disk uses the local driver, we use low-level routines as
 		// these are also able to handle symbolic links in case of doubt
-		$isLocalDisk = $defaultDisk->getDriver()->getAdapter() instanceof LocalAdapter;
+		$isLocalDisk = $defaultDisk->getAdapter() instanceof LocalFilesystemAdapter;
 		if ($isLocalDisk) {
 			foreach ($this->regularFilesOrSymbolicLinks as $fileOrLink) {
 				try {
-					$absolutePath = $defaultDisk->path($fileOrLink);
+					$absolutePath = Storage::path($fileOrLink);
```
nagmat84: No! There was a reason to use `$defaultDisk = SizeVariantNamingStrategy::getImageDisk();` here even though I admit that `defaultDisk` might better be named `defaultImageDisk`.

Same reason as above. `Storage::path` simply returns a path on the global default disk which is not necessarily identical to the disk which stores the images.

app/Image/FlysystemFile.php
```diff
@@ -178,6 +176,6 @@ public function toLocalFile(): NativeLocalFile
 			throw new MediaFileOperationException('file is not hosted locally');
 		}
 
-		return new NativeLocalFile($this->disk->path($this->relativePath));
+		return new NativeLocalFile(Storage::path($this->relativePath));
```
nagmat84: No! See above.

app/Image/StreamStatFilter.php
```diff
@@ -26,8 +26,10 @@ public function filter($in, $out, &$consumed, bool $closing): int
 	{
 		while ($bucket = stream_bucket_make_writeable($in)) {
 			$consumed += $bucket->datalen;
-			if ($this->params instanceof StreamStat) {
-				$this->params->bytes += $bucket->datalen;
+
+			$streamStat = $this->getStreamStats();
```
nagmat84: Why this change? If `append_stream_filter` would still receive a `StreamStat` object directly (instead of an array), this wouldn't be necessary. (Here and below).

app/Image/StreamStatFilter.php
```diff
@@ -60,11 +63,33 @@ public function onClose(): void
...
+		if (is_array($this->params)) {
+			for ($i = 0; $i < count($this->params); $i++) {
+				if ($this->params[$i] instanceof StreamStat) {
+					return $this->params[$i];
+				}
+			}
+		}
+
+		return null;
+	}
```
nagmat84: This is incredibly inefficient. The method `getStreamStats` is called for every processed chunk of a stream (i.e. something around every 4096 bytes of an image file). Iterating over an array of parameters every time in order to find the `StreamStat` is insane. At least this should be cached if using an array is absolutely unavoidable. But I don't get why this change is necessary at all.

app/Metadata/DiskUsage.php
```diff
@@ -56,7 +56,7 @@ public function getTotalSize(string $dir): int
 
 		if (is_dir($dir) === true) {
 			// If on a Unix Host (Linux, Mac OS)
-			if (!$this->is_win()) {
+			if (!$this->is_win() && Helpers::isExecAvailable()) {
```
nagmat84: Independent bug fix?

app/ModelFunctions/LogFunctions.php
```diff
@@ -37,7 +38,7 @@ private function interpolate(string $message, array $context = []): string
 	/**
 	 * Implements log so that AbstractLogger works.
 	 */
-	public function log($loglevel, $message, $context = [])
+	public function log($log, Stringable|string $message, array $context = []): void
```
nagmat84: Why has `$loglevel` be renamed to `$log`? The former is much more telling than the latter.

app/Models/Extensions/ConfigsHas.php
```diff
@@ -32,7 +34,7 @@ public static function hasExiftool(): bool
 		$has_exiftool = self::getValueAsInt('has_exiftool');
 
 		// value not yet set -> let's see if exiftool is available
-		if ($has_exiftool === 2) {
+		if ($has_exiftool === 2 && Helpers::isExecAvailable()) {
```
nagmat84: Independent bug fix?

app/Pipelines/AssertionValidator.php
```diff
@@ -0,0 +1,55 @@
...
+use Laragear\WebAuthn\Assertion\Validator\Pipes\CompileAuthenticatorData;
+use Laragear\WebAuthn\Assertion\Validator\Pipes\CompileClientDataJson;
+use Laragear\WebAuthn\Assertion\Validator\Pipes\IncrementCredentialCounter;
+use Laragear\WebAuthn\Assertion\Validator\Pipes\RetrieveChallenge;
+use Laragear\WebAuthn\Assertion\Validator\Pipes\RetrievesCredentialId;
+
+/**
+ * This validator is literally a copy of the one in WebAuthn,
+ * It is needed because Laragear makes use of is() which is wrong in Laravel
+ * We Update the CheckCredentialsForUser pipe for this reason.
```
nagmat84: Please make this comment more specific.

```php
 * This validator is a copy of {@link put link to specific class here}.
 * This copy is needed because {@link put link to specific method here} uses {@link put link to full path of is()-method} which is wrong in Laravel.
 * <Explain why it is wrong in Laravel>
 * For this reason, we replace {@link to original CheckCredentialForUser} by {@link to our replacement}.
```

config/app.php
```diff
@@ -201,45 +234,10 @@
...
-		'Blade' => Illuminate\Support\Facades\Blade::class,
-		'Broadcast' => Illuminate\Support\Facades\Broadcast::class,
-		'Bus' => Illuminate\Support\Facades\Bus::class,
-		'Cache' => Illuminate\Support\Facades\Cache::class,
-		'Config' => Illuminate\Support\Facades\Config::class,
-		'Cookie' => Illuminate\Support\Facades\Cookie::class,
-		'Crypt' => Illuminate\Support\Facades\Crypt::class,
-		'DB' => Illuminate\Support\Facades\DB::class,
+	'aliases' => Facade::defaultAliases()->merge([
+		// 'ExampleClass' => App\Example\ExampleClass::class,
```
nagmat84:
```suggestion
```

config/filesystems.php
```diff
@@ -41,12 +41,38 @@
...
+			'key' => env('AWS_ACCESS_KEY_ID'),
+			'secret' => env('AWS_SECRET_ACCESS_KEY'),
+			'region' => env('AWS_DEFAULT_REGION'),
+			'bucket' => env('AWS_BUCKET'),
+			'url' => env('AWS_URL'),
+			'endpoint' => env('AWS_ENDPOINT'),
+			'use_path_style_endpoint' => env('AWS_USE_PATH_STYLE_ENDPOINT', false),
+			'throw' => true,
+		],
+
```
nagmat84:
```suggestion
```

Don't add things to code which we do not use. We finally have had cleaned up that.

config/filesystems.php
```diff
@@ -41,12 +41,38 @@
...
 		// Lychee uses the disk "images" to store the media files
 		'images' => [
 			'driver' => 'local',
 			'root' => env('LYCHEE_UPLOADS', public_path('uploads/')),
 			'url' => env('LYCHEE_UPLOADS_URL', 'uploads/'),
-			'visibility' => env('LYCHEE_IMAGE_VISIBILITY', 'public'),
+			'visibility' => 'public',
```
nagmat84: Why that? We just made it configurable some PR ago.

config/filesystems.php
```diff
@@ -61,20 +87,6 @@
...
-		'images' => [
-			'driver' => 's3',
-			'key' => env('AWS_ACCESS_KEY_ID'),
-			'secret' => env('AWS_SECRET_ACCESS_KEY'),
-			'region' => env('AWS_DEFAULT_REGION', 'us-east-1'),
-			'bucket' => env('AWS_BUCKET'),
-			'url' => env('AWS_URL'),
-			'endpoint' => env('AWS_ENDPOINT'),
-		],*/
-
```
nagmat84: Why do you add unhelpful default examples by the framework but throw out the only example which might eventually turn out to be helpful?

database/migrations/2020_12_12_203831_create_web_authn_tables.php
```diff
@@ -15,34 +13,6 @@ class CreateWebAuthnTables extends Migration
...
-			// since it's very straightforward. You can change this for a plain string.
-			// It must be nullable because those old U2F keys do not use user handle.
-			$table->uuid('user_handle')->nullable();
-
-			$table->timestamps();
-			$table->softDeletes(WebAuthnCredential::DELETED_AT);
-			Configs::where('key', '=', 'username')->orWhere('key', '=', 'password')->update(['type_range' => STRING]);
-
-			$table->primary(['id', 'user_id']);
-		});
```
nagmat84: Why this change of a past migration?

database/migrations/2021_12_04_181200_refactor_models.php
```diff
@@ -280,7 +280,7 @@ public function up()
 		$this->createSizeVariantTableUp();
 		$this->createSymLinkTableUp();
 		$this->createRemainingForeignConstraints();
-		$this->createWebAuthnTableUp();
+		// $this->createWebAuthnTableUp();
```
nagmat84: If code isn't needed anymore or became even wrong/harmful/whatever, delete it properly. Or if there is a chance that his code might be needed again in the future, add a comment. But otherwise everyone will wonder in the future why this code has been commented out.

database/migrations/2021_12_04_181200_refactor_models.php
```diff
@@ -319,7 +319,7 @@ public function down()
 		$this->createUserAlbumTableDown();
 		$this->createPhotoTableDown();
 		$this->createSymLinkTableDown();
-		$this->createWebAuthnTableDown();
+		// $this->createWebAuthnTableDown();
```
nagmat84: See above.

database/migrations/2021_12_04_181200_refactor_models.php
```diff
@@ -986,40 +992,40 @@ private function renamePageContentTable(): void
...
+
+	// private function createWebAuthnTableUp(): void
+	// {
+	// 	$this->createWebAuthnTable(6);
+	// }
+
+	// private function createWebAuthnTableDown(): void
+	// {
+	// 	$this->createWebAuthnTable(0);
+	// }
```
nagmat84: See above.

database/migrations/2022_07_09_130303_create_webauthn_credentials.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+/**
+ * @see \Laragear\WebAuthn\Models\WebAuthnCredential
+ */
+return new class() extends Migration {
```
nagmat84: New syntax?!

database/migrations/2022_07_09_130303_create_webauthn_credentials.php
```diff
@@ -0,0 +1,79 @@
...
+	 * @return void
+	 */
+	public function up(): void
+	{
+		Schema::create('webauthn_credentials', static function (Blueprint $table): void {
+			static::defaultBlueprint($table);
+
+			// You may add here your own columns...
+			//
+			// $table->timestamp('last_login_at')->nullable();
```
nagmat84: Cleanup.

database/migrations/2022_07_09_130303_create_webauthn_credentials.php
```diff
@@ -0,0 +1,79 @@
...
+		// This is the public key the credential uses to verify the challenges.
+		$table->text('public_key');
+		// The attestation of the public key.
+		$table->string('attestation_format')->default('none');
+		// This would hold the certificate chain for other different attestation formats.
+		$table->json('certificates')->nullable();
+
+		// A way to disable the credential without deleting it.
+		$table->timestamp('disabled_at')->nullable();
+		$table->timestamps();
```
nagmat84: No, we mustn't use the methods provided by the framework to create the timestamp columns because the use the wrong types for certain DB engines.

nagmat84 (changes requested):
98 of 116 files.

app/Image/StreamStatFilter.php
ildyria: That change is necessary because `append_stream_filter` expects an array. :(

app/Http/Requests/Album/GetAlbumRequest.php
ildyria: I guess that was before I merge master. This changed line 25 from Album to AbstractAlbum :)

ildyria: Fixed [97cc858](https://github.com/LycheeOrg/Lychee/pull/1469/commits/97cc858f83753f0dee4d6a4527593bf38fc99b16)

app/Image/BinaryBlob.php
ildyria: From `use function Safe\stream_filter_append`:
```php
/**
 * Adds filtername to the list of filters
 * attached to stream.
 *
 * @param resource $stream The target stream.
 * @param string $filtername The filter name.
 * @param int $read_write By default, stream_filter_append will
 * attach the filter to the read filter chain
 * if the file was opened for reading (i.e. File Mode:
 * r, and/or +).  The filter
 * will also be attached to the write filter chain
 * if the file was opened for writing (i.e. File Mode:
 * w, a, and/or +).
 * STREAM_FILTER_READ,
 * STREAM_FILTER_WRITE, and/or
 * STREAM_FILTER_ALL can also be passed to the
 * read_write parameter to override this behavior.
 * @param array $params This filter will be added with the specified
 * params to the end of
 * the list and will therefore be called last during stream operations.
 * To add a filter to the beginning of the list, use
 * stream_filter_prepend.
 * @return resource Returns a resource on success. The resource can be
 * used to refer to this filter instance during a call to
 * stream_filter_remove.
 *
 * FALSE is returned if stream is not a resource or
 * if filtername cannot be located.
 * @throws StreamException
 *
 */
function stream_filter_append($stream, string $filtername, int $read_write = null, array $params = null)
{
    error_clear_last();
    if ($params !== null) {
        $safeResult = \stream_filter_append($stream, $filtername, $read_write, $params);
    } elseif ($read_write !== null) {
        $safeResult = \stream_filter_append($stream, $filtername, $read_write);
    } else {
        $safeResult = \stream_filter_append($stream, $filtername);
    }
    if ($safeResult === false) {
        throw StreamException::createFromPhpError();
    }
    return $safeResult;
}

```

ildyria: Created a PR here https://github.com/thecodingmachine/safe/pull/395
We will see.

app/Image/StreamStatFilter.php
ildyria: https://github.com/thecodingmachine/safe/pull/395

app/ModelFunctions/LogFunctions.php
ildyria: > Method `App\ModelFunctions\LogFunctions::log()` is not compatible with method `Psr\Log\AbstractLogger::log()`. intelephense(1038)

ildyria:
```php
<?php

namespace Psr\Log;

/**
 * Describes a logger instance.
 *
 * The message MUST be a string or object implementing __toString().
 *
 * The message MAY contain placeholders in the form: {foo} where foo
 * will be replaced by the context data in key "foo".
 *
 * The context array can contain arbitrary data. The only assumption that
 * can be made by implementors is that if an Exception instance is given
 * to produce a stack trace, it MUST be in a key named "exception".
 *
 * See https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md
 * for the full interface specification.
 */
interface LoggerInterface
{
...
    /**
     * Logs with an arbitrary level.
     *
     * @param mixed   $level
     * @param string|\Stringable $message
     * @param mixed[] $context
     *
     * @return void
     *
     * @throws \Psr\Log\InvalidArgumentException
     */
    public function log($level, string|\Stringable $message, array $context = []): void;
}
```

app/Http/Controllers/WebAuthn/WebAuthnLoginController.php
ildyria: Fixed [d2052dc](https://github.com/LycheeOrg/Lychee/pull/1469/commits/d2052dc2dbd87fe49b1bf75434ef40ed6ce22e29)

app/Http/Controllers/WebAuthn/WebAuthnManageController.php
ildyria: fixed [d2052dc](https://github.com/LycheeOrg/Lychee/pull/1469/commits/d2052dc2dbd87fe49b1bf75434ef40ed6ce22e29)

app/Http/Controllers/WebAuthn/WebAuthnRegisterController.php
ildyria: fixed [d2052dc](https://github.com/LycheeOrg/Lychee/pull/1469/commits/d2052dc2dbd87fe49b1bf75434ef40ed6ce22e29)

app/Http/Controllers/WebAuthn/WebAuthnRegisterController.php
ildyria: fixed [d2052dc](https://github.com/LycheeOrg/Lychee/pull/1469/commits/d2052dc2dbd87fe49b1bf75434ef40ed6ce22e29)

app/Metadata/DiskUsage.php
ildyria: See https://github.com/LycheeOrg/Lychee/pull/1540

app/Models/Extensions/ConfigsHas.php
ildyria: See https://github.com/LycheeOrg/Lychee/pull/1540

config/filesystems.php
ildyria: Fixed in [bb29f73](https://github.com/LycheeOrg/Lychee/pull/1469/commits/bb29f730217eb210ec0f538355080e183a74fe0c)

config/filesystems.php
ildyria: Fixed [bb29f73](https://github.com/LycheeOrg/Lychee/pull/1469/commits/bb29f730217eb210ec0f538355080e183a74fe0c)

config/filesystems.php
ildyria: Fixed [bb29f73](https://github.com/LycheeOrg/Lychee/pull/1469/commits/bb29f730217eb210ec0f538355080e183a74fe0c)

config/app.php
ildyria: Fixed [bb29f73](https://github.com/LycheeOrg/Lychee/pull/1469/commits/bb29f730217eb210ec0f538355080e183a74fe0c)

app/Pipelines/AssertionValidator.php
ildyria: Fixed [3338936](https://github.com/LycheeOrg/Lychee/pull/1469/commits/3338936b0ec2c21f32d512c53e849b7e74c04468)

database/migrations/2020_12_12_203831_create_web_authn_tables.php
ildyria: It is adding a table which will be later deleted. This was used by Larapass, but we are no longer using Larapass.

database/migrations/2022_07_09_130303_create_webauthn_credentials.php
ildyria: fixed [faf44a2](https://github.com/LycheeOrg/Lychee/pull/1469/commits/faf44a2f63086be3902560abe44d2c11a8daf791)

database/migrations/2022_07_09_130303_create_webauthn_credentials.php
ildyria: fixed [faf44a2](https://github.com/LycheeOrg/Lychee/pull/1469/commits/faf44a2f63086be3902560abe44d2c11a8daf791)

database/migrations/2021_12_04_181200_refactor_models.php
ildyria: fixed [faf44a2](https://github.com/LycheeOrg/Lychee/pull/1469/commits/faf44a2f63086be3902560abe44d2c11a8daf791)

database/migrations/2021_12_04_181200_refactor_models.php
ildyria: fixed [faf44a2](https://github.com/LycheeOrg/Lychee/pull/1469/commits/faf44a2f63086be3902560abe44d2c11a8daf791)

database/migrations/2020_12_12_203831_create_web_authn_tables.php
ildyria: fixed [faf44a2](https://github.com/LycheeOrg/Lychee/pull/1469/commits/faf44a2f63086be3902560abe44d2c11a8daf791)

database/migrations/2021_12_04_181200_refactor_models.php
ildyria: fixed [faf44a2](https://github.com/LycheeOrg/Lychee/pull/1469/commits/faf44a2f63086be3902560abe44d2c11a8daf791)

app/Actions/User/Notify.php
nagmat84: But have you tried, if it actually works? The error by Larastan/PHPStan seems to be a false positive.

The return type of `Eloquent::find` is a union type, see [Laravel API 9.x: `Illuminate/Database/Eloquent/Builder::find()`](https://laravel.com/api/9.x/Illuminate/Database/Eloquent/Builder.html#method_find):

 - `Builder::find()` may return a `Collection` if multiple results are found
 - `Builder::find()` returns a single instance of `Model` if queried for an ID as we do here

However, it seems that the Larastan stubs (wrongly) assume that `find` always returns a `Collection` and hence PHPStan is complaining about the additional brackets, because `[]` around a `Collection` would indeed make no sense as it would create a collection of a collection.

I haven't tried to run the code yet, but this is what I conjecture from code inspection.

app/Metadata/DiskUsage.php
nagmat84: I'll keep this conversation open until https://github.com/LycheeOrg/Lychee/pull/1540 has been merged into master and master again into this PR such that this change vanishes from the diff of this PR.

app/Pipelines/AssertionValidator.php
nagmat84: > in their brilliant stupidity

YMMD. :rofl:

database/migrations/2020_12_12_203831_create_web_authn_tables.php
nagmat84: > It is adding a table which will be later deleted. This was used by Larapass, but we are no longer using Larapass.

I suspected so, but I am unsure whether I like this kind of approach which plays around with old migrations. I would rather not to.

I get your principal idea. You want to improve the efficiency of the entire migration for users which will start with a clean install in the future. For those users it doesn't make much sense to first create an (empty) table `web_authn_credentials` which is later replaced by another (empty) table `webauthn_credentials`. However, we have this kind of situation in other places, too. The most extreme example of that is [`database/migrations/2021_12_04_181200_refactor_models.php`](https://github.com/LycheeOrg/Lychee/blob/52ea5b8861a7d34b2c5e26d9b5df11053568f774/database/migrations/2021_12_04_181200_refactor_models.php) which basically recreates the whole Lychee DB. For fresh installations it would be much more efficient to simply start with a DB script that creates the Lychee DB as it is supposed to be _now_ instead of running a long sequence of delta migrations.

If I remember correctly, @kamil4 or @d7415 exactly asked that question whether it is possible to have something like that some time ago. Unfortunately, this seems the approach being used by Laravel. Back then we answered the question definitely negative.

If we start to play around with old migrations we may not only consider fresh installations but we must also consider all corner cases like users which are in the middle between two changed migrations and then proceed with a migration which only "sees" one of the changes, etc. pp. I understand that you tried to take care of that by not only changing this migration, but also adopting all other migrations which depend on this one to deal with all possible cases.

But if we decide to go this way and do not only apply it to the current migration, but use it as a general design pattern, then we will soon end up in a hell of endless possibilities and nobody will be able to understand the migrations anymore. It will also become much harder to help people which are stuck in a migration, because there are simply more possibilities how the DB scheme may look like. (At the moment there is always exactly one for each completed migration.)

Hence, I am strongly in favour of _not_ touching old migrations unless we really need to, because the contain a bug for example. For the moment I would keep it like that.

In the long run, when the migration sequence becomes too annoying we should consider to ship out Lychee 5 and make a clean cut. We can then provide a 1st migration script which either creates the DB from scratch for all fresh installations or allows to migrate from the latest 4.x.y version. Moreover, all users who are not on the latest 4.x.y. version would be forced to first migrate to that before they can migrate to Lychee 5. This sort of clean cut would reduce the number of possibilities to two: either it is a fresh install or a migration from the latest Lychee 4.x.y. From there we would start with a new sequence of delta migrations.

I am curious to hear @kamil4 and @d7415 thoughts on that.

PS: Sorry for the long text.

database/migrations/2021_12_04_181200_refactor_models.php
nagmat84: I will keep this open for the moment until @kamil4 and @d7415 have given their thoughts on https://github.com/LycheeOrg/Lychee/pull/1469#discussion_r985245067.

I am not in favour of fiddling with old migrations for a little efficiency gain as I deem the risk of introducing new lurking bugs as too high. For fresh installations all the tables are empty anyway so it is not much what we gain here.

app/ModelFunctions/LogFunctions.php
nagmat84: > Method `App\ModelFunctions\LogFunctions::log()` is not compatible with method `Psr\Log\AbstractLogger::log()`.

This doesn't answer the question.

The signature of the parent class is

```php
 public function log($level, string|\Stringable $message, array $context = []): void
```

The previous signature of our method was

```php
 public function log($loglevel, string|\Stringable $message, array $context = []): void
```

and now you changed it to

```php
 public function log($log, string|\Stringable $message, array $context = []): void
```

The parameter's name in the parent class is just `$level`. For our class it has been `$loglevel`. OK, that has been an inconsistency, but this no explanation for changing it to `$log`. First of all, it does not solve the inconsistency (if this has been the goal) and as I said in the first comment, `$log` is much less telling than either `$loglevel` or `$level` because it can easily be confused with the actual log message.

routes/api.php
```diff
@@ -93,9 +93,20 @@
...
-Route::post('/WebAuthn::register/gen', [Administration\WebAuthController::class, 'generateRegistration']);
-Route::post('/WebAuthn::register', [Administration\WebAuthController::class, 'verifyRegistration']);
-Route::post('/WebAuthn::login/gen', [Administration\WebAuthController::class, 'generateAuthentication']);
-Route::post('/WebAuthn::login', [Administration\WebAuthController::class, 'verifyAuthentication']);
-Route::post('/WebAuthn::list', [Administration\WebAuthController::class, 'list'])->middleware(['admin']);
-Route::post('/WebAuthn::delete', [Administration\WebAuthController::class, 'delete'])->middleware(['admin']);
+// Route::post('/WebAuthn::register/gen', [Administration\WebAuthController::class, 'generateRegistration']);
+// Route::post('/WebAuthn::register', [Administration\WebAuthController::class, 'verifyRegistration']);
+// Route::post('/WebAuthn::login/gen', [Administration\WebAuthController::class, 'generateAuthentication']);
+// Route::post('/WebAuthn::login', [Administration\WebAuthController::class, 'verifyAuthentication']);
```
nagmat84: As always.

routes/api.php
```diff
@@ -93,9 +93,20 @@
...
+Route::post('/WebAuthn::list', [WebAuthn\WebAuthnManageController::class, 'list']);
+Route::post('/WebAuthn::delete', [WebAuthn\WebAuthnManageController::class, 'delete']);
+
+// WebAuthn Routes
+Route::post('/WebAuthn::register/options', [\App\Http\Controllers\WebAuthn\WebAuthnRegisterController::class, 'options'])
+	->name('webauthn.register.options');
```
nagmat84: Why is the comment placed between the 2nd and the 3rd route and not at the top, i.e. before the 1st route? I am unsure whether we need that comment at all, but if we put this comment here it should be placed at the top like a caption.

```suggestion
// WebAuthn Routes
Route::post('/WebAuthn::list', [WebAuthn\WebAuthnManageController::class, 'list']);
Route::post('/WebAuthn::delete', [WebAuthn\WebAuthnManageController::class, 'delete']);
Route::post('/WebAuthn::register/options', [\App\Http\Controllers\WebAuthn\WebAuthnRegisterController::class, 'options'])
	->name('webauthn.register.options');
```

tests/Feature/InstallTest.php
```diff
@@ -69,6 +69,7 @@ public function testInstall(): void
 			'albums',
 			'base_albums',
 			'web_authn_credentials',
+			'webauthn_credentials',
```
nagmat84: The first one should be removed, because it is replaced be the second. Right?

tests/Feature/CommandFixPermissionsTest.php
```diff
@@ -50,8 +50,8 @@ public function testFixPermissions(): void
 		])->assertSuccessful();
 
 		clearstatcache(true);
-		static::assertEquals(00664, fileperms($filePath) & 07777);
-		static::assertEquals(02775, fileperms($dirPath) & 07777);
+		static::assertEquals(00664, fileperms($filePath) & 07777, $filePath);
+		static::assertEquals(02775, fileperms($dirPath) & 07777, $dirPath);
```
nagmat84: Why this change? I am not sure if it even improves anything. If the 3rd parameter is not given, the error message in case of a failure defaults to `"Failed asserting that (2nd parameter) equals (value of 1st parameter)"`. While I agree that the error message could be improved and made more specific, the error message now is _only_ the path of the problematic file/directory without any indication what is even wrong. I wouldn't call this an improvement.

database/migrations/2020_12_12_203831_create_web_authn_tables.php
d7415: I would definitely like to trim some of the migrations. Ideally we could add a migration that laravel would recognise as a shortcut. but there doesn't seem to be a way to do that or similar.

I suppose we could cheat by adding a migration before that checks migration status, and if x, y and z haven't been run yet, does its thing and marks those as complete. Not sure if this would work though and it could easily go wrong.

I agree that changing existing migrations is messy, so I'm inclined to agree with the "wipe the slate clean with v5" approach.

database/migrations/2022_07_09_130303_create_webauthn_credentials.php
ildyria: fixed [faf44a2](https://github.com/LycheeOrg/Lychee/pull/1469/commits/faf44a2f63086be3902560abe44d2c11a8daf791)

app/ModelFunctions/LogFunctions.php
ildyria: Fixed [9f06381](https://github.com/LycheeOrg/Lychee/pull/1469/commits/9f063810edba917a0fa1b989812eb5616b21f4d8)

app/Image/StreamStatFilter.php
ildyria: Fixed [2d90746](https://github.com/LycheeOrg/Lychee/pull/1469/commits/2d9074661a26d88cde545f9dfd78ee16747e5e21)

app/Image/StreamStatFilter.php
ildyria: Fixed [2d90746](https://github.com/LycheeOrg/Lychee/pull/1469/commits/2d9074661a26d88cde545f9dfd78ee16747e5e21)

app/Image/BinaryBlob.php
ildyria: Fixed [2d90746](https://github.com/LycheeOrg/Lychee/pull/1469/commits/2d9074661a26d88cde545f9dfd78ee16747e5e21)

app/Metadata/DiskUsage.php
ildyria: Resolved.

app/Models/Extensions/ConfigsHas.php
ildyria: Resolved.

routes/api.php
ildyria: Fixed in [4cdfc8c](https://github.com/LycheeOrg/Lychee/pull/1469/commits/4cdfc8cfffc94a3b571ec8c68b2fa6063a0e5d1f)

routes/api.php
ildyria: Fixed in [4cdfc8c](https://github.com/LycheeOrg/Lychee/pull/1469/commits/4cdfc8cfffc94a3b571ec8c68b2fa6063a0e5d1f)

tests/Feature/CommandFixPermissionsTest.php
nagmat84: I assume this has been fixed in https://github.com/LycheeOrg/Lychee/pull/1469/commits/fae68b46c3e00709a6f54df5e1f9a284c14fe7d1, right?

app/Http/Controllers/Administration/SettingsController.php
nagmat84: Sure, but why did it work before? Did Laravel change the return type?

database/migrations/2021_12_04_181200_refactor_models.php
nagmat84: I am not sure if I like the fix of https://github.com/LycheeOrg/Lychee/pull/1469/commits/faf44a2f63086be3902560abe44d2c11a8daf791. Yes, it improves things in the sense that it removes code which has been commented out. But the whole change still touches old migrations which I would definitely not do. I would simply keep the old migrations as is, even if these migrations do things which turn out to be unnecessary later.

Also see  https://github.com/LycheeOrg/Lychee/pull/1469#discussion_r985269381.

database/migrations/2021_12_04_181200_refactor_models.php
nagmat84: See https://github.com/LycheeOrg/Lychee/pull/1469#discussion_r994333603

app/Actions/User/Notify.php
ildyria: Had a deeper look at it and agree with your point. Revered.

ildyria: Fixed in [25db8c3](https://github.com/LycheeOrg/Lychee/pull/1469/commits/25db8c3a0f74087e089b3d677ffea1c66e742b60)

app/Http/Controllers/Administration/SettingsController.php
ildyria: File system contract says:
```php
    /**
     * Write the contents of a file.
     *
     * @param  string  $path
     * @param  string|resource  $contents
     * @param  mixed  $options
     * @return bool
     */
    public function put($path, $contents, $options = []);
```
So I guess this is a problem of the phpstan/larastan...

ildyria: Correction, it is once again a stupid problem of Laravel not following their own contract...
```php
    /**
     * Write the contents of a file.
     *
     * @param  string  $path
     * @param  \Psr\Http\Message\StreamInterface|\Illuminate\Http\File|\Illuminate\Http\UploadedFile|string|resource  $contents
     * @param  mixed  $options
     * @return string|bool
     */
    public function put($path, $contents, $options = [])
    {
        $options = is_string($options)
                     ? ['visibility' => $options]
                     : (array) $options;

        // If the given contents is actually a file or uploaded file instance than we will
        // automatically store the file using a stream. This provides a convenient path
        // for the developer to store streams without managing them manually in code.
        if ($contents instanceof File ||
            $contents instanceof UploadedFile) {
            return $this->putFile($path, $contents, $options);
        }

        try {
            if ($contents instanceof StreamInterface) {
                $this->driver->writeStream($path, $contents->detach(), $options);

                return true;
            }

            is_resource($contents)
                ? $this->driver->writeStream($path, $contents, $options)
                : $this->driver->write($path, $contents, $options);
        } catch (UnableToWriteFile|UnableToSetVisibility $e) {
            throw_if($this->throwsExceptions(), $e);

            return false;
        }

        return true;
    }
```

tests/Feature/InstallTest.php
ildyria: yes.

tests/Feature/CommandFixPermissionsTest.php
ildyria: yes.

database/migrations/2021_12_04_181200_refactor_models.php
ildyria: fixed [3173156](https://github.com/LycheeOrg/Lychee/pull/1469/commits/317315682bd1bc5e5c95bc2f959adde8cff2d402)

database/migrations/2021_12_04_181200_refactor_models.php
ildyria: fixed [3173156](https://github.com/LycheeOrg/Lychee/pull/1469/commits/317315682bd1bc5e5c95bc2f959adde8cff2d402)

database/migrations/2020_12_12_203831_create_web_authn_tables.php
ildyria: fixed [3173156](https://github.com/LycheeOrg/Lychee/pull/1469/commits/317315682bd1bc5e5c95bc2f959adde8cff2d402)

database/migrations/2021_12_04_181200_refactor_models.php
ildyria: fixed [3173156](https://github.com/LycheeOrg/Lychee/pull/1469/commits/317315682bd1bc5e5c95bc2f959adde8cff2d402)

app/Image/FlysystemFile.php
ildyria: fixed [6a74bc3](https://github.com/LycheeOrg/Lychee/pull/1469/commits/6a74bc3634de4f36a162b57f875c42fb0965cd2c)

app/Image/FileDeleter.php
ildyria: fixed [6a74bc3](https://github.com/LycheeOrg/Lychee/pull/1469/commits/6a74bc3634de4f36a162b57f875c42fb0965cd2c)

app/Http/Middleware/LocalStorageOnly.php
ildyria: fixed [6a74bc3](https://github.com/LycheeOrg/Lychee/pull/1469/commits/6a74bc3634de4f36a162b57f875c42fb0965cd2c)

app/Actions/User/Notify.php
```diff
@@ -22,6 +22,10 @@ public function do(Photo $photo): void
 		$users = new Collection([User::query()->find(0)]);
 		$album = $photo->album;
 		if ($album !== null) {
+			// phpstan throws this error
+			// Parameter #1 ...$values of method Illuminate\Support\Collection<int,Illuminate\Database\Eloquent\Model|null>::push() expects Illuminate\Database\Eloquent\Model|null, Illuminate\Database\Eloquent\Collection given.
+			// However the signature is push(... $values) which also consider collection/array...
+			// @phpstan-ignore-next-line
 			$users->push($album->shared_with);
 			$users->push($album->owner);
```
nagmat84: PHPstan is right here. Allthough the signature is `push(...$values)`, this does not imply that you can pass an array directly. A variable-length parameter means that you can do something like `push($model1, $model2, $model3)` and then these tree models are available as as an array `$values` inside the function. However, if you call `push($models)` and `$models` is an array, then `$values` is an array with a single element which contains the array `$models`. That is not what we want here.

There are several solutions to this problem discussed in https://stackoverflow.com/questions/9045366/php-pass-array-as-variable-length-argument-list and https://stackoverflow.com/questions/744145/passing-an-array-as-arguments-not-an-array-in-php. I don't like neither of them. They are all ugly as hell.

Another solution would be to use [`Collection::concat`](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Collections/Collection.php#L935). Unfortunately, `concat` returns a new `Collection` and does not concatenate the collections in place, so we would need something like `$users = $users->concat($album->shared_width)`. But this is probably still the best solution.

app/Actions/User/Notify.php
```diff
@@ -22,6 +22,10 @@ public function do(Photo $photo): void
 		$users = new Collection([User::query()->find(0)]);
```
nagmat84: We discussed that line already and you had to add some PHPstan exception rules to make it work. Maybe, the best thing we can do here is to be overly explicit like this

```suggestion
		/** @var Collection<int, User> */
		$users = new Collection();
		$users->add(User::query()->sole(['id' => 0]));
```

That should remove the need for any PHPstan exception.

app/Http/Controllers/Administration/SettingsController.php
nagmat84: I see great opportunities for another hopeless PR towards Laravel. :rofl: But otherwise I deem this conversation resolved.

app/Image/FlysystemFile.php
```diff
@@ -60,16 +61,12 @@ public function write($stream, bool $collectStatistics = false): ?StreamStat
...
 			$streamStat = $collectStatistics ? static::appendStatFilter($stream) : null;
 
-			// TODO: `put` must be replaced by `writeStream` when Flysystem 2 is shipped with Laravel 9
-			// This will also be more consistent with `readStream`.
-			// Note that v1 also provides a method `writeStream`, but this is a misnomer.
-			// See: https://flysystem.thephpleague.com/v2/docs/what-is-new/
-			if (!$this->disk->put($this->relativePath, $stream)) {
-				throw new FlyException('Filesystem::put failed');
+			if (!$this->disk->writeStream($this->relativePath, $stream)) {
+				throw new FlySystemLycheeException('Filesystem::put failed');
```
nagmat84:
```suggestion
				throw new FlySystemLycheeException('Filesystem::writeStream failed');
```

app/Image/StreamStatFilter.php
```diff
@@ -26,6 +26,7 @@ public function filter($in, $out, &$consumed, bool $closing): int
 	{
 		while ($bucket = stream_bucket_make_writeable($in)) {
 			$consumed += $bucket->datalen;
+
```
nagmat84:
```suggestion
```

So that this file is removed from the diff.

database/migrations/2022_07_09_130303_create_webauthn_credentials.php
```diff
@@ -0,0 +1,111 @@
...
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up(): void
+	{
+		Schema::dropIfExists('web_authn_credentials');
+		Schema::create('webauthn_credentials', static function (Blueprint $table): void {
+			static::defaultBlueprint($table);
```
nagmat84, line 21: Just currious: Why this additional indirection? All our other migrations simply have the table definition in place. Why this change of heart?

Moreover, the function name `defaultBlueprint` is not telling anything actually.

ildyria: It is the result of :
`php artisan vendor:publish --provider="Laragear\WebAuthn\WebAuthnServiceProvider" --tag="migrations"`

nagmat84: OK. We can leave it like that.

app/Actions/User/Notify.php
ildyria: Fixed [33bc864](https://github.com/LycheeOrg/Lychee/pull/1469/commits/33bc8649ce7ab16428113bf4ecfe3b1cf7227bdc)

app/Actions/User/Notify.php
ildyria: Fixed [4b1ff0b](https://github.com/LycheeOrg/Lychee/pull/1469/commits/4b1ff0be6a835e98c9f532d00fcf75bc55ff2efb)

app/Image/StreamStatFilter.php
ildyria: Fixed [30ecdef](https://github.com/LycheeOrg/Lychee/pull/1469/commits/30ecdef753c7135dc6d34b141e672ef563eaba5e)

app/Image/FlysystemFile.php
ildyria: Fixed [30ecdef](https://github.com/LycheeOrg/Lychee/pull/1469/commits/30ecdef753c7135dc6d34b141e672ef563eaba5e)

app/Actions/Album/Delete.php
ildyria: if this is not here :
```
Caused by
TypeError: League\Flysystem\Filesystem::has(): Argument #1 ($location) must be of type string, null given, called in ~/Lychee/vendor/laravel/framework/src/Illuminate/Filesystem/FilesystemAdapter.php on line 180
```

nagmat84 (approved):

app/Actions/Import/FromUrl.php
```diff
@@ -49,6 +49,8 @@ public function do(array $urls, ?Album $album): Collection
 					// Silently do nothing, if `set_time_limit` is denied.
 				}
 
+				// If the component parameter is specified, this function returns a string (or int in case of PHP_URL_PORT )
```
qwerty287:
```suggestion
				// If the component parameter is specified, this function returns a string (or int in case of PHP_URL_PORT)
```

app/Image/DownloadedFile.php
```diff
@@ -27,6 +27,7 @@ class DownloadedFile extends TemporaryLocalFile
 	public function __construct(string $url)
 	{
 		try {
+			/** @var string $path : because we provide directly PHP_URL_PATH */
```
qwerty287: Here you use `:`, in another file (sorry, don't remember), you used `--`. Would be good to stay consistent here.

app/Image/FlysystemFile.php
```diff
@@ -2,18 +2,19 @@
...
+use League\Flysystem\FilesystemException;
+use League\Flysystem\Local\LocalFilesystemAdapter;
 use function Safe\fclose;
 
 /**
  * Class FlysystemFile.
  *
  * This class is based on legacy Flysystem v1 which ships with Laravel 8.
- * Laravel 9 will migrate to Flysystem v2 which provides a different and
+ * Laravel 9 will migrate to Flysystem v3 which provides a different and
```
qwerty287: I'd remove this comment. It's not important how this was implemented in Laravel 8.

qwerty287 (approved):
I added some comments about comments :)

app/Image/FlysystemFile.php
ildyria: fixed in [3d1ad0c](https://github.com/LycheeOrg/Lychee/pull/1469/commits/3d1ad0cdbb8bdbc9de354e26ca22e5b211056a8a)

app/Image/DownloadedFile.php
ildyria: Fixed in [3d1ad0c](https://github.com/LycheeOrg/Lychee/pull/1469/commits/3d1ad0cdbb8bdbc9de354e26ca22e5b211056a8a)

tests/Feature/SharingTestScenariosAbstract.php
```diff
@@ -57,28 +57,28 @@
  */
 abstract class SharingTestScenariosAbstract extends SharingTestBase
 {
-	/** @var bool the previously configured visibility for public photos */
+	/** @var bool -- the previously configured visibility for public photos */
```
nagmat84: No, I know that @qwerty287 asked you to make it consistent, because you used `:` as a separator and `--` in another place. I also had seen that, but it wasn't sufficiently important to me to ask for changing it.

But now you have changed it everywhere and now I have to ask you to revert it back, because it is not good. The PHPdoc comments don't use any separator at all. This way all our documents would start with an annoying `--` in the generated documentation. This is _not_ JsDoc where such a separator is actually required.

We should stick to the docs. See here: https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/var.html#var

This is not arbitrary. Please revert it everywhere.

nagmat84 (changes requested):

tests/Feature/SharingTestScenariosAbstract.php
ildyria: Done.

ildyria: See [765957d](https://github.com/LycheeOrg/Lychee/pull/1469/commits/765957d951b036678753a021369c8b40bbe65673)

app/Image/DownloadedFile.php
qwerty287: Can you remove the colon here as well, as you with the double-hyphen? That's the best solution because this also follows the docs then.

ildyria: It is already done. :')

qwerty287: Yeah, sorry

qwerty287 (approved):

qwerty287 (approved):

nagmat84 (approved):

PR #1471, opened 2022-08-20
Fix user test

nagmat84:
You might remember the sometimes failing test suite in the CI. As this problem just hit me locally, I guess I found the problem. This PR fixes it.

Currently, the test `UsersTest::testSetLogin` tried to set the admin login without ensuring that it has authenticated before. This sometimes works, if the test suite is till being authenticated as admin from a previous test, but it is not reliable, of course.

I also changed the pre-condition of the test to be more consistent and I renamed the test.

### DIFF
--- a/tests/Feature/UsersTest.php
+++ b/tests/Feature/UsersTest.php
@@ -28,7 +28,7 @@ class UsersTest extends TestCase
 {
 	use InteractWithSmartAlbums;
 
-	public function testSetAminLoginIfAdminUnconfigured(): void
+	public function testSetAdminLoginIfAdminUnconfigured(): void
 	{
 		/**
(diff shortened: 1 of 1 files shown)


tests/Feature/UsersTest.php
```diff
@@ -28,49 +28,30 @@ class UsersTest extends TestCase
 {
 	use InteractWithSmartAlbums;
 
-	public function testSetLogin(): void
+	public function testSetAminLoginIfAdminUnconfigured(): void
```
qwerty287:
```suggestion
	public function testSetAdminLoginIfAdminUnconfigured(): void
```

qwerty287 (approved):

PR #1475, opened 2022-08-21
Fixes #1473

### LINKED ISSUE #1473: Installer not updated for v4.5.3 directories
v4.5.3 [changed the directory structure of Lychee](https://github.com/LycheeOrg/Lychee/releases/tag/v4.5.3). The tests in the installer still test the old paths.

1. Start a clean installation of Lychee
2. Run the installer
3. Go to the directory tests.

**Screenshots**
2022-08-20-225813_751x711_scrot

nagmat84:
I decided to remove the tests for the various sub-folders below `uploads/` completely, because if they do not exist, they are created on the fly via Flysystem.

### DIFF
--- a/app/Actions/Import/Exec.php
+++ b/app/Actions/Import/Exec.php
@@ -128,10 +128,41 @@ class Exec
 			}
 
 			// Skip folders of Lychee
+			// Currently we must check for each directory which might be used
+			// by Lychee below `uploads/` individually, because the folder
+			// `uploads/import` is a potential source for imports and also
+			// placed below `uploads`.
+			// This is a design error and needs to be changed, at last when
+			// the media is stored remotely on a network storage such as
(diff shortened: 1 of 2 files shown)


d7415 (approved):
Looks sensible. Also tested!

Your comment looks a bit like https://lycheeorg.github.io/docs/advanced-setups.html#separating-data-from-the-code

nagmat84:
> Your comment looks a bit like https://lycheeorg.github.io/docs/advanced-setups.html#separating-data-from-the-code

With one tiny, but _extremely crucial_ difference. My comment says that the `import` folder must be pulled out from the directory where it is currently located, while the FAQ keeps it there. But the `import` folder is of a completely different nature than the other folders. The other folders contain files which are already part of the Lychee database, while `import` does not.

Issue #1462, opened 2022-08-16
[Change Proposal] Access Rights management: Users and Albums


ildyria:
Let's move out of the PR and discuss this here. As follows is the log of previous discussions.

### TL;DR: Proposition by @nagmat84 

- move protection settings `grants_full_photo`, `requires_link`, `is_downloadable`, `is_share_button_visible` from `base_albums` to `user_base_album`
- guest user becomes id = `null`
- GUI change:
   single dialog instead of two: A imagine a tabular like dialog with one row per user with whom the album is shared (including the anonymous user) and then a tuple of check boxes in columns for the various attributes. This means it would somewhat look similar to the layout we have for the user setting page.

@nagmat84 
> Let me sketch what I am thinking of:
> 
> I would like to have a unified approach for what is currently called protection settings (i.e. `is_public`, `is_downloadable`, etc.) and sharing photos and albums with authenticated users. It bothers me, that we can set attributes like `is_downloadable` on a per-album level if the album is public, but that we use a globally defined setting for that when the album is shared.
> 
> My idea is to have these settings on a per-share level, too. On a technical level, I would like to move the attributes `grants_full_photo`, `requires_link`, `is_downloadable`, `is_share_button_visible` from `base_albums` to `user_base_album`. In order to model a public album, I would insert a share with the `null` user into `user_base_album`. This makes a lot of sense, if you consider that `Auth::id` also returns `null` for the unauthenticated user. Hence, the boolean attribute `is_public` would be removed completely from `base_album` without replacement. An album is public if and only if it is shared with the `null` user. I assume this would also simplify the SQL queries, because the conditions become more consistent.
> 
> On the GUI level, we would only have a single dialog instead of two: A imagine a tabular like dialog with one row per user with whom the album is shared (including the anonymous user) and then a tuple of check boxes in columns for the various attributes. This means it would somewhat look similar to the layout we have for the user setting page.
> 
> In a second step I imagine the same approach, but for user groups. You can also share albums with groups and have the same options. All users are automatically members of the `null` group which by definition represents the group of authenticated users.
> 
> However, this endeavor is big. I would probably start with the users in one PR and then add groups in another PR. However, if we now make `is_public` tri-state, then we cannot do this in two separate PRs anymore, because this would mean temporarily losing a feature which we must not do.
> 
[...]

PR #1480, opened 2022-08-23
Shared albums should not consider the `require_link` property.

### LINKED ISSUE #1478: Shared albums which require a direct link for anonymous users are not visible to users with whom they are shared

I realized during testing that not all shared albums I shared with a certain user would show up in the gallery once logging in with the other user. This was strange as it was only specific albums that would not show, always the same ones across users. I copied the URL to check if the album would be accessible that way, and sure it was: just would not show up in the gallery alongside the other albums.

At this point I realized the common thing between the albums were that they were all set to "Public". Somehow this seems to interfere with Albums::get, which does not list these albums in the API result.

1. Create an album and share it with a user, confirm that it shows up in their gallery
2. Create another album, but set it to "Public" visibility (first three toggles on) before sharing it with an user
3. Confirm the album does not show up in that user's gallery, but is correctly accessible by its URL

### Browser and system

Lychee latest release built from Docker, running on LXC on Debian, up-to-date Firefox

ildyria:
Fixes #1478 

```php
/**
 *  - the user is the admin, or
 *  - the user is the owner, or
 *  - the album does not require a direct link and is shared with the user, or
 *  - the album does not require a direct link, is public and has no password set, or
 *  - the album does not require a direct link, is public and has been unlocked
 */
```

After further thoughts, this is pretty counter intuitive. We should have fixed that one a long time ago. :)


ildyria:
I am not sure if I missed some other conditions.

app/Policies/AlbumQueryPolicy.php
```diff
@@ -43,7 +43,7 @@ public function __construct(AlbumFactory $albumFactory, AlbumPolicy $albumPolicy
 	 *
 	 *  - the user is an admin
 	 *  - the user is the owner of the album
-	 *  - the album is shared with the user and the album does not require a direct link
+	 *  - the album is shared with the user
```
nagmat84, line 46: You have changed the comment of the method, but not the method itself, see lines 81-83 below. How so?

Maybe we should also have some tests which checks the new behavior.

nagmat84 (changes requested):
You missed the conditions in line 82 and 360.

Moreover, I am in strong favor of tests which checks the new behavior. At the moment, we seem to have reached a point which is rather bug-free with respect to security issues (i.e. accidentally leaking a photo which should stay hidden). Any change in the SQL queries might break that and hence, tests would be wonderful.

Or we could just wait until https://github.com/LycheeOrg/Lychee/issues/1462 is ready. As I detailed out in https://github.com/LycheeOrg/Lychee/issues/1478 there might actually be other users who rely on exactly this feature, i.e. that albums can also be hidden from users with whom they are shared. (Although, I lack the imagination of a meaningful use-case for that.)

I am still wondering if we should make "hidden" a per-share option. @kamil4 raised the same question in https://github.com/LycheeOrg/Lychee/pull/1357#issuecomment-1215763003, 3rd bullet point. This being said, I am not sure, if we should change the behavior now or simply wait until https://github.com/LycheeOrg/Lychee/issues/1462. At least, the implementation works as designed and intended.

app/Policies/AlbumQueryPolicy.php
ildyria: yeah I was thinking about adding some tests in that regard. At least to reproduce the bug and then kill it :)

ildyria:
As per requested finished the PR changes and added tests. :)

ildyria:
IDK what happened O.o but this is definitively not something I wanted to close. :'D

nagmat84 (approved):
LGTM, but untested. I only checked all the boolean conditions in the query (which was difficult enough :laughing:) and could not spot any remaining error.

@ildyria I assume you have tested it yourself and found no regression?

ildyria:
> @ildyria I assume you have tested it yourself and found no regression?

Yes.

ildyria:
Also the tests you added are making it pretty tight to create regressions :)

PR #1489, opened 2022-08-27
Drop page support

ildyria:
I took the difficult decision to drop the page support.
This functionality is not visible to the normal admin and one needs to be very familiar with the code to use it.
Maybe such can be added later but for now it is not a necessary feature of Lychee.


qwerty287 (approved):
Looks good :+1:

resources/views/landing.blade.php
```diff
@@ -18,7 +18,15 @@
     </div><!-- header -->
 
 
-    @include('includes.menu')
```
nagmat84, line 21: If the include is removed, the file `resources/view/includes/menu.blade.php` can be deleted as well.

nagmat84 (changes requested):

resources/views/landing.blade.php
ildyria: fixed with [0fe09e8](https://github.com/LycheeOrg/Lychee/pull/1489/commits/0fe09e8dd5763b1aa49aeb017d00f1ba09e8d8ad)

nagmat84 (approved):

qwerty287 (approved):
This is ready, or why don't we merge it?

ildyria:
First merge the install fix, release v4.6.0 and then merge all the waiting 4.6.1 pr. 🙂

⁣Télécharger BlueMail pour Android ​

Le 4 sept. 2022, 14:36, à 14:36, qwerty287 ***@***.***> a écrit:
***@***.*** approved this pull request.

qwerty287:
Yeah, I think I understood (unfortunately *after* I merged https://github.com/LycheeOrg/Lychee/pull/1490). 😬

PR #1491, opened 2022-08-27
Cleaning config table

ildyria:
The values `username` and `password` from the config table are no longer used.


qwerty287 (approved):

qwerty287 (changes requested):
Also wrong frontend branch...

qwerty287 (approved):

ildyria:
What the F is going on with the tests.

nagmat84:
Is this still WIP as the title suggests? Then this PR should be converted into a draft and reviews being invited after it has been completed. Or isn't this WIP and the title is just misleading?

ildyria:
> Is this still WIP as the title suggests? Then this PR should be converted into a draft and reviews being invited after it has been completed. Or isn't this WIP and the title is just misleading?

My mistake, no longer WIP because the last tests finally passes. I wanted to make sure it was not going to be merged until the tests pass (even though it was approved). :)

nagmat84:
> > Is this still WIP as the title suggests? Then this PR should be converted into a draft and reviews being invited after it has been completed. Or isn't this WIP and the title is just misleading?
> 
> My mistake, no longer WIP because the last tests finally passes. I wanted to make sure it was not going to be merged until the tests pass (even though it was approved). :)

You now that there is the option to mark PRs as draft for exactly this use case?

ildyria:
I know and I am using it in the `upgrade-laravel-9` one.
However I don't see any option to convert a PR back to a draft.

EDIT: Found it, just under reviewers. Not really visible. :(

app/Models/Configs.php
```diff
@@ -298,4 +298,12 @@ public function scopeAdmin(FixedQueryBuilder $query): FixedQueryBuilder
 	{
 		return $query->where('confidentiality', '<=', 3);
 	}
+
+	/**
+	 * Reset the cache.
+	 */
+	public static function dropCache(): void
```
nagmat84:
```suggestion
	public static function clearCache(): void
```

or

```suggestion
	public static function invalidateCache(): void
```

are probably the more frequently used terms.

However, I don't understand _why_ we suddenly need this method? How could the cache get out-of-sync with the DB?

database/migrations/2020_12_12_203153_migrate_admin_user.php
```diff
@@ -18,15 +18,13 @@ class MigrateAdminUser extends Migration
 	 */
 	public function up(): void
 	{
-		$user = new User();
+		$user = User::query()->findOrNew(0);
+		$user->incrementing = false; // disable auto-generation of ID
+		$user->id = 0;
+		Configs::dropCache();
```
nagmat84: Why do we need to clean the cache here? How should the cash become invalid?

In particular, this migration is form the year 2020. Why do we suddenly need this change?

database/migrations/2022_08_27_110209_drop_admin_user_config.php
```diff
@@ -0,0 +1,46 @@
...
+{
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		DB::table('configs')->where('key', '=', 'username')
+		->orWhere('key', '=', 'password')->delete();
```
nagmat84:
```suggestion
		DB::table('configs')
			->whereIn('key', ['username', 'password'])
			->delete();
```

This is nitpicking, but `whereIn` is more efficient than `orWhere`, especially with the silly MySQL query planner. `whereIn` allows to use an index, but if MySQL finds an `OR` in a query it stops trying to optimize the query or using an index, because MySQL cannot deal with `OR` in queries. :disappointed: 

But surely, this is irrelevant for a query which is only run at most once. :smile: 

Moreover, I believe the code is more succinct and cleaner.

tests/Feature/InstallTest.php
```diff
@@ -49,7 +50,7 @@ public function testInstall(): void
 		 * Clearing things up. We could do an Artisan migrate but this is more efficient.
 		 */
 
-		// The order is important: referring tables must be deleted first, referred tables last
+		Schema::disableForeignKeyConstraints();
```
nagmat84: Why this? How is this relevant for the PR?

tests/Feature/InstallTest.php
```diff
@@ -140,6 +142,7 @@ public function testInstall(): void
 		/**
 		 * We now should NOT be redirected.
 		 */
+		Configs::dropCache();
```
nagmat84: Why this change suddenly? How is this related to the PR?

nagmat84 (changes requested):

tests/Feature/InstallTest.php
ildyria: `Configs` is maintaining a static cache inside to avoid querying the table at each access and making it somewhat of a singleton.

During the execution of phpunit, the Configs singleton is never "cleaned", this is similar to sessions that are kept alive through multiple tests.

As a result because the Configs does not have username / password values from the start of the execution and not refreshed, the install/migration run is broken in the tests.

database/migrations/2020_12_12_203153_migrate_admin_user.php
ildyria: See https://github.com/LycheeOrg/Lychee/pull/1491#discussion_r956577674 .

tests/Feature/InstallTest.php
ildyria: I was trying an optimization to avoid having to disable one of the foreign key below.

tests/Feature/InstallTest.php
```diff
@@ -49,7 +50,6 @@ public function testInstall(): void
 		 * Clearing things up. We could do an Artisan migrate but this is more efficient.
 		 */
 
-		// The order is important: referring tables must be deleted first, referred tables last
```
nagmat84: Please restore this comment, too.

nagmat84 (changes requested):

tests/Feature/InstallTest.php
ildyria: fixed.

nagmat84 (approved):

PR #1495, opened 2022-08-29
fix Installation

ildyria:
- The Exception handling of Laravel changed with regard to the Encryption key... We now also look at the content of the message.
- When `.` write access was not present, `touch` was failing, preventing the redirection to install.
- Added a `QueryException` catch on the Installation check "config exists in DB" to ensure that this does not fail even if we don't have access to any DB.
- Require `database/database.sqlite` to be writable (makes it easier to used out of the box too).

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -222,7 +222,7 @@ class Handler extends ExceptionHandler
 	 */
 	protected function prepareResponse($request, \Throwable $e): SymfonyResponse
 	{
-		if (!$this->isHttpException($e) && config('app.debug') !== null) {
+		if (!$this->isHttpException($e) && config('app.debug') === true) {
 			return $this->toIlluminateResponse($this->convertExceptionToResponse($e), $e);
 		}
 
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -206,7 +206,7 @@ class Handler extends ExceptionHandler
 	 * decided that the client expects a HTML response, but _before_ the
 	 * actual work horse {@link Handler::renderHttpException} is called.
 	 *
-	 * This method is 99% identical to the parent method expects for a tiny
+	 * This method is 99% identical to the parent method except for a tiny
 	 * bug fix which adds the original exception to the encapsulating
 	 * `HttpException`.
 	 *
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -10,13 +10,60 @@ use App\Exceptions\Handlers\MigrationHandler;
 use App\Exceptions\Handlers\NoEncryptionKey;
 use App\Models\Logs;
 use Illuminate\Auth\AuthenticationException;
+use Illuminate\Contracts\Container\BindingResolutionException;
 use Illuminate\Contracts\Container\Container;
 use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
+use Illuminate\Http\Request;
 use Illuminate\Session\TokenMismatchException;
 use Illuminate\Support\Arr;
+use Psr\Container\ContainerExceptionInterface;
+use Psr\Container\NotFoundExceptionInterface;
 use Symfony\Component\HttpFoundation\Response as SymfonyResponse;
+use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\Exception\HttpExceptionInterface;
 
+/**
+ * Lychee's custom exception handler.
+ *
+ * While the overall architectural approach of the original exception handler
+ * of the framework is fine, the original exception handler is mostly broken
+ * when it comes to the details.
+ *
+ * The overall architectural approach is as follows:
+ *
+ *  1. Substitute or wrap certain exceptions by or into other exceptions
+ *     (i.e. via `mapException`)
+ *  2. Decide whether the client expects an HTML or JSON response
+ *  3. Convert (or "render") the exception into a response with said content
+ *     type
+ *
+ * However, there are two major issues with the original exception handler:
+ *
+ *  - Substitution of exception is not limited to `mapException` but happens
+ *    all the time which makes it hard to reliably predict what happens to
+ *    an exception when a method (other than `mapException`) is called.
+ *    One might end up with a different exception.
+ *    Moreover, not all of these substitution are sensible enough to add the
+ *    original exception as a predecessor to the new exception.
+ *  - A constant mix-up of the terms "HTTP" and "HTML".
+ *    The framework frequently uses the term "HTTP" as an antonym to "JSON"
+ *    when "HTML" would be rather appropriate.
+ *    For example like in `renderJsonResponse($e)` vs. `renderHttpResponse($e)`.
+ *    The latter is called, when an exception shall be converted into HTML.
+ *    But of course, a JSON response is also an HTTP response.
+ *    It seems as if the framework is not even aware of this confusion.
+ *
+ * 90% of this handler are bug fixes.
+ * This means, parent methods are not overwritten, because we need a special
+ * non-standard behaviour, but simply the _right_ behaviour.
+ * Unfortunately, this class cannot solve the unfortunate naming of some
+ * methods, but must stick to the names used by the parent class.
+ * Alternatively, this class could overwrite the entry method `render($e)`,
+ * re-implement everything which comes after that (even using better names)
+ * and let the rest of the parent class go down the drain.
+ * However, this bears the risk that some 3rd-party calls unfixed methods of
+ * the original exception handler.
+ */
 class Handler extends ExceptionHandler
 {
 	/**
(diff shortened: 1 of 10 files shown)


resources/views/install/permissions.blade.php
```diff
@@ -17,8 +17,21 @@
     @foreach ($permissions as $permission)
         <li class="list__item list__item--permissions">
         <span>{{ $permission['folder'] }}</span>
+        @if (count($permission['permission']) < 4)
+            @for ($i = 0;$i < 4 - count($permission['permission']); $i++)
+                <span class="perm float-right">
+                    &nbsp;
+                </span>
+            @endfor
+        @endif
```
nagmat84, line 26: What does this extra code do? What is its purpose? Why do we compare with 4?

app/Exceptions/Handlers/NoEncryptionKey.php
```diff
@@ -26,6 +26,9 @@ public function check(HttpException $e): bool
 			if ($e instanceof MissingAppKeyException) {
 				return true;
 			}
+			if ($e->getMessage() === 'No application encryption key has been specified.') {
```
nagmat84: I know that you write that Laravel 9 changed the Exception handling with regard to the encryption key. Could you be more specific about what has changed precisely?

I don't like to check for exception messages, because they are much more likely subject to change than the exception type.

app/Exceptions/Handlers/NoEncryptionKey.php
```diff
@@ -37,7 +40,11 @@ public function check(HttpException $e): bool
 	public function renderHttpException(SymfonyResponse $defaultResponse, HttpException $e): SymfonyResponse
 	{
 		try {
-			touch(base_path('.NO_SECURE_KEY'));
+			try {
+				touch(base_path('.NO_SECURE_KEY'));
+			} catch (Throwable) {
+				// do nothing
+			}
```
nagmat84: Please replace the obvious comment `// do nothing` by something more informative. In particular, I am wondering _why_ it is OK to not creating this file and what consequences it will have. We check for the non-existence/existence of that file somewhere else, if I remember correctly. Hence, failing to create this file should definitely have an effect and we should be aware of them.

app/Http/Middleware/Checks/IsInstalled.php
```diff
@@ -19,6 +21,8 @@ public function assert(): bool
 			return
 				!file_exists(base_path('.NO_SECURE_KEY')) &&
 				Schema::hasTable('configs');
+		} catch (QueryException $e) {
+			return !Str::contains($e->getMessage(), 'SQLSTATE[HY000] [1045]');
```
nagmat84: I can somehow understand that the method `assert` should not throw an exception, but always return either `true` or `false`. So I would have expected something like

```php
try {
    // do whatever needs to be done and return true or false
} catch (BindingResolutionException $e)  {
    // re-throw as this should never happen anyway, because this indicates a programming error
} catch {
    // return false for any other not rethrown message because we could obviously _not_ assert what we wanted to assert
    return false;
}
```

In particular, I do not understand why why the return value depends on whether the message contains `''SQLSTATE[HY000] [1045]"` or not. After googling, I found out that this message is thrown by MySQL if DB authentication failed.

First of all, this should be part of a comment above this line.

Secondly, there should be an explanation _why_ we return `true` if there has been any other SQL exception than an authentication problem.

nagmat84 (commented):
I am not very happy with this PR. Basically, because all changes lack a plausible rationale (inside the code) _why_ the changes are sane and _why_ the changes have no negative impact.

In particular, for me this PR actually feels like a hack of the sort "I don't know what is failing so let's suppress any errors for the lack of a better idea".

resources/views/install/permissions.blade.php
ildyria: It just makes sure that the database/database.sqlite line is pretty and align with the other `writable/exists/readable/executable` conditions.
As you like coherence, it does not make sense for a database file to be executable, thus this remove the check.
There are exactly 4 checks, this fills the rest with place holders.

app/Exceptions/Handlers/NoEncryptionKey.php
ildyria: Laravel 8.

Previously, Laravel was directly throwing the MissingAppKeyException, now if it is not an HttpException, the handler converts it into one, keeps the message AND DROP the previous exception.

Why did this behavior change on Laravel side? I don't know and I don't really care TBH.

app/Http/Middleware/Checks/IsInstalled.php
ildyria: I don't see what kind of other SQL exception we could have aside from this one.
I do not want to make the catch too broad in order to avoid useless redirection to Install.

kamil4 (changes requested):
So I tested this by applying the PR on top of the 4.6.0-RC3 release ZIP file. And I tested it how the ordinary new user would: by not following the manual but simply "trying things out".

There's one regression. Without this PR, I get a nice message:
```
503
../storage and sub-directories are not writable.
Please set the proper access rights.
```

With this PR, I get an empty page and a 500 error in the JS console.

`public/img/` is still required to be writable. This should be fixed.

Directory permission checks (part of `2022_07_13_174800_permission_test.php` I guess?) generate error messages:
```
Scanning: public/uploads
fileperms(): stat failed for public/uploads
Scanning: public/sym
fileperms(): stat failed for public/sym
```
I had a brief look into it and this is because the CWD is `Lychee/public`, whereas the command only works with `Lychee` as the CWD. I'm guessing the installer sets the CWD?

app/Exceptions/Handlers/NoEncryptionKey.php
nagmat84: > Previously, Laravel was directly throwing the `MissingAppKeyException`, now if it is not an HttpException, the handler converts it into one

Wait! Are you telling me that with the transition from Laravel 8 to 9 "something" inside the framework converts exceptions into `HttpException`? If yes, then this problem needs a much more thorough investigation, because there is much more code that checks exceptions types and relies on it. This would be a fundamental change and must be treated as such. This would also imply that the error reporting towards the JSON client might not work anymore the way as it has been used to. This cannot be brushed aside by a simple patch.

> Why did this behavior change on Laravel side? I don't know and I don't really care TBH.

That is exactly what I suspected and how that PR felt like.

ildyria: This IS STILL LARAVEL 8 !!! **Nothing related to Laravel 9.**

ildyria:
> There's one regression. Without this PR, I get a nice message:

I just tried and I still have it. O.o

If you do `make dist` and try with the Lychee.zip created, it should be working fine.

database/migrations/2022_07_13_174800_permission_test.php
```diff
@@ -12,7 +13,14 @@ class PermissionTest extends Migration
...
 	{
+		$is_public = Str::endsWith(getcwd(), 'public');
+		if ($is_public) {
+			chdir('..');
+		}
 		Artisan::call('lychee:fix-permissions', ['--dry-run' => 1]);
+		if ($is_public) {
+			chdir('public');
+		}
 	}
```
kamil4: This might be effective but it looks just nasty! Wouldn't it be better to fix `app/Console/Commands/FixPermissions.php` to use absolute pathnames instead (via `Storage::disk('images')` and `Storage::disk('symbolic')`, I guess)?

kamil4 (changes requested):
I won't pretend that I understand the subtleties of the changes to `NoEncryptionKey.php` and `IsInstalled.php`.  I have verified that they fix bugs we've observed...

I'm still getting error 500 when `storage` is not writable but, honestly, I can't explain it.  I created a zip file via `make dist` as you suggested but that didn't help.  I then tried to comment out all the changes brough via this PR and even *that* didn't help, which doesn't make any sense.  At this point I'm ready to give up on this particular mystery...

app/Exceptions/Handlers/NoEncryptionKey.php
nagmat84: I fixed it myself in https://github.com/LycheeOrg/Lychee/pull/1495/commits/27ad036e6a4496fec6849b2179809ab12bea4186. The actual fix is a one-liner. I also created a PR for the Laravel framework: https://github.com/laravel/framework/pull/43991.

The remaining changes are only a lot of additional comments. At some day I will completely replace this ugly class.

From my side, this conversation can be resolved.

database/migrations/2022_07_13_174800_permission_test.php
ildyria: Fixed with [adf57ba](https://github.com/LycheeOrg/Lychee/pull/1495/commits/adf57badf1e7e205591c09db2c5cae2e3a61ede7)

app/Exceptions/Handlers/NoEncryptionKey.php
nagmat84: I pushed a commit https://github.com/LycheeOrg/Lychee/pull/1495/commits/e055fe2839368d51cdf5de192ec0fd3628b009f7 which removes the need for this file `NO_SECURE_KEY` completely. I have never liked the approach which uses an extra file, because it may get out-of-sync with the actual installation state. Now even more, if we cannot even guarantee that the file is created reliably.

Instead https://github.com/LycheeOrg/Lychee/pull/1495/commits/e055fe2839368d51cdf5de192ec0fd3628b009f7 directly checks for the existence of the application key.

From my side, this conversation can be resolved.

app/Exceptions/Handler.php
```diff
@@ -153,12 +200,61 @@ protected function mapException(\Throwable $e): \Throwable
...
 
 	/**
-	 * Renders the given HttpException.
+	 * Prepare a response for the given exception.
+	 *
+	 * This method is called by the framework, _after_ the framework has
+	 * decided that the client expects a HTML response, but _before_ the
+	 * actual work horse {@link Handler::renderHttpException} is called.
+	 *
+	 * This method is 99% identical to the parent method expects for a tiny
```
nagmat84:
```suggestion
	 * This method is 99% identical to the parent method except for a tiny
```

Fixing my own typos :grinning:

nagmat84 (approved):
I am fine with it. (I should be after I changed things myself unless I am schizophrenic. :grinning:) I only found a type in my own comments.

ildyria:
I guess once @kamil4 is fine with it, it can be merged :)

kamil4 (approved):
I tested it and it appears to be working fine.

Issue #1481, opened 2022-08-23
[Enhancement] Giving security attributes better names


nagmat84:
**Intro**

In https://github.com/LycheeOrg/Lychee/pull/1443#issuecomment-1224418047 @ildyria asked why there is this seemingly duplication of sending certain security attributes to the front-end. The short answer is: because they only happen to have the same name (unfortunately), but have different semantical meanings.

This reminds me of something which has been bothering me for quite a while. The issue has some overlapping with https://github.com/LycheeOrg/Lychee/issues/1462 and https://github.com/LycheeOrg/Lychee/pull/1403, but it is worth to be considered on its own. I have been planning to give better names to some security attributes for a long time, but as I have been unable to come up with some better names on my own, I will put it to discussion.

**Problem**

The problem are code lines like 
```
lychee.may_upload = data.admin || data.may_upload
```
This is from the frontend, but it illustrates the problem. Here a user has the _right_ to upload into an album, if the user has the admin _capability_ or upload _capability_. The problem is that the term "upload" is used twice in two different meanings: on the left hand-side of the assignment it denotes the currently effective right in a specific use-case context, on the right hand-side it denotes a capability of a user. This bears some risk. In particular, it may lead to defective code if the wrong kind of some equivocally named attribute is used in the wrong place.

I also had a quick look at our new Laravel policies below `app/Policies` and my heart sunk, when I found some mix-up. At the moment, this is still a peace of cake, because we only have very limited multi-user support and hence the impact is manageable, if something gets wrong, but it will become a nightmare, if we get more and more fine-grained multi-user support.

**Definitions**

Let me first get some terms and definitions straight, before I get back to Lychee. (Sorry, I am a scholar.)

 - A _permission_ is something which is based on a relationship between subjects (i.e. user in our case) and objects (e.g. albums, photos, etc.). Permissions may be illustrated by an access control matrix which shows which specific subject has which permission on which specific object.
 - A _security attribute_ is directly attached to a subject or an object, i.e. it functionally depends on the subject or the object only, but not on the relation between them. In case of subjects _security attributes_ are also called _capabilities_ sometimes.
 - A _right_ is the effective outcome of a decision making _policy_ based on permissions and security attributes in a specific context.

_In summary:_ The _security attributes_ of object O_i, the _capabilities_ (=security attributes) of subject S_j and the _permissions_ of S_j for O_i are input to a _policy_ which determines whether S_j has the _right_ to perform a certain operation on O_i.

[...]

PR #1508, opened 2022-09-05
Ensure that Admin rights are overloading others

ildyria:
Currently we are fetching the `may_upload` and `is_locked` from user.
This results in inconsistent rights if the admin user row does not have those properties set properly.
The consequence of inconsistent rights are the removal of shortcuts from the front end (Rename, Upload, New etc.)

This PR provides a more flexible approach and uses the Gate to determine the rights.

This corrects a minor bug introduced with https://github.com/LycheeOrg/Lychee/commit/1c8eedb22f59b066d76395421f5b37a168db27e8


qwerty287 (approved):

kamil4 (approved):
The PR looks reasonable enough but, independently of it, wouldn't it be prudent to also fix the admin entry in the `users` table?

qwerty287 (approved):

nagmat84:
Please give me the chance to a look at this PR this evening first. I haven't had the time yet. I would like to check how it relates to https://github.com/LycheeOrg/Lychee/issues/1481.

I haven't read the diff yet, but from the description of the PR it sounds as if the true problem is actually a confusion of security attributes (aka user capabilities) and rights again and as if this PR was making this actually worse by covering the symptoms of problem instead of solving it from the ground.

ildyria:
@nagmat84 I didn't merge yet for this reason :)

Yes it is related to `lychee.may_upload = data.admin || data.may_upload` as you mentionned in #1481 
This PR can be seen as temporary fix to make the master branch working again in case of wrong attributes on admin.

I could also have used this PR to start working in the direction of #1481 but I prefer to create another one for that and get this one as a fix.

ildyria:
@nagmat84 started some refactoring work here: https://github.com/LycheeOrg/Lychee/tree/apply-rights-naming-convention

app/Http/Controllers/SessionController.php
```diff
@@ -73,9 +73,9 @@ public function init(): array
 				$user = Auth::user();
 				$return['user'] = $user?->toArray();
 				$return['rights'] = [
-					'is_admin' => Gate::check(UserPolicy::IS_ADMIN),
-					'is_locked' => $user?->is_locked ?? true,
-					'may_upload' => $user?->may_upload ?? false,
+					'is_admin' => Gate::check(UserPolicy::IS_ADMIN, User::class),
+					'is_locked' => !Gate::check(UserPolicy::CAN_EDIT_SETTINGS, User::class), // the use of the negation should be removed later
+					'may_upload' => Gate::check(UserPolicy::CAN_UPLOAD, User::class),
```
nagmat84, line 78: Actually, this doesn't make much sense.

`CAN_UPLOAD` is not a right on a user object (because one does not upload users), but rather a right on photos as well as albums.

Or in terms of https://github.com/LycheeOrg/Lychee/issues/1481#issuecomment-1229542549:

`may_upload` is a capability of the user and should stay there, but not be defined as the outcome of a policy. This capability contributes to rights like "may create album" or "may add photos". Only the latter are determined by a policy.

However, I also concede that the gate checks `UserPolicy::CAN_UPLOAD`, etc., have already been defined previously are are not introduced in by this PR. In other words, the actual fault has already happened earlier.

Hence, for the time being, I am fine with using this work-around here as a patch.

ildyria: https://github.com/LycheeOrg/Lychee/pull/1511 :) This aims to fix that.

PR #1510, opened 2022-09-06
Fix 1480 (Composer update) + do not upload artifact on branches

ildyria:
- fixes #1480
- update composer => some white spaces changes...
- Do not upload artifact on branches that are not master : it slows down by 7~8 minutes each workflow run...


qwerty287 (approved):

d7415:
Is it worth adding a `workflow_dispatch` action for this in case we want to create a zip for an arbitrary branch? (e.g. for non-shell users testing a fix)

PR #1511, opened 2022-09-07
Apply naming convention for Rights, Capabilites etc.

ildyria:
This PR builds on #1508 so merging this one will automatically merge the previous one.

#### Content: 

- **Drops the Admin Middleware**

- move authorization to **Gates on Requests**

- Increase number of Policy & granularity of Rights

- creates DTO for:
  -  user rights
  - settings rights
  - photo rights
  - album rights
  - album protection policy
  - user (for admin editing)

- renames capabilities with regard to #1481

- drops visibility for sharing links per album in favor of the already existing global setting.

- adds stricter tests on Locale and remove extra keys in other languages than English

##### TODO (maybe future PR?) :

- Photo protection policy ?
- make use of rights in front end (currently mostly depending of admin/user can upload/nothing).


app/Actions/Settings/SetLogin.php
```diff
@@ -27,8 +27,9 @@ public function do(string $username, string $password): User
 		$adminUser = User::query()->findOrFail(0);
 		$adminUser->username = $username;
 		$adminUser->password = Hash::make($password);
-		$adminUser->is_locked = false;
+		$adminUser->may_edit_own_settings = true;
```
nagmat84, line 30: Very verbose, but I like it. I only did not expected that from you. :laughing: I assume you would go for `may_selfedit` as in https://github.com/LycheeOrg/Lychee/issues/1481#issuecomment-1229542549.

app/DTO/UserCapabilities.php
```diff
@@ -0,0 +1,53 @@
...
+	 * Create from current user.
+	 *
+	 * @return UserCapabilities
+	 */
+	public static function ofCurrentUser(): UserCapabilities
+	{
+		return new UserCapabilities(
+			may_administrate: Gate::check(UserPolicy::IS_ADMIN, User::class),
+			may_upload: Gate::check(UserPolicy::MAY_UPLOAD, User::class),
+			may_edit_own_settings: Gate::check(UserPolicy::MAY_EDIT_OWN_SETTINGS, User::class)
```
nagmat84: This doesn't make much sense. The capabilities of a user are simply defined on that user object. Capabilities are not the result of a policy. The result of a policy are rights and capabilities are input to policies.

app/Http/Controllers/Administration/SharingController.php
```diff
@@ -38,7 +38,7 @@ public function list(ListShare $listShare): Shares
 		// TODO: move this to Request authorization.
 		// Note: This test is part of the request validation for the other
 		// methods of this class.
-		if (!Gate::check(UserPolicy::CAN_UPLOAD, User::class)) {
+		if (!Gate::check(UserPolicy::MAY_UPLOAD, User::class)) {
```
nagmat84: Same problem here. What you really wanted is a `AlbumPolicy::CAN_SHARE` and that one uses the `may_upload` capability of the user object.

app/Http/Controllers/SessionController.php
```diff
@@ -63,24 +62,16 @@ public function init(): array
 				// If the session is unauthenticated ('user' === null), but grants admin rights nonetheless,
 				// the front-end shows the dialog to create an admin account.
 				$return['user'] = null;
-				$return['rights'] = [
-					'is_admin' => true,
-					'is_locked' => false,
-					'may_upload' => true,
-				];
+				$return['rights'] = new UserCapabilities(true, true, true);
```
nagmat84: That is actually a literal contradiction. On the left hand side we have "rights", on the right hand side we have "capabilities".

The real question which must be solved first is what do we want to send to the frontend?

In the long run I am in favor of rights. Simply because this prevents us from repeating the logic of determining the rights in the frontend which we had to, if we were only sending over the capabilities. But rights are things like `CAN_SHARE_ALBUM` (not `MAY_UPLOAD`), `CAN_ADD_ALBUM` (not `MAY_UPLOAD`), etc.

That is why I already called the property `rights`. Moreover, it is not necessary to send the user capabilities a second time, because the user capabilities are already part of the `user` attribute which is also sent to the frontend.

app/Http/Controllers/SessionController.php
```diff
@@ -63,24 +62,16 @@ public function init(): array
...
 			} else {
 				/** @var User|null $user */
 				$user = Auth::user();
 				$return['user'] = $user?->toArray();
-				$return['rights'] = [
-					'is_admin' => Gate::check(UserPolicy::IS_ADMIN, User::class),
-					'is_locked' => !Gate::check(UserPolicy::CAN_EDIT_SETTINGS, User::class), // the use of the negation should be removed later
-					'may_upload' => Gate::check(UserPolicy::CAN_UPLOAD, User::class),
-				];
+				$return['rights'] = UserCapabilities::ofCurrentUser();
```
nagmat84: dito

app/Http/Requests/Settings/ChangeLoginRequest.php
```diff
@@ -26,7 +26,7 @@ class ChangeLoginRequest extends BaseApiRequest implements HasPassword
 	 */
 	public function authorize(): bool
 	{
-		return Gate::check(UserPolicy::CAN_EDIT_SETTINGS, User::class);
+		return Gate::check(UserPolicy::MAY_EDIT_OWN_SETTINGS, User::class);
```
nagmat84: Here, we would actually only need the _right_ `CAN_EDIT` and then pass the current user object to the policy like this

```
Gate::check(UserPolicy::CAN_EDIT, Auth::user())
```

The corresponding policy function would look like this

```php
class UserPolicy {
  function canEdit(User $user, User $userToBeEdited) {
    return $user->may_administrate || ($user->id === $userToBeEdited->id && $user->may_edit_own_settings);
  }
}
```

At first glance, this might look strange, but makes perfect sense if you think about it a second time. A policy returns a right. A right is always something between the currently authenticated user (`$user`) and an object which in this case "accidentally" happens to be another `User` object (`$userToBeEdited`). A policy must base its decision on user capabilities. Either the user has the `may_administrate` capabilities which allows the user to edit any user or the user has `may_edit_own_settings` capability and tries to edit its own user.

app/Http/Requests/Sharing/DeleteSharingRequest.php
```diff
@@ -22,7 +22,7 @@ class DeleteSharingRequest extends BaseApiRequest
 	 */
 	public function authorize(): bool
 	{
-		return Gate::check(UserPolicy::CAN_UPLOAD, User::class);
+		return Gate::check(UserPolicy::MAY_UPLOAD, User::class);
```
nagmat84: Again, the result of a policy has to be a right, in this case I would use the same as for adding shares, i.e. `Album::CAN_SHARE`.

app/Http/Requests/Sharing/SetSharingRequest.php
```diff
@@ -23,7 +23,7 @@ class SetSharingRequest extends BaseApiRequest implements HasAlbumIDs, HasUserID
 	 */
 	public function authorize(): bool
 	{
-		return Gate::check(UserPolicy::CAN_UPLOAD, User::class);
+		return Gate::check(UserPolicy::MAY_UPLOAD, User::class);
```
nagmat84: Again: `AlbumPolicy::CAN_SHARE`.

app/Http/Requests/User/SetEmailRequest.php
```diff
@@ -15,7 +15,7 @@ class SetEmailRequest extends BaseApiRequest
 
 	public function authorize(): bool
 	{
-		return Gate::check(UserPolicy::CAN_EDIT_SETTINGS, User::class);
+		return Gate::check(UserPolicy::MAY_EDIT_OWN_SETTINGS, User::class);
```
nagmat84: See some comments above. We need a right `UserPolicy::CAN_EDIT` here. At the moment, I would go with a single user-edit right. We can still add more fine granular rights like `UserPolicy::CAN_EDIT_EMAIL` later.

app/Locale/English.php
```diff
@@ -506,7 +506,7 @@ public function get_locale(): array
 			'CAMERA_DATE' => 'Camera date',
 			'NEW_PASSWORD' => 'new password',
 			'ALLOW_UPLOADS' => 'Allow uploads',
-			'RESTRICTED_ACCOUNT' => 'Restricted account',
+			'ALLOW_USER_SELF_EDIT' => 'Allow user modifications',
```
nagmat84: I don't like the translation. First and for all, it does not make clear that the capability is restricted to management of the _own_ user account. The way how it is written one could assume that it allows to manage any user account. Secondly, my mental picture of a "user modification" imagines a tattoo or piercing studio :rofl:

Maybe

```suggestion
			'ALLOW_USER_SELF_EDIT' => 'Allow self-management of user account',
```

?

app/Policies/AlbumPolicy.php
```diff
@@ -172,7 +172,7 @@ public function canDownload(?User $user, ?BaseAlbum $baseAlbum): bool
 	 */
 	public function canEdit(User $user, ?AbstractAlbum $album): bool
 	{
-		if (!$this->userPolicy->canUpload($user)) {
+		if (!$this->userPolicy->mayUpload($user)) {
```
nagmat84: Here we are inside a policy which decision for a right should be based on capabilities, security attributes and permissions, but not on rights from other policies.

```suggestion
		if (!$user->may_upload) {
```

Moreover you cannot upload a user, i.e. `UserPolicy::mayUpload` should simply not exist at all, because the user policy should returns rights on (other) user objects, not capabilities of that user.

The same holds for all the lines below in this class.

app/Policies/PhotoPolicy.php
```diff
@@ -161,7 +161,7 @@ public function canEdit(User $user, Photo $photo)
 	 */
 	public function canEditById(User $user, array $photoIDs): bool
 	{
-		if (!$this->userPolicy->canUpload($user)) {
+		if (!$this->userPolicy->mayUpload($user)) {
```
nagmat84: Same as for albums.

app/Policies/UserPolicy.php
```diff
@@ -10,8 +10,8 @@ class UserPolicy
 	use HandlesAuthorization;
 
 	public const IS_ADMIN = 'isAdmin';
-	public const CAN_UPLOAD = 'canUpload';
-	public const CAN_EDIT_SETTINGS = 'canEditSettings';
+	public const MAY_UPLOAD = 'mayUpload';
```
nagmat84: `MAY_UPLOAD` should not exist at all. A policy returns _rights_ not capabilities. Moreover, one cannot upload a user.

app/Policies/UserPolicy.php
```diff
@@ -10,8 +10,8 @@ class UserPolicy
...
+	public const MAY_EDIT_OWN_SETTINGS = 'mayEditOwnSettings';
```
nagmat84: We should only have a `CAN_EDIT` _right_.

app/Policies/UserPolicy.php
```diff
@@ -37,7 +37,7 @@ public function before(?User $user, $ability)
 	 */
 	public function isAdmin(?User $user): bool
 	{
-		return $user?->id === 0;
+		return $user?->may_administrate === true;
 	}
```
nagmat84: Should actually not exist at all.

app/Policies/UserPolicy.php
```diff
@@ -47,9 +47,9 @@ public function isAdmin(?User $user): bool
 	 *
 	 * @return bool
 	 */
-	public function canEditSettings(User $user): bool
+	public function mayEditOwnSettings(User $user): bool
 	{
-		return !$user->is_locked;
+		return $user->may_edit_own_settings;
 	}
```
nagmat84: I haven't tested it yet, but we should have something like

```php
public function canEdit(User $user, User $userToBeEdited): bool
{
    return $user->id === $userToBeEdited->id && $user->may_edit_own_settings;
}

We do not need the check for `is_administrate` as I wrote above, because this is already handled in `before()`.

File: app/Policies/UserPolicy.php [hunk 17]
```diff
@@ -59,7 +59,7 @@ public function canEditSettings(User $user): bool
 	 *
 	 * @return bool
 	 */
-	public function canUpload(User $user): bool
+	public function mayUpload(User $user): bool
 	{
 		return $user->may_upload;
 	}
```
nagmat84: Should not exist.

database/migrations/2022_09_07_141854_rename_capabilities.php
```diff
@@ -0,0 +1,58 @@
...
+	 */
+	public function up()
+	{
+		DB::beginTransaction();
+
+		// flip the locked value
+		$ids = DB::table('users')->where('is_locked', '=', true)->get('id')->pluck('id');
+		DB::table('users')->whereIn('id', $ids)->update(['is_locked' => 0]);
+		DB::table('users')->whereNotIn('id', $ids)->update(['is_locked' => 1]);
+		DB::commit();
```
nagmat84: Why that difficult? Why not simply

```suggestion
		DB::table('users')->update(['is_locked' => DB::raw('NOT is_locked'));
```

I haven't tried it, so maybe it does not work with SQLite or MySQL, but it should work in ANSI SQL.

nagmat84 (changes requested):
I appreciate the endeavor. However, this PR still mixes capabilities and rights. If we go for it (we should), we should try to get it correct this time.

ildyria:
And back to the drawing boards. :laughing:

app/Policies/UserPolicy.php
ildyria: But you still need a basic `admin` check for some parts. e.g. access to full settings.

app/Policies/UserPolicy.php
ildyria: I think this is over kill: the reason being that only admins can edit other users. This is handled by the `before()` .

ildyria:
This PR is getting "a bit" bigger than I expected. Sorry @nagmat84

ildyria:
I am not sure yet if this is 100% merge ready. I tried to cover quite a few of the corner cases.
The front-end still depends heavily on the global right to upload instead of per album... but I would rather do those change subsequently.

Nevertheless, as always I expect a lots of complaints of @nagmat84 so here we go for a first round of reviews. :D
Hence yes, @nagmat84, this is a review request. ;)

app/Actions/Album/SetProtectionPolicy.php
```diff
@@ -29,12 +29,12 @@ class SetProtectionPolicy extends Action
...
-		$album->requires_link = $protectionPolicy->requiresLink;
-		$album->is_nsfw = $protectionPolicy->isNSFW;
-		$album->is_downloadable = $protectionPolicy->isDownloadable;
-		$album->is_share_button_visible = $protectionPolicy->isShareButtonVisible;
+		$album->is_public = $protectionPolicy->is_public;
+		$album->is_link_required = $protectionPolicy->is_link_required;
+		$album->is_nsfw = $protectionPolicy->is_nsfw;
+		$album->is_share_button_visible = $protectionPolicy->is_share_button_visible;
+		$album->grant_download = $protectionPolicy->grant_download;
+		$album->grant_access_full_photo = $protectionPolicy->grant_access_full_photo;
```
nagmat84: I assume you have reordered and renamed the properties of the album to anticipate that the first four properties (`is...`) will remain security attributes of the album while the last two (`grant...`) will become permissions on a user-album relation in the future. Right? If so, we should add a comment here, because nobody will remember that consideration in two months.

Moreover, why did you remove the 3rd-person singular "s" from "grants"? We also use "is" and the album grant**s** access. I would revert that grammatical change.

```suggestion
		// Security attributes of the album itself independent of a particular user
		// Note: The first one (`is_public`) will become implicit in the future when the following three attributes are
		// move to a separate table for sharing albums with anonymous users
		$album->is_public = $protectionPolicy->is_public;
		$album->is_link_required = $protectionPolicy->is_link_required;
		$album->is_nsfw = $protectionPolicy->is_nsfw;
		$album->is_share_button_visible = $protectionPolicy->is_share_button_visible;
		// (Future) permissions on an album-user relation.
		// Note: For the time being these are still "globally" defined on the album for all users, but they will be
		// moved to a separate table for sharing albums with users.
		$album->grants_download = $protectionPolicy->grants_download;
		$album->grants_full_photo_access = $protectionPolicy->grant_full_photo_access;
```

app/Actions/Albums/Top.php
```diff
@@ -64,7 +64,7 @@ public function get(): TopAlbums
 		$smartAlbums = $this->albumFactory
 			->getAllBuiltInSmartAlbums(false)
 			->map(
-				fn ($smartAlbum) => Gate::check(AlbumPolicy::IS_VISIBLE, $smartAlbum) ? $smartAlbum : null
+				fn ($smartAlbum) => Gate::check(AlbumPolicy::CAN_SEE, $smartAlbum) ? $smartAlbum : null
```
nagmat84, line 67: I need to think some time, if I like this or should hate it.

This is perfectly fine with respect to the systematics to start the name of a right with `CAN...`. `CAN_VISIBLE` obviously makes no sense and `CAN_SEE` is in perfect line with `CAN_DOWNLOAD`, `CAN_ACCESS`, `CAN_ARCHIVE`, ..., etc. So actually I should like it.

But we also have a "visibility filter" in the query policy which is strongly related to this right and hence the verb should not change. :(

app/Actions/User/ResetAdmin.php
```diff
@@ -25,7 +25,8 @@ public function do(): void
 		$user->id = 0;
 		$user->username = '';
 		$user->password = '';
-		$user->is_locked = false;
+		$user->may_edit_own_settings = true;
+		$user->may_administrate = true;
```
nagmat84, line 29: Given this two lines, I need to share a (fun) story which happened to me recently during a discussion with a CC evaluator.

In CC language "administration" (or "administrate") is nothing technical, but conducted by the operating organization. For example, the HR department administrates matters related to employees and employments. "Configuration" (or "configure") is something which happens during manufacturing or procurement. For example, a firewall may have different form factors (19" rack mountable with 48 ports, table-top small form factor with 8 ports, etc.) but still be the same product from evaluation perspective with different configurations. However, settings which are changed by a privilege user role at runtime are _managed_.

We also used the terms in different meanings and the evaluator was completely derailed. He was completely puzzled how a user should be able to configure settings at runtime. (We used "configure settings" instead of "manage settings"). In the end we had to change our wording. Luckily we do not plan to get a CC evaluation for Lychee, otherwise it should probably be `may_self_manage` and `may_manage`.

app/Contracts/AbstractAlbum.php
```diff
@@ -26,9 +26,9 @@
  * @property string     $title
  * @property Collection $photos
  * @property Thumb|null $thumb
- * @property bool       $is_public
- * @property bool       $is_downloadable
- * @property bool       $is_share_button_visible
+ * @property bool       $is_public               // Not serialized
+ * @property bool       $grant_download          // Not serialized
+ * @property bool       $is_share_button_visible // Not serialized
```
nagmat84: I am not sure what you mean by "not serialized", but I will (hopefully) see later down this PR. (This comment is rather a reminder for myself to get back to this point later.)

But 3rd-person singular "s", please: `grants_download`.

app/DTO/AlbumDTO.php
```diff
@@ -0,0 +1,45 @@
...
+ * Data Transfer Object (DTO) for Albums.
+ *
+ * This allows us to decorate the Album with its associated current user rights.
+ */
+class AlbumDTO extends DTO
+{
+	public function __construct(
+		private AbstractAlbum $album
+	) {
+	}
```
nagmat84: I am not sure, if I like this new language construct called "constructor promotion". Do we really want to use it? At least the class properties should be properly defined inside the class. This makes it really hard to figure out which properties a class provides.

By the way: IMHO the class should be called `AbstractAlbumDTO` if it encapsulates an `AbstractAlbum`.

app/DTO/AlbumDTO.php
```diff
@@ -0,0 +1,45 @@
...
+
+use App\Contracts\AbstractAlbum;
+use App\Models\Album;
+
+/**
+ * Data Transfer Object (DTO) for Albums.
+ *
+ * This allows us to decorate the Album with its associated current user rights.
+ */
+class AlbumDTO extends DTO
```
nagmat84:
```suggestion
class AbstractAlbumDTO extends DTO
```

The class transfers an `AbstractAlbum`, then it should also be called like it.

app/DTO/AlbumDTO.php
```diff
@@ -0,0 +1,45 @@
...
+{
+	public function __construct(
+		private AbstractAlbum $album
+	) {
+	}
+
+	/**
+	 * Album with it's associated righs.
```
nagmat84:
```suggestion
	 * Abstract album with it's associated rights.
```

app/DTO/AlbumDTO.php
```diff
@@ -0,0 +1,45 @@
...
+			$albumDTO['albums'] = $this->album->children->map(fn (Album $a) => ((new AlbumDTO($a))->toArray()));
+		}
+
+		// add the rights
+		$albumDTO['rights'] = AlbumRights::ofAlbum($this->album);
+
+		// Provide the policies if the user can edit.
+		if ($albumDTO['rights']->can_edit) {
+			$albumDTO['policies'] = AlbumProtectionPolicy::ofAlbum($this->album);
+		}
```
nagmat84:
```suggestion
		$albumDTO['policies'] = $albumDTO['rights']->can_edit ? AlbumProtectionPolicy::ofAlbum($this->album) : null;
```

For future clients which are strongly typed, the DTO should always have the same properties and be identically structured. Hence,  `$albumDTO['policies']` should not be left undefined but properly set to `null`.

app/DTO/AlbumProtectionPolicy.php
```diff
@@ -2,50 +2,66 @@
...
+use App\Contracts\AbstractAlbum;
+use App\Models\Extensions\BaseAlbum;
+use App\SmartAlbums\BaseSmartAlbum;
 
+/**
+ * This defines the Protection policy.
+ *
+ * In other words this gives the info of the security attributes of an album.
+ * It should not be used in the front end to determine whether an action is doable or not (e.g. share link available).
+ */
```
nagmat84:
```suggestion
/**
 * This represents the Album Protection Policy.
 *
 * In other words this provides information on the security attributes of an album.
 * It should not be used in the front-end to determine whether an action is permitted or not (e.g. share link available).
 */
```

After I have read all the code of this class below, I am unsure what you want to achieve with the class and what its purpose should be. It feels only half-baked and inconsistent.

Previously, this DTO was only used to encapsulate and group together the various settings of a base album when the client attempted to set the values on the base album, i.e. the information flow was limited from client to backend and this DTO was only used in a very limited scope. Now, you seem to extend the intended use of this class to also transfer data from the backend to the client. That is fine, but we first need to answer the following (binary) question and then stick to that answer:

1. This DTO transfers the "raw", unmodified values of the Album Protection Policy as they are stored in the DB
2. This DTO transfers the "interpreted" values of the Album Protection Policy as they appear to be effective

It would have assumed that no. 1 is the correct answer. This would make a lot of sense to me. It is also in line with what you wrote in `AlbumDTO` about this object. The client must not use the values of this object to eventually determine whether a certain action is permitted. The client must consult the effective rights for that. This is also in line with the fact that this DTO is only sent to the client, if the client is allowed to edit the Album Protection Policy and then can use the values of this DTO to initialize the dialog. This approach would also be consistent with the previous use of this DTO to transfer these values from the client to the backend. These are all good reasons that speak in favor of approach no 1.

However, there are some aspects in this class which run contrary to this interpretation and indicate that you are trying to achieve no. 2. First of all, if no 1 was correct, why is there a need to define this class for smart albums? The Album Protection Policy of smart albums is not editable, but governed by global settings. Defining this class for smart albums as well would only make sense if an object of this class is supposed to send the effective rights to the client. For regular albums you also use the properties of the model to initialize the values of this class, e.g. `grant_access_full_photo: $abstractAlbum->grant_access_full_photo`. However, these attributes already return the interpreted, effective value not the value stored in the DB.

I still believe, you wanted to go for no. 1 and I am also convinced that this is the correct approach, but this is not what this class does consistently.

app/DTO/AlbumProtectionPolicy.php
```diff
@@ -2,50 +2,66 @@
...
-		bool $isNSFW,
-		bool $isDownloadable,
-		bool $isShareButtonVisible,
-		bool $grantsFullPhoto
+		public bool $is_public,
+		public bool $is_link_required,
+		public bool $is_nsfw,
+		public bool $is_share_button_visible,
+		public bool $grant_access_full_photo,
+		public bool $grant_download,
```
nagmat84:
```suggestion
		public bool $grants_full_photo_access,
		public bool $grants_download,
```

3rd-person-singular-s, please. I won't blue-pencil every occurrence of this during this review, because this is something which is much more easily corrected consistently in a proper IDE which supports refactoring.

app/DTO/AlbumProtectionPolicy.php
```diff
@@ -2,50 +2,66 @@
...
+		public bool $is_password_required = false,
```
nagmat84: I am still unsure if I like this kind of new, implicit definition of class properties. I am wondering what @kamil4 and @qwerty287 have to say about it.

This kind of shortcut might be fine for "simple" classes if _all_ class properties are implicitly defined by constructor arguments _and_ if _all_ constructor arguments define class properties. Note, that these are two different directions. But I am worried about the mixed case like this

```php
class MyClass {
    protected int $b;

    public function __construct(
        public string $a,
        protected int $c,
        bool $d,
    ) {
        $this->b = $d ? 1 : 0;
    }
}
```

In this case you have to consult both, the class definition and the constructor to determine which properties are defined on the class. IMHO, it makes the code much harder to read and understand for the questionable advantage of saving some typing effort.

app/DTO/AlbumProtectionPolicy.php
```diff
@@ -2,50 +2,66 @@
...
+	 * - Tag albums
+	 * - Smart albums
+	 * - Normal albums.
+	 *
+	 * @param AbstractAlbum $abstractAlbum
+	 *
+	 * @return AlbumProtectionPolicy|null
 	 */
-	public function toArray(): array
+	public static function ofAlbum(AbstractAlbum $abstractAlbum): AlbumProtectionPolicy|null
```
nagmat84: Given that comment and the code below, I rather prefer to have at least different methods for the different sub-classes of `AbstractAlbum` or even better have different DTO classes in line with the album hierarchy.

At least we should have methods like
 - `ofBaseAlbum(BaseAlbum $baseAlbum): AlbumProtectionPolicy`
 - `ofSmartAlbum(SmartAlbum $smartAlbum): AlbumProtectionPolicy`
This would also remove the need for `null` as a return type. I guess the `null`-case only exists to satisfy PHPStan, because you needed a default return type when none of the `instanceof`-conditions applies. 

Maybe it is even better to have distinct classes `SmartAlbumDTO`, `BaseAlbumDTO`, etc., in line with the album class hierarchy. I haven't though about this in full details.

However, this `instanceof` cascade below bears the risk to turn into spaghetti code. The whole point of a well-defined class hierarchy is to avoid the long `switch-case` constructs of the dark ages of programming when you had to mimic type inheritance by explicit type indicators. If one really need this `instanceof` cascade to get something done, then this is a code smell that something is wrong with the type hierarchy.

app/DTO/AlbumProtectionPolicy.php
```diff
@@ -2,50 +2,66 @@
...
+
+		if ($abstractAlbum instanceof BaseSmartAlbum) {
+			return new AlbumProtectionPolicy(
+				is_public: $abstractAlbum->is_public,
+				is_link_required: false,
+				is_nsfw: false,
+				is_share_button_visible: $abstractAlbum->is_share_button_visible,
+				is_password_required: false,
+				grant_access_full_photo: false,
+				grant_download: false,
```
nagmat84: Some of the hard-coded values are not correct for smart albums. For example, `grants_download` is eventually determined by the global configuration setting `Configs::getValueAsBool('downloadable')`.

app/DTO/DiagnosticInfo.php
```diff
@@ -2,45 +2,20 @@
 
 namespace App\DTO;
 
-class DiagnosticInfo extends DTO
+class DiagnosticInfo extends ArrayableDTO
```
nagmat84, line 5: I like the approach to use a common parent class which uses reflection to serialize the properties into an array. I also understand (somehow) the benefits of the benefit of declaring DTO properties via constructor promotion.

But maybe we should pull out this change into an independent PR. This independent PR could collect all changes due to this new approach which are not related to changes due to rights and capabilities. This independent PR could also host the discussion about constructor promotion in the mixed case which I initiated some comments above.

Moreover, I believe that this independent PR is something which we can decide faster than this whole rights and capability stuff.

app/DTO/ImageDimension.php
```diff
@@ -2,25 +2,11 @@
 
 namespace App\DTO;
 
-class ImageDimension extends DTO
+class ImageDimension extends ArrayableDTO
```
nagmat84, line 5: Independent PR?

app/DTO/LycheeGitInfo.php
```diff
@@ -4,33 +4,17 @@
 
 use function Safe\sprintf;
 
-class LycheeGitInfo extends DTO
+class LycheeGitInfo extends ArrayableDTO
```
nagmat84, line 7: Independent PR?

app/DTO/SortingCriterion.php
```diff
@@ -4,7 +4,7 @@
 
 use App\Exceptions\Internal\InvalidOrderDirectionException;
 
-abstract class SortingCriterion extends DTO
+class SortingCriterion extends ArrayableDTO
```
nagmat84, line 7: Independent PR?

app/DTO/User.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace App\DTO;
+
+use App\Models\User as UserModel;
+
+/**
+ * Data Transfer Object (DTO) to transmit the attributes of a User.
+ */
+class User extends ArrayableDTO
```
nagmat84, line 12: Why do we need this extra DTO? I get the point for albums and photos, because those DTOs decorate the actual album or photo with the currently effective rights of the authenticated user upon those objects. But nothing of this sort happens here. If I don't miss something, this DTO is simply a 1:1 copy of the user model.

app/DTO/UserRights.php
```diff
@@ -0,0 +1,36 @@
...
+use App\Contracts\AbstractAlbum;
+use App\Models\User;
+use App\Policies\AlbumPolicy;
+use App\Policies\UserPolicy;
+use Illuminate\Support\Facades\Gate;
+
+/**
+ * Data Transfer Object (DTO) to transmit the rights of a User.
+ */
+class UserRights extends ArrayableDTO
```
nagmat84: This DTO doesn't make sense to me (again). Rights are something which the currently authenticated user has on some other object "under investigation". This is true for `AlbumRights`, `PhotoRights` etc. They contain the effective rights of the currently authenticated user on that album or photo.

Following the same structure, `UserRights` has summarize the effective rights which the currently authenticated user has on some other user. (Remark: This other user can also be identical the authenticated user. in that case this object would represent the rights of the user on its own account.) In this kind of systematics, this object has to represent rights of a user on a user, for example: `CAN_CHANGE_PASSWORD`. This would make sense, e.g. can user X change the password of user Y. This right would be granted for administrators and if the authenticated user is identical to the user under investigation.

However, here we have things like "CAN_UPLOAD_ROOT". This is not a right of a user on another user.

app/DTO/Version.php
```diff
@@ -4,17 +4,13 @@
 
 use App\Exceptions\Internal\LycheeInvalidArgumentException;
 
-class Version extends DTO
+class Version extends ArrayableDTO
```
nagmat84, line 7: Independent PR?

nagmat84 (changes requested):
Reviewed 22 of 91 files.

app/DTO/AlbumProtectionPolicy.php
ildyria: > why is there a need to define this class for smart albums

To provide a "review" of what the current access possibilities are for Smart Albums?

app/DTO/AlbumProtectionPolicy.php
qwerty287: I agree here with @nagmat84, but here everything's fine according to your definition of a "simple" class, but I agree that mixed property definitions shouldn't be used.

app/Http/Controllers/SessionController.php
```diff
@@ -63,24 +62,16 @@ public function init(): array
 				// If the session is unauthenticated ('user' === null), but grants admin rights nonetheless,
 				// the front-end shows the dialog to create an admin account.
 				$return['user'] = null;
-				$return['rights'] = [
-					'is_admin' => true,
-					'is_locked' => false,
-					'may_upload' => true,
-				];
+				$return['rights'] = new UserRights(true, true, true);
```
nagmat84: This is the issue which still bothers me most, because the attributes here are still not rights, but capabilities.

I know that I named this objects `'rights'` and that it was me who put the user capabilities into it mostly because I did not want to change too much here and in the frontend at the time when I cleaned up the `init()` method. But now, when were are cleaning up the naming convention for rights, capabilities etc. what this PR intends to we should get it right.

Rights are something based on the relation between the currently authenticated user and another object. As I already said above rights of a user on a user would be things like `MAY_CHANGE_PASSWORD`, i.e. may user X change the password of user Y.

When I added this line I envisioned a two-level array like this:

```php
$return['rights'] = [
    'root_album' => [
        can_edit => true,
        can_upload => true,
        can_download => true,
        /* more rights on root album */
    ],
    'settings' => [
        'can_edit' => true,
        /* more detailed rights on settings in the future */
    ],
   /* more objects on which a user can have rights */
];
```

This approach would stress the relational nature of rights and clearly, things like `may_administrate` would not be among them.

Given the new development above, I wonder whether we still need this entry at all or if we are just fine the new DTO above. So lets step through the three attributes we currently have in `UserRights`:

 - `can_upload_root`: I assume we don't need this here at all. The same way as albums and photos are wrapped into a DTO `AlbumDTO` and `PhotoDTO` we could as well wrap the root album into one which augments the rights of the user on the root album.
 - `can_edit_own_settings`: Not needed for the same reason. Instead of returning the user model one line above we should return a DTO which wraps the user model and augments rights of the current user on that user (which happens to be the user itself). At the moment I see to actual user rights (on users): `can_edit_all` and `can_edit_password`.
 - `can_administrate`: This one is a little problematic right now without changing too much in the frontend as well. But basically it could and should be solved in the same spirit. For the time being (i.e. without changing too much in the frontend) we may as well drop this one (for now) and directly use the capability `may_administrate` of the user model in the frontend. (This is also not the perfect solution, but still better than calling this one rights here.)

------------------------------

However, there is one more problem to solve and this is how we express rights on `null` models. Currently, we need this in two cases: for the admin user as long as no real admin user exists (i.e. after installation) and for the root album. Currently, the DTOs above (i.e. `AlbumDTO` and `PhotoDTO` take a non-null model and _add_ the attribute `rights` to it. This approach does not work if the model itself is `null`. Hence, maybe a more sustainable approach would be not to _add_ the attribute, but place it next to the model, i.e. something like that

```php
class ModelWithRighsDTO extends ArrayableDTO {
    public ?Model $model;
    public ModelRights $rights;

    public function __construct(?Model $model = null) {
        $this->model = $model;
        $this->rights = new ModelRights::ofModel($model);
    }
}
```

This would also allow to report rights on `null` models to the frontend and solve a lot problems.

app/Http/Livewire/Sidebar.php
```diff
@@ -72,8 +73,8 @@ public function generateAlbumStructure()
 		$share = new \stdClass();
 		$share->title = Lang::get('ALBUM_SHARING');
 		$_public = $this->album->is_public ? Lang::get('ALBUM_SHR_YES') : Lang::get('ALBUM_SHR_NO');
-		$_hidden = $this->album->requires_link ? Lang::get('ALBUM_SHR_YES') : Lang::get('ALBUM_SHR_NO'); // TODO : double check;
-		$_downloadable = $this->album->is_downloadable ? Lang::get('ALBUM_SHR_YES') : Lang::get('ALBUM_SHR_NO');
+		$_hidden = $this->album->is_link_required ? Lang::get('ALBUM_SHR_YES') : Lang::get('ALBUM_SHR_NO'); // TODO : double check;
+		$_downloadable = $this->album->grant_download ? Lang::get('ALBUM_SHR_YES') : Lang::get('ALBUM_SHR_NO');
```
nagmat84:
```suggestion
		$_downloadable = $this->album->grants_download ? Lang::get('ALBUM_SHR_YES') : Lang::get('ALBUM_SHR_NO');
```

app/Http/Requests/Album/SetAlbumCoverRequest.php
```diff
@@ -26,7 +26,7 @@ class SetAlbumCoverRequest extends BaseApiRequest implements HasAlbum, HasPhoto
 	 */
 	public function authorize(): bool
 	{
-		return Gate::check(AlbumPolicy::CAN_EDIT, [AbstractAlbum::class, $this->album]) && Gate::check(PhotoPolicy::IS_VISIBLE, $this->photo);
+		return Gate::check(AlbumPolicy::CAN_EDIT, [AbstractAlbum::class, $this->album]) && Gate::check(PhotoPolicy::CAN_SEE, $this->photo);
```
nagmat84, line 29: I am find with changing `IS_VISIBLE` to `CAN_SEE` for now. I wouldn't do anything more. But whenever I see things like this, I feel the urge to come up with a proper photo right `CAN_EDIT`.

app/Http/Requests/Album/SetAlbumProtectionPolicyRequest.php
```diff
@@ -30,12 +37,12 @@ public function rules(): array
...
-			AlbumProtectionPolicy::IS_PUBLIC_ATTRIBUTE => 'required|boolean',
-			AlbumProtectionPolicy::REQUIRES_LINK_ATTRIBUTE => 'required|boolean',
-			AlbumProtectionPolicy::IS_NSFW_ATTRIBUTE => 'required|boolean',
-			AlbumProtectionPolicy::IS_DOWNLOADABLE_ATTRIBUTE => 'required|boolean',
-			AlbumProtectionPolicy::IS_SHARE_BUTTON_VISIBLE_ATTRIBUTE => 'required|boolean',
-			AlbumProtectionPolicy::GRANTS_FULL_PHOTO_ATTRIBUTE => 'required|boolean',
+			self::IS_PUBLIC_ATTRIBUTE => 'required|boolean',
+			self::IS_LINK_REQUIRED_ATTRIBUTE => 'required|boolean',
+			self::IS_NSFW_ATTRIBUTE => 'required|boolean',
+			self::GRANT_DOWNLOAD_ATTRIBUTE => 'required|boolean',
```
nagmat84:
```suggestion
			self::GRANTS_DOWNLOAD_ATTRIBUTE => 'required|boolean',
```

app/Http/Requests/Album/SetAlbumProtectionPolicyRequest.php
```diff
@@ -30,12 +37,12 @@ public function rules(): array
...
+			self::IS_SHARE_BUTTON_VISIBLE_ATTRIBUTE => 'required|boolean',
+			self::GRANT_ACCESS_FULL_PHOTO_ATTRIBUTE => 'required|boolean',
```
nagmat84:
```suggestion
			self::GRANTS_FULL_PHOTO_ACCESS_ATTRIBUTE => 'required|boolean',
```

app/Http/Requests/Settings/ChangeLoginRequest.php
```diff
@@ -26,7 +26,7 @@ class ChangeLoginRequest extends BaseApiRequest implements HasPassword
 	 */
 	public function authorize(): bool
 	{
-		return Gate::check(UserPolicy::CAN_EDIT_SETTINGS, User::class);
+		return Gate::check(UserPolicy::CAN_EDIT_OWN_SETTINGS, User::class);
```
nagmat84: In order to follow the general systematics we need something like this

```php
return Gate::check(UserPolicy::CAN_EDIT_SETTINGS, [User::class, Auth::user()]);
```

This might look a little bit weird on the first glance, but makes a lot of sense if one keeps in mind that the currently authenticated user has rights on some object (and this object happens to be the user itself). The `UserPolicy` has to look like this

```php
class UserPolicy {
    public function canEditSettings(User $user, User $otherUser): void
    {
        // Bypass for administrators is granted by `before`, so we only need to deal with the case of self-editing here
        return $user->id === $otherUser->id; 
    }
}

File: app/Http/Requests/User/SetEmailRequest.php [hunk 45]
```diff
@@ -15,7 +15,7 @@ class SetEmailRequest extends BaseApiRequest
 
 	public function authorize(): bool
 	{
-		return Gate::check(UserPolicy::CAN_EDIT_SETTINGS, User::class);
+		return Gate::check(UserPolicy::CAN_EDIT_OWN_SETTINGS, User::class);
```
nagmat84: See some comments above. In order to stay systematic we should use
```php
return Gate::check(UserPolicy::CAN_EDIT_SETTINGS, [User::class, Auth::user()]);
```

app/Locale/ChineseSimplified.php
```diff
@@ -513,7 +513,7 @@ public function get_locale(): array
 			'CAMERA_DATE' => 'Camera date',
 			'NEW_PASSWORD' => 'new password',
 			'ALLOW_UPLOADS' => 'Allow uploads',
-			'RESTRICTED_ACCOUNT' => 'Restricted account',
+			'ALLOW_USER_SELF_EDIT' => 'Allow self-management of user account',
```
nagmat84, line 506: You used "manage". Our CC evaluator (at work) would be pleased. :heart_eyes: :smile:

app/Models/BaseAlbumImpl.php
```diff
@@ -91,11 +91,11 @@
...
  * @property User                       $owner
  * @property bool                       $is_public
- * @property bool                       $grants_full_photo
- * @property bool                       $requires_link
- * @property bool                       $is_downloadable
+ * @property bool                       $is_link_required
  * @property bool                       $is_share_button_visible
  * @property bool                       $is_nsfw
+ * @property bool                       $grant_access_full_photo
+ * @property bool                       $grant_download
```
nagmat84:
```suggestion
 * @property bool                       $grants_full_photo_access
 * @property bool                       $grants_download
```

app/Models/BaseAlbumImpl.php
```diff
@@ -144,15 +144,17 @@ class BaseAlbumImpl extends Model implements HasRandomID
 		'title' => null, // Sic! `title` is actually non-nullable, but using `null` here forces the caller to actually set a title before saving.
 		'description' => null,
 		'owner_id' => 0,
+		'sorting_col' => null,
+		'sorting_order' => null,
+		// Security attributes
+		'is_nsfw' => false, // this is a special case.
```
nagmat84: Please be more specific in the comment. Why is this a special case? What makes it special?

app/Models/BaseAlbumImpl.php
```diff
@@ -144,15 +144,17 @@ class BaseAlbumImpl extends Model implements HasRandomID
...
-		'is_downloadable' => false,
+		'is_link_required' => false,
 		'is_share_button_visible' => false,
-		'is_nsfw' => false,
 		'password' => null,
-		'sorting_col' => null,
-		'sorting_order' => null,
+		// Permissions
+		'grant_access_full_photo' => true,
+		'grant_download' => false,
```
nagmat84:
```suggestion
		'grants_full_photo_access' => true,
		'grants_download' => false,
```

app/Models/BaseAlbumImpl.php
```diff
@@ -180,6 +182,14 @@ class BaseAlbumImpl extends Model implements HasRandomID
 		'password',
 		'sorting_col',   // serialize DTO `order` instead
 		'sorting_order', // serialize DTO `order` instead
+
+		// Security attributes attributes
```
nagmat84:
```suggestion
		// Security attributes
```

app/Models/BaseAlbumImpl.php
```diff
@@ -180,6 +182,14 @@ class BaseAlbumImpl extends Model implements HasRandomID
...
+		'is_public',
+		'is_link_required',
+		'is_share_button_visible',
+		// Permissions taken care by Rights
```
nagmat84: What do you want to tell us with that comment? At least the proposition "of" is missing here but still I don't get the intention of the comment.

app/Models/BaseAlbumImpl.php
```diff
@@ -180,6 +182,14 @@ class BaseAlbumImpl extends Model implements HasRandomID
...
+		'grant_download',
+		'grant_access_full_photo',
```
nagmat84:
```suggestion
		'grants_download',
		'grants_full_photo_access',
```

app/Models/Extensions/BaseAlbum.php
```diff
@@ -26,11 +26,11 @@
  * @property Carbon                     $updated_at
  * @property string|null                $description
  * @property bool                       $is_nsfw
- * @property bool                       $grants_full_photo
+ * @property bool                       $grant_access_full_photo
```
nagmat84:
```suggestion
 * @property bool                       $grants_full_photo_access
```

app/Models/Photo.php
```diff
@@ -72,8 +71,7 @@
  * @property Album|null   $album
  * @property User         $owner
  * @property SizeVariants $size_variants
- * @property bool         $is_downloadable
- * @property bool         $is_share_button_visible
+ * @property bool         $grant_download
```
nagmat84:
```suggestion
 * @property bool         $grants_download
```

app/Models/Photo.php
```diff
@@ -104,7 +102,7 @@ class Photo extends Model implements HasRandomID
 		'taken_at' => DateTimeWithTimezoneCast::class,
 		'live_photo_full_path' => MustNotSetCast::class . ':live_photo_short_path',
 		'live_photo_url' => MustNotSetCast::class . ':live_photo_short_path',
-		'is_downloadable' => MustNotSetCast::class,
+		'grant_download' => MustNotSetCast::class,
```
nagmat84:
```suggestion
		'grants_download' => MustNotSetCast::class,
```

app/Models/Photo.php
```diff
@@ -135,8 +133,9 @@ class Photo extends Model implements HasRandomID
 	 */
 	protected $appends = [
 		'live_photo_url',
-		'is_downloadable',
-		'is_share_button_visible',
+		// No longer provided because parts of Rights DTO
+		// 'grant_download',
+		// 'is_share_button_visible',
```
nagmat84:
```suggestion
```

Just remove outdated and broken code, don't comment it out. In two month everybody is wondering _why_ it has been commented out and whether this is a temporary fix only and the code is still there because it might come back.

app/Models/Photo.php
```diff
@@ -438,17 +399,13 @@ public function toArray(): array
 			$result['is_public'] = $result['is_public'] === true ? 1 : 0;
 		}
 
+		$result['rights'] = PhotoRights::ofPhoto($this);
```
nagmat84: `PhotoDTO` also adds `$result['rights']`, correct? Hence, this one is actually superfluous and I would also call it bad design because it unconditionally adds things to the serialized `Photo` model which are not truly part of the model.

I know the currently effective rights are required below to decide whether the `'original'` size variant must be unset from the returned array, but I would not add the rights to the array here.

app/Models/Photo.php
```diff
@@ -438,17 +399,13 @@ public function toArray(): array
...
+
 		// Downgrades the accessible resolution of a photo
-		// The decision logic here is a merge of three formerly independent
-		// (and slightly different) approaches
 		if (
-			!Gate::check(PhotoPolicy::IS_OWNER, $this) &&
+			!$result['rights']->can_access_full_photo &&
```
nagmat84:
```suggestion
			!Gate::check(PhotoPolicy::CAN_ACCESS_FULL_PHOTO, [Photo::class, $this]) &&
```

because `$result['rights']` should not be added to the result (see above).

app/Models/User.php
```diff
@@ -139,7 +146,7 @@ public function username(): string
 	 *
 	 * @return string
 	 */
-	public function name(): string
+	public function getNameAttribute(): string
```
nagmat84: I guess this change is a false positive as this should already have been fixed by https://github.com/LycheeOrg/Lychee/pull/1525. I guess this PR needs another merge with the current `master` branch?

app/Policies/AlbumPolicy.php
```diff
@@ -119,31 +142,56 @@ public function canAccess(?User $user, ?AbstractAlbum $album): bool
...
 	 *
 	 * @throws ConfigurationKeyMissingException
 	 */
-	public function canDownload(?User $user, ?BaseAlbum $baseAlbum): bool
+	public function canDownload(?User $user, ?AbstractAlbum $abstractAlbum): bool
 	{
-		if ($baseAlbum === null) {
-			return Configs::getValueAsBool('downloadable');
+		$default = Configs::getValueAsBool('downloadable');
+		if ($abstractAlbum === null) {
```
nagmat84, line 144:
```suggestion
		// The root album always uses the global setting
		if ($abstractAlbum === null) {
```

app/Policies/AlbumPolicy.php
```diff
@@ -119,31 +142,56 @@ public function canAccess(?User $user, ?AbstractAlbum $album): bool
...
+	 * @param User               $user
+	 * @param AbstractAlbum|null $abstractAlbum
+	 *
+	 * @return bool
+	 *
+	 * @throws ConfigurationKeyMissingException
+	 */
+	public function canUpload(User $user, ?AbstractAlbum $abstractAlbum = null): bool
+	{
+		// If base album is null, we consider root
```
nagmat84:
```suggestion
		// The upload right on the root album is directly determined by the user's capabilities.
```

app/Policies/AlbumPolicy.php
```diff
@@ -183,24 +231,6 @@ public function canEdit(User $user, ?AbstractAlbum $album): bool
 			($album instanceof BaseAlbum && $this->isOwner($user, $album));
```
nagmat84:
```suggestion
			($album instanceof BaseAlbum && $album->owner->id === $user->id);
```

I would actually try to throw away the method `PhotoPolicy::isOwner` or at least make it not public. `isOwner` is not a right, hence it should not be provided publicly by this class. On top it is such a trivial method that it barely makes sense to have `isOwner` as an internal helper method.

app/Policies/AlbumPolicy.php
```diff
@@ -243,6 +273,72 @@ public function canEditById(User $user, array $albumIDs): bool
...
+
+		if ($user?->may_upload !== true) {
+			return false;
+		}
+
+		if (in_array($abstractAlbum->id, AlbumFactory::BUILTIN_SMARTS, true)) {
+			return false;
+		}
+
+		return $abstractAlbum instanceof BaseAlbum && $this->isOwner($user, $abstractAlbum);
```
nagmat84, line 279:
```suggestion
		return $abstractAlbum instanceof BaseAlbum && $abstractAlbum->owner->id === $user?->id);
```

As above:

> I would actually try to throw away the method `PhotoPolicy::isOwner` or at least make it not public. `isOwner` is not a right, hence it should not be provided publicly by this class. On top it is such a trivial method that it barely makes sense to have `isOwner` as an internal helper method.

app/Policies/AlbumPolicy.php
```diff
@@ -243,6 +273,72 @@ public function canEditById(User $user, array $albumIDs): bool
...
+	 * @param AbstractAlbum $abstractAlbum
+	 *
+	 * @return bool
+	 *
+	 * @throws ConfigurationKeyMissingException
+	 */
+	public function canShareByLink(?User $user, ?AbstractAlbum $abstractAlbum): bool
+	{
+		if ($abstractAlbum instanceof BaseAlbum
+			&& $this->isOwner($user, $abstractAlbum)
```
nagmat84:
```suggestion
			&& $abstractAlbum->owner->id === $user?->id
```

As above:

> I would actually try to throw away the method `PhotoPolicy::isOwner` or at least make it not public. `isOwner` is not a right, hence it should not be provided publicly by this class. On top it is such a trivial method that it barely makes sense to have `isOwner` as an internal helper method.

nagmat84 (commented):
57 of 91 files.

app/Policies/AlbumPolicy.php
```diff
@@ -27,11 +27,16 @@ class AlbumPolicy
...
+
+	public const CAN_SEE = 'canSee';
 	public const CAN_ACCESS = 'canAccess';
 	public const CAN_DOWNLOAD = 'canDownload';
+	public const CAN_UPLOAD = 'canUpload';
 	public const CAN_EDIT = 'canEdit';
-	public const IS_VISIBLE = 'isVisible';
 	public const CAN_EDIT_ID = 'canEditById';
+	public const CAN_SHARE_WITH_USERS = 'canShareWithUsers';
+	public const CAN_SHARE_BY_LINK = 'canShareByLink';
```
nagmat84: Maybe we should rename that to something like `CAN_SEE_SHARE_LINKS` or `CAN_SEE_DIRECT_LINKS_FOR_SHARING`. The point is we cannot actually forbid someone to share links and the links are straightforward if you know the ID of the photo or album. This whole right is more or less only a convenience feature whether the user sees the button in the toolbar and has the option to easily copy the links.

I also thought that we already had agreed upon making this a global configuration options which controls the whole UI at once whether the button is shown in the toolbar or not. It seems to be an overkill to have this on a per-album/per-photo basis given the fact that this isn't a real security feature.

app/Policies/AlbumQueryPolicy.php
```diff
@@ -71,7 +71,7 @@ public function applyVisibilityFilter(AlbumBuilder|FixedQueryBuilder $query): Al
 			$query2
 				->where(
 					fn (AlbumBuilder|TagAlbumBuilder $q) => $q
-						->where('base_albums.requires_link', '=', false)
+						->where('base_albums.is_link_required', '=', false)
```
nagmat84, line 73: I am not going to check every occurrence of `requires_link` myself. I assume that you did a search+replace and spotted all occurrences.

app/Policies/PhotoPolicy.php
```diff
@@ -18,10 +20,13 @@ class PhotoPolicy
 
 	// constants to be used in GATE
 	public const IS_OWNER = 'isOwner';
-	public const IS_VISIBLE = 'isVisible';
+
+	public const CAN_SEE = 'canSee';
 	public const CAN_DOWNLOAD = 'canDownload';
 	public const CAN_EDIT = 'canEdit';
 	public const CAN_EDIT_ID = 'canEditById';
+	public const CAN_SHARE_BY_LINK = 'canShareByLink';
```
nagmat84: See equivalent comment on albums.

app/Policies/UserPolicy.php
nagmat84: This should only happen inside the policies. Outside the policy there should no need to check this, because code outside the policies should query the rights as determined by the policies. And within the policies it is fine to query the user's capability directly, i.e. use `$user->may_administrate`.

app/Policies/UserPolicy.php
```diff
@@ -37,30 +21,33 @@ public function before(?User $user, $ability)
...
 	 * @param User $user
 	 *
 	 * @return bool
 	 */
-	public function canUpload(User $user): bool
+	public function canEditOwnSettings(User $user): bool
 	{
-		return $user->may_upload;
+		return $user->may_edit_own_settings;
 	}
```
nagmat84: I still believe we should have

```php
public function canEdit(User $user, User $otherUser): bool {
    return $user->may_edit_own_settings && $user->id === $otherUser->id;
}
```

Moreover, we probably need a right for creating/deleting users like this

```php
public function canCreateOrDelete(User $user): bool {
    // Note, the administrator is already handled in the `before()` method and every one else is not allowed to create/delete users.
    return false;
}
```

app/Policies/UserPolicy.php
```diff
@@ -37,30 +21,33 @@ public function before(?User $user, $ability)
...
+	 * @param string    $ability
 	 *
-	 * @return bool
+	 * @return void|bool
 	 */
-	public function canEditSettings(User $user): bool
+	public function before(?User $user, $ability)
 	{
-		return !$user->is_locked;
+		if ($this->isAdmin($user)) {
```
nagmat84: As we are inside a policy, it is fine to query the capabilities of a user directly

```suggestion
		if ($user?->may_administrate === true)) {
```

This gives us the chance to remove the `isAdmin` method which does not return a right.

app/SmartAlbums/BaseSmartAlbum.php
```diff
@@ -40,9 +40,9 @@ abstract class BaseSmartAlbum implements AbstractAlbum
 	protected PhotoQueryPolicy $photoQueryPolicy;
 	protected string $id;
 	protected string $title;
-	protected bool $isPublic;
-	protected bool $isDownloadable;
-	protected bool $isShareButtonVisible;
+	protected bool $grant_download;
```
nagmat84:
```suggestion
	protected bool $grants_download;
```

app/SmartAlbums/BaseSmartAlbum.php
```diff
@@ -40,9 +40,9 @@ abstract class BaseSmartAlbum implements AbstractAlbum
...
+	protected bool $is_public;
+	protected bool $is_share_button_visible;
```
nagmat84: I am not 100% sure anymore, but I thought we said that we wanted to make this a global option only (also see previous remarks).

app/SmartAlbums/BaseSmartAlbum.php
```diff
@@ -51,15 +51,15 @@ abstract class BaseSmartAlbum implements AbstractAlbum
...
 	{
 		try {
 			$this->photoQueryPolicy = resolve(PhotoQueryPolicy::class);
 			$this->id = $id;
 			$this->title = $title;
-			$this->isPublic = $isPublic;
-			$this->isDownloadable = Configs::getValueAsBool('downloadable');
-			$this->isShareButtonVisible = Configs::getValueAsBool('share_button_visible');
+			$this->is_public = $is_public;
+			$this->grant_download = Configs::getValueAsBool('downloadable');
```
nagmat84:
```suggestion
			$this->grants_download = Configs::getValueAsBool('downloadable');
```

app/SmartAlbums/BaseSmartAlbum.php
```diff
@@ -51,15 +51,15 @@ abstract class BaseSmartAlbum implements AbstractAlbum
...
+			$this->is_share_button_visible = Configs::getValueAsBool('share_button_visible');
```
nagmat84: See above.

app/SmartAlbums/BaseSmartAlbum.php
```diff
@@ -145,13 +145,11 @@ public function toArray(): array
...
 		//     only used within a listing of sub-albums.
-
 		$result = [
 			'id' => $this->id,
 			'title' => $this->title,
-			'is_public' => $this->isPublic,
-			'is_downloadable' => $this->isDownloadable,
-			'is_share_button_visible' => $this->isShareButtonVisible,
+			// 'is_public' => $this->isPublic,
+			// 'is_share_button_visible' => $this->isShareButtonVisible,
```
nagmat84:
```suggestion
```

Outdated code which will never come back should be removed properly.

app/SmartAlbums/Utils/MimicModel.php
```diff
@@ -62,6 +63,7 @@ public function toJson($options = 0): string
 	 */
 	public function __get(string $key)
 	{
+		Debugbar::warning($key);
```
nagmat84: Why? Is this temporary?

app/SmartAlbums/Utils/MimicModel.php
```diff
@@ -72,10 +74,12 @@ public function __get(string $key)
 
 		if (method_exists($this, $getter)) {
 			return $this->{$getter}(); // @phpstan-ignore-line, PhpStan does not like variadic calls
+		} elseif (property_exists($this, $key)) {
+			return $this->{$key}; // @phpstan-ignore-line, PhpStan does not like variadic calls
```
nagmat84, line 76:
```php
public function canEdit(User $user, User $otherUser): bool {
    return $user->may_edit_own_settings && $user->id === $otherUser->id;
}
```

database/migrations/2022_09_07_141854_rename_capabilities.php
```diff
@@ -0,0 +1,50 @@
...
+	 * @return void
+	 */
+	public function up()
+	{
+		// flip the locked value
+		DB::table('users')->update(['is_locked' => DB::raw('NOT is_locked')]);
+
+		// rename the column
+		Schema::table('users', function (Blueprint $table) {
+			$table->renameColumn('is_locked', 'may_edit_own_settings');
```
nagmat84, line 22: I need to check whether renaming a columns breaks foreign constraints for SQLite. I have that feeling that renaming was mimicked by deleting and re-creating the table and forgetting foreign key constraints on that way.

database/migrations/2022_09_12_175257_rename_attributes_grants.php
```diff
@@ -0,0 +1,56 @@
...
+	public function up()
+	{
+		if (Schema::connection(null)->getConnection()->getDriverName() === 'sqlite') {
+			Schema::disableForeignKeyConstraints();
+		}
+		Schema::table('base_albums', function (Blueprint $table) {
+			$table->renameColumn('requires_link', 'is_link_required');
+		});
+		Schema::table('base_albums', function (Blueprint $table) {
+			$table->renameColumn('is_downloadable', 'grant_download');
```
nagmat84:
```suggestion
			$table->renameColumn('is_downloadable', 'grants_download');
```

database/migrations/2022_09_12_175257_rename_attributes_grants.php
```diff
@@ -0,0 +1,56 @@
...
+		});
+		Schema::table('base_albums', function (Blueprint $table) {
+			$table->renameColumn('grants_full_photo', 'grant_access_full_photo');
```
nagmat84:
```suggestion
			$table->renameColumn('grants_full_photo', 'grants_full_photo_access');
```

database/migrations/2022_09_12_175257_rename_attributes_grants.php
```diff
@@ -0,0 +1,56 @@
...
+	public function down()
+	{
+		if (Schema::connection(null)->getConnection()->getDriverName() === 'sqlite') {
+			Schema::disableForeignKeyConstraints();
+		}
+		Schema::table('base_albums', function (Blueprint $table) {
+			$table->renameColumn('is_link_required', 'requires_link');
+		});
+		Schema::table('base_albums', function (Blueprint $table) {
+			$table->renameColumn('grant_download', 'is_downloadable');
```
nagmat84:
```suggestion
			$table->renameColumn('grants_download', 'is_downloadable');
```

database/migrations/2022_09_12_175257_rename_attributes_grants.php
```diff
@@ -0,0 +1,56 @@
...
+		});
+		Schema::table('base_albums', function (Blueprint $table) {
+			$table->renameColumn('grant_access_full_photo', 'grants_full_photo');
```
nagmat84:
```suggestion
			$table->renameColumn('grants_full_photo_access', 'grants_full_photo');
```

nagmat84 (commented):
74 of 91 files. That is everything except for the tests.

ildyria:
Rebased on top of #1529

app/Actions/Album/SetProtectionPolicy.php
ildyria: fixed. Can be resolved.

app/Contracts/AbstractAlbum.php
ildyria: added more details. Hopefully can be resolved.

app/DTO/AlbumDTO.php
ildyria: Fixed.

app/Models/User.php
ildyria: Well this is surprising. I thought I rebased on Master.

app/Policies/AlbumPolicy.php
ildyria: made it private.

app/Policies/AlbumPolicy.php
ildyria: made it private.

app/Policies/AlbumPolicy.php
ildyria: as above. made it private.

app/DTO/AlbumProtectionPolicy.php
```diff
@@ -2,50 +2,93 @@
...
-	 * {@inheritDoc}
+	 * TODO: To be removed.
+	 *
+	 * @param AbstractAlbum $abstractAlbum
+	 *
+	 * @return AlbumProtectionPolicy
+	 *
+	 * @throws LycheeLogicException
+	 */
+	public static function ofAlbum(AbstractAlbum $abstractAlbum): AlbumProtectionPolicy
```
ildyria: This will need to be removed. I know.

ildyria:
@nagmat84 I would suggest you have another look. There should not be many changes left to be done hopefully.
I tried to go in most corner cases.

app/DTO/AlbumProtectionPolicy.php
ildyria: done.

app/DTO/AlbumProtectionPolicy.php
ildyria: fixed.

app/DTO/AlbumProtectionPolicy.php
ildyria: fixed.

ildyria:
@nagmat84 I tried to cover all your comments. :)

So far, the only part missing is on `CAN_EDIT_OWN_SETTINGS`.

@kamil4 @qwerty287 @d7415 Enjoy the review. :)
https://github.com/LycheeOrg/Lychee-front/pull/320 < also needed.

ildyria:
Considering closing this, squash and open a new PR in order to have a clean history and remove resolved comments. :')

nagmat84:
> Considering closing this, squash and open a new PR in order to have a clean history and remove resolved comments. :')

I don't mind either way. But in case you decide to "restart", I would not review this PR but wait until you have decided.

ildyria:
> > Considering closing this, squash and open a new PR in order to have a clean history and remove resolved comments. :')
> 
> I don't mind either way. But in case you decide to "restart", I would not review this PR but wait until you have decided.

I'll do that then.

PR #1516, opened 2022-09-10
Fixes #1506

### LINKED ISSUE #1506: 500 HttpException while upgrading Lychee from 4.5.3 to 4.6.0
*A clear and concise description of what the bug is.*

I was in the process of  upgrading Lychee 4.5.3 to 4.6.0 on 7.9.2009 (Core) with PHP 8.0, MYSQL 5.7, Apache 2.4 using the manual update method mentioned in docs : 

https://lycheeorg.github.io/docs/update.html#update-manually

However after migrating the user and successful migration completion . 

When I visit the URL  i.e : 

http://domain.com/{{lychee-folder}}/public/gallery

The following is thrown i.e : 

`500 HttpException Undefined variable $locale (View: /{{PATH}}/{{TO}}/{{LYCHEE}}/resources/views/gallery.blade.php)`

Any way to resolve this? . Is it a known issue?

1. Try to upgrade Lychee using the following docs https://lycheeorg.github.io/docs/update.html#update-manually
2.  Migrate user entering the user credentials.
3.  Migration successful
4.  Visit the URL http://domain.com/{{lychee-folder}}/public/gallery
5. `500 HttpException Undefined variable $locale (View: /{{PATH}}/{{TO}}/{{LYCHEE}}/resources/views/gallery.blade.php)` is thrown.

**Screenshots**
image

Output from `php artisan lychee:diagnostics`

```
System Information
------------------
Lychee Version (release):        4.6.0
DB Version:                      4.6.0

PHP Version:                     8.0.16
Max uploaded file size:          20M
Max post size:                   8M
Max execution time:              0
MySQL Version:                   5.7.39

Imagick Available:               -
Imagick Version:                 -
GD Version:                      bundled (2.1.0 compatible)

Config Information
------------------
version:                         040600
sorting_albums_col:              max_taken_at
skip_duplicates:                 0
landing_owner:                   John Smith
landing_title:                   John Smith
landing_subtitle:                Cats, Dogs & Humans Photography
landing_facebook:                https://www.facebook.com/JohnSmith
landing_flickr:                  https://www.flickr.com/JohnSmith
landing_twitter:                 https://www.twitter.com/JohnSmith
landing_instagram:               https://instagram.com/JohnSmith
landing_youtube:                 https://www.youtube.com/JohnSmith
landing_background:              dist/cat.jpg
site_title:                      Lychee v4
site_copyright_begin:            2019
site_copyright_end:              2019
additional_footer_text:
upload_processing_limit:         4
zip_deflate_level:               6
```

### Browser and system

Mozilla Firefox 104.0.1 (64-bit)
CentOS Linux release 7.9.2009 (Core)

nagmat84:
Fixes #1506.

Occasionally, updating Lychee fails because users explicitly cached the Lychee/Laravel configuration, forget about it and after a package update strange things start to happen, because the Laravel caches still contain outdated configurations or source files.

This PR prevent this from happening by running `artisan optimize --clever --dont-ask=assume-no` during update.

This PR also puts the original `optimize` command provided by the framework on steroids. The extended `optimize` command allows to only build a cache if the user has explicitly chosen to do so in the past. From the comment above the new command:

```php
/**
 * Improves the original "optimize" command provided by the framework.
 *
 * There are three improvements:
 *
 *  1) This command explicitly clears the old cache before building a new one.
 *     Note that rebuilding a new cache is not sufficient, because this
 *     does *not* overwrite the entire cache but some leftovers may remain.
 *     This actually looks like an oversight and bug in the original command.
 *  2) This command adds a "clever" mode which only rebuilds a new cache if
 *     a previous cache has existed.
 *     We use this in our install/update scripts in order to rebuild the cache
 *     after installation/update without enforcing to use a cache for everyone.
 *  3) This command adds a confirmation, if the user requests to build a cache
 *     for non-productive environments as this is most likely an error and
 *     undesired.
 *     The confirmation can be skipped by pre-selecting the answer via a
 *     command line option.
 */
```


ildyria:
You could check the existence of `vendor/bin/phpunit` or `vendor/bin/phpstan` to determine if dev is being used or prod. :)

nagmat84:
I know, I could, but I actually don't feel like doing it. What I meant is, there is no option to do it _inside_ Composer using the means offered by Composer. 

First of all, it would require just another script and I feel like we already have too many scripts (in too many different script languages) and I don't want to add another one onto that pile. Secondly, using the existence of a particular file which "accidentally" happens to exist as a development dependency as an indicator whether Composer is in development mode or not feels kind of nasty workaround. In particular, it feels like something which might break at some point and then nobody knows why.

So, yes I wanted to fix #1506 with as little complications as possible. I don't mind if this PR gets dropped and #1506 stays open or gets closed as "won't fix". In particular, as the users brought themselves into that situation in the first place. I also don't mind if someone extends this PR and fixes #1506 in a better way.

kamil4 (approved):
LGTM (untested).

kamil4:
At least the `post-merge` is a script already. Does `php artisan optimize` create some files with predictable names? We could run `optimize:clear` followed by `optimize` only if the files in question already exist...

nagmat84:
Actually, `opimize` is a meta command to cache the routes, the configuration and the views. We could probably check the existence of

 - `$this->laravel->getCachedConfigPath()` for the cached configuration (see [`Illuminate\Foundation\Console\ConfigCacheCommand:60`](https://github.com/laravel/framework/blob/762b917f0bc57f9e9280980e1df037064ab49659/src/Illuminate/Foundation/Console/ConfigCacheCommand.php#L60)), and
- `$this->laravel->getCachedRoutesPath()` for the cached routes (see [`Illuminate\Foundation\Console\RouteCacheCommand:66`](https://github.com/laravel/framework/blob/762b917f0bc57f9e9280980e1df037064ab49659/src/Illuminate/Foundation/Console/RouteCacheCommand.php#L66))

inside our own console command. Checking for compiled views seems to a little bit more complicated as there does not seem to be a single file, but at least one file for each compiled view.

app/Console/Commands/Optimize.php
```diff
@@ -0,0 +1,86 @@
...
+	 * @return void
+	 *
+	 * @throws ConsoleException
+	 * @throws BindingResolutionException
+	 * @throws NotFoundExceptionInterface
+	 * @throws ContainerExceptionInterface
+	 */
+	public function handle(): void
+	{
+		$shallBeClever = $this->hasOption('clever') && $this->option('no-clever') === true;
```
ildyria: `$this->option('no-clever')` ??

nagmat84: I was coding while riding the train and I had to switch connection so I was in a hurry to commit and push my changes. 😆

nagmat84:
@ildyria and @kamil4: You have already approved the PR but I would like to ask you to re-review it again as it has changed significantly. As suggested by @kamil4 in https://github.com/LycheeOrg/Lychee/pull/1516#issuecomment-1243054247 and discussed in https://github.com/LycheeOrg/Lychee/pull/1516#issuecomment-1244006596 I decided to improve the original Laravel command `optimize`.

ildyria (approved):
LGTM.

app/Console/Commands/Optimize.php
```diff
@@ -0,0 +1,112 @@
...
+
+	/**
+	 * Checks whether Lychee is running in a non-production environment.
+	 *
+	 * Note, this method deliberately tends to `true` in case of doubt.
+	 * This means if anything indicates that the setup might be used for
+	 * developing or testing purposes, the result is `true`.
+	 * Such indicators are the environment setting, enabled debug mode or
+	 * debug bar, installed PhpUnit or PhpStan.
+	 * If we are not in production mode, this command ask for confirmation,
```
kamil4:
```suggestion
	 * If we are not in production mode, this command asks for confirmation,
```

app/Console/Commands/Optimize.php
```diff
@@ -0,0 +1,112 @@
...
+	 * and we rather ask one time too often than not.
+	 *
+	 * @return bool `true`, if the
```
kamil4: Unfinished sentence.

app/Console/Commands/Optimize.php
```diff
@@ -0,0 +1,112 @@
...
+		$hasPreviousCache = file_exists($this->laravel->getCachedConfigPath()) || file_exists($this->laravel->getCachedRoutesPath());
+
+		$this->call('optimize:clear');
+
+		if ($shallBeClever && !$hasPreviousCache) {
+			return;
+		}
+
+		if ($this->isNonProductive()) {
+			$this->alert('Application not in Production!');
```
kamil4, line 71:
```suggestion
			$this->alert('Application not in production!');
```

Any particular reason for that capitalization?

app/Console/Commands/Optimize.php
```diff
@@ -0,0 +1,112 @@
...
+
+			$hasConfirmed = $confirmationDefault ?? $this->confirm('Do you really wish to run this command?');
+
+			if (!$hasConfirmed) {
+				$this->comment('Command Canceled!');
```
kamil4, line 76:
```suggestion
				$this->comment('Command canceled!');
```

Same question...

composer.json
```diff
@@ -106,6 +106,7 @@
         ],
         "post-autoload-dump": [
             "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
+            "@php artisan optimize --clever --dont-confirm=assume-no",
```
kamil4, line 109: Will this command be run if updating via GUI? If so, what happens with the prompt then?

scripts/post-merge
```diff
@@ -26,6 +26,8 @@ else
         composer install --no-dev --prefer-dist
     fi
     if [ -f ".env" ]; then
+        echo "php artisan optimize --clever --dont-confirm=assume-no"
```
kamil4, line 29: Same question: will this command be run if updating via GUI? If so, what happens with the prompt then?

kamil4 (approved):

app/Console/Commands/Optimize.php
nagmat84: Yes, to be consistent with the original Laravel Artisan commands. See https://github.com/laravel/framework/blob/abd4787bcaf0bb55dd4be35bf15f8985ee06bbcc/src/Illuminate/Console/ConfirmableTrait.php#L16.

But I had the same thought as you.

app/Console/Commands/Optimize.php
nagmat84: Same argument as in https://github.com/LycheeOrg/Lychee/pull/1516/files#r982537576.

composer.json
nagmat84: Yes, the command in run if updating via the GUI, as we must clear the cache in that case as well. I would say, we have to do that in that case _in particular_, because users who run the Web Updater are probably more likely unable to clean the cache via CLI manually.

The prompt is skipped because we use `--dont-confirm=assume-no`. Actually, that is the whole purpose why I added this extension to the original command such that we can skip the prompt for non-interactive use.

scripts/post-merge
nagmat84: See https://github.com/LycheeOrg/Lychee/pull/1516/files#r982539157.

app/Console/Commands/Optimize.php
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee/pull/1516/commits/b2bd466a07a99b79b721a5ef91e01b6f01696491.

composer.json
kamil4: I missed that that option is effectively a tri-state. I _was_ wondering about the weird naming; now I get it.

PR #1519, opened 2022-09-15
Adopt new modern dialog

nagmat84:
This is mostly the frontend sync for (https://github.com/LycheeOrg/Lychee-front/pull/322.

On top it allows for easier and efficient queries of shares. Previously, `Share::list` returned all existing shares and the fronend had to sort out what it really needed. Now, it is possible to optionally sent a user ID or album ID to `Share::list` and the filtering is already performed on the backend. This is not only much more efficient, but also simplifies the frontend code.


ildyria (approved):
LGTM. Untested.

app/Http/Requests/Contracts/HasOptionalUser.php
```diff
@@ -0,0 +1,23 @@
...
+	 *
+	 * This method is called `user2`, because Laravel already defines
+	 * {@link \Illuminate\Http\Request::user()} which returns the user which
+	 * is currently authenticated within the HTTP session.
+	 * This method returns another user object which is explicitly part of the
+	 * request.
+	 *
+	 * @return User|null
+	 */
+	public function user2(): ?User;
```
kamil4, line 22: I'll sound like @nagmat84 but this name (`user2`) strikes me as ugly. Shouldn't it be called `optionalUser` or `requestUser` or some such?

kamil4 (approved):
I'm basically OK with this part (untested). I still need to go through the front end changes, though.

app/Http/Requests/Contracts/HasOptionalUser.php
ildyria: rather than `optionalUser` or `requestUser` I would go for `modifiedUser`.

- _optional_ does not make sense on the `HasUser` interface which extends this one.
- request still induces ambiguity: is it the user _doing_ the request or the user _modified by_ the request?

nagmat84: The term `user2` hasn't been introduced by this PR. It has been there before. Of course, this is not an argument to not try to improve it, but I don't have any good idea. @ildyria already gave the same thoughts I had.

We could call it `userToBeModified` (very clumsy, I don't like it, but unambitious). `additionalUser` but I don't think this is better than `user2`.

nagmat84: @kamil4 If your are fine with this, you should resolve the conversation.

PR #1522, opened 2022-09-19
Merge `gallery`, `view` and `frame` Into a Unified Frontend

nagmat84:
The primary purpose of this PR is to sync https://github.com/LycheeOrg/Lychee-front/pull/323, but is also has some merits on its own:

 - Cleans up some of the configuration options
 - Removes and simplifies everything which is not needed anymore on the backend, because the frontend takes care of those things on its own
 - It also removes the left-over of the "page" mode which has already been removed by https://github.com/LycheeOrg/Lychee/pull/1489, but failed to clean up completely.


ildyria:
I am not sure yet about that way of dropping `/frame`, but I am curious of what it will look like.

resources/views/landing.blade.php
```diff
@@ -1,55 +1,54 @@
...
+					<li>
+						<div class="overlay"></div>
+						<img src="{{ $infos['background'] }}" alt="">
+					</li>
+				</ul>
+			</div>
+		</div>
+		@include('includes.footer')
+		<script type="text/javascript" src="{{ Helpers::cacheBusting('dist/landing.js') }}"></script>
+		<script src="dist/Larapass.js"></script>
```
ildyria: This one can be removed: Larapass will never be used in the landing page. ;)

resources/views/landing.blade.php
```diff
@@ -1,55 +1,54 @@
...
-            <ul>
-                <li>
-                    <div class="overlay"></div>
-                    <img src="{{ $infos['background'] }}" alt="">
-                </li>
-            </ul>
-        </div>
-    </div>
-
-    <div id="home_socials" class="animate animate-up">
```
ildyria, line 50: nuking socials from the landing page?

ildyria: My mistake, they are now in the footer. :+1:

resources/views/landing.blade.php
nagmat84: Correct, I will remove it. (Although it has always been there.)

nagmat84: Fixed in https://github.com/LycheeOrg/Lychee/pull/1522/commits/2f61e97d1f057a8559bcc6233ec54ce8dc353caa.

public/frontend.html
```diff
@@ -0,0 +1,375 @@
+<!DOCTYPE html>
```
qwerty287: Why is this not in `dist`? It's auto-generated, so I'd say it should be there.

nagmat84: No strong opinion about that. My idea was that is plays the same role for the frontend as `index.html` for the backend. Both serve as the entry point to the respective application. Hence, I thought the files should be on the same hierarchy level.

I must admit, that I don't understand the systematics of our `public` folder at all. (Maybe there isn't one.) Firstly, we have folder `css`, `fonts`, `img`, `js`. Hence one could assume that the files are categorized according to their type. Secondly, we have a folder `installer` which contains all assets (styles, fonts and images) for the installer. This implies an organization according to the "module". But the installer is the only one which is organized like that. Thirdly, we have folders `dist` and `src`. `dist` contains nearly everything of the frontend. `src` contains JS files whose purpose I don't understand.

So if anyone could shed some light onto how it is _supposed_ to be organized, I am happy to move the file where ever it belongs.

qwerty287: The files in `css` and `js` are the compiled variants for Livewire, I think. The .gitignore of `src` states: `Ignore everything in this directory, it's a placeholder for js source files needes for js.map files of leaflet.Markercluster plugin`.

But I agree with you. We need to find a better solution for this `public` folder, but I would put this file into `dist` because it's generated together with the JS / CSS files.

database/migrations/2022_09_18_143201_make_wui_settings_public.php
```diff
@@ -0,0 +1,223 @@
...
+			->where('key', '=', 'site_copyright_begin')
+			->update(['confidentiality' => 0]);
+		DB::table('configs')
+			->where('key', '=', 'site_copyright_end')
+			->update(['confidentiality' => 0]);
+
+		// Remove NSFW text setting which is replaced by a localized text
+		DB::table('configs')
+			->where('key', '=', 'nsfw_warning_text')
+			->delete();
```
kamil4: I'm not sure about this one. I mean, it's nice that the _default_ is localized now; that's an improvement. But you are effectively taking away from the users the ability to customize that text to their liking. We have received at least one report from somebody who found the available space (200 characters?) too small, so clearly there is interest in customizing it.

I would suggest to keep a config option (perhaps renamed) but make it empty by default, in which case the localized text is used.

app/Http/Controllers/ViewController.php
```diff
@@ -18,33 +15,17 @@ class ViewController extends Controller
...
-			return view('view', [
-				'locale' => $lang,
-				'url' => $url,
-				'photo' => $photo,
-				'picture' => $picture,
-				'title' => $title,
-				'rss_enable' => $rss_enable,
-			]);
-		} catch (BindingResolutionException|ContainerExceptionInterface $e) {
+			return redirect('/#view/' . $photo->id);
```
kamil4: I would like to have @ildyria's input on this. When I first asked -- years ago -- about why we even have this separate "view" mode, I was told it was to support linking to Lychee content from Facebook, Twitter, and such. From what I remember (I probably forgot at least some details), Twitter doesn't like `#` in the URLs. Importantly, the old view blade also statically embeds metadata in HTML: https://github.com/LycheeOrg/Lychee/blob/b4cbfc4fe4aeab21b95ad14df666960ac2b2be9a/resources/views/view.blade.php#L12-L29
This is important because social media sites won't run the JS of the page. I don't see any code in this PR that would do anything of this sort, and viewing the network traffic in the JS browser console I see a redirect followed by the loading of `frontend.html`.

routes/web.php
```diff
@@ -43,5 +43,4 @@
 
 Route::get('/view', [ViewController::class, 'view'])->name('view')->middleware(['redirect-legacy-id']);
 Route::get('/demo', [DemoController::class, 'js']);
-Route::get('/frame', [FrameController::class, 'init'])->name('frame')->middleware(['migration:complete']);
-
+Route::redirect('/frame', '/#frame')->name('frame')->middleware(['migration:complete']);
```
kamil4: This absolute redirect prevents the frame mode from working for me (I don't have Lychee installed at the root URL). I'll push a fix for this (and a related issue on the landing page).

kamil4 (dismissed):

app/Http/Controllers/ViewController.php
ildyria: @kamil4 there is no possibility at the moment to support such metadata. While it is a nice to have for photos, I think it is even more important for albums. That is for me one of the strongest argument in favor of Livewire front-end.

kamil4: Yes, what we currently have is incomplete, but isn't this PR moving us a step _back_ in that respect? Personally I don't use the view mode for anything so I don't care but I worry if there may be many users that will be negatively affected by such a change?

database/migrations/2022_09_18_143201_make_wui_settings_public.php
nagmat84: You got me. I know there was another thing which I wanted to pull out of this PR as an individual PR and this one was it.

Actually, I had a totally different reason why I removed this config option. That the NSFW text is now localized is just a nice side effect. I am aware of the problem that this takes away a feature and why I wanted it to make an indepedent PR.

In order to avoid code injection attacks, I attempted to only use the JS property `textContent` and not `innerHTML`. Previously, this configuration option contained HTML code like `<h1>...</h1><p>...</p><p>...</p>`. I wanted to get rid of that. That is also why the locale file now contains three distinct entries: one with the text content for each of these HTML elements.

The current HTML code looks like this
```html
<!-- NSFW Warning -->
<div id="sensitive_warning" class="overlay-container"><h1></h1><p></p><p></p></div>
```
and already predefines the three child elements. The locale defines three items
```php
'NSFW_BANNER_HEADER' => 'Sensitive content',
'NSFW_BANNER_EXPL' => 'This album contains sensitive content which some people may find offensive or disturbing.',
'NSFW_BANNER_CONSENT_TAP' => 'Tap to consent.',
```
and the JS code looks like
```js
const nsfwBanner = document.querySelector("#sensitive_warning");
nsfwBanner.children.item(0).textContent = lychee.locale["NSFW_BANNER_HEADER"];
nsfwBanner.children.item(1).textContent = lychee.locale["NSFW_BANNER_EXPL"];
nsfwBanner.children.item(2).textContent = lychee.locale["NSFW_BANNER_CONSENT_TAP"];
```

If we keep the configuration option, then I assume it should only be a single configuration option as it has been before. (I do not assume that we would make three configuration options out of it.) I further assume this single configuration option should probably support HTML markup in order to provide the most flexibility. This could be a potential gateway for injection attacks although one can argue that the configuration option is managed by administrators and administrators should now what they are doing. But in theory, an administrator could put a link to a JS script file into that configuration option in good faith and then the JS script might be susceptible to attacks putting the whole Lychee application at risk.

Of course, alternatively, we could provide _three_ distinct configuration options (one for the header, one for the text, one for the action) which only allow to change the text, but which are otherwise restricted to the given structure.

app/Http/Controllers/ViewController.php
nagmat84: There are two aspects which we need to distinguish here.

> [The separate "view" mode] was to support linking to Lychee content from Facebook, Twitter, and such. [...] Twitter doesn't like # in the URLs.

That is correct, but it still works. The "external" link is still `http://<base-url>/view?p=<id>`, hence the URL which a user would publish on Twitter is without a hashtag `#`. A request to this link is then redirected to `http://<base-url>/#view/<id>`. This is fine for Twitter & Co. It is the same kind of redirection an from `/frame` to `/#frame`, see your comment below.

I.e., except for the issue that I probably forgot to consider Lychee installations at a non-root level, we should be fine here.

> Importantly, the old view blade also statically embeds metadata in HTML. [...] This is important because social media sites won't run the JS of the page. [...] Yes, what we currently have is incomplete, but isn't this PR moving us a step back in that respect?

You are right and I had the same concerns as you, but I was unable to find a good solution for that problem. Hence, I simply gave up. Honestly, I have no idea how modern single-page JS web applications solve this problem. Maybe the old predication that social media and search engines don't run JS isn't true anymore? (I haven't done a web search for it, it is just a guess.)

A week back when I had a look at Webpack to check out whether Webpack could be a replacement for Gulp, I discovered something totally freaky. An front JS application built with Webpack does _everything_ (I mean literally _everything_) in JS. Webpack only generates the most basic skeleton for a HTML page even without any `<link>` in the header to CSS and then does _everything_ in JS. Even if your web application uses SCSS or SASS which is compiled into CSS stylesheets, Webpack takes that, goes one step further and transforms it into JS code. For me this seems totally crazy. Have a look here: https://webpack.js.org/concepts/loaders/#configuration In order to include SASS into your Webpack application, three loaders are involved: a) a SASS loader which transforms SASS into CSS, b) a CSS loader which transforms CSS into something intermediate, and c) a "style" loader which generates JS code. In https://webpack.js.org/concepts/loaders/#loader-features it clearly says, _loaders can be chained. Each loader in the chain applies transformations to the processed resource. [...] Finally, webpack expects JavaScript to be returned by the last loader in the chain._ I also took the Guide https://webpack.js.org/guides/ until and including Asset Management. In https://webpack.js.org/guides/asset-management/#loading-css you find this

> Open up `dist/index.html` in your browser again and you should see that _Hello webpack_ is now styled in red. To see what webpack did, inspect the page (**don't view the page source, as it won't show you the result**, because the `<style>` tag is **dynamically created by JavaScript**)

Sorry, for the detour. But I thought it would be nice to know. By the way, same holds for the `meta` tags on a page for a Webpack application. A page generated with Webpack does not contain any static meta information at all, but everything is generated dynamically. Back to the problem at hand. I am willingly and open to implement the same thing for our current frontend. It shouldn't be too complicated to update the `<meta>` tags in the HTML header dynamically upon loading an image or album. We already have to deal with the toolbar and other stuff, hence swapping `<meta>` tags dynamically just fits in.

Yes, I also have been under the impression that web crawlers (search engines, social media platforms, etc.) do not execute JS code but only have a look at the static HTML code. I don't know how they do it, but somehow it seems to work. Unfortunately, I don't have any hard evidence or can confirm that.

> That is for me one of the strongest argument in favor of Livewire front-end.

Well, Livewire is built on Webpack. Hence, this is doing the same crazy stuff. You don't get a static HTML page, but everything is loaded on-demand and everything is styled, manipulated, etc. during JS runtime.

nagmat84: I guess I found the answer. The source is not very reliable, because it is from a manager magazine on SEO (search engine optimization), but it makes somehow sense and fits into the picture.

Since 2019, Google and Bing run every page through Google's V8 engine and _execute all JS code_ before analyzing the page and before caching a rendered a preview of the page. The take away message from that manager magazine is that you must not forbid search engines to access your JS code as otherwise they were not able to correctly analyze your SPA. I I guess that explains it.

However, I don't know what Twitter & Co do, but I assume they do similar things as the face the same challenges.

ildyria: > Well, Livewire is built on Webpack. Hence, this is doing the same crazy stuff. You don't get a static HTML page, but everything is loaded on-demand and everything is styled, manipulated, etc. during JS runtime.

As far as I know, the first page is loaded from html. The rest however is loaded on demand, but I may be wrong.

nagmat84: Here is another somewhat shady source https://www.impactplus.com/blog/google-publishes-new-developer-doc-on-javascript-seo-basics with a reference to a twitter message from Google also dating back to July 2019. The page says

> And with how widespread front-end JavaScript frameworks and Single Page Applications (SPAs) are becoming, this guide update is Google’s attempt to clarify how best to make your site search engine optimized.
> [...]
> You may not think of your website as being a “web app” [...]. But many web services and coding frameworks are now bringing functionality like pre-fetching links (so linked pages can load faster) and only reloading content that changes between pages (again — loading pages faster!). [...] These have typically been in the realm of “web applications” in the past, but they carry a lot of benefits for standard websites as well. So, unless your website is incredibly barebones, it can probably be considered a web app.

And then an explanation follows what Google is doing.

nagmat84: Found the Google reference itself: https://developers.google.com/search/docs/crawling-indexing/javascript/javascript-seo-basics#titles-and-snippets

> Unique, descriptive `<title>` elements and helpful meta descriptions help users to quickly identify the best result for their goal and we explain what makes [good <title> element](https://developers.google.com/search/docs/appearance/title-link#page-titles) and [meta descriptions](https://developers.google.com/search/docs/appearance/snippet#meta-descriptions) in our guidelines.
>
> **You can use JavaScript to set or change the meta description as well as the `<title>` element.**

nagmat84: Put everything into independent issues:

 - https://github.com/LycheeOrg/Lychee-front/issues/341
 - https://github.com/LycheeOrg/Lychee-front/issues/342
 - https://github.com/LycheeOrg/Lychee-front/issues/343

nagmat84: https://github.com/LycheeOrg/Lychee-front/pull/323/commits/ef5535132726f65cbb330dfae2ce84ccfd9f0bff on the frontend should fix the first issue.

database/migrations/2022_09_18_143201_make_wui_settings_public.php
nagmat84: Fixed it in https://github.com/LycheeOrg/Lychee/pull/1522/commits/6b8f27f83a09eea221aee87053d84a66537bdfd6 and https://github.com/LycheeOrg/Lychee-front/pull/323/commits/b6b162e13f9ce0212c278a32381e44bc060753c2 at the risk of a potential code injection.

public/frontend.html
nagmat84: Fixed it in https://github.com/LycheeOrg/Lychee/pull/1522/commits/368ee0e8056235ca40c47cdbc70874cf7f191509 and in https://github.com/LycheeOrg/Lychee-front/pull/323/commits/c170c2f441318754b9ff0ba51c0f637a067eed7e.

app/Http/Controllers/ViewController.php
nagmat84: After I added RSS support back, I noticed that the lack of client-side navigation which uses URLs with proper paths instead of fragments has more severe effects than I hoped for (@kamil4 you were absolutely right with your concerns). In particular, Twitter does not seem to handle the redirection very well. Hence, I had to (partially) revert some of the changes of this PR in https://github.com/LycheeOrg/Lychee/pull/1522/commits/f3f2f6c9ea82c46a27598fc622ce57cd440add68 and https://github.com/LycheeOrg/Lychee-front/pull/323.

 - For the time being we need to keep the different HTTP controller methods for `view` and `frame` to mimick URLs with proper paths on the backend. So you should have another look at [`app/Http/Controllers/IndexController.php`](https://github.com/LycheeOrg/Lychee/blob/merge_modes/app/Http/Controllers/IndexController.php). Instead of redirecting to URLs with fragments, the frontend is served directly.
 - I had to re-introduce a Blade template [`/resources/views/frontend.blade.php`](https://github.com/LycheeOrg/Lychee/blob/merge_modes/resources/views/frontend.blade.php). Nonetheless, this still simplifies matters a lot. Only the HTTP headers a pre-rendered by the backend, the actual body is still provided by the frontend repository
 - I had to extend the "god method" [`lychee.load`](https://github.com/LycheeOrg/Lychee-front/blob/merge_modes/scripts/main/lychee.js#L905-L924) (in the frontend) to be even more god-like in order to partially add support for client-side routing based on the path component (required for `/frame` and `/view`). Proper support for path-based, client-side routing is something which I had planned for a later PR, but now it has became necessary earlier. Frankly, I "hacked" the two necessary edge cases for `/frame` and `/view` into `lychee.load` more or less, but it works at last.

routes/web.php
nagmat84: I assume this conversation has become obsolete due to https://github.com/LycheeOrg/Lychee/commit/f3f2f6c9ea82c46a27598fc622ce57cd440add68 and https://github.com/LycheeOrg/Lychee-front/pull/323, see https://github.com/LycheeOrg/Lychee/pull/1522#discussion_r1002749404.

ildyria (approved):
LGTM, tested.

ildyria:
@kamil4  Need your checks. :)
@nagmat84 resolved conflicts.

PR #1525, opened 2022-09-22
Two small (but important) fixes

nagmat84:
This PR fixes to bugs I came across while I had been working on https://github.com/LycheeOrg/Lychee/pull/1522.

 1. A newer version of some package does not call the method `User::name()` on the user model anymore, but assumes, that `name` is an attribute and calls `User::$name`. Hence, we need the accessor function `User::getNameAttribute()` to please Laravel.
 2. The migration `2019_10_03_214750_frame_refresh_in_sec` does not use DB methods only, but depends on particular productive code which breaks when this productive code changes.

**Appeal**

We should really stop to use production code in our migrations. That is a bad thing to do. The migrations will exist "forever" in some sense and even very old migrations must still be runnable in the future. If they use production code and make certain assumptions about that, this means we can never change that code without breaking those migrations. And nobody feels like fixing ancient migrations.

This particular issue took me two hours to debug in the step debugger. I encountered really weird and seemingly unrelated error messages like `ConfigurationMissingException: No such key map_provider` for a lot of seemingly random configuration settings. I then noticed, that I only had about 60 configuration items in my test DB instead of the usual 90. "Something" was deleting about 30 configuration items. Then I found out that it happened during the Unit Tests when the migration was tested and finally I found the wrongdoer in that particular migration.

### DIFF
--- a/app/Http/Livewire/Sidebar.php
+++ b/app/Http/Livewire/Sidebar.php
@@ -84,7 +84,7 @@ class Sidebar extends Component
 			['head' => Lang::get('ALBUM_PASSWORD'), 'value' => $_password],
 		];
 		if ($this->album->owner_id !== null) {
-			$share->content[] = ['head' => Lang::get('ALBUM_OWNER'), 'value' => $this->album->owner->name()];
+			$share->content[] = ['head' => Lang::get('ALBUM_OWNER'), 'value' => $this->album->owner->name];
 		}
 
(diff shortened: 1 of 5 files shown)


ildyria (approved):

kamil4 (approved):
I believe I wrote that migration in the first place. In my defense, that was in 2019, and I simply didn't realize the consequences at the time...

PR #1528, opened 2022-09-25
Improve & fix CSP

ildyria:
After further inspection on the findings of @maggick , it appears that CSP was disabled...
This updates the config file and aims to fix it.


config/secure-headers.php
```diff
@@ -316,47 +316,227 @@
...
+
+		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/manifest-src
+		'manifest-src' => [
+		],
+
+		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/media-src
+		'media-src' => [
+			'self' => true,
+			'allow' => [
+				'blob:',
```
d7415: Do we also want `env('LYCHEE_UPLOADS_URL', 'https://lycheeorg.github.io/'),` here?

config/secure-headers.php
```diff
@@ -316,47 +316,227 @@
...
+		'frame-ancestors' => [
+		],
+
+		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src
+		'frame-src' => [
+		],
+
+		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src
+		'img-src' => [
+		],
```
d7415: This appears to be redefined below. If I'm not missing something we should either remove this or merge them here.

config/secure-headers.php
```diff
@@ -316,47 +316,227 @@
...
+		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/manifest-src
+		'manifest-src' => [
+		],
+
+		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/media-src
+		'media-src' => [
+			'self' => true,
+			'allow' => [
+				'blob:',
+			],
```
nagmat84:
```suggestion
			'allow' => [],
```

I hope this removes the annoying empty `blob:` element from the the header. Currently it is

```
default-src 'self' blob:;
```

The change should give us

```
default-src 'self';
```

config/secure-headers.php
```diff
@@ -316,47 +316,227 @@
...
+
+			// Enable `strict-dynamic` will *ignore* `self`, `unsafe-inline`,
+			// `allow` and `schemes`. You can find more information from:
+			// https://www.w3.org/TR/CSP3/#strict-dynamic-usage
+			'strict-dynamic' => false,
+
 			'hashes' => [
 				'sha256' => [
+					// 'sha256-hash-value-with-base64-encode',
 					'8bLztrDF3NUpheSuvAzpebgX1DpPJEfhmUHKTwGF4qA=',
```
nagmat84: Where does this hash come from? It results into the following header

```
script-src 'self' 'report-sample' 'sha256-8bLztrDF3NUpheSuvAzpebgX1DpPJEfhmUHKTwGF4qA=' https://www.dropbox.com/static/api/1/dropins.js;
```

According to the docs, this is an alternative way to allow a script without denoting its source by giving its hash, i.e. the script with this hash value is allowed. Do we need it? If yes, we should add a comment whose hash value is this and how this hash is re-computed whenever the script to which is hash belongs is updated.

Otherwise we should remove it.

config/secure-headers.php
```diff
@@ -380,38 +560,17 @@
 				'blob:',
```
nagmat84: Same thing here. If we don't use `data:` and `blob:` we should remove it.

config/secure-headers.php
```diff
@@ -380,38 +560,17 @@
...
-		'frame-ancestors' => [],
-		'frame-src' => [],
-		'manifest-src' => [],
-		'media-src' => [
+
+		'default-src' => [
 			'self' => true,
 			'allow' => [
 				'blob:',
 			],
```
nagmat84:
```suggestion
			'allow' => [],
```

nagmat84 (changes requested):

nagmat84:
As @d7415 has already pointed out, there are some duplicate key entries in this config file. This should be cleaned up.

Screenshot_20220925_153536

nagmat84:
And we should sync the current `master` of the frontend with this PR as well such that the latest security fixes https://github.com/LycheeOrg/Lychee-front/pull/324 and https://github.com/LycheeOrg/Lychee-front/pull/325 which have already been merged go live.

config/secure-headers.php
d7415: It was added in https://github.com/LycheeOrg/Lychee/commit/3e6cda2cb725b5f0affbfbbbc68cbe22e7dcb1b4. Not sure what for though.

config/secure-headers.php
ildyria: Removed in [3f5583f](https://github.com/LycheeOrg/Lychee/pull/1528/commits/3f5583f3f4300551dbb90229ed1d7e908202272d)

config/secure-headers.php
ildyria: Merged in [3f5583f](https://github.com/LycheeOrg/Lychee/pull/1528/commits/3f5583f3f4300551dbb90229ed1d7e908202272d)

config/secure-headers.php
ildyria: Removed in [3f5583f](https://github.com/LycheeOrg/Lychee/pull/1528/commits/3f5583f3f4300551dbb90229ed1d7e908202272d)

config/secure-headers.php
ildyria: I tried to remove it, didn't seem to break anything. So I guess it will stay removed.

ildyria: Removed in [3f5583f](https://github.com/LycheeOrg/Lychee/pull/1528/commits/3f5583f3f4300551dbb90229ed1d7e908202272d)

config/secure-headers.php
ildyria: Removed in [3f5583f](https://github.com/LycheeOrg/Lychee/pull/1528/commits/3f5583f3f4300551dbb90229ed1d7e908202272d)

config/secure-headers.php
ildyria: Removed in [3f5583f](https://github.com/LycheeOrg/Lychee/pull/1528/commits/3f5583f3f4300551dbb90229ed1d7e908202272d)

ildyria:
> As @d7415 has already pointed out, there are some duplicate key entries in this config file. This should be cleaned up.
> 
> Screenshot_20220925_153536

Fixed in [887f731](https://github.com/LycheeOrg/Lychee/pull/1528/commits/887f731)

config/secure-headers.php
d7415: I think we need `data:`

d7415: I can't find the source, but there's a gif defined that way in `main.js`

config/secure-headers.php
```diff
@@ -375,43 +390,165 @@
...
+		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src-elem
+		'script-src-elem' => [
+		],
+
+		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src
+		'style-src' => [
 			'self' => true,
-			'allow' => [
-				'blob:',
+			'unsafe-inline' => true, // We sadly need this one
```
nagmat84: We should make the comment more specific. Nobody will know _why_ we need it in two month from now. I would also remove the appraisement ("sadly"). Better something like

```php
// TODO: Set this to `false` after the inline style in <reference here> has been moved to a style file
```

nagmat84 (approved):
Only one slight change, but otherwise it looks good to me. I haven't tested it again, because I am on a business trip and I am reading and typing from my smartphone. So I also skimmed the changes only. But I assume that all issues have been addressed. Hence, I don't want to block this PR with my previous review any longer.

config/secure-headers.php
d7415: Agreed re more specific, but we can't be *that* specific.

Maybe...
```suggestion
			'unsafe-inline' => true, // We sadly need this one due to styles applied by javascript
```

nagmat84: I assumed that only inline styles are problematic as we have in

https://github.com/LycheeOrg/Lychee-front/blob/692dbacabb739756a2027c8f95f7e9b59e93dc16/scripts/main/settings.js#L472-L473

JavaScript should only add or remove style classes as in

https://github.com/LycheeOrg/Lychee-front/blob/692dbacabb739756a2027c8f95f7e9b59e93dc16/scripts/main/settings.js#L220

I thought that this would be fine and does not count as "inline style". If this is indeed the case, there are really only a handful of inline styles which need to be removed. (Actually, I am only aware of the single one above )

d7415: > JavaScript should only add or remove style classes as in

I was basing this on @ildyria's  response on gitter, but there do appear to be some direct references to the `style` property and a few inline classes if you `git grep style` (when at a pc).

nagmat84: OK, I did not want to imply that the reference which I have been aware of was the only one. I only wanted to point out that JavaScript wasn't a general problem as long as JS stuck to only modify style classes.

So maybe something like this

```
// We need this one due to direct styles (not just style classes) applied by javascript
```

config/secure-headers.php
nagmat84: I am not sure what @d7415 wanted to imply with his first comment and am not sure whether we should _add_ `LYCHEE_UPLOADS_URL`. Consider the case that the backend is running on a different server than the upload path resides on and that the images as well as videos (in case of life photos) are hosted below a different domain. (I know, this is currently still not supported.) In this case it does not suffice to allow `'self'` as the origin for images and media.

However, I do agree that `'https://lycheeorg.github.io/'` shouldn't be the default.

config/secure-headers.php
nagmat84: Fixed this one myself in https://github.com/LycheeOrg/Lychee/pull/1528/commits/757a77e7b077955aef2afbd8e4829ce1b38aee20.

config/secure-headers.php
d7415: I meant that it *was* present for images (it's been removed, which I'm not convinced by, but it has) and so the same should apply for media if videos are stored there. I assume the default was set because of the demo, but I think that case should be covered by `self`.

If `LYCHEE_UPLOADS_URL` isn't allowed in either section until we add (proper) support for media on another server, I'm fine with that, but if it's in one it should be in both.

config/secure-headers.php
d7415: Unresolving this until someone can confirm where that `data:` gif is from and that we definitely don't use it.

config/secure-headers.php
nagmat84: Strongly in favor of consistency.

I am mostly indifferent whether it should be added or removed.

d7415: It's currently in neither, so that sounds good enough for both of us.

nagmat84: Then I resolve this one.

config/secure-headers.php
nagmat84: I haven't found any `data:` URL in `/resources` nor `Lychee-front/scripts`. So it is definitely not used by us directly. The `data:` scheme was added by [commit de5da01c](https://github.com/LycheeOrg/Lychee/pull/339) and it commented with

> Allow image to be directly encoded at the img source parameter

See

https://github.com/LycheeOrg/Lychee/blob/de5da01ca5a8e39ef82630ec9e174fb26b12d684/config/secure-headers.php#L380-L386

So it might be internally used by the 3rd-party library which renders the map. The only frontend code which was introduced by the corresponding frontend PR https://github.com/LycheeOrg/Lychee-front/pull/144 and might be suspect to using `data:` URLs is

https://github.com/LycheeOrg/Lychee-front/blob/3b71fe061dceaeaa507a0ad252d7f7351a7e15aa/scripts/main/view.js#L735-L754

This snippet of code renders the tiny map in the sidebar of a geo-tagged photo and adds the direction cone as an map overlay. I guess we can just test whether this part still works if `data:` URI are disabled. If it does, we should be fine.

d7415: Actually it looks like it's only a 1x1px gif. It looks like openstreetmap might use this, and some libraries/etc use it to assist with lazy loading of images.

Based on that I don't mind removing it, so long as we remember that weird behaviour may be because of that. IIRC it should show up in the console when items are blocked by CSP, which may help to confirm.

d7415: Oh, just saw your commit. I'll resolve this :P

d7415 (approved):

PR #1529, opened 2022-09-25
Arrayable DTO using reflection

ildyria:


ildyria:
@nagmat84 
Yes, this makes use of Constructor property promotion.

But I would only use them where it makes sense and does not hinder the readability. Mixing promoted and normal properties as described [here](https://stitcher.io/blog/constructor-promotion-in-php-8#combining-promoted--and-normal-properties) is a big no-no and I would be more than happy to have a phpstan rules which forbids that. :)

app/DTO/ArrayableDTO.php
```diff
@@ -0,0 +1,31 @@
...
+	 *
+	 * @return array the serialized properties of this object
+	 */
+	public function toArray(): array
+	{
+		$result = [];
+		$cls = new ReflectionClass($this);
+		$props = $cls->getProperties(ReflectionProperty::IS_PUBLIC);
+		foreach ($props as $prop) {
+			$result[$prop->getName()] = $prop->getValue($this);
```
nagmat84: I wonder if this works in all cases, in particular when the type of the value is not a primitive type, but an object. Maybe this implementation currently works, because all child classes use primitive types only.

But I would have expected something along the following lines for a general approach:

```php
$value = $prop->getValue($this);
if (is_object($value)) {
  if ($value instanceof Arrayable) {
    $value = $value->toArray();
  } elseif (/* some other class type for which we know how to convert */) {
    // convert
  } else {
    throw new \InvalidArgumentException("We don't know how to convert :-(");
  }
$result[$prop->getName()] = $value;
```

nagmat84 (commented):
I am indifferent about this PR. Except for the one (optional) change request below, I am generally fine with merging it.

I am neither refusing nor approving it. Mostly, because I would like to hear @qwerty287 and/or @kamil4 opinion about implicitly declaring class properties via promoting constructors. See https://github.com/LycheeOrg/Lychee/pull/1511#discussion_r979385789 for some thoughts about that. Even if we forbid the mixed case, I am still not sure if I like it. But that is probably just me getting old.

I will leave approval to @qwerty287 or @kamil4.

qwerty287:
About the implicit property declarations, I don't have a strong opinion if they are not mixed up with explicitly declared ones. From my side, these changes are fine.

app/DTO/ArrayableDTO.php
ildyria: > Maybe this implementation currently works, because all child classes use primitive types only.

I did not consider other cases, otherwise I would have simplied other DTO using collections for example.

ildyria:
Adding the suggestion of @nagmat84 and simplifying some other DTO. :)

app/DTO/ArrayableDTO.php
```diff
@@ -0,0 +1,42 @@
...
+		$props = $cls->getProperties(ReflectionProperty::IS_PUBLIC);
+		foreach ($props as $prop) {
+			$value = $prop->getValue($this);
+			if (is_object($value)) {
+				if ($value instanceof Arrayable) {
+					$value = $value->toArray();
+				} else {
+					throw new LycheeLogicException(sprintf("We don't know how to %s.", get_class($value)));
+				}
+			}
```
nagmat84: I suspect that we need to consider at least one more case to make this class truly general. But again it is an optional CR and I am not sure if we need it. If the value itself is an array its values might not be primitive values but be objects which need to be converted into arrays as well.

So something along the lines

```php
if (is_array($value)) {
	foreach ($value as &$entry) {
		// we pass $entry by reference to convert objects in place
		if(is_object($entry) {
			if ($entry instanceof Arrayable) {
				$entry = $entry->toArray();
			} else {
				throw new LycheeLogicException('...');
			}
		}
	}
}
```

app/DTO/ArrayableDTO.php
```diff
@@ -0,0 +1,42 @@
...
+		$result = [];
+		$cls = new ReflectionClass($this);
+		$props = $cls->getProperties(ReflectionProperty::IS_PUBLIC);
+		foreach ($props as $prop) {
+			$value = $prop->getValue($this);
+			if (is_object($value)) {
+				if ($value instanceof Arrayable) {
+					$value = $value->toArray();
+				} else {
+					throw new LycheeLogicException(sprintf("We don't know how to %s.", get_class($value)));
```
nagmat84: This exception message is missing a verb. Moreover, my proposal "we don't know how" was meant sloppily not as a serious exception message.
```suggestion
					throw new LycheeLogicException(sprintf("Unable to convert %s into an array", get_class($value)));
```

app/DTO/ArrayableDTO.php
ildyria: > I suspect that we need to consider at least one more case to make this class truly general.

I did not consider the "_array of object_" because I consider it to be the responsibility of the array provider if an array is passed directly.

kamil4 (approved):
I was surprised by this PR but that's because I've never heard of constructor property promotion before so I couldn't understand how it works. But since it appears to be a language idiom specifically designed to cut down on the boilerplate code, I see no reason why we _shouldn't_ use it in situations where it indeed makes for fewer lines of code...

qwerty287 (approved):

PR #1540, opened 2022-09-29
add exec check

ildyria:
Add missing safety check on exec.

### DIFF
--- a/app/Models/Extensions/ConfigsHas.php
+++ b/app/Models/Extensions/ConfigsHas.php
@@ -33,15 +33,20 @@ trait ConfigsHas
 		$has_exiftool = self::getValueAsInt('has_exiftool');
 
 		// value not yet set -> let's see if exiftool is available
-		if ($has_exiftool === 2 && Helpers::isExecAvailable()) {
-			try {
-				$cmd_output = exec('command -v exiftool');
-			} catch (\Exception $e) {
-				$cmd_output = false;
-				Handler::reportSafely(new ExternalComponentMissingException('could not find exiftool; `has_exiftool` will be set to 0', $e));
+		if ($has_exiftool === 2) {
+			if (Helpers::isExecAvailable()) {
+				try {
+					$cmd_output = exec('command -v exiftool');
+				} catch (\Exception $e) {
+					$cmd_output = false;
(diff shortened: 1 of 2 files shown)


qwerty287 (approved):

app/Models/Extensions/ConfigsHas.php
```diff
@@ -32,7 +33,7 @@ public static function hasExiftool(): bool
 		$has_exiftool = self::getValueAsInt('has_exiftool');
 
 		// value not yet set -> let's see if exiftool is available
-		if ($has_exiftool === 2) {
+		if ($has_exiftool === 2 && Helpers::isExecAvailable()) {
```
nagmat84: I am not sure, if this is the way how it should be. With this change, `$has_exiftool` independently remains `2` if `exec` is not available.

Personally, I don't know why we have a configuration setting `$has_exiftool` at all and why we do not simply auto-detect the availability of `exiftool` each time we need it. However, we do have the configuration option `$has_exiftool` and it is designed to cache the detected environment once. Hence, we should stay in that line unless we explicitly decide against it.

This being said I would have expected something like

```php
if ($has_exiftool === 2) {
    if (Helpers::isExecAvailable()) {
        // try to detected exiftool as before
    } else {
        $has_exiftool = 0;
    }
    // save detected setting for `$has_exiftool` as before
}
```

nagmat84 (changes requested):

app/Models/Extensions/ConfigsHas.php
kamil4: As a developer I _do _ find `$has_exiftool` (as well as `$has_ffmpeg` and `$imagick`) very useful as they let me easily test different code paths. That said, perhaps the names of these variables are misnomers. Something like `$use_exiftool` would better reflect what's actually being meant.

nagmat84: > As a developer I _do _ find `$has_exiftool` (as well as `$has_ffmpeg` and `$imagick`) very useful as they let me easily test different code paths.

Agreed, but this is neither an argument for nor against my change request.

ildyria: Fixed [8bd5599](https://github.com/LycheeOrg/Lychee/pull/1540/commits/8bd5599f974218dfed8bde2e7ad0cfea61944b87)

nagmat84 (approved):
Not tested but LGTM.

PR #1542, opened 2022-10-02
i18n(EN) grammar and syntax detected by Grammarly

joebordes:


d7415 (approved):
Some good, some I don't care, some I mildly disagree with, but it looks like a general improvement.

qwerty287 (approved):

PR #1494, opened 2022-08-28
Consolidate localization

nagmat84:
While I have been working on some other parts, I realized that our localization was out-of-sync (mostly between the frontend and the backend). For example the frontend used `USER_DELETED` but the backend provided a constant `USER_REMOVED`. Hence, the localization did not work. Then I realized, that the order of variables was not consistent between the different languages, etc., pp. I freaked out and so I decided to clean it up from the scratch.

 * _All_ localization files of the backend `app/Locale` a are now 513 lines long and have the identical constants in identical positions :nerd_face: 
   _Note:_ I did not aim for a logic or semantic ordering, I mostly took the English localization as the ground truth and re-sorted the other files accordingly. There does not seem to be a reasonable ordering, so be it.
 * I checked which constants are actually used by the frontend but have missing either in the backend or the frontend, I added these constants
 * I checked which constants are were not used anymore, I removed these variables.

While I was on it, I improved typography as well :nerd_face: :

 * Proper quotation marks for the various languages, cp. https://en.wikipedia.org/wiki/Quotation_mark#Summary_table
 * Correct question and exclamation marks in Chinese; Chinese uses fixed-width marks which have different unicode char point than western symbols
 * Proper apostrophe to indicate left-out characters
 * Correct French spacing with non-breaking narrow space before '?' and '!' as well as non-breaking normal-width space before ':'
 * Correct horizontal ellipsis instead of three dots

PR #1546, opened 2022-10-03
Russian translations

sabian:
A little bit of translations into Russian


nagmat84:
Thanks for your contribution! This is highly appreciated. Unfortunately, there is another PR https://github.com/LycheeOrg/Lychee/pull/1494 pending which cleans up a lot of the translations. Moreover some of the original English string have changed. This means some of your work might be lost.

For example, currently we have

```php
'PHOTO_HIDDEN_EXPL' => 'Only people with the direct link can view this photo.'
```

but this will change to

```php
'PHOTO_HIDDEN_EXPL' => 'Anonymous users need a direct link to view this photo.'
```

to make it more precise.  I can read Russian and hence I don't know what your translation

```php
'PHOTO_HIDDEN_EXPL' => 'Только люди, у которых есть прямая ссылка, могут просмотреть эту фотографию.',
```

says literally, but I guess it is based on the old version.

May I ask you to have a look at https://github.com/LycheeOrg/Lychee/pull/1494 and create a PR against that PR?

PR #1551, opened 2022-10-08
Add csp exceptions

kamil4:
Allows a handful of inline event handlers and a single inline script via CSP hashes.  I'm guessing that at least `upload.check()` won't be needed when @nagmat84's improved basicModal gets merged.

Removes one inline script that hasn't been needed for a good while, actually.

While with these changes Lychee works as expected again under both Firefox and Chrome, in either case there are warnings in the JS console. Firefox doesn't recognize the `unsafe-hashes` keyword, which is needed to get Chrome to work as expected. Chrome, on the other hand, complains about a number of permission policies: `ambient-light-sensor`, `battery`, and more. I can't help feeling that we are perhaps operating too close to the bleeding edge when it comes to our use of CSP...

Also updates php-exif to 0.7.12 in `composer.json` (it was already that way in `composer.lock`).

### DIFF
--- a/resources/views/update/error.blade.php
+++ b/resources/views/update/error.blade.php
@@ -103,6 +103,8 @@
             </form>
         </div>
     </div>
+	<!-- Do not change even a single character in the block below without
+	     also updating the checksum in config/secure-headers.php! -->
 	<script>
 	document.addEventListener("DOMContentLoaded", function(event) {
 		document.querySelector("form").addEventListener("submit", function(e){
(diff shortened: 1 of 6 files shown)


resources/views/update/error.blade.php
```diff
@@ -103,6 +103,8 @@
             </form>
         </div>
     </div>
+	<!-- Do not change even a single character in the block below without
+	     also updating the checksum in config/secure-headers.php! -->
```
ildyria, line 107: :+1:

ildyria (approved):
Tested & LGTM

PR #1539, opened 2022-09-29
Apply rights naming convention

ildyria:
~~This PR builds on top of ~~#1508~~ #1529 so merging this one will automatically merge the previous one.~~

#### Content: 

- **Drops the Admin Middleware**

- move authorization to **Gates on Requests**

- Increase number of Policy & granularity of Rights

- creates DTO for:
  -  user rights
  - settings rights
  - photo rights
  - album rights
  - album protection policy
  - user (for admin editing)

- renames capabilities with regard to #1481

- drops visibility for sharing links per album in favor of the already existing global setting.

- ~~adds stricter tests on Locale and remove extra keys in other languages than English~~

##### TODO (maybe future PR?) :

- Photo protection policy ?
- make use of rights in front end (currently mostly depending of admin/user can upload/nothing).

Closes https://github.com/LycheeOrg/Lychee/issues/1363

PR #1557, opened 2022-10-16
Fix locale

ildyria:
Add tests to remove overflowing-unused translations.


nagmat84:
Could we wait with this PR until after https://github.com/LycheeOrg/Lychee/pull/1494 has been merged? I guess a lot of it will be solved by then and merging this PR now only makes it harder to get https://github.com/LycheeOrg/Lychee/pull/1494 out of the door. https://github.com/LycheeOrg/Lychee/pull/1494 is only waiting for https://github.com/LycheeOrg/Lychee-front/pull/322 to be merged.

ildyria:
@nagmat84 this PR just makes a minor changes to some of the local and add tests to ensure consistency. I fail to see how this impacts much #1494 which is in draft mode. :)
Furthermore, this is more like a PR in advance of your future comments on #1539 where you are going to for an external PR on some sections. :D

tests/Feature/LangTest.php
```diff
@@ -35,12 +41,48 @@ public function testLang(): void
...
 			}
 		}
 
 		static::assertEquals('en', Lang::get_code());
 		static::assertEquals('OK', Lang::get('SUCCESS'));
+		$this->assertFalse($failed);
+	}
+
+	/**
+	 * Test Languages are strictly.
```
kamil4, line 59: This comment doesn't seem to explain anything?

tests/Feature/LangTest.php
```diff
@@ -35,12 +41,48 @@ public function testLang(): void
...
+		$full = Lang::get_lang();
+		$lang_available = Lang::get_lang_available();
+
+		foreach ($lang_available as $code) {
+			$lang_test = Lang::factory()->make($code);
+			$keys = array_keys($lang_test->get_locale());
+
+			foreach ($keys as $key) {
+				try {
+					static::assertArrayHasKey($key, $full, 'Language ' . $lang_test->code() . ' as too many keys.');
```
kamil4, line 78:
```suggestion
					static::assertArrayHasKey($key, $full, 'Language ' . $lang_test->code() . ' has too many keys.');
```

Would it make sense to also print the extra key? Or is that done already?

kamil4 (changes requested):

nagmat84:
> https://github.com/LycheeOrg/Lychee/pull/1494 which is in draft mode. :)

This PR is only in draft mode as a safety measure such you do not accidentally merge it.

It already has been approved by you and @kamil4 and I am not doing anything on it but keeping it up-to-date with current changes. The obstacle is that https://github.com/LycheeOrg/Lychee/pull/1494 uses unicode char points which our current frontend can not properly handle in all places. Hence, https://github.com/LycheeOrg/Lychee/pull/1519 needs to be merged first, but then there is nothing else which prevents https://github.com/LycheeOrg/Lychee/pull/1494 from being merged, too.

app/Locale/Slovak.php
```diff
@@ -303,7 +302,6 @@ public function get_locale(): array
 			'SUCCESS' => 'OK',
 			'RETRY' => 'Opakovať',
 
-			'SETTINGS_WARNING' => 'Zmena rozšírených nastavení môže mať negatívny vplyv na stabilitu, bezpečnosť a rýchlosť tejto aplikácie. Upravujte len to, o čom presne viete, čo robíte.',
```
nagmat84, line 306: This one is actually used in `view.js` by `view.full_settings.content.init`. We must not delete it.

app/Locale/Slovak.php
```diff
@@ -312,8 +310,6 @@ public function get_locale(): array
 			'SETTINGS_SUCCESS_IMAGE_OVERLAY' => 'EXIF-Overlay nastavenia aktualizované',
 			'SETTINGS_SUCCESS_PUBLIC_SEARCH' => 'Verejné vyhľadávanie bolo aktualizované',
 			'SETTINGS_SUCCESS_LICENSE' => 'Prednastavená licencia aktualizovaná',
-			'SETTINGS_SUCCESS_CSS' => 'CSS aktualizované',
-			'SETTINGS_SUCCESS_UPDATE' => 'Nastavenia úspešne aktualizované',
```
nagmat84, line 316: Same thing here, see `settings.js` method `settings.changeCSS`. Don't delete it.

app/Locale/Slovak.php
```diff
@@ -385,8 +381,6 @@ public function get_locale(): array
 			'LANG_TEXT' => 'Zmeniť jazyk Lychee na:',
 			'LANG_TITLE' => 'Zmena jazyka',
 
-			'CSS_TEXT' => 'Vlastné CSS:',
-			'CSS_TITLE' => 'CSS zmeniť',
```
nagmat84, line 389: And again. See `view.js` method `view.settings.content.setCSS`.

tests/Feature/LangTest.php
```diff
@@ -35,12 +41,48 @@ public function testLang(): void
...
+				} catch (ExpectationFailedException $e) {
+					$msgSection->writeln('<comment>Error:</comment> ' . $e->getMessage());
+					$failed = true;
+				}
 			}
 		}
 
 		static::assertEquals('en', Lang::get_code());
 		static::assertEquals('OK', Lang::get('SUCCESS'));
+		$this->assertFalse($failed);
```
nagmat84, line 55: Error: Static method 'assertFalse' should not be called dynamically 

```suggestion
		static::assertFalse($failed);
```

Which is also rather obvious compared with the two lines above.

tests/Feature/LangTest.php
```diff
@@ -35,12 +41,48 @@ public function testLang(): void
...
+			$lang_test = Lang::factory()->make($code);
+			$keys = array_keys($lang_test->get_locale());
+
+			foreach ($keys as $key) {
+				try {
+					static::assertArrayHasKey($key, $full, 'Language ' . $lang_test->code() . ' as too many keys.');
+				} catch (ExpectationFailedException $e) {
+					$msgSection->writeln('<comment>Error:</comment> ' . $e->getMessage());
+					$failed = true;
+				}
```
nagmat84, line 82: Instead of looping across all the keys we could simply use the PHP native method [`array_diff_key`](https://www.php.net/manual/en/function.array-diff-key.php). It should make the code a little bit more efficient (not that I care such much about it in the test), but it should also make the code much more clearer and easier to grasp.

Moreover, this test method and the test method above could be combined into one. Basically, they do the same thing just with swapped roles. So that is the pseudo code:

```php
public function testLanguageConsistency(): void {
	$englishDictionary = Lang::get_lang();
	$availableDictionaries = Lang::get_lang_available();
	$failed = false;
	foreach ($availableDictionaries as $locale) {
		$dictionary = Lang::factory()->make($locale);
		$missingKeys = array_diff_key(englishDictionary, $dictionary);
		$extraKeys = array_diff_key($dictionary, englishDictionary);
		if (size($missingKeys) !== 0) {
			$msgSection->writeln(sprintf('<comment>Error:</comment> Locale %s misses the following keys: %s', $locale, implode($missingKeys, ', ')));
			$falied = true;
		}
		if (size($extraKeys) !== 0) {
			$msgSection->writeln(sprintf('<comment>Error:</comment> Locale %s has the following extra keys: %s', $locale, implode($extraKeys, ', ')));
			$falied = true;
		}
	}
	static::assertFalse($failed);
}
```

nagmat84 (changes requested):
I am in strict favor to not change the current locale files. I just jumped into the middle of them and first six keys which have been removed are actually used currently. I stopped checking at this point. A simple `grep` did the job.

Moreover, this clean-up is already made in https://github.com/LycheeOrg/Lychee/pull/1494 which has already been approved.

However, I am fine with improving the test for language consistency.

ildyria:
> I am in strict favor to not change the current locale files. I just jumped into the middle of them and first six keys which have been removed are actually used currently. I stopped checking at this point. A simple `grep` did the job.
> 
> Moreover, this clean-up is already made in #1494 which has already been approved.
> 
> However, I am fine with improving the test for language consistency.

Then I will close that one and directly open a PR on #1494 But then it will have to be merged before #1539 :)

PR #1558, opened 2022-10-17
Add option to skip diagnostic checks

### LINKED ISSUE #1555: [Enhancement] Option to skip 2775 permission checks in diagnostics
My installation is using S3 storage and when I run diagnostics, the script can take literally hours to check the 2775 permissions in my gallery. The reason is because S3 storage badly under-perform with operations like listing.

Considering that this permission is not implemented in a S3 bucket anyways, I would like to skip this part of the script.

For now, I can't run diagnostics at all.

qwerty287:
Closes https://github.com/LycheeOrg/Lychee/issues/1555

1. Allows to add a new env var with class names of checks that will be skipped.
2. Allows to define as many `--skip` options you want containing class names of checks (e.g. `TimezoneCheck`) and skip the check then. May be helpful if you have checks that would take a long time to run (just as described in #1555)


ildyria:
One one hand, I would say LGTM.
On the other hand I would say that the permission check should be skipped in the case of S3.

d7415:
> On the other hand I would say that the permission check should be skipped in the case of S3.

The way that works at the moment (mounting S3 as a local filesystem), we have no way of knowing.

kamil4:
Ideally there should also be a way to skip the tests for the Diagnostics initiated by the frontend. So maybe this should be a config option that works for both the frontend and the commandline and the `--skip` could then override it?

qwerty287:
@kamil4 I implemented this option in https://github.com/LycheeOrg/Lychee/pull/1558/commits/ef390b5b5cf1d99b973e9b2f51d465c9b9854469.

ildyria (approved):
LGTM.

I thought you meant a DB config, but indeed in the `.env` it makes more sense.

config/app.php
```diff
@@ -120,6 +120,16 @@
...
 
+	/*
+	|--------------------------------------------------------------------------
+	| Skip diagnostics checks
+	|--------------------------------------------------------------------------
+	|
+	| Allows to define class names of diagnostics checks that will be skipped.
+	|
+	*/
+	'skip_diagnostics_checks' => explode(',', env('SKIP_DIAGNOSTICS_CHECKS', '')),
```
ildyria, line 131: Please add `#SKIP_DIAGNOSTICS_CHECKS=` and short description in `.env.example` otherwise this option is invisible to the user. :)

ildyria (changes requested):
Please add `#SKIP_DIAGNOSTICS_CHECKS=` and short description in `.env.example` :)

kamil4 (approved):
LGTM (untested). Indeed I did mean a config in the DB, but one in `.env` probably makes more sense (and I agree with @ildyria that there should be an example).

ildyria (approved):
LGTM.

PR #1560, opened 2022-10-18
Add psd support if Imagick is available

### LINKED ISSUE #1554: RAW images support
Lychee looks great. I am wondering, however: Is there any way to support .psd files as well? 

We usually like to bundle our screenshots and png files with our actual .psd files. It would be annoying to have to decouple them. 

Ran a search here and don't see much commentary on this. Apologies if I missed something.

ildyria:
Fixes #1554


app/Image/MediaFile.php
```diff
@@ -199,6 +215,24 @@ public function getOriginalBasename(): string
...
+	 * @return string[]
+	 */
+	protected static function getSupportedImageMimeTypes(): array
+	{
+		if (self::$cachedSupportedImageMimeTypes === null) {
+			if (Configs::hasImagick()) {
+				self::$cachedSupportedImageMimeTypes = array_merge(self::SUPPORTED_IMAGE_MIME_TYPES, self::SUPPORTED_IMAGICK_MIME_TYPES);
+			} else {
+				self::$cachedSupportedImageMimeTypes = self::SUPPORTED_IMAGE_MIME_TYPES;
+			}
```
nagmat84: For better readability, I would prefer the tertiary operator here, but no strong opinion. Just a thought.

tests/TestCase.php
```diff
@@ -33,6 +33,8 @@ abstract class TestCase extends BaseTestCase
 	public const MIME_TYPE_IMG_GIF = 'image/gif';
 	public const MIME_TYPE_IMG_JPEG = 'image/jpeg';
 	public const MIME_TYPE_IMG_PNG = 'image/png';
+	public const MIME_TYPE_IMG_PSD = 'image/vnd.adobe.photoshop';
+	public const MIME_TYPE_APP_PSD = 'application/vnd.adobe.photoshop';
```
nagmat84: Please keep the list in alphabetic order.

nagmat84 (commented):
I haven't tested it, but it looks good and I don't see anything which could potentially fail. Only two minor suggestions.

I wonder if we should extend the approach to any image file format which is supported by Imagick and is somewhat common, not only PSD. See https://imagemagick.org/script/formats.php#supported for a complete list. I don't suggest to add all of them, but what is about Gimp file format (as the open source alternative to PSD), etc.?

**@ildyria Did you test what happens, if you display such an image in the browser?**

I am afraid of the following problem. Depending on the resolution, etc., the JS frontend may attempt to display the original image size in the photo view for a non-raw image. Previously, when we only considered JPEG, PNG, GIF, etc. as "not raw" we could be pretty sure that any recent browser was able to render the image. For "raw" images the JS frontend falls back to one of the medium sizes (if available) or shows a placeholder.

After this change, PSD is considered "not raw". This could lead to the problem that JS frontend tries to display that one in the browser. I doubt that this will work.

ildyria:
> @ildyria Did you test what happens, if you display such an image in the browser?

Currently it is returning the .psd as small/medium if the image is too small, and as a result is giving a broken image. :(

ildyria:
At the moment I am struggling with the tests. phpunit dies in segfault and I can't find why. It works totally fine directly in browser. :(

ildyria:
Finally got the tests to work.

kamil4:
> After this change, PSD is considered "not raw". This could lead to the problem that JS frontend tries to display that one in the browser. I doubt that this will work.

Yeah, I worry that for this reason this might be a wrong approach. It feels to me that we _should_ be treating them as raw. Maybe there should just be some sort of "default raw formats" list so that for well known formats the user doesn't need to add them via settings.

d7415:
> I wonder if we should extend the approach to any image file format which is supported by Imagick and is somewhat common, not only PSD

:+1:

> Yeah, I worry that for this reason this might be a wrong approach. It feels to me that we _should_ be treating them as raw. Maybe there should just be some sort of "default raw formats" list so that for well known formats the user doesn't need to add them via settings.

:+1:

ildyria:
I am closing this as this creates problems with Lychee if the PSD is too small in size to generate a medium file (and I am too lazy to fix the code to handle that edge case).

PR #1563, opened 2022-10-20
File cleanups and remove Homestead

qwerty287:
- delete travis file
- delete codecov upload script
- delete `aliases` file
- rename `readme.md` -> `README.md`
- remove `contrib` make targets
- remove Homestead files + dependency


ildyria:
I don't think we want to remove `server.php` :/

makefile
```diff
@@ -38,8 +38,7 @@ dist-gen: clean composer
 	@cp -r composer.lock                    Lychee
 	@cp -r index.php                        Lychee
 	@cp -r LICENSE                          Lychee
-	@cp -r readme.md                        Lychee
-	@cp -r server.php                       Lychee
```
ildyria, line 42: I am not sure about that line.

nagmat84 (approved):
That was an easy PR. Deleted things are always good things :smile:

I haven't test it. @qwerty287 I assume you did? I haven't checked where `scripts/after.sh` and `scripts/codecov.sh` might have been used or if they are still used somewhere. Have you `grep`ed for them? If they are called nowhere and our CI is still running, I guess it is fine to delete them.

ildyria:
`after.sh` may be used by docker but I'm not sure. @d7415  ?

makefile
nagmat84: Why not? What do mean by "your are not sure"?

qwerty287: Yes, this was accidentally changed.

qwerty287:
@ildyria I don't think so, searched for it in the docker repo, no result. This file was added for Homestead.

qwerty287:
And codecov.sh was only used for Travis CI we don't use anymore.

makefile
nagmat84: Why, I thought it was done on purpose. Why do we want to include `server.php` into the distribution? As far as I understand it, the file is only useful during development if you run a minimalistic webserver from within Artisan. That is also the reason why I asked what @ildyria meant by "he is not sure".

qwerty287: We can definitely remove this file within the upgrade to Laravel 9 because it's not included in the default template (https://github.com/laravel/laravel) anymore.

nagmat84: Then I would keep it the way how it has been initially in this PR, i.e. revert https://github.com/LycheeOrg/Lychee/pull/1563/commits/3d7da4b0b00d111d75a6afe7bd28875be8201baa, unless @ildyria can explain why the file is needed.

d7415:
> `after.sh` may be used by docker but I'm not sure. @d7415 ?

I'm not at a PC, but I'm pretty sure it doesn't.

makefile
ildyria: Just delete it (but also the actual server.php file if it is not in the Laravel 9 repo).

qwerty287: I would keep it and delete it in #1469 and just change the makefile here.

ildyria: also works for me.

qwerty287:
@ildyria Can I go on with merging?

PR #1564, opened 2022-10-21
Add option to append tags

### LINKED ISSUE #654: [Enhancement] Add a checkbox to overwrite or add tags when selecting mulitple images
Steps to reproduce:

- Tag some pictures and add multiple tags to them.

image

- Now, select multiple pictures and click on tag selected:

image
image

- Click on the initial picture and you'll see the old tags gone.

image

As you can see, the old tags have all been replaced when you add new tags. It should've kept the old tags, otherwise, it'll be really hard to be able to tag pictures.

Thanks

qwerty287:
https://github.com/LycheeOrg/Lychee-front/pull/339

Also contains sync for https://github.com/LycheeOrg/Lychee-front/pull/336 and https://github.com/LycheeOrg/Lychee-front/pull/337

Closes #654


app/Http/Requests/Photo/SetPhotosTagsRequest.php
```diff
@@ -17,12 +17,17 @@ class SetPhotosTagsRequest extends BaseApiRequest implements HasPhotos, HasTags
...
+
+	public bool $override;
+
 	/**
 	 * {@inheritDoc}
 	 */
 	public function rules(): array
 	{
 		return [
+			self::OVERRIDE_ATTRIBUTE => 'boolean',
```
nagmat84:
```suggestion
			self::OVERRIDE_ATTRIBUTE => 'required|boolean',
```
?

app/Http/Requests/Photo/SetPhotosTagsRequest.php
```diff
@@ -37,5 +42,6 @@ protected function processValidatedValues(array $values, array $files): void
 	{
 		$this->photos = Photo::query()->findOrFail($values[HasPhotos::PHOTO_IDS_ATTRIBUTE]);
 		$this->tags = $values[HasTags::TAGS_ATTRIBUTE];
+		$this->override = $values[self::OVERRIDE_ATTRIBUTE] ?? false;
```
nagmat84: If we require the attribute `override` above, we don't need to define a default value here and I don't see any reason why we should not required it.

app/Http/Requests/Photo/SetPhotosTagsRequest.php
```diff
@@ -17,12 +17,17 @@ class SetPhotosTagsRequest extends BaseApiRequest implements HasPhotos, HasTags
 	use HasTagsTrait;
 	use AuthorizeCanEditPhotosTrait;
 
+	public const OVERRIDE_ATTRIBUTE = 'override';
```
nagmat84: In order to make it make it more explicit that this attribute is a boolean value, I would prefer a name like `shallOverride` (or `shall_override`) or similar. Only `override` is too untelling for my taste.

app/Http/Requests/Photo/SetPhotosTagsRequest.php
```diff
@@ -17,12 +17,17 @@ class SetPhotosTagsRequest extends BaseApiRequest implements HasPhotos, HasTags
...
+
+	public bool $override;
```
nagmat84: Same thing here.
```suggestion
	public bool $shallOverride;
```

for example.

tests/Feature/Lib/PhotosUnitTest.php
```diff
@@ -204,18 +204,21 @@ public function set_star(
...
 	 * @param string[]    $ids
 	 * @param string[]    $tags
+	 * @param bool        $override
 	 * @param int         $expectedStatusCode
 	 * @param string|null $assertSee
 	 */
 	public function set_tag(
 		array $ids,
 		array $tags,
+		bool $override = true,
```
nagmat84, line 214: Same here
```suggestion
		bool $shallOverride = true,
```

app/Locale/ChineseSimplified.php
```diff
@@ -297,6 +297,7 @@ public function get_locale(): array
 			'ERROR_SEARCH_DEACTIVATED' => '搜索功能已在设置中停用。',
 			'SUCCESS' => 'OK',
 			'RETRY' => '重试',
+			'OVERRIDE' => 'Override',
```
nagmat84, line 300: I still have to check with the frontend where we use this label, but in order to prevent clash of names in the future I would prefer a key which is more specific than only `OVERRIDE`. I would expect something like `CHECKBOX_OVERRIDE_LABEL` or whatever is appropriate. (I haven't had a look at the frontend yet, hence I don't know where this key is used.)

nagmat84 (changes requested):

nagmat84 (approved):

app/Http/Requests/Photo/SetPhotosTagsRequest.php
```diff
@@ -17,12 +17,17 @@ class SetPhotosTagsRequest extends BaseApiRequest implements HasPhotos, HasTags
 	use HasTagsTrait;
 	use AuthorizeCanEditPhotosTrait;
 
+	public const SHALL_OVERRIDE_ATTRIBUTE = 'shall_override';
```
ildyria, line 20: For the sake of consistency, I think a trait `HasShallOverrideTrait` should be created.

kamil4 (changes requested):
The localization files are missing the TAGS_OVERRIDE_INFO key.

qwerty287:
@kamil4 Looks like I pushed while you reviewed :) https://github.com/LycheeOrg/Lychee/pull/1564/commits/2a881988cd21c23fe9d08083216ffe796cc84eec

kamil4:
> The localization files are missing the TAGS_OVERRIDE_INFO key.

Never mind...

kamil4 (approved):

app/Http/Requests/Photo/SetPhotosTagsRequest.php
qwerty287: But there are other requests like `DeleteSharingRequest` that also take it directly.

nagmat84: I have no real opinion about that. I typically start creating a trait if at least three but at minimum two different classes use the same thing. `shall_override` is very unique to the `SetPhotosTagsRequest` (at least for the moment).

Personally, I would not create a trait for it. This only adds an additional step of indirection and complexity without saving any lines of code. But that is a question of personal taste.

PR #1569, opened 2022-10-27
[Bug][Regression] Fix dir permissions

nagmat84:
The upgrade from Laravel 8 to 9 and the implicit upgrade from Flystem 1 to 3, introduced a regression regarding file permissions. There are two independent problems: one is ultimately fixed by this PR for the other this PR contains a workaround.

 * Besides the configuration parameter `visibility` Flysystem 3 requires a new parameter `directory_visibility` which - nomen est omen - defines the permissions for newly created directories. If this option is not set (or equals `null`), Flysystem does not fall back to `visibility` (as one could expect) but uses an internal default value.
 * The second problem is an upstream regression between Flysystem 1 and 3 (actually between 2 and 3) which makes Flysystem behave inconsistent with respect to the umask value. The upstream bug is reported at https://github.com/thephpleague/flysystem/issues/1584. As a workaround this bugfix temporarily sets the umask value to zero.

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -47,6 +47,7 @@ return [
 			'root' => env('LYCHEE_UPLOADS', public_path('uploads/')),
 			'url' => env('LYCHEE_UPLOADS_URL', 'uploads/'),
 			'visibility' => env('LYCHEE_IMAGE_VISIBILITY', 'public'),
+			'directory_visibility' => env('LYCHEE_IMAGE_VISIBILITY', 'public'),
 			'permissions' => [
 				'file' => [
 					'world' => 00666,
--- a/.env.example
+++ b/.env.example
@@ -45,16 +45,6 @@ DB_LOG_SQL=false
 # Some rare setups may require directories and files to be world writeable.
 # In this case, use "world" here.
 # USE WITH PRECAUTIONS: world writeable files and folders may be a SECURITY RISK.
-# Note at the time of writing, the Flysystem package v1 has a bug which
-# prevents the "world" preset from working.
-# This is a temporary problem and will be solved after
-# https://github.com/thephpleague/flysystem/pull/1523 will have been merged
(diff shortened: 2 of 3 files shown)


kamil4 (approved):
Looks good (untested).

Wasn't there also a temporary comment somewhere that was meant to be removed once we moved to Laravel v9? In `.env.example`, I think? This might be a good PR to attach it to...

config/filesystems.php
```diff
@@ -47,6 +47,7 @@
 			'root' => env('LYCHEE_UPLOADS', public_path('uploads/')),
 			'url' => env('LYCHEE_UPLOADS_URL', 'uploads/'),
 			'visibility' => env('LYCHEE_IMAGE_VISIBILITY', 'public'),
+			'directory_visibility' => env('LYCHEE_IMAGE_VISIBILITY', 'public'),
```
ildyria, line 50: wouldn't we want this to be in the `.env.example` too ? :thinking:

ildyria (approved):
LGTM (untested)

config/filesystems.php
nagmat84: I don't understand what you mean. This PR adds a missing entry to the configuration file. The environment variable `LYCHEE_IMAGE_VISIBILITY` is the same as before. So there is nothing to change with respect to `.env.example`.

nagmat84:
> Wasn't there also a temporary comment somewhere that was meant to be removed once we moved to Laravel v9? In `.env.example`, I think? This might be a good PR to attach it to...

Right. Thank you for reminding me. I removed the outdated comment in https://github.com/LycheeOrg/Lychee/pull/1569/commits/9814d42abd249cc4838380378b40a07baa5519fd.

config/filesystems.php
ildyria: Oh, my mistake, I thought it was a different value than for `visibility`.

PR #1571, opened 2022-10-28
composer update

### LINKED ISSUE #1548: [Enhancement] Metadata Exif description field
*Explain in a few words which functionality or improvements you would like to see in Lychee.*
I'm just trialling Lychee at the moment. I notice that when uploading images the "description" field doesn't get imported even though it exists in the original file. Neither does it seem to get exported when downloading the file.

Is this intentional, or a bug? Any chance of a fix?

Or am I doing something wrong?

ildyria:
Fixes #1548 properly.

I would say that this should be the last PR before 4.6.2 :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,7 @@
         "laravel/framework": "^9.2",
         "livewire/livewire": "^2.7",
         "lychee-org/nestedset": "^6",
-        "lychee-org/php-exif": "^0.7.13",
+        "lychee-org/php-exif": "^0.7.14",
         "maennchen/zipstream-php": "^2.1",
         "php-ffmpeg/php-ffmpeg": "^1.0",
         "php-http/guzzle7-adapter": "^1.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "1061802c57a1df0955dfc63f1c9156c9",
+    "content-hash": "f203d1710ac0667bf2309911235d2243",
     "packages": [
         {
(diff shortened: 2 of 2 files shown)


nagmat84 (approved):

kamil4 (commented):
What about changes to `composer.json`?

ildyria:
@kamil4 Fixed.

composer.json
```diff
@@ -53,7 +53,7 @@
         "laravel/framework": "^9.2",
         "livewire/livewire": "^2.7",
         "lychee-org/nestedset": "^6",
-        "lychee-org/php-exif": "^0.7.13",
+        "lychee-org/php-exif": "^0.7.14",
```
nagmat84, line 56: Just our of curiosity: Why do we always update the `composer.json`, too, even if we only increase patch levels?

IMHO, this is an unnecessary step. `composer update` creates a `composer.lock` which resolves all dependencies to its most recent version. IMHO, `composer.json` only defines a lower limit. This means `lychee-org/php-exif": "^0.7.13` also includes 0.7.14 if this version is published.

d7415: True, but why not? It's a minimal step (compared to opening and reviewing a PR, at least) and may catch some edge cases.

kamil4: I didn't realize that `^0.7.13` includes 0.7.14 (I've never actually bothered to find out what this syntax means -- where is it documented?). That's why I keep complaining about it, because I assume that @ildyria simply keeps forgetting to commit `composer.json` :smiley:.

nagmat84: https://getcomposer.org/doc/articles/versions.md#caret-version-range-

ildyria: One of the reason why is that we enforce the bug fix of #1548 rather than depending on the `.lock` especially if someone decides to stupidly do `composer update --prefer-lowest`.

PR #1572, opened 2022-10-28
Enable Php 8.2 in CICD

ildyria:
Will officially be released in less than a 1 month, let's see what is breaking...


nagmat84 (approved):
Obviously, CICD works and nothing breaks. So I would say, we are good.

PR #1578, opened 2022-11-05
fix smart albums rights

ildyria:
Sync front.


qwerty287 (approved):

nagmat84 (changes requested):
I only request changes, because @kamil4 discovered some more bugs (e.g. the share album icon bug) which also affect current master. I would like to fix those first on the frontend before we sync it.

nagmat84 (approved):

PR #1583, opened 2022-11-07
Config rename

ildyria:


kamil4:
I've been keeping my distance from the whole naming discussion because currently I have neither the time nor, frankly, the patience for it, but when I see a separate PR of manageable length I can't help but think about it. So, at the danger of engaging in bikeshedding, here are my resolutely uninformed thoughts:
* Yes, the old names (`full_photo`, `downloadable`) are awful.
* The new names (`grants_*`) are better.
* In this particular context though, are they descriptive enough? These config variables are used only in the relatively rare cases where the settings of the parent album do not apply. So wouldn't it be better to call them something like `grants_full_photo_access_fallback` and `grants_download_fallback` to indicate their limited role?
* There are at least two more config variables used in a similar context: `share_button_visible` and `public_photos_hidden`. Wouldn't this be the right opportunity to unify their naming as well?

ildyria:
> `grants_full_photo_access_fallback` and `grants_download_fallback` to indicate their limited role?

Then `grants_full_photo_access_default` and `grants_download_default` would be better then IMHO. :)

> There are at least two more config variables used in a similar context: `share_button_visible` and `public_photos_hidden`. Wouldn't this be the right opportunity to unify their naming as well?

Sure can do. :)
In the meantime, may I interested you in #1584 :D

qwerty287 (approved):

qwerty287:
What happened to
> There are at least two more config variables used in a similar context: share_button_visible and public_photos_hidden. Wouldn't this be the right opportunity to unify their naming as well?

?

PR #1590, opened 2022-11-12
Fix list of shares

nagmat84:
This fixes the list of shares (from the left menu) for non-admin, authenticated users. As @kamil4 reported [on Gitter at Nov 7th 2022 18:20](https://gitter.im/LycheeOrg/Developers?at=63693e5e21df5f7a54da0ec2) non-admin user receive a `403 Forbidden` response. This is a regression due to https://github.com/LycheeOrg/Lychee/pull/1519.

Before https://github.com/LycheeOrg/Lychee/pull/1519 the request `Sharing::list` had unconditionally returned a list of all active shares for all users and all albums as well as a list of all existing albums and all users. The the frontend filtered the received list for the the album or user of interest. The frontend code was not only ugly as hell, but the whole approach had been inefficient. There is no reason to query and transmit everything only to throw way 97% on the frontend. Hence, https://github.com/LycheeOrg/Lychee/pull/1519 introduced two query parameters `albumID` and `userID` to enable filtering on the backend.

However, I went one step further and thought it would clever to forbid unrestricted queries `Share::list` for non-admin users. This was fine when one used the dialog from the toolbar Sharing, but it did no occur to me that non-admin users can also open the page with all sharings from the left menu.

I also discovered related, lurking bug. Do you remember that @kamil4 once said that he doesn't like the attribute name `user2` on a request? Here, this actually gave leeway to a trap. The request `Sharing::list` is one of the rare instances with actually _three_ users being involved: the authenticated user who makes the request, the user who owns the album and the user with whom the album is shared.`user2` was used for the last two, but surely they are different. I renamed them to `owner` and `participant`.

### DIFF
--- a/app/Actions/Sharing/ListShare.php
+++ b/app/Actions/Sharing/ListShare.php
@@ -12,17 +12,24 @@ use Illuminate\Support\Facades\DB;
 class ListShare
 {
 	/**
-	 * @param User|null      $user
-	 * @param BaseAlbum|null $baseAlbum
+	 * Returns a list of shares optionally filtered by the passed attributes.
+	 *
+	 * @param User|null      $participant the optional user who participates
+	 *                                    in a sharing, i.e. the user with
+	 *                                    whom albums are shared
+	 * @param User|null      $owner       the optional owner of the albums
+	 *                                    which are shared
+	 * @param BaseAlbum|null $baseAlbum   the optional album which is shared
 	 *
 	 * @return Shares
 	 *
 	 * @throws QueryBuilderException
 	 */
-	public function do(?User $user, ?BaseAlbum $baseAlbum): Shares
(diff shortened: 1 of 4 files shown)


app/Actions/Sharing/ListShare.php
```diff
@@ -64,10 +76,15 @@ public function do(?User $user, ?BaseAlbum $baseAlbum): Shares
 				unset($album->parent);
 			});
 
-			$users = DB::table('users')
-				->select(['id', 'username'])
-				->where('id', '>', 0)
-				->orderBy('username', 'ASC')
+			// Existing users with whom an album can be shared optionally
```
ildyria, line 79: Is this filtering really useful ?

ildyria (approved):
Not super convinced by some of the additional filtering.
Nevertheless if it fixes the bug then LGTM.

PR #1591, opened 2022-11-16
Add API to change upload date & fix API token auth

### LINKED ISSUE #984: [Enhancement] API to set photo upload date
Hello! Thanks for all the awesome work. I love Lychee so far.

I'm working on a script to import photos into Lychee, and right now, all imported photos have the same upload date (now). I would like to be able to set the upload date to a different date, which I happen to have data for. Note that this is still different than the takestamp. I have no issue with that, but that is not what I'm looking for.

Thanks!

qwerty287:
Closes #984 

Adds `Photo::setUploadDate` and fixes a bug of the API token auth that tried to read the token from the header called `token` while it should read `Authorization`.

Associated front-end https://github.com/LycheeOrg/Lychee-front/pull/353


tests/Feature/Lib/PhotosUnitTest.php
```diff
@@ -278,6 +278,32 @@ public function set_license(
 		}
 	}
 
+	/**
+	 * Set license.
```
ildyria:
```suggestion
	 * Set upload date.
```

app/Http/Requests/Contracts/HasDate.php
```diff
@@ -0,0 +1,15 @@
...
+use Illuminate\Support\Carbon;
+
+interface HasDate
+{
+	public const DATE_ATTRIBUTE = 'date';
+
+	/**
+	 * @return Carbon|null
+	 */
+	public function requestDate(): ?Carbon;
```
ildyria, line 14: Why `requestDate()` and not just `date()` ?

qwerty287: Because it is already a function.

ildyria (approved):
LGTM.
@qwerty287  feel free to review the front-end I proposed.

qwerty287:
Thanks, please don't merge yet. There's an improvement I'd like to add to this. If the new upload date is  after the current `updated_at`, this should be changed as well.

app/Http/Controllers/PhotoController.php
```diff
@@ -287,6 +287,9 @@ public function setLicense(SetPhotoLicenseRequest $request): void
 	public function setUploadDate(SetPhotoUploadDate $request): void
 	{
 		$request->photo()->created_at = $request->requestDate();
+		if ($request->photo()->updated_at->isBefore($request->requestDate())) {
```
ildyria: This is unnecessary. When doing `->save()` on a model, laravel will update the value of `updated_at` to the timestamp of the operation.
I just checked.

ildyria (approved):
LGTM.
Tested.

Issue #1553, opened 2022-10-12
"SQLSTATE[42S22]: Column not found: 1054 Unknown column 'owner_id' in 'field list'" after (successful?) update


eip618:
```
2022-10-12 04:19:52 UTC -- error   -- PDO::prepare -- 368 -- SQLSTATE[42S22]: Column not found: 1054 Unknown column 'owner_id' in 'field list'
2022-10-12 04:19:52 UTC -- error   -- App\Actions\Diagnostics\Checks\MissingUserCheck::check -- 13 -- SQLSTATE[42S22]: Column not found: 1054 Unknown column 'owner_id' in 'field list'; caused by
2022-10-12 04:19:52 UTC -- error   -- Illuminate\Database\Connection::runQueryCallback -- 712 -- SQLSTATE[42S22]: Column not found: 1054 Unknown column 'owner_id' in 'field list' (SQL: select `owner_id` from `albums` group by `owner_id`); caused by
2022-10-12 04:19:52 UTC -- error   -- App\Actions\Diagnostics\Checks\MissingUserCheck::check -- 13 -- SQLSTATE[42S22]: Column not found: 1054 Unknown column 'owner_id' in 'field list' (SQL: select `owner_id` from `albums` group by `owner_id`); caused by
2022-10-12 04:19:32 UTC -- error   -- Illuminate\Routing\AbstractRouteCollection::handleMatchedRoute -- 43 -- 
2022-10-12 04:19:32 UTC -- error   -- App\Http\Middleware\FixStatusCode::handle -- 31 -- ; caused by
2022-10-12 04:19:21 UTC -- error   -- PDO::prepare -- 368 -- SQLSTATE[42S22]: Column not found: 1054 Unknown column 'owner_id' in 'field list'
2022-10-12 04:19:21 UTC -- error   -- App\Actions\Diagnostics\Checks\MissingUserCheck::check -- 13 -- SQLSTATE[42S22]: Column not found: 1054 Unknown column 'owner_id' in 'field list'; caused by
2022-10-12 04:19:21 UTC -- error   -- Illuminate\Database\Connection::runQueryCallback -- 712 -- SQLSTATE[42S22]: Column not found: 1054 Unknown column 'owner_id' in 'field list' (SQL: select `owner_id` from `albums` group by `owner_id`); caused by
[...]

PR #1595, opened 2022-11-21
refactor diagnostics to use Laravel pipelines

ildyria:
Diagnostics no longer uses auto discovery factory.
This should solve the problem of users just over-writting their configuration when updating Lychee.
See : https://github.com/LycheeOrg/Lychee/issues/1553#issuecomment-1321234867

There is very minimal code change, mostly moving files & matching the Laravel pipeline structure.


eip618:
> Diagnostics no longer uses auto discovery factory. This should solve the problem of users just over-writting their configuration when updating Lychee. See : [#1553 (comment)](https://github.com/LycheeOrg/Lychee/issues/1553#issuecomment-1321234867)

did i do the upgrade wrong?

ildyria:
> did i do the upgrade wrong?

Not really. Per the [manual update instruction](https://lycheeorg.github.io/docs/update.html#update-manually) it could be said you did it write by simply over-writing all the folders. The problem is that it does not erases already existing files and as a result the automatic detection for diagnostic was working on older files.
Hence this proposal to hard code the diagnostics and remove the automated part.

qwerty287 (approved):
Did not review the moved files, and I don't know about laravel's pipeline system, but I think this looks good.

ildyria:
> I don't know about laravel's pipeline system, but I think this looks good.

This is a nice example :  https://martinjoo.dev/laravel-pipelines

PR #1601, opened 2022-11-26
Add config option to set default album visibility

### LINKED ISSUE #1363: [Enhancement] Public by default
Would it be possible to have pictures and albums be public by default? I use Lychee as a public photo album, and it's quite nuisance to manually set everything to public one-by-one.

qwerty287:
Requires #1539 

Closes #1363


app/Actions/Album/Create.php
```diff
@@ -25,6 +27,46 @@ public function create(string $title, ?Album $parentAlbum): Album
...
+			default:
+				$protectionPolicy = new AlbumProtectionPolicy(
+					is_public: false,
+					is_link_required: false,
+					is_nsfw: false,
+					grants_full_photo_access: true,
+					grants_download: true,
+					is_password_required: false,
+				);
+				break;
```
qwerty287: The reason why the CI fails is that I don't know how I should set the default values (except `is_public`) here. What should I use?

ildyria: I would say that in the spirit of the PR, if `is_public` is true, then `is_link_required` should be false.

`is_nsfw` => I would default it to false, but we could add a Config for that.
`grants_download` => There is probably a Config default for that.
 `grants_full_photo_access` => idem.
 `is_password_required` is obviously false.

ildyria: Similarly to `ofBaseAlbum` I would add `ofDefaultPublic` and `ofDefaultPrivate` to avoid duplication.

PR #1602, opened 2022-11-26
Create admin during installation, allow multiple admins

qwerty287:
Requires #1539 to be merged.

- create the admin user during installation
- add command to create new users
- allow multiple admins (in theory, untested)

TODO

- [x] add installation step to create user
- [x] add command to create user
- [x] remove related legacy code
- [x] fix code and tests to not longer assert that admin is user with ID 0
- [x] fix tests (does not find tables `users` and `sym_links` !?)
- [x] increment all user IDs (no longer use ID 0 to fix Laravel issues)
- [x] Add command to reset the password of a user (useful if you are locked out of admin)
- [x] remove old code
- [x] Password confirmation check on admin creation.


database/migrations/2022_11_26_183251_increment_user_i_ds.php
```diff
@@ -0,0 +1,63 @@
...
+{
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up(): void
+	{
+		/** @var User $user */
+		$user = User::find(0);
```
ildyria: I am not sure if we should use `User::` or `DB::table`
The later would make us agnostic from the production code which is likely to be better.

qwerty287:
Somebody has an idea why PostgreSQL tests fail? The exception has this message:
```
DETAIL:  Key (id)=(1) already exists. (SQL: insert into "users" ("may_upload", "may_edit_own_settings", "may_administrate", "username", "password", "updated_at", "created_at") values (0, 1, 0, Test user, $2y$04$ijCmj/NhCDBeNB6zmQ/sVONTKQ/SoCq615bMw3UIwMuGtkCN67bAu, 2022-11-30 16:58:19.479180, 2022-11-30 16:58:19.479180) returning "id")#0 /home/runner/work/Lychee/Lychee/vendor/laravel/framework/src/Illuminate/Database/Connection.php(720): Illuminate\Database\Connection->runQueryCallback()
```

But this user is created in https://github.com/LycheeOrg/Lychee/blob/8870d96f76bb9f75cfe97f95018551e042e8411e/app/Actions/User/Create.php#L22-L35 I didn't change (except the debugging `echo`s).
Why doesn't this auto-increment the ID?

app/Policies/SettingsPolicy.php
```diff
@@ -39,7 +38,7 @@ public function canEdit(User $user): bool
 	 */
 	public function canSeeLogs(?User $user): bool
 	{
-		return AdminAuthentication::isAdminNotRegistered();
+		return $user !== null && $user->may_administrate;
```
ildyria:
```suggestion
		return $user?->may_administrate === true;
```

ildyria:
Have you considered the following case:
- user download lychee
- user run artisan migrate
- user opens Lychee in their browser.

What happens ?

qwerty287:
Nothing happens, you just can't log in. You have to use the CLI command to create a new user (or via web install).

ildyria:
I am wondering if we should not add a 5th step in the install process (i.e. after migration) where the user register the admin account, and if the admin user is not created, we are redirected to this page.

qwerty287:
I'm not sure, it's better to not allow a public registration in any state *except* the installation phase (which is completed at this point). I'd keep it similar to what this PR is currently doing, and let the user either create the admin during web install or using the CLI.

app/Http/Controllers/Administration/DiagnosticsController.php
```diff
@@ -29,7 +28,7 @@ class DiagnosticsController extends Controller
 	 */
 	private function isAuthorized(): bool
 	{
-		return Gate::check(SettingsPolicy::CAN_SEE_DIAGNOSTICS, Configs::class) || AdminAuthentication::isAdminNotRegistered();
```
ildyria, line 32: I still think it is important to access the diagnostics if no admins are registered.

qwerty287: Not sure because there's always an admin user or you can access the CLI.

tests/Feature/InstallTest.php
```diff
@@ -148,8 +146,24 @@ public function testInstall(): void
...
+		if ($admin === null) {
+			$admin = new User();
+			$admin->incrementing = false;
+			$admin->id = 1;
+			$admin->may_upload = true;
+			$admin->may_edit_own_settings = true;
+			$admin->may_administrate = true;
+			$admin->username = 'admin';
+			$admin->password = Hash::make('password');
+			$admin->save();
```
qwerty287, line 181: Needs sequence update if postgres

phpstan.neon
```diff
@@ -33,6 +33,8 @@ parameters:
 		- '#Dynamic call to static method (Illuminate\\Database\\Query\\Builder|Illuminate\\Database\\Eloquent\\(Builder|Relations\\.*)|App\\Models\\Extensions\\FixedQueryBuilder)(<.*>)?::without\(\)#'
 		- '#Dynamic call to static method (Illuminate\\Database\\Query\\Builder|Illuminate\\Database\\Eloquent\\(Builder|Relations\\.*)|App\\Models\\Extensions\\FixedQueryBuilder)(<.*>)?::count\(\).#'
 		- '#Dynamic call to static method (Illuminate\\Database\\Query\\Builder|Illuminate\\Database\\Eloquent\\(Builder|Relations\\.*)|App\\Models\\Extensions\\FixedQueryBuilder)(<.*>)?::update\(\).#'
+		- '#Dynamic call to static method Illuminate\\Http\\Request::validate#'
+		- '#Dynamic call to static method App\Models\Extensions\FixedQueryBuilder<TModelClass of Illuminate\Database\Eloquent\Model>::orderByDesc().#'
```
ildyria:
```suggestion
		- '#Dynamic call to static method App\\Models\\Extensions\\FixedQueryBuilder<.*>::orderByDesc\(\).#'
```

app/Console/Commands/ResetAdmin.php
```diff
@@ -1,72 +0,0 @@
-<?php
```
ildyria: I am wondering if it would not be better to keep that file but have it print "use php artisan lychee:update_user  <username> <password> instead or lychee:create_user  <username> <password> --may_administrate".

qwerty287: Good idea 👍

app/Console/Commands/CreateUser.php
```diff
@@ -0,0 +1,82 @@
...
+
+		$mayAdministrate = $count < 1 || $this->option('may-administrate') === true;
+		$mayEditOwnSettings = $mayAdministrate || $this->option('may-edit-own-settings') === true;
+		$mayUpload = $mayAdministrate || $this->option('may-upload') === true;
+
+		$user = $this->create->do($username, $password, $mayUpload, $mayEditOwnSettings);
+		$user->may_administrate = $mayAdministrate;
+		$user->save();
+
+		$this->line('Successfully created user ' . $username);
```
ildyria:
```suggestion
		$this->line(sprintf('Successfully created%s user %s ', $mayAdministrate ? ' admin' : '', $username));
```

app/Http/Controllers/Install/MigrationController.php
```diff
@@ -26,8 +32,13 @@ class MigrationController extends Controller
 	 *
 	 * @throws FrameworkException
 	 */
-	public function view(): View
+	public function view(Request $request): View
```
ildyria: Why not use a proper `InstallAdminRequest` object ?

database/migrations/2022_12_10_183251_increment_user_i_ds.php
```diff
@@ -0,0 +1,73 @@
...
+		/** @var App\Models\User|null $admin */
+		$admin = DB::table('users')->find(0);
+		if ($admin !== null && ($admin->username === '' || $admin->password === '')) {
+			// The admin user (id 0) has never set a username and password, so we remove it.
+			// This should only happen on a completely new installation where the admin user is created by the
+			// MigrateAdminUser migration and the user has never logged in.
+			DB::table('users')->delete(0);
+		}
+		/** @var App\Models\User $user */
+		foreach (User::query()->orderByDesc('id')->get() as $user) {
```
ildyria:
```suggestion
		foreach (DB::table('users')->query()->orderByDesc('id')->get() as $user) {
```

database/migrations/2022_12_10_183251_increment_user_i_ds.php
```diff
@@ -0,0 +1,73 @@
...
+			$oldID = $user->id;
+			$newID = $oldID + 1;
+			$user->id = $newID;
+			$user->incrementing = false;
+			$user->save();
```
ildyria: Use a proper DB query here instead of `->save()` ?

tests/Feature/InstallTest.php
```diff
@@ -125,7 +125,7 @@ public function testInstall(): void
 		/**
 		 * apply migration.
 		 */
-		$response = $this->get('install/migrate');
+		$response = $this->post('install/migrate', ['username' => 'admin', 'password' => 'password']);
 		$this->assertOk($response);
 		$response->assertViewIs('install.migrate');
```
ildyria, line 131: Assertion needed : that there exists an admin user at this point and that we can log to it with `admin` and `password`.

ildyria:
I am wondering if we shouldn't create an additional page AFTER the migration page in the install process rather than having it be at the same time of the migration.

tests/Feature/InstallTest.php
qwerty287: All the tables are deleted before, so there's nothing in there. This is the first operation after migration, so there's can't be a user in the DB.

qwerty287:
I understand your issue with it, but I think it's better to do it in one step because you can't really detect an installation after it's migrated. As I stated already above, I don't think there should be a possibility to create a new admin user except in the installation phase which is ended after the migration step (and the only way to check installation would be a config). Also, I don't think you should be able to use parts of the web installer (the user creation) if you're running the CLI installation. This would require an additional config option to check if the installation was done via web or CLI.

tests/Feature/InstallTest.php
qwerty287: I think I got you wrong, this is not a *login* mechanism, but a *registration*. A new user is created with `admin`/`password` by this request after the migration.

ildyria: Yes. That is what I would like this assertion to check. that the DB actually do contains an admin user with the said username and password. :)

qwerty287: Ah afterwards :) yeah good idea

ildyria:
At initial admin user creation, there need to be a password validation check, namely a second password field to confirm that the password is entered properly as chosen (to avoid typos mistakes etc).

Ideally there needs to be a visual feedback that the button has been clicked.
See in: `resources/views/update/error.blade.php` (does not need to be the same but similar)
```html
	<!-- Do not change even a single character in the block below without
	     also updating the checksum in config/secure-headers.php! -->
	<script>
	document.addEventListener("DOMContentLoaded", function(event) {
		document.querySelector("form").addEventListener("submit", function(e){
			document.querySelector("form").hidden = true;
			var text = document.createElement("div");
			text.innerHTML = "Migration started. <b>DO NOT REFRESH THE PAGE</b>.";
			document.querySelector(".form").appendChild(text);
			// e.preventDefault();    //stop form from submitting
		});
	});
	</script>
```

app/Http/Controllers/Administration/UsersController.php
```diff
@@ -71,6 +73,9 @@ public function save(SetUserSettingsRequest $request, Save $save): void
 	 */
 	public function delete(DeleteUserRequest $request): void
 	{
+		if ($request->user2()->id === Auth::id()) {
```
ildyria, line 76: Probably should be moved in the `DeleteUserRequest` validation step :)

app/Http/Requests/Install/InstallMigrationRequest.php
```diff
@@ -0,0 +1,44 @@
...
+class InstallMigrationRequest extends FormRequest implements HasUsername, HasPassword
+{
+	use HasUsernameTrait;
+	use HasPasswordTrait;
+
+	public function rules(): array
+	{
+		return [
+			RequestAttribute::USERNAME_ATTRIBUTE => ['required', new UsernameRule()],
+			RequestAttribute::PASSWORD_ATTRIBUTE => ['required', new PasswordRule(false)],
```
ildyria: Add password validation attribute :)

database/migrations/2022_12_10_183251_increment_user_i_ds.php
```diff
@@ -0,0 +1,71 @@
...
+	public function up(): void
+	{
+		Schema::disableForeignKeyConstraints();
+		/** @var App\Models\User|null $admin */
+		$admin = DB::table('users')->find(0);
+		if ($admin !== null && ($admin->username === '' || $admin->password === '')) {
+			// The admin user (id 0) has never set a username and password, so we remove it.
+			// This should only happen on a completely new installation where the admin user is created by the
+			// MigrateAdminUser migration and the user has never logged in.
+			DB::table('users')->delete(0);
```
ildyria:
```suggestion
			DB::table('users')->where('id', '=', 0)->delete();
```
The result is the same, but it is more explicit of what is going on.

qwerty287:
> Ideally there needs to be a visual feedback that the button has been clicked.
See in: resources/views/update/error.blade.php (does not need to be the same but similar)

I'd do this in another PR and get this one out first,

database/migrations/2022_12_10_183251_increment_user_i_ds.php
```diff
@@ -26,7 +26,7 @@ public function up(): void
 		foreach (DB::table('users')->orderByDesc('id')->get() as $user) {
 			$oldID = $user->id;
 			$newID = $oldID + 1;
-			DB::table('users')->find($oldID)->update(['id' => $newID]);
```
ildyria: `find()` was returning an `object` without methods. Resulting to this error:
`Call to undefined method stdClass::update()`

qwerty287: Thanks!

ildyria (approved):
LGTM.

PR #1603, opened 2022-11-26
separate classes & fix php-stan stricter complaints

ildyria:
Just prepare code for https://github.com/LycheeOrg/phpstan-lychee/pull/1


qwerty287 (approved):
I'm fine with these changes, however, I'm not sure if the others agree here. Especially the `RequestsAttribute` class and the separate `RandomID` class just for `const`s might be discussable.

ildyria:
> I'm fine with these changes, however, I'm not sure if the others agree here. Especially the `RequestsAttribute` class and the separate `RandomID` class just for `const`s might be discussable.

This change is mostly to comply to the no constants in Interfaces.

One of the argument for `RequestsAttribute` class is that it unifies the constants, and thus make sure that all requests are using the same pattern.

As for RandomID, pretty much same reason.

PR #1605, opened 2022-11-27
more coverage

ildyria:


qwerty287 (approved):
I'm not sure if this is the best idea to increase code coverage, but I'm fine with it :)

ildyria:
Those are extreme edge cases where we need to throw exceptions or modify significantly the server on which Lychee run.
That's why I did it as such. :(

PR #1609, opened 2022-11-27
fixes 1608

ildyria:
Fixes #1608 

The problem was that when loading a cover the associated relation to `size_variants` as not eager loaded.
It is unclear why this behaviour appears. I could reproduce it with my live DB but not on my test bench.
Will possibly need more investigation (to figure out what is the triggering cause).

**This patch proposes to always eager load the size-variants when fetching a Picture model (which is pretty much always the case anyway).**

While this code is no longer necessary, it still helps to decrease the memory load and the number of models hydrated.
```php
->withOnly(['size_variants' => (fn (HasMany $r) => self::sizeVariantsFilter($r))])
```

As follows is the trace of the error.
```
{
    "message": "Attempted to lazy load [size_variants] on model [App\\Models\\Photo] but lazy loading is disabled.",
    "exception": "Illuminate\\Database\\LazyLoadingViolationException",
    "file": "~/Lychee/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php",
    "line": 563,
    "trace": [
        {
            "file": "~/Lychee/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php",
            "line": 522,
            "function": "handleLazyLoadingViolation",
            "class": "Illuminate\\Database\\Eloquent\\Model",
            "type": "->"
        },
        {
            "file": "~/Lychee/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php",
            "line": 453,
            "function": "getRelationValue",
            "class": "Illuminate\\Database\\Eloquent\\Model",
            "type": "->"
        },
        {
            "file": "~/Lychee/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php",
            "line": 2218,
            "function": "getAttribute",
            "class": "Illuminate\\Database\\Eloquent\\Model",
            "type": "->"
        },
        {
            "file": "~/Lychee/app/Models/Extensions/Thumb.php",
            "line": 83,
            "function": "__get",
            "class": "Illuminate\\Database\\Eloquent\\Model",
            "type": "->"
        },
        {
            "file": "~/Lychee/app/Relations/HasAlbumThumb.php",
            "line": 276,
            "function": "createFromPhoto",
            "class": "App\\Models\\Extensions\\Thumb",
            "type": "::"
        },
        {
            "file": "~/Lychee/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php",
[...]

Issue #1608, opened 2022-11-27
LazyLoadingViolationException on /api/Albums::get


AnnoyingTechnology:
### Detailed description of the problem [REQUIRED]

An XHR request provokes a 500 Error with the following reponse
```json
{
    "message": "Server Error",
    "exception": "LazyLoadingViolationException"
}
```

But the Exception is caught somewhere so nothing gets logged to `laravel.log`, which doesn't me to debug further.

### Steps to reproduce the issue

1. Install Lychee with php 8.1
2. Open it

### Output of the diagnostics  [REQUIRED]

```
Diagnostics
-----------
Warning: /var/www/domains/REDACTED/public/uploads/ is owned by group www-data, but should be owned by one out of root
Warning: /var/www/domains/REDACTED/public/uploads/thumb2x is owned by group www-data, but should be owned by one out of root
Warning: /var/www/domains/REDACTED/public/uploads/small2x is owned by group www-data, but should be owned by one out of root
Warning: /var/www/domains/REDACTED/public/uploads/medium is owned by group www-data, but should be owned by one out of root
Warning: /var/www/domains/REDACTED/public/uploads/original is owned by group www-data, but should be owned by one out of root
Warning: 7 more directories with wrong owner
Warning: Dropbox import not working. dropbox_key is empty.
Warning: You may experience problems when uploading a photo of large size. Take a look in the FAQ for details.
Warning: You may experience problems when uploading a photo of large size. Take a look in the FAQ for details.
Warning: Default timezone not properly set; you might experience strange results when importing photos without explicit EXIF timezone

System Information
------------------
Lychee Version (git):                    master (1aa57e0) -- Up to date (1 minutes ago).
DB Version:                              4.6.2

composer install:                        --no-dev
APP_ENV:                                 production
APP_DEBUG:                               false

System:                                  Linux
PHP Version:                             8.1.13
PHP User agent:                          Lychee/4 (https://lycheeorg.github.io/)
Timezone:                                UTC
Max uploaded file size:                  2M
Max post size:                           8M
Max execution time:                      0
SQLite Version:                          3.34.1

exec() Available:                        yes
Imagick Available:                       1
Imagick Enabled:                         1
Imagick Version:                         1691
GD Version:                              2.3.3

Config Information
------------------
version:                                 040602
check_for_updates:                       0
sorting_photos_col:                      taken_at
sorting_photos_order:                    ASC
sorting_albums_col:                      max_taken_at
[...]

PR #1610, opened 2022-11-28
Fixes 1609 and 1608

ildyria:
Remove the previous fix of #1609 use correct eager loading


qwerty287:
I'd say we can revert #1609 once we have another fix. It should stay in master for now to not break functionality.

ildyria:
This PR now reverts #1609 and provides proper fix for #1608.

qwerty287 (approved):
Looks good 👍

PR #1627, opened 2022-12-08
remove use of Models in migrations

ildyria:
Avoid using models in the migrations files.
Move to the DB facade instead.


ildyria:
Ready for review :)

database/migrations/2022_11_27_143608_bump_version040603.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
-use App\Models\Configs;
 use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
 
 class BumpVersion040603 extends Migration
```
qwerty287: Can we be consistent about the class definition style? Either named or unnamed classes, but not both if possible. I.e. change this to `return new class() extends Migration`

ildyria: The new migration created by Laravel are all named with `return new class() extends Migration`

I would avoid renaming old migrations as this may trigger a re-run.

qwerty287: I don't think so, Laravel's migrations are based on the filenames AFAIK. Maybe somebody can test this

qwerty287 (approved):
Untested.

database/migrations/2022_11_27_143608_bump_version040603.php
ildyria: I just tested and can confirm your statement.

ildyria:
@qwerty287 due to the large amount of change as of the refactoring to anonymous classes I will have to trouble you again.

qwerty287 (approved):

PR #1629, opened 2022-12-09
Configurable display of the number of subalbums

evoludolab:
This pull request complements LycheeOrg/Lychee-front#363 to display the number of subalbums on subalbum icon (more details there).

### DIFF
--- a/app/Http/Controllers/Administration/SettingsController.php
+++ b/app/Http/Controllers/Administration/SettingsController.php
@@ -372,25 +372,6 @@ class SettingsController extends Controller
 	}
 
 	/**
-	 * Enable display of number of subalbums on subalbum icon.
-	 *
-	 * @param Request $request
-	 *
-	 * @return void
-	 *
-	 * @throws InvalidConfigOption
-	 * @throws BadRequestException
-	 */
-	public function setSubalbumCount(Request $request): void
-	{
-		$request->validate(['show_num_albums' => 'required|boolean']);
-		Configs::set(
-			'show_num_albums',
-			(int) $request->boolean('show_num_albums')
-		);
-	}
-
-	/**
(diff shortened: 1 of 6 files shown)


ildyria:
I do appreciate your work, unfortunately your formula is wrong: it leaks the number of sub-albums (even if hidden).
That is  something I would like to avoid being revealed. However you can probably use the children queries with visibility restrictions and apply a count on those. :)

evoludolab:
@ildyria ...and I was naively thinking that I found an elegant way to efficiently determine the number of subalbums without another database query 😳 Well, I guess it does all that but with the drawback you point out. Even though it does not matter/apply in my use case, I fully agree that this info should not be revealed and, maybe even more importantly, it would be very confusing because the number on the icon would not necessarily match up with the number of subalbums that a user sees.
Your comment was additionally very helpful and possibly saved me a lot of work because I was toying with the thought of doing some bookkeeping to keep track of the number of photos in an album to avoid additional DB queries. Now it is clear that there is no way that this can work 😉 
I will need your comments to sink in some more to determine whether it's worth attempting to fix this PR or better start fresh and maybe do a bigger PR that includes the photo count. Feel free to close this but I for one will it leave it open for now.

evoludolab:
BTW. after giving all this some more thought, I just realized that the problem you mention also applies to the current approach to display a folder badge. More specifically, the variable 'has_albums' uses the isLeaf() method in NodeTrait.php, which works the same way as in my PR (in fact, that's what inspired me). Of course this is only noticeable if _all_ subalbums are hidden for a particular user. In that case they would nevertheless see the folder badge on the enclosing album.

evoludolab:
Time to close this in favour of a new PR.

PR #1634, opened 2022-12-12
Store user CSS in DB instead of file

qwerty287:
Advantages:
- easier code (it's much easier to fetch a DB config entry instead of a file)
- less code (except migration)
- no longer requires `public/dist` to be writable
- more consistent (in code)
- less files to backup (only DB and `public/uploads`)


ildyria:
Do you think 200 characters is enough for a full CSS file?
@SerenaButler

SerenaButler:
> Do you think 200 characters is enough for a full CSS file? @SerenaButler

Absolute No. My css has around 24.000 chars ... yes it is sloppy and redundant :-P

Apart from that I don't really like the idea putting css into the db. It's wrong there imo.

qwerty287:
I understand the issues here, maybe it's also not the correct place. However, I'm not sure if putting it into the `public/dist` as it's now is the best idea.

PR #1635, opened 2022-12-12
add downloads shield

ildyria:


d7415 (approved):
Why not.

Aside: I find it weird that shields.io have this (which GitHub doesn't make obvious) but not clones, which are shown in Insights.

qwerty287 (approved):

PR #1649, opened 2022-12-22
Remove and clean up unused files

qwerty287:
- remove unused font files
- convert `cat.jpg` to WEBP (much smaller size, about 1MB less at all)
- removed JS source maps: If you don't think this is good, I can undo it. However, I'm not sure if we should keep them because they are limited to two source maps, while the `frontend.js` has six references. In addition, this copied the *full source code* of the leaflet.markercluster library, which is definitely useless.


ildyria (approved):
Tested. LGTM.

PR #1655, opened 2022-12-23
Add number of foreign key check and listing option

ildyria:
Add in the diagnostic:
- the current foreign keys in the DB (disabled by default).
- a count of the foreign keys in the DB


ildyria:
I want to check something and I would be happy to get this merged so I can use it in #1539  and check the migration see if there are any foreign keys dropped. :)

ildyria:
Unfortunately, I do not have a DB on postgresql to test and find the correct query :(

qwerty287 (approved):
I also don't have a postgres DB, but it looks fine. If possible, we should add PgSQL support ASAP.

PR #1660, opened 2022-12-24
New install page for admin + redirection

ildyria:
- add install step in the web installer screen.
- redirect to that page if an admin user does not exists
- add various securities.

### DIFF
--- a/app/Http/Controllers/Install/SetUpAdminController.php
+++ b/app/Http/Controllers/Install/SetUpAdminController.php
@@ -11,10 +11,15 @@ use Illuminate\Routing\Controller;
 use Illuminate\Support\Facades\Hash;
 
 /**
- * Class MigrationController.
+ * Class SetUpAdminController.
  */
 class SetUpAdminController extends Controller
 {
+	/**
+	 * Called on GET request.
+	 *
+	 * @return View
+	 */
 	public function init(): View
 	{
 		return view('install.setup-admin',
@@ -26,7 +31,8 @@ class SetUpAdminController extends Controller
 	}
 
 	/**
-	 * Migrates the Lychee DB and generates a new API key.
+	 * Set up the admin user.
+	 * Called on POST request.
 	 *
 	 * @return View
 	 *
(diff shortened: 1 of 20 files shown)


app/Http/Controllers/Install/SetUpAdminController.php
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * Set up the admin user.
+	 * Called on POST request.
+	 *
+	 * @return View
+	 *
+	 * @throws FrameworkException
+	 */
+	public function view(SetUpAdminRequest $request): View
```
qwerty287: Don't call this `view`. The other method should be `view`, and this one something like `create` or `post`.

resources/views/install/setup-admin.blade.php
```diff
@@ -0,0 +1,46 @@
...
+			<div class="buttons-container">
+				<button type="submit" class="button float-right" >
+					<i class="fa fa-check fa-fw" aria-hidden="true"></i>
+					Create admin account
+					<i class="fa fa-angle-double-right fa-fw" aria-hidden="true"></i>
+				</button>
+			</div>
+		</form>
+	@else
+		<strong>Admin account has been created.</strong>
```
qwerty287: Move to new blade template if it's successful.

app/Http/Controllers/Install/SetUpAdminController.php
ildyria: Fixed [9ff238b](https://github.com/LycheeOrg/Lychee/pull/1660/commits/9ff238b976bd10c36d8dbc7f98e9d8696abd80d8).

resources/views/install/setup-admin.blade.php
ildyria: Fixed [9ff238b](https://github.com/LycheeOrg/Lychee/pull/1660/commits/9ff238b976bd10c36d8dbc7f98e9d8696abd80d8).

qwerty287 (approved):
Looks good, but didn't test myself.

ildyria:
> Looks good, but didn't test myself.

I did :)

ildyria:
When no admin exists, we are redirected here:
image

When DB is not writable or for some other reason:
image

On success:
image

When trying to access again:
image

qwerty287 (approved):

PR #1668, opened 2022-12-26
DTO → Resources

### LINKED ISSUE #1642: [Enhancement] Use Laravel resources instead of homemade DTO
See https://laravel.com/docs/9.x/eloquent-resources

Laravel Daily video:
[IMAGE ALT TEXT](http://www.youtube.com/watch?v=sAscQPIm_jY&t=200s)

While this would imply (yet again) a significant refactoring,
it would clean up and clarify what the expected returns are.

ildyria:
Fixes #1642 

Leverage of Laravel Resources for a more explicit attributes in responses. No hidden surprises.

- [x] : add json missing tests for smart albums as we no longer return null.


Makefile
```diff
@@ -102,3 +103,11 @@ gen_major:
 
 release_major: gen_major
 	git commit -m "bump to version $(shell cat version.md)"
+
+TESTS_PHP := $(shell find tests/Feature -name "*Test.php" -printf "%f\n")
+TEST_DONE := $(addprefix build/,$(TESTS_PHP:.php=.done))
```
qwerty287, line 108: What's this for? Can you explain a bit please?

ildyria: When I was debugging each tests, rather than executing the full test suite, I was running `make all_tests`.

`TESTS_PHP` contains a list of all the tests files.
`TEST_DONE` generates a list of `build/testName.done` from `TESTS_PHP`

The recipe `all_tests` depends on all the `TEST_DONE` files and thus generates them.
those are generated with the recipe:
```makefile
build/%.done: tests/Feature/%.php
	vendor/bin/phpunit --filter $* && touch build/$*.done
```
which will run phpunit only for that specific test file and generate a done on success.

As a result, by doing `make all_tests` I will be executing all the tests, sequentially. This allows me to go through all the tests one by one and debug them each one at time rather than all in one go.
Modifying a test file also immediately triggers rebuilding the associated done file.

app/Models/Extensions/ToArrayThrowsNotImplemented.php
```diff
@@ -0,0 +1,19 @@
...
+
+use App\Exceptions\Internal\NotImplementedException;
+
+/**
+ * Trait ToArrayThrowsNotImplemented.
+ *
+ * Now that we use Resources toArray should no longer be used.
+ * Throw an exception if we encounter this function in the code.
+ */
+trait ToArrayThrowsNotImplemented
```
qwerty287, line 13: Is this really necessary? Calling a non-existing function should throw an exception anyways, and I don't see why we need something like this.

ildyria: Unfortunately yes. The reason is that `Model` are of the type `Arrayable`, hence provide the method `toArray` by default which will convert the attributes except those mentioned in `$hidden`.
In theory, the method provided by this trait should not be reached in any code path, but I would like to really ensure this does not happen at all in a hidden array cast that we did not think of. :)

qwerty287 (approved):
Tested, seems working.

ildyria:
Fun fact, this fixes the broken front-end when doing searches.

PR #1698, opened 2023-01-08
Fix 2fa tests

ildyria:
This will make sure that 2FA is properly executed.
Values have been generated locally from a yubikey 4.


qwerty287 (approved):

ildyria:
Will need to fix it a bit more :')

ildyria:
Now all tests should pass. :)

PR #1714, opened 2023-01-16
Improve tests to also use users.

### LINKED ISSUE #634: [Enhancement] Add Regression tests.

ildyria:
Fixes #634  - Last tests missing

**Still unsure we want those changes as they complexify the tests quite a bit...**

Change can be summarized as follows:
tests requiring a logged user via `setUp` are now using an abstract method which provides which user to run.
The rest of those change is to accomodate this abstract method.

### DIFF
--- a/tests/Feature/Abstract/AbstractPhotosAddHandlerGDTest.php
+++ b/tests/Feature/Abstract/AbstractPhotosAddHandlerGDTest.php
@@ -10,16 +10,17 @@
  * @noinspection PhpUnhandledExceptionInspection
  */
 
-namespace Tests\Feature;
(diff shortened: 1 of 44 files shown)


qwerty287 (approved):
Looks good to me.

ildyria:
Decided against it as it complexifies the test significantly and thus decrease the ability to track down what went wrong.
Tests should be easy to reproduce & understand, this is not something that does that.

PR #1724, opened 2023-01-21
Add configuration option to set auth guard driver

r7l:
Lychee extends the auth guard driver to be able to use session and token at the same time. It also defaults to this option by setting it to: `session-or-token` but does not provide an option for users to set this driver when needed.

Using tokens breaks Lychee when running behind a basic auth from Apache / Nginx with a user not related to Lychee and will only display a `BadRequestHeaderException` error.

This PR adds an option called `LYCHEE_AUTH_GUARD` which will default to `session-or-token` as it is the default already. It allows users with select between `session`, `token` and `session-or-token`.

I will add another PR for the docs.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -60,6 +60,16 @@ DB_LIST_FOREIGN_KEYS=true
 # LYCHEE_DIST_URL="dist/"
 # LYCHEE_SYM_URL="sym/"
 
+# Authentication guard. This defines how the users are actually retrieved
+# out of your database or other storage mechanisms used by this application
+# to persist your user's data.
+# Possible values are:
+#
+# - session: user is retrieved from session
+# - token: user is retrieved from token like basic auth
+# - session-or-token: uses both (the default)
+# LYCHEE_AUTH_GUARD="session-or-token"
+
 CACHE_DRIVER=file
 SESSION_DRIVER=file
 SESSION_LIFETIME=120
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -60,6 +60,16 @@ DB_LIST_FOREIGN_KEYS=true
 # LYCHEE_DIST_URL="dist/"
 # LYCHEE_SYM_URL="sym/"
 
+# Authentication guard. This defines how the users are actually retrieved
+# out of your database or other storage mechanisms used by this application
+# to persist your user's data.
+# Possible values are:
+#
+# - session: user is retrieved from session
+# - token: user is retrieved from token like basic auth
```
qwerty287: I don't think recommending `token` is a good idea. It would be better to remove it in favor of `session` and `session-or-token`.

If there are only two possibilities left, we can also move it to a boolean variable like `ENABLE_TOKEN_AUTH` which can then also be used to hide the UI elements.

r7l: It is already recommended in [auth.php](https://github.com/LycheeOrg/Lychee/blob/master/config/auth.php#L33) but it seems to be a default text from Laravel as i can see the same text in other projects as well. I don't mind the options and possibilities as long as it does allow to disable tokens.

If you have it a boolean it would require some logic in auth.php to be added. My PR currently just allows any of the existing settings to be set.

qwerty287: You don't need that much logic: `env('ENABLE_TOKEN_AUTH', true) ? 'session-or-token' : 'session'`

qwerty287 (approved):

ildyria (approved):
LGTM.

PR #1730, opened 2023-01-23
Fix wording for authentication guard option

r7l:
As discussed [here](https://github.com/LycheeOrg/LycheeOrg.github.io/pull/85), the wording was approved by mistake. This PR will fix it.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -60,8 +60,7 @@ DB_LIST_FOREIGN_KEYS=true
 # LYCHEE_DIST_URL="dist/"
 # LYCHEE_SYM_URL="sym/"
 
-# Support for token based authentication like basic auth. Enabled by default.
-# Disable when running Lychee behind 3rd party managed authentication.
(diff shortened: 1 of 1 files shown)


.env.example
```diff
@@ -60,8 +60,7 @@ DB_LIST_FOREIGN_KEYS=true
 # LYCHEE_DIST_URL="dist/"
 # LYCHEE_SYM_URL="sym/"
 
-# Support for token based authentication like basic auth. Enabled by default.
-# Disable when running Lychee behind 3rd party managed authentication.
+# Support for token based authentication used by API requests. Enabled by default.
 # ENABLE_TOKEN_AUTH="true"
```
d7415: Question for the room: For consistency, should we drop the quotes?
```suggestion
# ENABLE_TOKEN_AUTH=true
```

r7l: You're right. That was a leftover from being a string value at first.

ildyria (approved):
LGTM.

Issue #1732, opened 2023-01-25
[Enhancement] Consistent URLs when Landing Page is disabled


nicokaiser:
When the Landing Page is enabled, the Gallery and all its URLs are `https://example.com/gallery#xyzzy` and `https://example.com/gallery#xyzzy/aabbcc`.
When the Landing Page is disabled, the URLs are `https://example.com/#xyzzy` and `https://example.com/#xyzzy/aabbcc`, but this is only due to the IndexController also delivering the Gallery in this case. The `gallery#` URLs still seem to be the canonical URLs.

Also, redirects are going to the `gallery#` versions. `https://example.com/r/xyzzy` -> `https://example.com/gallery#xyzzy`
When enabling the Landing Page at a later time, previous URLs like `https://example.com/#xyzzy` do not work anymore.

Wouldn't it be more consistent to have all URLs generated by the frontend include the `gallery` prefix? There are two options:

1. Currently IndexController renders the frontend when Landing Page is disabled. This could be replaced by a `redirect('gallery')`. This would be the "cleanest" way to achieve this, however when using this option, there is not real "homepage" anymore (because `/` redirects to `/gallery`). (The return type of the method also needs to be adjusted)

https://github.com/LycheeOrg/Lychee/blob/dc33883d76a3d76bf961ca29c9cacc108b7fc9d8/app/Http/Controllers/IndexController.php#L68

```diff
- 			return $this->frontend();
+ 			return redirect('gallery');
```

or 2. Change `lychee.goto` so it includes the prefix:

https://github.com/LycheeOrg/Lychee-front/blob/5ade25d897bc04df9cab15bda0dc0fb6c5e0b16a/scripts/main/lychee.js#L756

```diff
- 	url = "#" + (url !== null ? url : "");
+ 	url = "gallery#" + (url !== null ? url : "");
```

I'm not sure if the latter has implications on other URLs generated using `lychee.goto`, however I guess (!) this method is only used to generate album or photo URLs.

ildyria:
Will be fixed with the Livewire front-end I am working on.
This will be made consistent.

nicokaiser:
Great to hear! Anything I can help with on the Livewire task?

ildyria:
I am currently doing the header modal forms and will be working on the sidebar editing soon after, though I am considering adding a drop down-setting effect on the Album page to give a view of all modifiable settings in one go.

[...]

PR #1738, opened 2023-01-26
Fix URLs in RSS feeds (#1732)

nicokaiser:
URLs in RSS feeds are "https://example.com/#xyzzy/aabbcc".

This works in default installations, however the canonical URL to albums (and photos) starts with "/gallery", so these URLs should be
"https://example.com/gallery#xyzzy/aabbcc".

When using the landing page, URLs even _need_ to include "gallery", otherweise all of the RSS URLs only show the landing page.


nicokaiser:
See also [Lychee-front#379](https://github.com/LycheeOrg/Lychee-front/pull/379)

qwerty287 (approved):

Issue #1740, opened 2023-01-28
[Enhancement] Set theme-color


nicokaiser:
This is a small and subtle tweak, but it looks very good on platforms where the browser honors the `theme-color` meta property.

```html
<meta name="theme-color" content="#222">
```

`#222` is the color on the top of the toolbar gradient, so by setting the theme-color to this value, the toolbar transitions smoothely into the browser's frame, which makes the gradient look much better.

This can be seen on the right side of this screenshot (it is even better visible on mobile devices like Mobile Safari on iOS):

![theme-color](https://user-images.githubusercontent.com/238631/215284082-8ffa8168-8baa-45ad-bc87-ec7000b02c44.jpg)

nicokaiser:
![theme-color-ios](https://user-images.githubusercontent.com/238631/215284787-661823b7-9061-4ea2-883b-a6eebe9f0f03.jpg)

PR #1741, opened 2023-01-28
Set theme-color on frontend template (#1740)

nicokaiser:
See #1740.


ildyria (approved):

ildyria:
> See #1740.

In the future, instead of _See xxx_ use :
> Fixes #1740

That way, when we merge it automatically closes the associated Issue. :)

PR #1743, opened 2023-01-30
Livewire: Landing page

nicokaiser:
Initial implementation of the Landing page.

- The footer does not work (you changed the CSS classes, I could not figure out how the flex classes need to be set)
- The Gallery link should probably use `PageMode::GALLERY->value` instead of hardcoded `gallery`
- The Landing page must not be accessible when disabled (currently it is still there under `/livewire/landing`, which must be avoided)

I'm not entirely sure why the `PageMode` construct is needed, as opposed to proper routes (but this may be because I don't have any experience in Livewire yet).


ildyria:
> I'm not entirely sure why the `PageMode` construct is needed, as opposed to proper routes (but this may be because I don't have any experience in Livewire yet).

The reason is because we want to be Single Page Application. Everything goes via the `Index.php` livewire component.

ildyria:
I checked the previous version, and you are right. Need to rethink the landing page integration.

PR #1756, opened 2023-02-09
Fixes locale not working (#1754)

### LINKED ISSUE #1754: Language support stopped working

Frontend languages stopped working

1. Change Lychee language to something other than "en"
2. Open Lychee
3. Lychee is still english

I think this might got broken with https://github.com/LycheeOrg/Lychee/pull/1668 but I am not sure. It has worked some versions ago, but I cannot really tell when.

ildyria:
Fixes #1754 

When moving to Resources, this was done at the same time as migrating the locale to `lang/`, this induced a small bug.
Should now be resolved.

### DIFF
--- a/app/Http/Resources/InitResource.php
+++ b/app/Http/Resources/InitResource.php
@@ -34,7 +34,7 @@ class InitResource extends JsonResource
 			$gitHubVersion->hydrate();
 		}
 
-		// we also return the local
+		// we also return the locale
 		$locale = include base_path('lang/' . app()->getLocale() . '/lychee.php');
 
 		return [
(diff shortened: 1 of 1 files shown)


app/Http/Resources/InitResource.php
```diff
@@ -34,12 +34,16 @@ public function toArray($request): array
 			$gitHubVersion->hydrate();
 		}
 
+		// we also return the local
```
qwerty287:
```suggestion
		// we also return the locale
```

qwerty287 (approved):

PR #1757, opened 2023-02-09
Add WebAuthn.js to frontend.js (#1710)

### LINKED ISSUE #1710: WebAuthn.js does not work on older Safari versions
Since (about) version 4.7, Lychee does not work in older Safari browsers. Safari <= 13 is present on older iOS versions, and most probably some older android devices are also affected.

The culprit seems to be the new WebAuthn.js, which uses private class methods, and which is directly committed into the dist folder (and thus not being transpiled by the Babel task in Lychee-front.

Has this been a deliberate choice, or should the files in dist also be somewhere else and included in the transpilation step?

I‘d happy to help with a PR, but am not yet familiar with the interweaved structure of Lychee and Lychee-front…

nicokaiser:
WebAuthn is transpiled and compiled into frontend.js and not included separately. Reduces clutter and additional requests.

Requires https://github.com/LycheeOrg/Lychee-front/pull/381 to be merged first.

Fixes #1710


qwerty287 (approved):
Besides the missing frontend sync (and the issues there) LGTM

qwerty287:
See #1769

PR #1758, opened 2023-02-09
Set noindex for error responses

### LINKED ISSUE #342: [Enhancement] SEO optimization 2/3 - Set `<meta name="robots" content="noindex">` for 404 responses
Currently, our web frontend is not very SEO friendly. It is entirely written in JS, which is fine for modern web crawlers, but it violates some best practices.

**Enhancement:** If the frontend attempts to load an album or photo which does not exist, the frontend currently only shows the `404` response from the backend as an error message to the user. On top, the JS code should also add a `<meta name="robots" content="noindex">` to error pages using JavaScript.

See: https://developers.google.com/search/docs/crawling-indexing/javascript/javascript-seo-basics#avoid-soft-404s.

nicokaiser:
Fixes https://github.com/LycheeOrg/Lychee-front/issues/342

### DIFF
--- a/resources/views/error/error.blade.php
+++ b/resources/views/error/error.blade.php
@@ -3,8 +3,9 @@
     <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1">
+        <meta name="robots" content="noindex">
 
-        <title>Error </title>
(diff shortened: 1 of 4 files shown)


ildyria (approved):
LGTM.

PR #1764, opened 2023-02-15
Migrate to Laravel 10

ildyria:


app/Actions/Diagnostics/Pipes/Infos/SystemInfo.php
```diff
@@ -24,8 +24,14 @@ public function handle(array &$data, \Closure $next): array
 			};
 
 			$dbtype = $sql[0];
-			$results = DB::select(DB::raw($sql[1]));
```
ildyria, line 27: https://laravel.com/docs/10.x/upgrade#database-expressions
`DB::raw()` type is no longer compatible with `DB::select()`

app/Models/Configs.php
```diff
@@ -221,6 +221,7 @@ public static function getValueAsBool(string $key): bool
 	 */
 	public static function getValueAsEnum(string $key, string $type): \BackedEnum|null
 	{
+		// @phpstan-ignore-next-line
```
ildyria, line 225: Due to typehinting on getValueAsEnum, it is assumed that `$type` will be a BackedEnum, this is not necessarily the case... in practice.

d7415: Is it worth putting this comment in the file instead?

ildyria: Fixed [7f6ffca](https://github.com/LycheeOrg/Lychee/pull/1764/commits/7f6ffcaf4f3eb412af1e97be894ae779df82dd85).

composer.json
```diff
@@ -45,10 +49,10 @@
         "doctrine/dbal": "^3.1",
         "geocoder-php/cache-provider": "^4.3",
         "geocoder-php/nominatim-provider": "^5.5",
-        "laragear/webauthn": "1.x-dev",
-        "laravel/framework": "^9.2",
-        "lychee-org/nestedset": "^6",
-        "lychee-org/php-exif": "^0.8",
+        "laragear/webauthn": "laravel-10-dev",
```
qwerty287: There's a release now: https://github.com/Laragear/WebAuthn/releases/tag/v1.2.0

ildyria: Yeah I really appreciate that this maintainer pretty much use my PR as template and just rewrite everything...
I still need to investigate why some parts of U2F are not working. I suspect it is because it is limited to a subset of cryptographic key types (namely RSA and no ECDSA).

ildyria:
Let's get this done. :)

composer.json
ildyria: @qwerty287 Fixed in [1b99e6a](https://github.com/LycheeOrg/Lychee/pull/1764/commits/1b99e6ac6b1553dfd80db47526d98f364f7110cf)

qwerty287 (approved):
Tested, seems working.

PR #1766, opened 2023-02-22
Add setting `auto_fix_orientation` to enable auto image rotation

wladif:
Currently, imported images are automatically rotated even when the editor_enabled is disabled, the automatic rotation takes place as part of the image normalization process.
 This feels a bit unintuitive and this PR fixes it by taking into account this setting when normalizing the image.

### DIFF
--- a/app/Actions/Photo/Strategies/AddStandaloneStrategy.php
+++ b/app/Actions/Photo/Strategies/AddStandaloneStrategy.php
@@ -21,6 +21,7 @@ use App\Image\Handlers\GoogleMotionPictureHandler;
 use App\Image\Handlers\ImageHandler;
 use App\Image\Handlers\VideoHandler;
 use App\Image\StreamStat;
+use App\Models\Configs;
 use App\Models\Photo;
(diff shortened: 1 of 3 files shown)


ildyria:
No, that **setting is to allow the manual rotations buttons in the header.**
It is completely unrelated to automatic rotation on upload.

A better approach would be to create an `auto_fix_orientation` setting (defaulted to 1) in the Image Processing category and use it instead.

You can use `php artisan make:migration add_auto_fix_orientation_setting` to create a stub for the database migration.

wladif:
I've avoided adding another setting because to me it made more sense that if I, as a user, can't edit the image the system should not be allowed to do it either, this will not be the case with an extra setting (not without extra logic anyway). 
I'll update the PR to add the extra setting. Thanks!

ildyria:
@wladif I fixed the formatting issue (`make formatting`) could you confirm this is ready for review ? :)

ildyria (approved):
LGTM.

wladif:
@ildyria Thanks for fixing the formatting. Yes, it's ready for review.

PR #1769, opened 2023-03-04
Compile WebAuthn.js into frontend.js

qwerty287:
Recreating #1757 with synced frontend.

Credits @nicokaiser 

(The frontend.js file has been compressed thus there are 14000 deleted lines)


ildyria (approved):
LGTM.

PR #1773, opened 2023-03-08
`composer update`

qwerty287:


ildyria:
For phpstan, add 
```php
/** @var NsCollection<Album> $albums */
/** @var ?NsCollection<Album> $sharedAlbums */
```
before line 
```php
list($albums, $sharedAlbums) = $albums->partition(fn (Album $album) => $album->owner_id === $userID);
```
The type inference by partition is not correct.

ildyria (approved):
LGTM once the phpstan problem is fixed.

PR #1785, opened 2023-04-06
Add jobs processing

ildyria:
Instead of processing live, we now support the use of Jobs & Queues.
This ensure faster response when jobs are enabled.

This pull request adds a new config `use_job_queues` with default to `0`.
By setting it to `1`, import will be added to a processing queue.

Such jobs on the queue can then be executed by running `php artisan queue:work` on server side.


qwerty287 (approved):
Seems good, untested

qwerty287:
Maybe we can add some tests for this

ildyria:
> Maybe we can add some tests for this

Done.

I actually noticed that we can just switch everything to jobs if we are using the `sync` for queue connection. The problem is that the code is no longer dependent if it is a duplicate or not. :/

qwerty287 (approved):

PR #1789, opened 2023-04-10
doc blocks + fix deprecation utf8_encoding

ildyria:


app/Models/BaseAlbumImpl.php
```diff
@@ -100,10 +100,14 @@
  * @property bool                       $grants_full_photo_access
  * @property bool                       $grants_download
  * @property Collection                 $shared_with
+ * @property int|null                   $shared_with_count
  * @property string|null                $password
  * @property bool                       $is_password_required
  * @property PhotoSortingCriterion|null $sorting
  * @property AlbumProtectionPolicy      $policy
+ * @property int                        $is_share_button_visible  // NOT USED
```
qwerty287, line 108: Why do we have this?

ildyria: I just know that it should be removed as we are no longer using it in our settings and we just use a global setting everywhere.
The update of doc block is mostly a commit of `php artisan ide_helper:models`.

qwerty287: Can you just remove it or is it still necessary?

ildyria: I think there was a risk of column removal which would mess up with the sqlite DB. :/

qwerty287 (approved):

PR #1795, opened 2023-04-15
Hardcode types rather than @return in migrations

ildyria:
Small code style.


qwerty287 (approved):
Thanks, I would have asked you to split this out anyways 😂

PR #1819, opened 2023-05-03
[Enhancement] Make it possible to fetch random image without needing to be public & starred

### LINKED ISSUE #1814: [Enhancement] Make it possible to fetch random image without needing to be public & starred
Make it configurable if the random method should include all accessible photos.

mingan666:
Closes #1814 

Adds new configuration for the album id from which random images are chosen.
The new config will default to "starred" which gives the same functionality as before.


ildyria (approved):
LGTM.

on4r4p:
Im sorry but i dont get how does this work ?
frame is enable in my settings but there is no slideshow when i go to starred

ildyria:
You need to go to `http://<your-lychee>/frame`

Issue #1729, opened 2023-01-23
[Enhancement] Use LastModified date when no exif data are found


ildyria:
Original issue: #1727

![See here](https://user-images.githubusercontent.com/59219907/214068700-18ce9ef4-49ba-40cb-8cd4-685e4e5c2123.png)

When uploading a file, we have access to extra meta-data. We can use those to fill up dates when provided.

- [x] : Fetch data and replace when no Exif data are provided
- [x] : Add config option to enable / disable this functionality.

wladif:
@ildyria I have some free time this days and I could try to implement this. Is it ok with you?

ildyria:
Go for it ! :) 
It is written as "project for voluneers" so pretty much up for grab ! :)

wladif:
@ildyria With the current php-exif lib I'm having a hard time figuring out if an images has Exif data or not. As far as I can tell, when Imagemagick is used, the library will always return some an Exif instance, even if the image doesn't have exif. I find this to be a bit confusing. Is this something intended or is there a plan to refactor it? How can I check if a file has actual exif data or not ? Thanks

ildyria:
Thanks for pointing that one out. I fixed the return type of the reader: https://github.com/LycheeOrg/php-exif/pull/62 to reflect this behavior.

Yes this behavior is working as expected. Exif data is also size etc...
What matters in your case is whether some of the data in the Exif object are there or not.
See here https://github.com/LycheeOrg/php-exif/blob/master/lib/PHPExif/Exif.php

wladif:
@ildyria There is already a config for this functionality. The config was included in #1821, it's called: `use_last_modified_date_when_no_exif_date`.

PR #1821, opened 2023-05-04
Add backend implementation to use file's last modified time

wladif:
This PR adds the remaining back-end logic for using the file's last modified time when there's no Exif creation time thus closing https://github.com/LycheeOrg/Lychee/issues/1729.


app/Http/Requests/Photo/AddPhotoRequest.php
```diff
@@ -34,11 +35,22 @@ protected function processValidatedValues(array $values, array $files): void
 		$this->album = $albumID === null ?
 			null :
 			$this->albumFactory->findAbstractAlbumOrFail($albumID);
+		$this->fileLastModifiedTime = $this->getFileLastModifiedTimeInSeconds($values);
```
ildyria:
```suggestion
		// Get the File Last Modified attributes in seconds instead of milliseconds.
		$this->fileLastModifiedTime = $values[RequestAttribute::FILE_LAST_MODIFIED_TIME] / 1000;
```

app/Http/Requests/Photo/AddPhotoRequest.php
```diff
@@ -34,11 +35,22 @@ protected function processValidatedValues(array $values, array $files): void
...
 	{
 		return $this->file;
 	}
+
+	public function fileLastModifiedTime(): int
+	{
+		return $this->fileLastModifiedTime;
+	}
+
+	private function getFileLastModifiedTimeInSeconds(array $values): int
```
ildyria: This function is not really necessary.

tests/Feature/LibUnitTests/PhotosUnitTest.php
```diff
@@ -46,6 +46,7 @@ public function upload(
 		$response = $this->testCase->post(
 			'/api/Photo::add', [
 				'albumID' => $albumID,
+				'fileLastModifiedTime' => 1678824303000,
```
ildyria: what happens when this value is 0 ?
Another test to add is to also check if we can have an image without that creation date exif data and check that the associated value is as expected.

app/Actions/Photo/Create.php
```diff
@@ -48,15 +48,16 @@ public function __construct(?ImportMode $importMode, int $intendedOwnerId)
...
+	 * @param int                $fileLastModifiedTime the timestamp to use if there's no creation date in Exif
+	 * @param AbstractAlbum|null $album                the targeted parent album
 	 *
 	 * @return Photo the newly created or updated photo
 	 *
 	 * @throws ModelNotFoundException
 	 * @throws LycheeException
 	 */
-	public function add(NativeLocalFile $sourceFile, ?AbstractAlbum $album = null): Photo
+	public function add(NativeLocalFile $sourceFile, int $fileLastModifiedTime, ?AbstractAlbum $album = null): Photo
```
ildyria: I may be wrong but I think the function should be defined as :
```suggestion
	public function add(NativeLocalFile $sourceFile, ?int $fileLastModifiedTime = null, ?AbstractAlbum $album = null): Photo
```
And in the first line inside the function:
```php
if ($fileLastModifiedTime === null) {
  // Retrieve the filemtime() from NativeLocalFile
}
```

Nevertheless, I would make sure that $album is actually needed to be optional. From all the call I saw in that PR, it was always provided. This leads me to think that the order of arguments should be switched to:
1. `NativeLocalFile`
2. `?AbstractAlbum`
3. `int` or `?int = null`
The reason being that it would follow the same convention as for `ProcessImageJob` constructor.

wladif: Do you mean $fileLastModifiedTime ? Because $album is already optional...

tests/Feature/LibUnitTests/PhotosUnitTest.php
wladif: Sure, I'll add one. The expected behavior would be that in case of 0, the create date will be 1st January, 1970 (because we're processing the fileLastModifiedTime as an UNIX timestamp).

app/Actions/Photo/Create.php
ildyria: yes, but I am not sure that $album is actually really optional. It seems to me that in every call add() it is already defined.

wladif: I can try to make it mandatory, but I would do it in a different PR as it's not really related to the enhancement. For now I've made fileLastModifiedTime optional.

ildyria: https://github.com/LycheeOrg/Lychee/pull/1822 :)

app/Console/Commands/ExifLens.php
```diff
@@ -70,7 +71,7 @@ public function handle(): int
 			foreach ($photos as $photo) {
 				try {
 					$localFile = $photo->size_variants->getOriginal()->getFile()->toLocalFile();
-					$info = Extractor::createFromFile($localFile);
+					$info = Extractor::createFromFile($localFile, filemtime($localFile->getRealPath()));
```
ildyria, line 74: no longer necessary :)

app/Actions/Photo/Create.php
wladif: @ildyria I do have a question regarding #1822: didn't you just removed the default value? $album is still of type ?AbstractAlbum which is still nullable (and optional)? Sorry if this sounds dumb, but my PHP knowledge is not very strong... Thanks!

ildyria: There are no dub questions. :)
`?AbstractAlbum` is still nullable but no longer optional.
```php
// This is valid php
function method(String $arg1, ?String $arg2, ?String $arg3 = 'default'): void {
	echo $arg1;
	echo "\n";
	echo $arg2 === null ? 'NULL' : $arg2;
	echo "\n";
	echo $arg3 === null ? 'NULL' : $arg3;
};
```

You can call it like this:
```php
method('arg1', 'arg2', 'arg3');
/** This will print
 * arg1
 * arg2
 * arg3
 */

method('arg1', 'arg2', null);
/** This will print
 * arg1
 * arg2
 * NULL
 */

method('arg1', null, 'arg3');
/** This will print
 * arg1
 * NULL
 * arg3
 */
```

And because the last parameter is optional = it has a default value.
```php
method('arg1', 'arg2');
/** This will print
 * arg1
 * arg2
 * default
 */


method('arg1', null);
/** This will print
 * arg1
 * NULL
 * default
 */
```

// This is not valid:
// method('arg1');

However this is not a valid php definition:
```php
// This is valid php
function method(String $arg1, ?String $arg2 = 'defaultarg2', ?String $arg3): void {
	echo $arg1;
	echo "\n";
	echo $arg2 === null ? 'NULL' : $arg2;
	echo "\n";
	echo $arg3 === null ? 'NULL' : $arg3;
};
```
Optional parameters must be defined after mandatory ones.

Hope this clarifies it?

ildyria (approved):
LGTM.
Did some very very minor changes to align with the repo (e.g. test constants ordered alphabetically).

app/Actions/Photo/Create.php
wladif: Things are more clear now! Thank you for the explanation!

PR #1828, opened 2023-05-07
Fix left/right in photo view of tag albums

ildyria:
Someone reported on Gitter that the slideshow was not working on Tag Albums.
This should fix it.

### DIFF
--- a/app/Http/Resources/Models/TagAlbumResource.php
+++ b/app/Http/Resources/Models/TagAlbumResource.php
@@ -3,6 +3,7 @@
 namespace App\Http\Resources\Models;
 
 use App\DTO\AlbumProtectionPolicy;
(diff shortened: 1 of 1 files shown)


qwerty287 (approved):
Test to prevent regression?

ildyria:
> Test to prevent regression?

I can try to create a new PR for that.

jinwater88:
Is the memory function in iPhone photo album the same as the slide show here? What technology is used to achieve it? Moviepy?

ildyria:
> Is the memory function in iPhone photo album the same as the slide show here? What technology is used to achieve it? Moviepy?

I don't have an iphone so I am not sure what you are talking about. Could you clarify your question?

PR #1832, opened 2023-05-10
Update version of phpcs

ildyria:


qwerty287 (approved):

qwerty287:
Maybe we could try to determine the version dynamically using something like `jq`.

PR #1846, opened 2023-05-15
Remove homemade Log solution, add support log-viewer

### LINKED ISSUE #1843: [Enhancement] Log migration
Considering nuking the logs table (remnant from v3) and migrating all the logs to the Laravel log system.

This would also allow us to use https://github.com/opcodesio/log-viewer

Two step migration:
- [x] : Integrate [log-viewer](https://github.com/opcodesio/log-viewer)
- [x] : ~~Migrate the log data.~~

ildyria:
Fixes #1843 

Because https://github.com/opcodesio/log-viewer spammed their interface with "support me", buy me coffee etc... We had to fork their solution : See [here](https://github.com/LycheeOrg/log-viewer).
Quite unfortunate as I would have let it in if they were a bit more discreet.

This does not contain a Migration that removes the Log table. Rationale being that some user may want to be able to still access those data in a near future. We can consider dropping those at the next minor version.


qwerty287 (approved):
Did some basic tests, seems to work. Code looks good, but it seems like a migration is missing that removes the `logs` table. Or was this intentional?

ildyria:
> This does not contain a Migration that removes the Log table. Rationale being that some user may want to be able to still access those data in a near future. We can consider dropping those at the next minor version.

:D

qwerty287:
Sorry, I should read the description...

PR #1861, opened 2023-05-21
remove hashes from CSP when using log-viewer

ildyria:


ildyria:
Tested with Docker deployments of zip archive for clean environment. Everything works (finally).

qwerty287 (approved):

Issue #1652, opened 2022-12-23
[Enhancement] import from URL: get type from Content-Type header


qwerty287:
*Explain in a few words which functionality or improvements you would like to see in Lychee.*
Try to import an image without `.jpg` or whatever extension. It will fail because it can't get the type of it. It should also check the Content-Type header.

ildyria:
Seems this is down thanks to @wladif #1873

PR #1863, opened 2023-05-25
Add a sample image without extension that will be used in tests.

wladif:
Part of the fix for #1652. Prerequisite for the PR #1864 .


qwerty287 (approved):
I'm fine with this, but if it's unused for now I'd rather do it in #1864 directly.

wladif:
> I'm fine with this, but if it's unused for now I'd rather do it in #1864 directly.

I would prefer the same, but I don't know an easy way to make the test pass on both builds for PR & for merge into master (with just one PR and an image stored in this repository...)
I was inspired by [this](https://github.com/LycheeOrg/Lychee/blob/a3423b67c8e1e2f965bfde5714b69b3ba01f50a2/tests/Feature/Constants/TestConstants.php#LL19C5-L19C5) which I believe points to the master branch.

PR #1870, opened 2023-06-06
feat(#1869): Allow script-src and connect-src configuration

### LINKED ISSUE #259: [Enhancement] Counters for Number of Views

### LINKED ISSUE #1869: [Enhancement] Allow dynamic configuration of secure headers
As discussed in #259 Lychee already provides support for custom JavaScript, which makes it possible to add e.g. matomo tracking.

The only problem left is the fact that the security headers prevent loading an external javascript file.

For this use case, it would be enough to add an env var default for [securure-headers/script-src/allow[]](https://github.com/LycheeOrg/Lychee/blob/1a0b83107c8ffdcc228dce58b38f886639b13b48/config/secure-headers.php#L470)

I would be more than happy to create a PR for that, but there are a few questions to this from my side:

- Are there any other values that would be benefital to make them configurable?
- What should the env var look like to keep it consistent with other env vars for config (my suggestion would be: SECURE_HEADERS__*SCRIPT_SRC_ALLOW*)

If thats implemented, would it be also an option to make it part of the documentation? E. g. small section "How to integrate Matomo, Google Analytics". Can also provide the content for that :)

timo-reymann:
Fixes #1869
Fixes #259

As discussed in [LycheeOrg/Lychee#1896](https://github.com/LycheeOrg/Lychee/issues/1869), this enables modification of the Content Security Policy, allowing users to integrate Trackers as Matomo, Google Analytics, etc. as well as loading JavaScript from external sources.

This is compatible with previous versions and the setting doesn't need to be set explicitly with an env var.

The following example JavaScript works with the new settings (Matomo tracker):

```js
var _paq = window._paq = window._paq || [];
_paq.push(['trackPageView']);
_paq.push(['enableLinkTracking']);
(function() {
  var u="https://analytics.example.com/";
  _paq.push(['setTrackerUrl', u+'matomo.php']);
  _paq.push(['setSiteId', '<site-id>']);
  var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
  g.async=true; 
  g.src=u+'matomo.js'; 
  s.parentNode.insertBefore(g,s);
})();

navigation.addEventListener("navigate", function(e) {
  _paq.push(['setDocumentTitle', document.title]);
  _paq.push(['trackPageView']);
});
```

Assuming the following env vars are set:

```env
SECURITY_HEADER_SCRIPT_SRC_ALLOW=https://analytics.example.com/matomo.js
SECURITY_HEADER_CSP_CONNECT_SRC=https://analytics.example.com/matomo.js
```

Also supports multiple values separated by spaces.


ildyria (approved):
LGTM.

ildyria (approved):
LGTM.

PR #1873, opened 2023-06-09
Support files without extensions in importFromUrl

wladif:
This fixes the enhancement  #1652. 
It extends the importFromUrl logic to support importing of images from URLs without any extension.

**Caveats**:
1) the constant `SAMPLE_DOWNLOAD_TIFF_WITHOUT_EXTENSION` needs to be updated to point to Lychee repo once this PR is merged. I'll do in a future PR.
2) Unfortunately, Github sets the content type to `application/octet-stream` when file doesn't have an extension, this means that the branch `elseif (isset($originalMimeType) && self::isSupportedMimeType($originalMimeType))` is not covered by current tests. If you know a provider/way to host the sample files (`mongolia` & `tiff`) without extension so that the web server will set the Content Type correctly (e.g. `image/jpeg`), please let me know and I'll add the tests. Thanks!


app/Image/Files/BaseMediaFile.php
```diff
@@ -76,6 +76,24 @@ abstract class BaseMediaFile extends AbstractBinaryBlob implements MediaFile
 		'application/octet-stream', // Some mp4 files; will be corrected by the metadata extractor
 	];
 
+	public const MIME_TYPES_TO_FILE_EXTENSIONS = [
+		'image/gif' => '.gif',
+		'image/jpeg' => 'jpg',
```
ildyria:
```suggestion
		'image/jpeg' => '.jpg',
```

app/Image/Files/BaseMediaFile.php
```diff
@@ -326,4 +344,32 @@ public static function assertIsSupportedOrAcceptedFileExtension(string $extensio
...
+	 *
+	 * @return string the default file extension for the given MIME type
+	 */
+	public static function getDefaultFileExtensionForMimeType(string $mimeType): string
+	{
+		if (array_key_exists(strtolower($mimeType), self::MIME_TYPES_TO_FILE_EXTENSIONS)) {
+			return self::MIME_TYPES_TO_FILE_EXTENSIONS[strtolower($mimeType)];
+		} else {
+			return '';
+		}
```
ildyria: I think instead of the if you can do:
```php
return self::MIME_TYPES_TO_FILE_EXTENSIONS[strtolower($mimeType)] ?? '';
```

app/Image/Files/BaseMediaFile.php
```diff
@@ -326,4 +344,32 @@ public static function assertIsSupportedOrAcceptedFileExtension(string $extensio
 			throw new MediaFileUnsupportedException(MediaFileUnsupportedException::DEFAULT_MESSAGE . ' (bad extension: ' . $extension . ')');
 		}
 	}
+
+	/**
```
ildyria, line 348: If those two functions are only going to be used in Downloaded files, it makes more sense to place them there, idem for the `MIME_TYPES_TO_FILE_EXTENSIONS`

ildyria:
To make it easier for you in the future, I suggest when you want to add a feature to create a new branch instead of doing it on master on your side :)
A bit of doc is here: https://lycheeorg.github.io/docs/contributions.html#how-to-properly-submit-a-pull-request-to-lychee

wladif:
Cool! Thanks! Is there any issue that I still need to address?

ildyria (approved):

PR #1875, opened 2023-06-12
add support for 418 for dubious queries

ildyria:
The goal of this is to flag queries which are just bots trying to scan for vulnerabilities.

Combined with Fail2Ban, we can now ban IP then to avoid being hit uselessly by such visitors.

read more here:
https://aureola.codes/en/blog/2021/how-stop-vulnaribility-scanners-laravel-fail2ban

`/etc/fail2ban/filter.d/nginx-honeypot.conf`
```
[Definition]
failregex = ^<HOST>.*"(GET|POST).*" (418) .*$
ignoreregex =
```

`/etc/fail2ban/jail.d/honeypot.conf`
```
[nginx-honeypot]
enabled = true
filter = nginx-honeypot
port = http,https
logpath = /var/log/nginx/access.log
maxretry = 1
```
Of course in the case of Apache we need to check the respective log.

As said in the article:

> What you can achieve is that you get **less unwanted traffic**. No more and no less. Also, you may be able to prevent vulnerabilities from being found that would otherwise have been discovered.

### DIFF
--- a/app/Http/Controllers/HoneyPotController.php
+++ b/app/Http/Controllers/HoneyPotController.php
@@ -53,7 +53,7 @@ class HoneyPotController extends Controller
 	 * using abort(404) does not give the info which path was called.
 	 * This could be very useful when debugging.
 	 * By throwing a proper exception we preserve this info.
-	 * Not that this will generate a log line of type ERROR.
+	 * This will generate a log line of type ERROR.
 	 *
 	 * @param string $path called
 	 *
@@ -69,7 +69,7 @@ class HoneyPotController extends Controller
 	/**
 	 * Similar to abort(404), abort(418) does not give info.
 	 * It is more interesting to raise a proper exception.
-	 * By having a proper exception we are also able to decrease the severity.
+	 * By having a proper exception we are also able to decrease the severity to NOTICE.
 	 *
 	 * @param string $path called
 	 *
--- a/app/Http/Controllers/HoneyPotController.php
+++ b/app/Http/Controllers/HoneyPotController.php
@@ -16,6 +16,11 @@ class HoneyPotController extends Controller
 {
 	public function __invoke(string $path = ''): void
 	{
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/HoneyPotController.php
```diff
@@ -0,0 +1,84 @@
...
+		$honeypot_xpaths_array_suffix = config('honeypot.xpaths.suffix', []);
+
+		foreach ($honeypot_xpaths_array_prefix as $prefix) {
+			foreach ($honeypot_xpaths_array_suffix as $suffix) {
+				$honeypot_paths_array[] = $prefix . '.' . $suffix;
+			}
+		}
+
+		// Check if Honey is available
+		if (config('honeypot.enabled', true) !== true) {
```
qwerty287: Would be better to only do these array operations if it's enabled. I.e. put all into an if statement

config/honeypot.php
```diff
@@ -0,0 +1,64 @@
...
+	 * Honey.
+	 *
+	 * Set of possible path.
+	 * Those will be concatenated into a regex.
+	 */
+	'paths' => [
+		'.env',
+		'.git/config',
+		'.git/HEAD',
+		'.well-known/security.txt',
```
qwerty287, line 19: Not really related: We could actually add a [security.txt](https://securitytxt.org) to Lychee to easily allow researchers to report vulns to us.

qwerty287 (commented):
I don't really like the matching system, but I don't have any other idea

app/Http/Controllers/HoneyPotController.php
ildyria: fixed with [abf5c42](https://github.com/LycheeOrg/Lychee/pull/1875/commits/abf5c42429edc243676d5dee40a7f658f0298464)

qwerty287 (approved):

PR #1881, opened 2023-06-17
reorder logging to make it less poluted

ildyria:
Reorder logging files so that they are less poluted.

### DIFF
--- a/config/logging.php
+++ b/config/logging.php
@@ -48,28 +48,7 @@ return [
 	'channels' => [
 		'stack' => [
 			'driver' => 'stack',
-			'channels' => ['debug-daily', 'notice', 'warning', 'error', ],
-		],
-
-		// By the way...
-		'notice' => [
-			'path' => storage_path('logs/notice.log'),
-			'driver' => 'daily',
-			'level' => 'notice',
-		],
-
-		// Something may have gone wrong
-		'warning' => [
-			'path' => storage_path('logs/warning.log'),
-			'driver' => 'single',
-			'level' => 'warning',
-		],
-
-		// Something went wrong
-		'error' => [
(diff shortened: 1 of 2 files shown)


config/logging.php
```diff
@@ -48,36 +48,38 @@
...
 			'driver' => 'stack',
-			'channels' => ['debug-daily', 'notice', 'warning', 'error', ],
+			'channels' => ['debug-daily', 'error', 'warning',  'notice'],
 		],
 
-		// By the way...
-		'notice' => [
-			'path' => storage_path('logs/notice.log'),
+		// Whatever debug log is needed
+		// Nostly SQL requests
```
d7415:
```suggestion
		// Mostly SQL requests
```

config/logging.php
```diff
@@ -48,36 +48,38 @@
 	'channels' => [
 		'stack' => [
 			'driver' => 'stack',
-			'channels' => ['debug-daily', 'notice', 'warning', 'error', ],
+			'channels' => ['debug-daily', 'error', 'warning',  'notice'],
```
d7415, line 51: I haven't actually checked how this part works, but wouldn't we want debug at the notice end?

ildyria: Not really.
- Debug logs everything
- then you hit Error, if the message is an error it stops here
- then you hit Warning, same principle
- then you hit notice.

If you go the other way, the problem is that Warning will also capture Error messages. which makes it very noisy.
And you can't block propagation on Warning because Error will also be stopped.

d7415: Ah, I see. :+1:

d7415 (approved):
As ever, untested, but looks good.

ildyria:
I tested locally :)

PR #1882, opened 2023-06-17
Add csp rules for api

mingan666:
This adds some new CSP rules needed to be able to view the API docs.


ildyria (approved):
LGTM.

config/secure-headers.php
```diff
@@ -558,6 +561,10 @@
 		'style-src' => [
 			'self' => true,
 			'unsafe-inline' => true, // We need this one due to direct styles (not just style classes) applied by JavaScript
+			'allow' => [
+				'https://unpkg.com/@stoplight/elements/styles.min.css',
+				// 'url',
```
d7415:
```suggestion
```

d7415:
Is this only relevant for dev builds? If so, do we really want to add this noise for everyone?

ildyria:
> Is this only relevant for dev builds? If so, do we really want to add this noise for everyone?

It is not much noise though.

But a full proper approach would be checking the route and disabling the CSP on the fly (see e.g. for log-viewer).
On this specific case, we are a bit less constrained and can directly white list 2 urls instead of doing a full disable.

ildyria:
@d7415 better ?

ildyria:
Made the pull request cleaner:
- no more hacky functions
- use middleware to disable the CSP when necessary
- enable CSP by default, disable programatically.

qwerty287 (approved):
Untested

d7415 (approved):
Only skimmed, but it feels neater than the previous options.

PR #1896, opened 2023-06-23
Only apply explain when requested

ildyria:


d7415 (approved):
Looks reasonable. Just a speed boost?

qwerty287 (approved):

ildyria:
> Looks reasonable. Just a speed boost?

more than that, for some reason there is a query that makes it not happy in the tests on MariaDB.
So I am disabling it for now so that tests are green again.

PR #1898, opened 2023-06-24
Improve explain on db_logging

ildyria:


tests/Feature/WebAuthTest.php
```diff
@@ -30,6 +30,7 @@ public function setUp(): void
 	{
 		parent::setUp();
 		$this->setUpRequiresEmptyWebAuthnCredentials();
+		config(['app.url' => 'https://localhost']);
```
ildyria, line 33: This allows to just run tests locally without having to think whether the `.env` matches on not.

qwerty287 (approved):

PR #1899, opened 2023-06-24
Improved speed on global table

ildyria:
100x speed improvements.

PR #1907, opened 2023-06-28
v4.9.4

ildyria:
#1899 gives a significant speed boost.
This is something that is very much needed for some users (e.g. me).


qwerty287 (approved):
After https://github.com/LycheeOrg/Lychee/pull/1908

PR #1971, opened 2023-08-07
Enable video thumbnail by default

Lingxi-Li:
There is no video thumbnail out of the box and [the related doc](https://lycheeorg.github.io/docs/faq_troubleshooting.html#why-dont-my-videos-have-thumbnails) to enable it is outdated. The PR makes video thumbnail generation default behavior.

### DIFF
--- a/app/Image/Handlers/GoogleMotionPictureHandler.php
+++ b/app/Image/Handlers/GoogleMotionPictureHandler.php
@@ -77,8 +77,8 @@ class GoogleMotionPictureHandler extends VideoHandler
 			$file->close();
 
 			$ffmpeg = FFMpeg::create([
-				'ffmpeg.binaries' => '/usr/bin/ffmpeg',
-				'ffprobe.binaries' => '/usr/bin/ffprobe',
+				'ffmpeg.binaries' => Configs::getValueAsString('ffmpeg_path'),
+				'ffprobe.binaries' => Configs::getValueAsString('ffprobe_path'),
 			]);
 			$audioOrVideo = $ffmpeg->open($this->workingCopy->getRealPath());
 			if ($audioOrVideo instanceof Video) {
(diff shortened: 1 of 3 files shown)


ildyria (approved):
LGTM, but what if ffmpeg and ffprobe at not at this path ? 

also according to:
https://github.com/PHP-FFMpeg/PHP-FFMpeg#how-this-library-works

This should not be necessary. Wouldn't it be better to first check whether they are in the path and then if not hardcode the paths instead ?
You can also use a database migration and access the values with Configs::getValueAsString() if you want to make those path configurable (possibly a better option). :)

Lingxi-Li:
Agree that there are better well-designed alternatives than this quick turnaround. Yet, I don't have any PHP knowledge, so won't go any further.

My installation is set up from the master branch. Per my experience, video thumbnail begins to work only after

1. editing the `Lychee/vendor/lychee-org/php-exif/lib/PHPExif/Adapter/FFprobe.php` file and
2. performing the changes in this PR.

I think "failing fast right at start" is better than "feature malfunctioning silently with no obvious clue". The latter usually leads to more difficult and longer troubleshooting sessions. So I consider this quick turnaround an improvement over existing code.

Lingxi-Li:
@ildyria Looks like the fixer PR automation is broken. I manually added a commit to fix the code style.

ildyria:
I fixed your PR.
Could you do a `git pull` and then do `artisan migrate` and try to see if the thumbnail generation is working ?

ildyria:
@qwerty287 given that I contaminated this PR, could you review it too please ?

Lingxi-Li:
@ildyria I did a brand-new installation with the changes and video thumbnails are working out of the box 😃

qwerty287 (approved):

PR #1977, opened 2023-08-11
Hungarian language added

KnauszFerenc:
I added a full Hungarian translate


ildyria (approved):
I don't know anything about Hungarian but that looks like an eastern Europe language so LGTM.

flaki:
In case you be wondering I checked the strings and this is an excellent quality translation (I am also Hungarian), thank you Ferenc great job!

PR #1985, opened 2023-08-21
Fixes #1751 - Add error thrown if APP_URL does not match current url

### LINKED ISSUE #1751: WebAuthn/U2F: Attestation Error: Relying Party ID not scoped to current

WebAuthn does not work.

1. Log in
2. Open Settings -> U2F
3. Click "Register new device"
4. Have the browser generate a new passkey
5. WebAuthn::register API call returns Error 422

```json
{"message":"Attestation Error: Relying Party ID not scoped to current.","errors":{"attestation":["Attestation Error: Relying Party ID not scoped to current."]}}
```

### Browser and system

Chrome on macOS

The documentations does not say anything about the U2F feature, so it should be either fixed (and "just work"), or removed.

ildyria:
Fixes #1751

This aims to avoid some of the confusions _"why is U2F not working?"_.

### DIFF
--- a/app/Actions/Diagnostics/Pipes/Checks/AppUrlMatchCheck.php
+++ b/app/Actions/Diagnostics/Pipes/Checks/AppUrlMatchCheck.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace App\Actions\Diagnostics\Pipes\Checks;
+
+use App\Contracts\DiagnosticPipe;
+
+class AppUrlMatchCheck implements DiagnosticPipe
+{
+	/**
+	 * {@inheritDoc}
+	 */
+	public function handle(array &$data, \Closure $next): array
+	{
+		if (config('app.url') !== request()->httpHost())
+		{
+			$data[] = 'Error: APP_URL does not match the current url. This will break U2F authentication. Please update APP_URL to reflect this change.';	
+		}
+
+		return $next($data);
+	}
+}
--- a/app/Actions/Diagnostics/Pipes/Checks/AppUrlMatchCheck.php
+++ b/app/Actions/Diagnostics/Pipes/Checks/AppUrlMatchCheck.php
@@ -11,9 +11,12 @@ class AppUrlMatchCheck implements DiagnosticPipe
 	 */
 	public function handle(array &$data, \Closure $next): array
 	{
-		if (config('app.url') !== request()->httpHost() &&
-			config('app.url') !== request()->schemeAndHttpHost()) {
-			$data[] = 'Error: APP_URL does not match the current url. This will break U2F authentication. Please update APP_URL to reflect this change.';
+		$config_url = config('app.url');
+		// http:// is 7 characters.
+		if (strpos($config_url, '/', 8) !== false) {
+			$data[] = 'Warning: APP_URL contains a sub-path. This may impact your WebAuthn authentication.';
+		} else if ($config_url !== request()->httpHost() && $config_url !== request()->schemeAndHttpHost()) {
+			$data[] = 'Error: APP_URL does not match the current url. This will break WebAuthn authentication. Please update APP_URL to reflect this change.';
 		}
 
 		return $next($data);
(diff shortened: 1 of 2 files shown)


qwerty287 (commented):
What happens if setting APP_URL with a subpath? Is this allowed or would this be a wrong configuration?

ildyria:
> What happens if setting APP_URL with a subpath? Is this allowed or would this be a wrong configuration?

Good point, I adapted the code to throw a warning in such cases:
[8320d16](https://github.com/LycheeOrg/Lychee/pull/1985/commits/8320d168436cf830789994205b55b2e3aa7ce5e6)

app/Actions/Diagnostics/Pipes/Checks/AppUrlMatchCheck.php
```diff
@@ -0,0 +1,24 @@
...
+
+class AppUrlMatchCheck implements DiagnosticPipe
+{
+	/**
+	 * {@inheritDoc}
+	 */
+	public function handle(array &$data, \Closure $next): array
+	{
+		$config_url = config('app.url');
+		// http:// is 7 characters.
```
d7415:
```suggestion
		// https:// is 8 characters.
```
As we encourage https it may be best to specify that in this comment? We are meant to put the https there, right? (genuine question..)
A quick look at the [php `strpos` docs](https://www.php.net/manual/en/function.strpos.php) suggests the test below still covers this case anyway.

ildyria: I tested with https:// :')

d7415 (approved):
I wonder whether it's worth adding validation for subpaths (or the host of `APP_URL`s containing them), but that can always be added later.

Untested, as usual.

qwerty287:
Not sure if  you understood my point, but my question actually was: What happens with *both* domain and subpath? I.e. `APP_URL=https://example.com/xzy`. Is this supported/allowed by laravel?

d7415:
> Is this supported/allowed by laravel?

At a glance, it looks like it is.

ildyria:
https://www.w3.org/TR/webauthn-2/#rp-id

> The [RP ID](https://www.w3.org/TR/webauthn-2/#rp-id) of a [public key credential](https://www.w3.org/TR/webauthn-2/#public-key-credential) determines its scope. I.e., it determines the set of origins on which the public key credential may be exercised, as follows:
> - The [RP ID](https://www.w3.org/TR/webauthn-2/#rp-id) must be equal to the [origin](https://www.w3.org/TR/webauthn-2/#determines-the-set-of-origins-on-which-the-public-key-credential-may-be-exercised)'s [effective domain](https://html.spec.whatwg.org/multipage/origin.html#concept-origin-effective-domain), or a [registrable domain suffix](https://html.spec.whatwg.org/multipage/origin.html#is-a-registrable-domain-suffix-of-or-is-equal-to) of the [origin](https://www.w3.org/TR/webauthn-2/#determines-the-set-of-origins-on-which-the-public-key-credential-may-be-exercised)'s [effective domain](https://html.spec.whatwg.org/multipage/origin.html#concept-origin-effective-domain).
> - The [origin](https://www.w3.org/TR/webauthn-2/#determines-the-set-of-origins-on-which-the-public-key-credential-may-be-exercised)'s [scheme](https://url.spec.whatwg.org/#concept-url-scheme) must be https.
> - The [origin](https://www.w3.org/TR/webauthn-2/#determines-the-set-of-origins-on-which-the-public-key-credential-may-be-exercised)'s [port](https://url.spec.whatwg.org/#concept-url-port) is unrestricted.

> For example, given a [Relying Party](https://www.w3.org/TR/webauthn-2/#relying-party) whose origin is https://login.example.com:1337, then the following [RP ID](https://www.w3.org/TR/webauthn-2/#rp-id)s are valid: login.example.com (default) and example.com, but not m.login.example.com and not com.

qwerty287 (approved):

Issue #1988, opened 2023-08-23
having trouble migrating to new host


stanelie:
### Detailed description of the problem [REQUIRED]
*Trying to move to a new host, I tried a bunch of stuff, none of it worked. The most common error, after copying the latest lychee files to the new host, cloning the mysql database and copying over my public/upload folder as well as the .env file, I get redirected to a /migrate folder that does not exist. Thoughts?*

### Steps to reproduce the issue
**Steps to reproduce the behavior:**
N/A

**Screenshots**
*If applicable, add screenshots to help explain your problem.*

### Output of the diagnostics  [REQUIRED]
`*/crashplan/www/lychee# php artisan lychee:diagnostics



In Connection.php line 795:
                                                                                                                                                      
  SQLSTATE[42S22]: Column not found: 1054 Unknown column 'may_administrate' in 'where clause' (Connection: mysql, SQL: select count(*) as aggregate   
  from `users` where `may_administrate` = 1)                                                                                                          
                                                                                                                                                      

In Connection.php line 416:
                                                                                               
  SQLSTATE[42S22]: Column not found: 1054 Unknown column 'may_administrate' in 'where clause'  
                                                                                               *`

### Browser and system
Any browser, running Lychee on Ubuntu 20.04, php 8.1, mysql 8.0.34-0ubuntu0.22.04.1 (Ubuntu)

stanelie:
While I'm here, what is the correct way to install or migrate this app? The documentation is a bit fuzzy on this : I could download the code from github and just copy it over my install, but this never worked for me, or there is also this thing with composer with which I am unfamiliar with and don't really know how to use correctly, and also a .deb package...?

stanelie:
Made some progress on my own, will come back with next issue.

PR #1990, opened 2023-08-24
fix complaints in Diagnostics when no migrations has been run

ildyria:
Fixes complaint seen in #1988
>Output of the diagnostics [REQUIRED]
>```
> /crashplan/www/lychee# php artisan lychee:diagnostics
>
>In Connection.php line 795:
>
>SQLSTATE[42S22]: Column not found: 1054 Unknown column 'may_administrate' in 'where clause' (Connection: mysql, >SQL: select count(*) as aggregate
>from users where may_administrate = 1)
>
>In Connection.php line 416:
>
>SQLSTATE[42S22]: Column not found: 1054 Unknown column 'may_administrate' in 'where clause'
>```

Basically skip any check that requires a DB interaction.

### DIFF
--- a/app/Actions/Diagnostics/Pipes/Checks/UpdatableCheck.php
+++ b/app/Actions/Diagnostics/Pipes/Checks/UpdatableCheck.php
@@ -11,6 +11,7 @@ use App\Facades\Helpers;
 use App\Metadata\Versions\GitHubVersion;
 use App\Metadata\Versions\InstalledVersion;
 use App\Models\Configs;
+use Illuminate\Support\Facades\Schema;
 use function Safe\exec;
 
 /**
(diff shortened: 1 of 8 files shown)


app/Actions/Diagnostics/Pipes/Checks/UpdatableCheck.php
```diff
@@ -64,6 +65,9 @@ public static function assertUpdatability(): void
 			return;
 			// @codeCoverageIgnoreEnd
 		}
+		if (!Schema::hasTable('configs')) {
+			throw new ConfigurationException('Migration is not run');
```
qwerty287, line 69: Why do we throw an error here and just return `$next(...)` on the other ones?

ildyria: Because this is not the `handle()` function ?

qwerty287: Ah sorry, didn't see

qwerty287 (approved):

PR #1995, opened 2023-08-29
Fix #1784 - certain hidden diacritics in album titles

### LINKED ISSUE #1784: uppercase letters with a circumflex don't fit into the overlay div
image
Hello, it seems like uppercase letters with a circumflex above don't fit into the overlay div.
Disabling it fixed it, but it is not an option for long album titles. 
For anyone who would like to test, you can copy the problematic characters: 
"Č Ė Š Ū Ž"

kfpv:
Added some top padding to album titles in gallery to not cut off top diacritics in uppercase letters


qwerty287:
Please submit this to the correct file in https://github.com/LycheeOrg/Lychee-front which is our frontend repository.

PR #2015, opened 2023-09-16
replace layout and overlay to proper Enum types

ildyria:
The layout was using int: 0,1,2.

This created some weird php trickery where we could not make the difference between:
```php
$arr = [
	'0' => 'translation 0',
	'1' => 'translation 1',
	'2' => 'translation 2'
];
```
and 
```php
$arr = [
	'translation 0',
	'translation 1',
	'translation 2'
];
```

Due to type inference, php would assume that the arrays are the same, while the intent is not there.
The keys in that specific case are the intended config option and the value are the translations.
This is breaking and making code complex on Livewire front-end.

By changing this to proper Enum types, we are making the configuration cleaner.

Because this is a migration change. This will in 4.12.0


qwerty287 (approved):
There's no PR for the frontend yet?

ildyria:
https://github.com/LycheeOrg/Lychee-front/pull/409/files

PR #2017, opened 2023-09-16
Unify photoId to photoID and albumId to albumID when found

ildyria:
cleaner code.

### DIFF
--- a/app/Factories/AlbumFactory.php
+++ b/app/Factories/AlbumFactory.php
@@ -42,7 +42,7 @@ class AlbumFactory
 	 * Returns an existing instance of an album with the given ID or fails
 	 * with an exception.
 	 *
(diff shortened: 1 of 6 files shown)


qwerty287:
We could apply the same for terms like `URL`, `PHP` etc. (if there are any)

ildyria:
> We could apply the same for terms like `URL`, `PHP` etc. (if there are any)

I did not look for those, as I was working on the Livewire part. I was mostly concerned with the dispatch messages format between of components. It uses an array, and I wanted to be sure that the keys were all unified to avoid nasty surprises.

Also the attributes in the DB are like album_id ... snake_case.
This made this change a bit risky. Had to double check everything.

qwerty287 (approved):

PR #1303, opened 2022-05-05
Enable full support of Livewire + Tailwind: New front-end

ildyria:
# WORK IN PROGRESS &mdash; Just ignore :sweat_smile: 

## Setup for testing

After switching to this branch.
```sh
# Update the vendor files.
composer install 

# Update the database, there are new constants being used.
php artisan migrate 

# Set up the JS dependencies.
npm install

# Spawn the Vite server for hot reload pages on changes (recommended for development).
npm run dev
# or compile files (recommended for production).
npm run build
```

Then you can access the front-end at the following url:
`https://yourlychee.example.org/livewire`
Or by setting `LIVEWIRE_ENABLED=false` in your `.env`, in this case, the front-end will switch to Livewire completely.

## Fixes

Fixes #12 &mdash; :heavy_check_mark:
Fixes #383 &mdash; :heavy_check_mark: 
Fixes #725 &mdash; :heavy_check_mark:
Fixes #748 &mdash; :heavy_check_mark: (need localization)
Fixes #828 &mdash; :heavy_check_mark:
Closes #845 &mdash; :heavy_check_mark:
Fixes #857 &mdash; :heavy_check_mark:
Fixes #1053 &mdash; :heavy_check_mark: (no more public smart album)
Fixes #1211 &mdash; :heavy_check_mark:
Fixes #1249 &mdash; :heavy_check_mark:
Fixes #1189 &mdash; :heavy_check_mark:
Fixes #1253 &mdash; :heavy_check_mark:
Fixes #1462 &mdash; :heavy_check_mark:
Fixes #1676 &mdash; :heavy_check_mark:
Fixes #1704  &mdash; :heavy_check_mark:
Fixes #1720 &mdash; :heavy_check_mark:
Fixes #1732 &mdash; :heavy_check_mark:
Fixes #1891 &mdash; :heavy_check_mark:
Fixes #1948 &mdash; :heavy_check_mark:
Fixes #1973 &mdash; :heavy_check_mark:
Fixes #1989 &mdash; :heavy_check_mark:
Fixes https://github.com/LycheeOrg/Lychee-front/issues/199  &mdash; :heavy_check_mark:
Fixes #2011 &mdash; :heavy_check_mark:

Fixes https://github.com/LycheeOrg/Lychee-front/issues/90 &mdash; :heavy_check_mark:
Fixes https://github.com/LycheeOrg/Lychee-front/issues/104 &mdash; :heavy_check_mark:
[...]

PR #2018, opened 2023-09-17
Unique constraint for config keys

qwerty287:
From #1303 

This is standalone, so it's better to take it out of it and separately merge it


ildyria (approved):

ildyria:
> From #1303
> 
> This is standalone, so it's better to take it out of it and separately merge it

I was actually planning to do it. :laughing:

PR #2019, opened 2023-09-17
Update composer (include breaking)

qwerty287:


ildyria:
image
The naming of the archive seems to be broken though :(

`Album__getArchive`
and `Album__getArchive (1)`
is what I get now. 
No zip extension.

qwerty287:
This is very weird. In PHP everything seems working - the filename is set correctly in the header. However, the filename does not occur in firefox' network console. If I execute the same via curl, the filename is returned correctly.

qwerty287:
Hmm checked again it does not work with curl too. If I use `curl -I` it works though :thinking: I'll do some more debugging

qwerty287:
Found it. ZipStream got a way to automatically set HTTP headers which is enabled by default and overwrites out headers. I disabled it.

ildyria (approved):
LGTM.

PR #2025, opened 2023-09-22
License as enum type

ildyria:
Use proper enum type instead of array of values..

Extracted from Livewire branch :)


qwerty287 (approved):

SerenaButler:
I got an error on my gallery page: "CC-BY-NC-ND" is not a valid backing value for enum "App\Enum\LicenseType"

I added `"case CC_BY_NC_ND = 'CC-BY-NC-ND';`" and "`self::CC_BY_NC_ND->value => 'CC Attribution-NonCommercial-NoDerivatives',`" and it is working again.

PR #2066, opened 2023-12-02
Add optimize database call

ildyria:
Add optimization call so that Large databases can be optimized from the webinterface via an authenticated Get request.

### DIFF
--- a/app/Actions/Db/OptimizeDb.php
+++ b/app/Actions/Db/OptimizeDb.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace App\Actions\Db;
+
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+class OptimizeDb
+{
+	public function do(): array
+	{
+		$ret = ['Optimizing Database.'];
+
+		$connection = Schema::connection(null)->getConnection();
+		$driverName = $connection->getDriverName();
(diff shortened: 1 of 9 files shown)


app/Actions/Db/OptimizeDb.php
```diff
@@ -0,0 +1,58 @@
...
+			'pgsql' => 'PostgreSQL detected.',
+			'sqlite' => 'SQLite detected.',
+			default => 'Warning:Unknown DBMS; doing nothing.',
+		};
+
+		$sql = match ($driverName) {
+			'mysql' => 'OPTIMIZE TABLE ',
+			'pgsql' => 'VACUUM(FULL, ANALYZE)',
+			'sqlite' => 'VACUUM',
+			default => 'NOTHING',
```
qwerty287: Can you use `null` or something like this?

ildyria: Fixed with [4e7408f](https://github.com/LycheeOrg/Lychee/pull/2066/commits/4e7408f8978a647048db6f9b70cebd40678a37a0) and  [f9df33a](https://github.com/LycheeOrg/Lychee/pull/2066/commits/f9df33aa7d5a0dc5850c7e42842e3d709dd925fe).

qwerty287 (approved):

PR #2071, opened 2023-12-18
Drops support for singular public photos in search.

ildyria:
This does two things:
- rename the `public_search` config to `search_public` (foreshadowing others  configs in the search side)
- drops the `public_photos_hidden` config => Reason why 2 tests files are dropped as they are no longer applicable.


ildyria:
Note that this will most likely require a minor version bump.

qwerty287 (approved):

PR #2072, opened 2023-12-19
Add configuration check between int and positive (>0)

ildyria:
Just something that can be added on stand alone instead of the Livewire PR.

### DIFF
--- a/app/Models/Configs.php
+++ b/app/Models/Configs.php
@@ -56,6 +56,7 @@ class Configs extends Model
 	use ThrowsConsistentExceptions;
 
 	protected const INT = 'int';
+	protected const POSTIIVE = 'positive';
 	protected const STRING = 'string';
 	protected const STRING_REQ = 'string_required';
 	protected const BOOL = '0|1';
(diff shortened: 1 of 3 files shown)


app/Models/Configs.php
```diff
@@ -122,7 +123,12 @@ public function sanity(?string $candidateValue, ?string $message_template = null
 			case self::INT:
 				// we make sure that we only have digits in the chosen value.
 				if (!ctype_digit(strval($candidateValue))) {
-					$message = sprintf($message_template, 'positive integer');
+					$message = sprintf($message_template, 'positive integer or 0');
+				}
+				break;
+			case self::POSTIIVE:
+				if (!ctype_digit(strval($candidateValue)) || intval($candidateValue, 10) === 0) {
```
qwerty287, line 130: Should we also check for negative?

ildyria: I don't think so, there is no real case where we need config values which are negative.

qwerty287 (approved):

Issue #2104, opened 2023-12-26
Post-5.0.0: Upgrade instructions (and possibly post-merge hook) missing npm commands


TwizzyDizzy:
### Detailed description of the problem [REQUIRED]
The documentation on updating [using git](https://lycheeorg.github.io/docs/update.html#update-using-git) Lychee has not been updated to reflect changes to version 5.0.0. It seems though, installation instructions already reflect the use of npm: https://lycheeorg.github.io/docs/#installing-lychee

It seems what is missing from the update instructions is an `npm install` and and `npm run build` in order to `vite build` and generate the assets.

And if I'm not mistaken, this is also missing in the post-merge hook in the `./scripts/post-merge` hook.

If you need any additional information (and also if you want a PR for the documentation), please let me know.

Cheers
Thomas

### Steps to reproduce the issue
-

**Screenshots**
-

### Output of the diagnostics  [REQUIRED]
Without this step in the update procedure, the following error will appear in `/path/to/lychee/storage/logs/errors.log` (reflected as a status code 500 error when accessing Lychee): 

```
[2023-12-26 14:39:56] local.ERROR: Illuminate\Foundation\Vite::manifest:708 Vite manifest not found at: /path/to/lychee/public/build/manifest.json
App\Http\Middleware\MigrationStatus::handle:55 Vite manifest not found at: /path/to/lychee/public/build/manifest.json; caused by
Livewire\Mechanisms\ExtendBlade\ExtendedCompilerEngine::handleViewException:708 Vite manifest not found at: /path/to/lychee/public/build/manifest.json (View: /path/to/lychee/resources/views/components/layouts/app.blade.php); caused by
App\Http\Middleware\MigrationStatus::handle:55 Vite manifest not found at: /path/to/lychee/public/build/manifest.json (View: /path/to/lychee/resources/views/components/layouts/app.blade.php); caused by
Livewire\Mechanisms\ExtendBlade\ExtendedCompilerEngine::handleViewException:708 Vite manifest not found at: /path/to/lychee/public/build/manifest.json (View: /path/to/lychee/resources/views/components/layouts/app.blade.php) (View: /path/to/lychee/resources/views/components/layouts/app.blade.php); caused by
[...]

PR #2109, opened 2023-12-26
Modify post-merge script to reflect 5.0.0 build changes

TwizzyDizzy:
This fixes the post-merge script to reflect build changes needed in 5.0.0.

As discussed in https://github.com/LycheeOrg/Lychee/issues/2104

Cheers
Thomas

### DIFF
--- a/scripts/post-merge
+++ b/scripts/post-merge
@@ -18,12 +18,10 @@ else
         composer install
         if command -v npm > /dev/null; then
             printf "\n${ORANGE}npm detected${NO_COLOR}\n"
-            if  [ -f "package.json" ]; then
-                echo "npm install --no-audit --no-fund"
-                npm install --no-audit --no-fund
-            else
-                printf "${GREEN}no package.json found${NO_COLOR}\n"
-            fi
+            echo "npm install --no-audit --no-fund"
+            npm install --no-audit --no-fund
+        else
+            echo "npm not found. Development will likely not be possible."
         fi
     else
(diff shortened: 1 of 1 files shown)


scripts/post-merge
```diff
@@ -29,6 +29,17 @@ else
...
         composer install --no-dev --prefer-dist
+        if command -v npm > /dev/null; then
+            printf "\n${ORANGE}npm detected${NO_COLOR}\n"
+            if  [ -f "package.json" ]; then
+                echo "npm install --no-audit --no-fund"
+                npm install --no-audit --no-fund
+                echo "npm run build"
+                npm run build
+            else
+                printf "${GREEN}no package.json found${NO_COLOR}\n"
```
qwerty287: You can remove that `else` branch.

TwizzyDizzy: I just jumped back to one commit before the "if -f package.json" was added and back then, it indeed seems to not have been part of the repo at this point.

So I'm getting you right in that you want to remove this condition due to the fact that `package.json` is now part of the repo, correct?

If that's the case, then I agree. And I would then add, that line 21-26 (`master`) can be adjusted in the same way.

Cheers
Thomas

ildyria: > So I'm getting you right in that you want to remove this condition due to the fact that package.json is now part of the repo, correct?

yes.

> If that's the case, then I agree. And I would then add, that line 21-26 (master) can be adjusted in the same way.

Indeed.

ildyria (approved):
LGTM.

PR #2114, opened 2023-12-27
fly.io stuff for demo.

ildyria:
Files necessary for the demo available here: https://lychee-demo.fly.dev/


qwerty287 (approved):

Dockerfile
```diff
@@ -0,0 +1,96 @@
...
+        ln -sf /etc/nginx/sites-available/default-octane /etc/nginx/sites-enabled/default; \
+    fi
+
+# Multi-stage build: Build static assets
+# This allows us to not include Node within the final container
+FROM node:${NODE_VERSION} as node_modules_go_brrr
+
+RUN mkdir /app
+
+RUN mkdir -p  /app
```
d7415:
```suggestion
```

This is repeated above

kamil4:
Wouldn't it make more sense for the demo site to be in its own repo, rather than in the main development repo? I guess it's not a big deal if it's not part of the official release tarballs, but still...

ildyria:
No strong opinion.

d7415:
> Wouldn't it make more sense for the demo site to be in its own repo, rather than in the main development repo? I guess it's not a big deal if it's not part of the official release tarballs, but still...

That would make sense. Would save putting the images in the main git repo too.

PR #2124, opened 2023-12-28
Avoid more issues because people can't read release notes...

ildyria:
image

the `class="hidden"` will automatically hide this message if css is loaded correctly. :sunglasses:

### DIFF
--- a/resources/views/components/warning-misconfiguration.blade.php
+++ b/resources/views/components/warning-misconfiguration.blade.php
@@ -8,9 +8,10 @@
 			<pre style="font-size: 16px; display:inline-block; margin: 0;">APP_URL=https://lychee.example.com</pre>
 		</li>
 		<li style="font-size: 20px; margin-bottom: 10px;">if you are working behind a reverse proxy, that
-			<pre style="font-size: 16px; display:inline-block; margin: 0;">TRUSTED_PROXIES</pre> is set.<br>
+			<pre style="font-size: 16px; display:inline-block; margin: 0;">TRUSTED_PROXIES</pre> is set to the forwarding ip.<br>
 			For example:
-			<pre style="font-size: 16px; display:inline-block; margin: 0;">TRUSTED_PROXIES=*</pre>.
+			<pre style="font-size: 16px; display:inline-block; margin: 0;">TRUSTED_PROXIES=*</pre><br>
+			Note that the wildcard value (<pre style="font-size: 16px; display:inline-block; margin: 0;">*</pre>) is a very <b>insecure</b> option and not recommended.
 		</li>
 		<li style="font-size: 20px; margin-bottom: 10px;">if you are working behind an apache reverse proxy, that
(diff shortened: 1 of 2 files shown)


qwerty287 (approved):

resources/views/components/warning-misconfiguration.blade.php
```diff
@@ -0,0 +1,27 @@
...
+	<ul>
+		<li style="font-size: 20px; margin-bottom: 10px;">your
+			<pre style="font-size: 16px; display:inline-block; margin: 0;">APP_URL</pre> is properly set to the correct url.<br>
+			For example:
+			<pre style="font-size: 16px; display:inline-block; margin: 0;">APP_URL=https://lychee.example.com</pre>
+		</li>
+		<li style="font-size: 20px; margin-bottom: 10px;">if you are working behind a reverse proxy, that
+			<pre style="font-size: 16px; display:inline-block; margin: 0;">TRUSTED_PROXIES</pre> is set.<br>
+			For example:
+			<pre style="font-size: 16px; display:inline-block; margin: 0;">TRUSTED_PROXIES=*</pre>.
```
d7415: Please don't recommend the wildcard.
(In practice it's probably not major for users in that configuration, as the real reverse proxy should strip anything bad, but still)

ildyria: What would you suggest then ? :/

d7415: Personally I probably wouldn't include an example :man_shrugging: 

I just don't like including hacky workarounds in official documentation.

d7415:
This feels like a lot to include that should never be seen, but I can see why.

I'm also not super convinced by the inline styles, but they are enabled in our CSP, so they shouldn't break anything.

resources/views/components/warning-misconfiguration.blade.php
```diff
@@ -8,9 +8,10 @@
 			<pre style="font-size: 16px; display:inline-block; margin: 0;">APP_URL=https://lychee.example.com</pre>
 		</li>
 		<li style="font-size: 20px; margin-bottom: 10px;">if you are working behind a reverse proxy, that
-			<pre style="font-size: 16px; display:inline-block; margin: 0;">TRUSTED_PROXIES</pre> is set.<br>
+			<pre style="font-size: 16px; display:inline-block; margin: 0;">TRUSTED_PROXIES</pre> is set to the forwarding ip.<br>
 			For example:
-			<pre style="font-size: 16px; display:inline-block; margin: 0;">TRUSTED_PROXIES=*</pre>.
+			<pre style="font-size: 16px; display:inline-block; margin: 0;">TRUSTED_PROXIES=*</pre><br>
+			Note that the wildcard value (<pre style="font-size: 16px; display:inline-block; margin: 0;">*</pre>) is a very <b>insecure</b> option and not recommended.
```
d7415, line 14: :+1:

PR #2135, opened 2023-12-30
Description should be desc for overlay

ildyria:


d7415 (approved):
Not tested, but matches what's in settings and looks complete.

qwerty287 (approved):

PR #2138, opened 2023-12-30
Allow different aspect ratios for album thumbs (+ per album setting)

ildyria:


kamil4:
I haven't tried this (I have yet to update my over a year old Lychee instances, let alone to v5...) but I have a more... philosophical question: why are album thumbs being treated so differently from picture thumbs?

If I remember correctly, "back in the day" all thumbs in Lychee were square. I believe @ildyria then added support for aspect-preserving thumbs, but for pictures only, not for albums.

Now this PR extends album thumbs support from square-only to six arbitrary aspect ratios, but still without adding support for aspect-preserving thumbs to match what's been the case for pictures for many years now.

Why do we continue treating album thumbs so differently from picture thumbs?

On a more technical note, I haven't looked at the code too closely, but at least in the past, square thumbs were actually generated during photo import and, looking at the demo site, that still seems to be the case. Clearly the code doesn't extend that to five more aspect ratios, so how does this work in practice? Do we rely on the web browser (and CSS) to do the cropping? And is it cropping from the aspect-preserving thumb or from the square one? It _should_ be using the aspect-preserving one; otherwise if one selects, say, 3/2 aspect, for a landscape picture the shown thumb would be a crop of a crop, i.e., both horizontal and vertical borders would be cut off -- but is that what the code currently does? Also, if we are to rely on the web browser to do the cropping, shouldn't Lychee stop generating the square thumbs altogether?

ildyria:
> why are album thumbs being treated so differently from picture thumbs?

For multiple reasons, but the mains ones are historic ones.

- Albums have always been square with the hover 3 slides. Images never had that effect.
- Photo display got augmented with Justified layout and now since v5 with Masonry and Grid (same as square but keeping aspect~ish).
- I don't think the albums are doing well in a Justified or Masonry layout as it makes it a bit messier. In this specific case, they are so visually different that you can't get confused.

This result in quite different code paths (especially for the choice of the thumbs).

> On a more technical note, I haven't looked at the code too closely, but at least in the past, square thumbs were actually generated during photo import and, looking at the demo site, that still seems to be the case. Clearly the code doesn't extend that to five more aspect ratios, so how does this work in practice? Do we rely on the web browser (and CSS) to do the cropping? 

It is currently using CSS 3 [aspect-ratio property](https://caniuse.com/?search=aspect-ratio) on the containing `<span>` + `h-full` `w-full` on the `` were we also have [object-fit property](https://caniuse.com/?search=object-fit) set to `cover`.
This takes care for forcing an aspect-ratio + cropping.

> And is it cropping from the aspect-preserving thumb or from the square one?

Currently square one.

> It should be using the aspect-preserving one; otherwise if one selects, say, 3/2 aspect, for a landscape picture the shown thumb would be a crop of a crop, i.e., both horizontal and vertical borders would be cut off -- but is that what the code currently does?

We could be using small indeed instead, I have been starting thinking about that since I changed the code for the square layout of photos to use small instead of thumbs, simplifying the code path.

> Also, if we are to rely on the web browser to do the cropping, shouldn't Lychee stop generating the square thumbs altogether?

That is also something I was considering yesterday, square thumbs are only being used on the left-right (previous-next) buttons + listings of albums when using move.

ildyria:
Some samples.

Default:
image

2/3:
image

3/2:
image

ildyria:
> It _should_ be using the aspect-preserving one; otherwise if one selects, say, 3/2 aspect, for a landscape picture the shown thumb would be a crop of a crop, i.e., both horizontal and vertical borders would be cut off -- but is that what the code currently does?

I guess that would induce changes here:
https://github.com/LycheeOrg/Lychee/blob/master/app/Relations/HasAlbumThumb.php#L42
https://github.com/LycheeOrg/Lychee/blob/master/app/Models/Extensions/Thumb.php#L39

I don't have a strong opinion on that.

app/Enum/AspectRatioType.php
```diff
@@ -0,0 +1,47 @@
...
+	{
+		return [
+			self::aspect5by4->value => '5/4 (instagram portrait)',
+			self::aspect4by5->value => '4/5 (instagram landscape)',
+			self::aspect2by3->value => '2/3 (portrait)',
+			self::aspect3by2->value => '3/2 (landscape)',
+			self::aspect1by1->value => 'square',
+			self::aspect1byx9->value => '16/9 (landscape)',
+		];
+	}
```
qwerty287: I wouldn't call this "localized". Move to locales instead?

ildyria: sure can do.

ildyria: Done. [3ef4883](https://github.com/LycheeOrg/Lychee/pull/2138/commits/3ef4883acd9a283fc37ef52e3081fb6f5d7830ae)

ildyria:
Added localization + fixed the thumbs to not use the square but the small and small2x if available.
Should be ready for review.

app/Models/Extensions/Thumb.php
```diff
@@ -81,11 +86,16 @@ public static function createFromQueryable(Relation|Builder $photoQueryable, Sor
 	 */
 	public static function createFromPhoto(?Photo $photo): ?Thumb
 	{
-		$thumb = $photo?->size_variants->getThumb();
+		$thumb = config('app.livewire', true) === true
+			? $photo?->size_variants->getSmall() ?? $photo?->size_variants->getThumb()
+			: $photo?->size_variants->getThumb();
```
kamil4: Oh gosh, a nested ternary, with no brackets no less! :scream: Would the following achieve the same effect?

```suggestion
		$thumb = config('app.livewire', true) === true && $photo?->size_variants->getSmall()
		    ? $photo?->size_variants->getSmall()
		    : $photo?->size_variants->getThumb();
```

app/Models/Extensions/Thumb.php
```diff
@@ -81,11 +86,16 @@ public static function createFromQueryable(Relation|Builder $photoQueryable, Sor
...
 		if ($thumb === null) {
 			return null;
 		}
-		$thumb2x = $photo->size_variants->getThumb2x();
+
+		$thumb2x = config('app.livewire', true) === true
+			? $photo?->size_variants->getSmall2x() // We do note provide '?? sv->Thumb2x()' because if messes up on retina display
```
kamil4:
```suggestion
			? $photo?->size_variants->getSmall2x() // We do not provide '?? sv->Thumb2x()' because it messes up on retina display
```

kamil4: I think it should actually be:

```suggestion
		$thumb2x = config('app.livewire', true) === true && $photo?->size_variants->getSmall()
			? $photo?->size_variants->getSmall2x()
```

No, `getSmall()` is not a typo -- it should repeat the same condition as when selecting `$thumb`; that way we ensure that `$thumb` and `$thumb2x` are of the same aspect ratio...

ildyria: Fixed [736d5ee](https://github.com/LycheeOrg/Lychee/pull/2138/commits/736d5ee064ca089c9b8cb3d05de30f8d8dc563be)

ildyria: Fixed [736d5ee](https://github.com/LycheeOrg/Lychee/pull/2138/commits/736d5ee064ca089c9b8cb3d05de30f8d8dc563be)

app/Models/Extensions/Thumb.php
ildyria: Fixed [736d5ee](https://github.com/LycheeOrg/Lychee/pull/2138/commits/736d5ee064ca089c9b8cb3d05de30f8d8dc563be)

app/Models/Extensions/Thumb.php
```diff
@@ -79,24 +79,38 @@ public static function createFromQueryable(Relation|Builder $photoQueryable, Sor
...
+			? $photo->size_variants->getSmall2x()
 			: $photo?->size_variants->getThumb2x();
 
+		/**
+		 * TODO: Code for later when Livewire is the only front-end.
+		 */
+		// $thumb = $photo?->size_variants->getSmall() ?? $photo?->size_variants->getThumb();
+		// if ($thumb === null) {
+		// 	return null;
+		// }
```
kamil4, line 108: Yeah, this is... more subtle than it may seem.

From what I remember, (square) thumbs are _guaranteed_ to be generated -- if necessary, through upscaling. small/small2x are not if the original is of a particularly low resolution.

Though even that "guaranteed" above is actually not always true -- thumbs might be missing for video files and I think also for password-protected albums. So the frontend (do we still call it so?) most likely already has code to handle such cases correctly.

lang/en/aspect_ratio.php
```diff
@@ -6,8 +6,8 @@
 	| Aspect Ratio lines.
 	|--------------------------------------------------------------------------
 	*/
-	'5by4' => '5/4 (instagram portrait)',
-	'4by5' => '4/5 (instagram landscape)',
+	'5by4' => '5/4 (instagram landscape)',
+	'4by5' => '5/4 (instagram portrait)',
```
kamil4: I have to admit that I have no idea why they would both be called 5/4???

kamil4 (approved):
One comment, one puzzling question, but overall no serious concerns.

lang/en/aspect_ratio.php
ildyria: oups.

PR #2155, opened 2024-01-03
Fix WebAuthn not working

### LINKED ISSUE #2154: WebAuthn Option Disappeared in Login screen in v5
After upgrading from 4.13.0 to 5.0.0, I can no longer see webauthn option in login popup. In 4.13.0, I used to see a key icon in the top left corner of the login popup and was able to login with biometric mechanism, but that button is no longer there and I can't find it anywhere else.

Tried it again after upgrading from 5.0.0 to 5.0.1, and to 5.0.2 and it's still not there.

### Browser and system
Chrome and Safari on macOS Sonoma.

ildyria:
Fixes #2154

### DIFF
--- a/lang/de/lychee.php
+++ b/lang/de/lychee.php
@@ -315,6 +315,7 @@ return [
 	'U2F_AUTHENTIFICATION_SUCCESS' => 'Authentifizierung erfolgreich!',
 	'U2F_CREDENTIALS' => 'Anmeldedaten',
 	'U2F_CREDENTIALS_DELETED' => 'Anmeldedaten gelöscht!',
+	'U2F_LOGIN' => 'Login with WebAuthn',
 
 	'NEW_PHOTOS_NOTIFICATION' => 'E-Mails für neue Fotos senden',
 	'SETTINGS_SUCCESS_NEW_PHOTOS_NOTIFICATION' => 'Benachrichtigung für neue Fotos aktualisiert',
--- a/lang/de/lychee.php
+++ b/lang/de/lychee.php
@@ -315,7 +315,7 @@ return [
 	'U2F_AUTHENTIFICATION_SUCCESS' => 'Authentifizierung erfolgreich!',
 	'U2F_CREDENTIALS' => 'Anmeldedaten',
 	'U2F_CREDENTIALS_DELETED' => 'Anmeldedaten gelöscht!',
-	'U2F_LOGIN' => 'Login with WebAuthn',
+	'U2F_LOGIN' => 'Log in with WebAuthn',
 
 	'NEW_PHOTOS_NOTIFICATION' => 'E-Mails für neue Fotos senden',
 	'SETTINGS_SUCCESS_NEW_PHOTOS_NOTIFICATION' => 'Benachrichtigung für neue Fotos aktualisiert',
--- a/lang/de/lychee.php
+++ b/lang/de/lychee.php
@@ -315,7 +315,7 @@ return [
 	'U2F_AUTHENTIFICATION_SUCCESS' => 'Authentifizierung erfolgreich!',
 	'U2F_CREDENTIALS' => 'Anmeldedaten',
 	'U2F_CREDENTIALS_DELETED' => 'Anmeldedaten gelöscht!',
-	'U2F_LOGIN' => 'Log in with WebAuthn',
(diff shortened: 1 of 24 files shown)


lang/en/lychee.php
```diff
@@ -315,6 +315,7 @@
 	'U2F_AUTHENTIFICATION_SUCCESS' => 'Authentication successful!',
 	'U2F_CREDENTIALS' => 'Credentials',
 	'U2F_CREDENTIALS_DELETED' => 'Credentials deleted!',
+	'U2F_LOGIN' => 'Login with WebAuthn',
```
kamil4: With my "grammar Nazi" hat on, I believe this should actually be:

```suggestion
	'U2F_LOGIN' => 'Log in with WebAuthn',
```

(verb vs noun)

resources/js/data/webauthn/loginWebAuthn.ts
```diff
@@ -48,5 +49,29 @@ export const loginWebAuthn = (Alpine: Alpine) =>
 					})
 					.catch(() => alpine.$dispatch("notify", [{ type: "error", msg: this.error_msg }]));
 			},
+
+			handleKeyUp(event: KeyboardEvent): void {
+				const skipped = ["TEXTAREA", "INPUT", "SELECT"];
+
+				if (document.activeElement !== null && skipped.includes(document.activeElement.nodeName)) {
+					console.log("WAuthn skipped: " + document.activeElement.nodeName);
```
kamil4: A leftover?

resources/views/components/webauthn/login.blade.php
```diff
@@ -2,8 +2,8 @@
     bg-black/80 z-5 fixed flex items-center justify-center w-full h-full top-0 left-0 box-border opacity-100"
     data-closable="true"
     x-data="loginWebAuthn('{{ __("lychee.U2F_AUTHENTIFICATION_SUCCESS") }}', '{{ __("lychee.ERROR_TEXT") }}')"
-    @keydown.window="if (event.key === 'k') { webAuthnOpen = true }"
-    @keydown.escape.window="webAuthnOpen = false"
+    @keyup.window="handleKeyUp(event)"
+    x-on:webauthn-open.document="webAuthnOpen = true; console.log('open')"
```
kamil4: Another leftover debug log?

resources/js/data/webauthn/loginWebAuthn.ts
ildyria: not really. But I can remove it :)

resources/views/components/webauthn/login.blade.php
ildyria: that one yes. :)

resources/js/data/webauthn/loginWebAuthn.ts
kamil4: I think it depends what triggers it (and right now I don't even pretend to understand that).

If it's more of an assertion that Just Shouldn't Happen (TM) then I think it's fine to keep it (perhaps after replacing "WAuthn" with "WebAuthn", to be consistent). If on the other hand this message triggers during regular use, then I would remove it -- I don't think we should be printing to the debug log unless something rather unexpected is happening.

ildyria: That small if statement prevents to trigger keybinding behaviors when you are text areas etc...

lang/en/lychee.php
ildyria: Fixed in [cba3849](https://github.com/LycheeOrg/Lychee/pull/2155/commits/cba3849efb6514c55b72d865a60d1961525506de)

lang/de/lychee.php
```diff
@@ -315,6 +315,7 @@
 	'U2F_AUTHENTIFICATION_SUCCESS' => 'Authentifizierung erfolgreich!',
 	'U2F_CREDENTIALS' => 'Anmeldedaten',
 	'U2F_CREDENTIALS_DELETED' => 'Anmeldedaten gelöscht!',
+	'U2F_LOGIN' => 'Log in with WebAuthn',
```
qwerty287:
```suggestion
	'U2F_LOGIN' => 'Mit WebAuthn anmelden',
```

qwerty287 (approved):

PR #2161, opened 2024-01-04
Update Readme, add theme repository, optimize ImageMagick

tinohager:


ildyria (approved):

ildyria:
If you could think of something for https://github.com/LycheeOrg/LycheeOrg.github.io that would also be great :)

PR #2171, opened 2024-01-05
Remove text-neutral for easier configuration of themes

ildyria:


resources/views/components/forms/dropdown.blade.php
```diff
@@ -4,7 +4,7 @@
 after:pointer-events-none {{ $class }}">
     <select class="m-0 py-1 w-[120%] text-text-main-0 bg-transparent text-2xs px-2" {{ $attributes }}>
     @foreach($options as $key => $option)
-        <option class="text-neutral-800"
+        <option class="text-text-main-800"
```
tinohager, line 7: looks wrong for me?

`text-text-main-800`
`text-main-800`

ildyria: That's is because the color is called `text-main-800`. With tailwind it is `text-[color]` e.g. `text-neutral-400` to colorize text with `neutral-400`. We could use neutral etc.. but by doing so, it makes it easier to specify the text color separately from changing color for the full UI.

tinohager: that's because you named the variable main, isn't it? I find it somehow strange when 2x text appears

```css
.text-text-main-0 {
    color: var(--text-main-0);
}
```

For style customization, it would also be more convenient if there was a general style that could be overwritten. <br>
`lychee-h1` instead of `text-sm text-text-main-0 font-bold text-ellipsis`

ildyria: > For style customization, it would also be more convenient if there was a general style that could be overwritten.
> 
> `lychee-h1` instead of `text-sm text-text-main-0 font-bold text-ellipsis`

Yes and no, you can now override specific parts e.g. disable font-bold everywhere.
I decided to go for utility first as it prevents having to add a lot of css classes etc... https://tailwindcss.com/docs/utility-first

I think adding empty class names in strategic places is a better approach, it does not cost much on the dist.css file and would allow easier targeting of elements.

If you feel some parts of the layout are difficult to access via css, please open a pull request modifying the corresponding blade template. But it would be great if the amount of those change needed could be limited.
e.g. the following would be fine:
```html
<div class="overlay-thumb">
<h1 class="text-sm text-text-main-0 font-bold text-ellipsis">O Fortuna<h1>
Velut luna
Statu variabilis
</div>
```

while this wouldn't:
```html
<div class="overlay-thumb">
<h1 class="overlay-thumb-h1 text-sm text-text-main-0 font-bold text-ellipsis">O Fortuna<h1>
Velut luna
Statu variabilis
</div>
```
As it would be redundant.

I did not remove all the ID and class for this specific reason: to make it still not too difficult to create a theme.

kamil4 (approved):
Not the most exciting PR I've seen :wink:.

PR #2182, opened 2024-01-08
make APP_URL optional again

### LINKED ISSUE #2189: allow multiple virtual domains in APP_URL
I have set APP_URL to https://photo.examle.org and everything is working properly. Unfortunately, my page is also accessible via https://photos.example.org, https://fotos.example.org and https://foto.example.org. Is it possible to configure APP_URL accordingly (i.e., with a comma separated list).?

ildyria:
Fixes #2189

But only for displaying pictures.

### DIFF
--- a/app/View/Components/Gallery/Album/Thumbs/AlbumThumb.php
+++ b/app/View/Components/Gallery/Album/Thumbs/AlbumThumb.php
@@ -27,7 +27,7 @@ class AlbumThumb extends Component
 			$this->src = URL::asset('img/no_images.svg');
 
 			if ($thumb !== '') {
-				$this->dataSrc = URL::asset($thumb);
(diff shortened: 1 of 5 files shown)


qwerty287:
Why is this necessary? Wouldn't enforcing be better?

ildyria:
It is necessary for those who want to use WebAuthn or (Oauth in the future).
But the combination with dir makes it possible to not require it anymore (and **should limit the number of complaints "my pictures do not display anymore"**)

It will also somewhat allow this: https://github.com/LycheeOrg/Lychee/issues/2189 : being able to host on multiple urls (though once again, WebAuthn & Oauth won't work).

qwerty287 (approved):

PR #2200, opened 2024-01-13
Improve diagnostics with count of thumbs that can be regenerated

ildyria:
@LudovicRousseau 
Would you be able to test it on your install ? 
https://patch-diff.githubusercontent.com/raw/LycheeOrg/Lychee/pull/2200.patch < here is the patch.

Locally I get 
```
Info: Found 5 small that could be generated
       You can use `php artisan lychee:generate_thumbs small` to generate them.
```

### DIFF
--- a/app/Actions/Diagnostics/Pipes/Checks/SmallMediumExistsCheck.php
+++ b/app/Actions/Diagnostics/Pipes/Checks/SmallMediumExistsCheck.php
@@ -6,6 +6,7 @@ use App\Contracts\DiagnosticPipe;
 use App\Enum\SizeVariantType;
 use App\Image\SizeVariantDimensionHelpers;
 use App\Models\SizeVariant;
+use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Schema;
 
 /**
@@ -37,7 +38,8 @@ class SmallMediumExistsCheck implements DiagnosticPipe
 
 		$svHelpers = new SizeVariantDimensionHelpers();
 
-		$result = SizeVariant::query()
+		/** @var object{num_small:int,num_medium:int,num_small2x:int,num_medium2x:int,max_num_small:int,max_num_medium:int,max_num_small2x:int,max_num_medium2x:int} $result */
+		$result = DB::query()
 		->selectSub(
 			SizeVariant::query()
 			->selectRaw('COUNT(*)')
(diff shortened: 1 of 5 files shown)


LudovicRousseau:
Application of the patch fails. I do not have a `phpstan.neon` file. So I skipped this file.
In diagnostics I now get:
```
        Info: Found 610 medium that could be generated
              You can use `php artisan lychee:generate_thumbs medium` to generate them.
```

I then run: `php artisan lychee:generate_thumbs medium` that creates some new medium images.
In diagnostics I then get:
```
        Info: Found 545 medium that could be generated
              You can use `php artisan lychee:generate_thumbs medium` to generate them.
```
I then run: `php artisan lychee:generate_thumbs medium` that creates some new medium images.
In diagnostics I then get:
```
        Info: Found 501 medium that could be generated
              You can use `php artisan lychee:generate_thumbs medium` to generate them.
```
etc...

I am now at `Found 461 medium that could be generated`.

Maybe the command could generate ALL the medium images?

Since I started the `php artisan ...` from my shell the files are created as my user `rousseau` instead of `www-data`.
In diagnostics I get:
```
Warning: /home/rousseau/Serveurs_web/images.xxxx.fr/Lychee/public/uploads/medium/8d has permissions 0775, but should have 2775
Error: /home/rousseau/Serveurs_web/images.xxxx.fr/Lychee/public/uploads/medium/8d is not writable by www-data
[...]
```
Solution: `sudo chgrp -R www-data medium`

I discover that the files in `big/` are NOT in sub-folders. And in `medium/` I see sub-directories like `medium/fd/54/`. I guess you changed the way to store files to increase performance. Very good!
Do you plan to also upgrade the way the files are stored in `big/`? That could be done in the "Maintenance" page.

ildyria:
Actually there is no `big` anymore... It is `original` now. :)
What is in `big` is legacy uploads.

ildyria:
@LudovicRousseau I improved the feedback for the thumb generation, note that you will still be blocked by the timeout (which is probably why it was generating ~50 images instead of generating the default 100 of them).

LudovicRousseau:
> Actually there is no `big` anymore... It is `original` now. :) What is in `big` is legacy uploads.

Ah OK.
Maybe I should re-upload my albums then.

ildyria:
> Ah OK. Maybe I should re-upload my albums then.

You don't need to, all new pictures will be place in `original`. The optimization is mostly if you are browsing the images in your file explorer.

app/Actions/Diagnostics/Pipes/Checks/SmallMediumExistsCheck.php
```diff
@@ -0,0 +1,135 @@
...
+			->selectRaw('COUNT(*)')
+			->where('type', '=', SizeVariantType::MEDIUM2X),
+			self::NUM_MEDIUM2X
+		)
+		->selectSub(
+			SizeVariant::query()
+			->selectRaw('COUNT(*)')
+			->where(fn ($q) => $q
+				->when($svHelpers->getMaxWidth(SizeVariantType::SMALL) !== 0, fn ($q1) => $q1->where('width', '>', $svHelpers->getMaxWidth(SizeVariantType::SMALL)))
+				->when($svHelpers->getMaxHeight(SizeVariantType::SMALL) !== 0, fn ($q2) => $q2->orWhere('height', '>', $svHelpers->getMaxHeight(SizeVariantType::SMALL)))
```
kamil4, line 72: Just wondering if that `orWhere` in the second line will behave as expected when `where` is missing (i.e., if width in the first line is `0`)? It won't generate a syntactically invalid SQL query or something?

(I know -- I should test; I'm just not there yet)

Same question with respect to the other three size variants below...

ildyria: I would believe that Laravel is smart enough to resolve that. :)

qwerty287 (approved):
untested

ildyria:
> untested

I tested it locally and with @LudovicRousseau  :)

PR #2210, opened 2024-01-18
Fix Diagnostics not visible when migrations are pending

ildyria:


app/Metadata/Versions/FileVersion.php
```diff
@@ -44,7 +45,7 @@ public function __construct()
 	 */
 	public function hydrate(bool $withRemote = true, bool $useCache = true): void
 	{
-		if ($withRemote) {
+		if ($withRemote && Schema::hasTable('configs')) {
```
kamil4, line 48: Is that needed in case of an incomplete installation?

app/Metadata/Versions/FileVersion.php
```diff
@@ -44,7 +45,7 @@ public function __construct()
...
 			$updateReques = resolve(UpdateRequest::class);
 			$json = $updateReques->get_json($useCache);
```
kamil4:
```suggestion
			$updateRequest = resolve(UpdateRequest::class);
			$json = $updateRequest->get_json($useCache);
```

kamil4 (approved):
LGTM (not tested).

app/Metadata/Versions/FileVersion.php
ildyria: yes, otherwise it fails with key not found in config.

PR #2217, opened 2024-01-20
Fix import from Dropbox from Livewire side

### LINKED ISSUE #2215: Import from Dropbox does nothing, how to debug?
I had an install of Lychee v4 working with dropbox, but due to some issues with my webhost I ended up making a new Lychee v5.1.0 instance using the official docker image and was just going to re-import all of my photos from Dropbox.

I generated a new API key for my dropbox app that had worked previously and entered it into the config.

Now when I click "+" and then "Import from Dropbox" nothing happens.

Using the web console I see no network activity, no script errors, etc, just nothing.

I used Dropbox's API tester to make sure my auth key worked and it does. 

Something I noticed is all of the Lychee docs refer to a 'Drop ins" key but it doesn't appear that dropbox offers the ability to create a drop ins key anymore just a general API key?

Is there a way to debug where this is breaking down?

1. Go to "+"
2. Click on "Import from dropbox"
(No action happens from here)

NOTE: Instance is behind a load balancer, so the instance listens on port 80/http but the app_url refers to the load balancer listing at 443/https which is why the two APP_URL errors appear.  It does not seem to affect the serving of images.

   Diagnostics
    -----------
        Warning: zend.assertions is disabled although Lychee is in debug mode. For easier debugging code generation for assertions should be enabled.
        Error: APP_URL (https://lychee.***************aphy.com) does not match the current url (http://lychee.***************aphy.com).
               This will break WebAuthn authentication.
        Error: APP_URL (https://lychee.***************aphy.com) does not match the current url (http://lychee.***************aphy.com).
               This will prevent images from being properly displayed.
        Info: Latest version of PHP is 8.3
        Warning: git (software) is not available.

### Browser and system
MacOS Firefox 121 and Chrome 120 for client, oracle linux + docker for server.

ildyria:
Fixes #2215 

The full component was not implemented...
Similar to v4 front-end, it fetches the URL of the images and re-use the Import by URL.

Review shouldn't be too hard. This pull request also highlight the interactions between AlpineJS & Livewire.


resources/views/livewire/components/modal.blade.php
```diff
@@ -1,7 +1,7 @@
 <div>
     @if ($isOpen)
         <div class="basicModalContainer transition-opacity duration-1000 ease-in animate-fadeIn
-					bg-black/80 z-50 fixed flex items-center justify-center w-full
```
ildyria, line 4: This is necessary because otherwise notifications are below the backdrop.

resources/views/livewire/pages/settings.blade.php
```diff
@@ -5,7 +5,7 @@
     </x-header.bar>
 	<div class="overflow-x-clip overflow-y-auto h-[calc(100vh-56px)]">
 		<div class="settings_view w-10/12 max-w-xl text-text-main-400 text-sm mx-auto">
-			<livewire:forms.settings.base.string-setting key="set-dropbox-key" description="DROPBOX_TEXT"
+			<livewire:forms.settings.base.password-setting key="set-dropbox-key" description="DROPBOX_TEXT"
```
ildyria, line 8: This is to avoid revealing the api key while going to the settings menu.

qwerty287 (approved):
Code looks good, but untested, I don't use dropbox

ildyria:
I don't normally use Dropbox, but I tested it locally and it was working fine assuming that the Dropbox settings are properly set.

PR #2221, opened 2024-01-20
Remove is_public & deprecated Public smart album

### LINKED ISSUE #2224: Remove deadcode (is_public on photos & public smart album)

ildyria:


kamil4 (approved):
LGTM (not tested).

What does it do to the old front end? I see an exception being thrown in routes... Will things continue to work so long as the feature is not used?

ildyria:
> LGTM (not tested).
> 
> What does it do to the old front end? I see an exception being thrown in routes... Will things continue to work so long as the feature is not used?

the public album disappear.
the set public slider on the visibility panel of pictures will throw the aforementioned exception (with the error message visible in the bar).

PR #2226, opened 2024-01-21
Anonymize the paths in the diagnostics

ildyria:
Similar to urls, just to make it easier for people to copy paste their diagnostics while preserving privacy.


kamil4 (approved):
LGTM (not tested)

PR #2232, opened 2024-01-22
Add missing secure header for redirection

ildyria:
While testing with dev, the CSP is disabled...
Obviously when switching to prod we get a complaint...


d7415 (approved):
Untested

PR #2242, opened 2024-01-23
No more album ID within the history

ildyria:
Redundant info which could simply be hard coded...


kamil4 (approved):
LGTM (in some free time I will need to learn what this whole job history table is about, as I don't remember it :wink:).

PR #2245, opened 2024-01-25
fix CVE-2020-8203

ildyria:
https://github.com/advisories/GHSA-p6mc-m468-83gw/dependabot

See more here: https://github.com/fi3ework/vite-plugin-checker/issues/297

> See [CVE-2020-8203](https://github.com/advisories/GHSA-p6mc-m468-83gw/dependabot), which comes from lodash.pick; the locked version used by this plugin is considered vulnerable.
>
> This is not a direct impact on my project; we had vite-plugin-checker installed but were not using it, and **our resolution was to remove the plugin**. The likelihood of this being a real issue is also close to zero because of the context where vite-plugin-checker is run…but some of us are beholden to SOC2 or similar acronyms.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,6 @@
                 "ts-loader": "^9.5.1",
                 "typescript": "^5.3.3",
                 "vite": "^5.0.12",
-                "vite-plugin-checker": "^0.6.2",
                 "vite-plugin-commonjs": "^0.10.1"
             },
             "devDependencies": {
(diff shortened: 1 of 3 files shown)


d7415 (approved):
If we don't need it, by all means get rid.

Not tested

ildyria:
> If we don't need it, by all means get rid.

I don't really needed it, it would be a nice to have. The checker pretty much breaks the build JS if Ts is not happy.
By default Vite does a transpilation from Ts to Js, this checker adds the check at build time.

PR #2246, opened 2024-01-26
Avoid white flash on v4 redirection

ildyria:


d7415 (commented):
It makes the redirect look like the rest of Lychee to avoid the flash?

ildyria:
> It makes the redirect look like the rest of Lychee to avoid the flash?

yep.

d7415 (approved):

PR #2251, opened 2024-01-27
Improve diagnostics by adding check of temporary file systems

ildyria:


app/Actions/Diagnostics/Pipes/Checks/BasicPermissionCheck.php
```diff
@@ -300,7 +304,7 @@ private function anonymize(string $path): string
 			$this->realPaths[] = public_path();
 			$this->anonymizePaths[] = Helpers::censor(public_path(), 0.2);
 			$this->realPaths[] = storage_path();
-			$this->anonymizePaths[] = Helpers::censor(storage_path(), 0.2);
+			$this->anonymizePaths[] = Helpers::censor(storage_path(), 0.4);
```
kamil4, line 307: Why is this one "special"?

kamil4 (approved):
LGTM (not tested).

app/Actions/Diagnostics/Pipes/Checks/BasicPermissionCheck.php
ildyria: Because it masked part of the storage path, making it hard to understand exactly which folder was mentionned. :/

PR #2257, opened 2024-01-28
Re-arrange commands

ildyria:
Nothing much going on there. Just moving files around to make easier to guess their concern.


kamil4 (approved):
LGTM (not tested).

PR #2269, opened 2024-01-31
Mark some classes are readonly

ildyria:
Just smaller code base, stricter code.

### DIFF
--- a/app/Actions/Photo/Archive.php
+++ b/app/Actions/Photo/Archive.php
@@ -90,8 +90,8 @@ class Archive
 
 		$responseGenerator = function () use ($archiveFileInfo) {
 			$outputStream = fopen('php://output', 'wb');
-			stream_copy_to_stream($archiveFileInfo->getFile()->read(), $outputStream);
-			$archiveFileInfo->getFile()->close();
+			stream_copy_to_stream($archiveFileInfo->file->read(), $outputStream);
+			$archiveFileInfo->file->close();
 			fclose($outputStream);
 		};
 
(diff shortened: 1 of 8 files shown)


app/Actions/Photo/Archive.php
```diff
@@ -90,8 +90,8 @@ protected function file(Photo $photo, DownloadVariantType $downloadVariant): Str
 
 		$responseGenerator = function () use ($archiveFileInfo) {
 			$outputStream = fopen('php://output', 'wb');
-			stream_copy_to_stream($archiveFileInfo->getFile()->read(), $outputStream);
-			$archiveFileInfo->getFile()->close();
+			stream_copy_to_stream($archiveFileInfo->file->read(), $outputStream);
```
hereldar, line 93: I think that the `ArchiveFileInfo::$file` property is protected, so it cannot be accesed here.

PhpStan should alert about this type of problems. I don't know if it didn't check this PR or it's level is too low.

hereldar (commented):
Most of the changes look good for me, I have only commented a little error.

app/Actions/Photo/Archive.php
ildyria: I had a doubt, but because ArchiveFileInfo is now `final` and `readonly` I switched it to public.
https://github.com/LycheeOrg/Lychee/blob/9deb15b5a4f5fbeb5a47d9681c13542109afa9db/app/Actions/Photo/Extensions/ArchiveFileInfo.php#L50-L54

hereldar: Read-only properties have some limitations due to they do not ensure values are inmutable.

They only ensure that is not possible to assign a new value.

<https://php.watch/versions/8.1/readonly#limitations>

Using getters avoids that problem for strings and arrays (as the are copied), but not for objects (as they are passed bay reference).

That said, mos of the cases you can use  public read-only properties as replacement of getters with no problems.

hereldar (approved):

d7415 (approved):
Based mostly on @hereldar's review

PR #2270, opened 2024-01-31
Improve coverage

ildyria:
- removing dead code, and non-attainable lines...

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,6 +1,9 @@
 ---
 codecov:
   require_ci_to_pass: true
+  notify:
+    after_n_builds: 9
+    wait_for_ci: true
 comment:
   behavior: default
   require_base: false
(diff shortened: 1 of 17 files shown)


codecov.yml
```diff
@@ -1,6 +1,9 @@
 ---
 codecov:
   require_ci_to_pass: true
+  notify:
+    after_n_builds: 9
```
d7415, line 5: Is this to stop the "This build decreases coverage by 0.000000000000007%" messages?

d7415 (dismissed):
Doesn't look like it was used (and you should know)

codecov.yml
ildyria: YES!

ildyria:
Mostly adding some code ignore in un-testable corners + fix codecov warnings a bit too early...

d7415 (approved):
I suppose the desired approach is that we create tests that throw these errors, but that's not going to be straightforward for all of them.

ildyria:
> I suppose the desired approach is that we create tests that throw these errors, but that's not going to be straightforward for all of them.

good luck with that, especially the ones where you need to not have posix not enabled... Those somewhat requires to have their own ubuntu stack etc. :(

PR #2277, opened 2024-02-01
Improved mobile UI

ildyria:


ildyria:
Testing stuff... though ideally it would be better to directly build a docker image out of the branch with https://github.com/LycheeOrg/Lychee-Docker/pull/185

PR #2282, opened 2024-02-05
Add support for Feature flags.

ildyria:
This pull request reuse a similar interface of [pennant](https://laravel.com/docs/10.x/pennant) but in a much lighter version.

:warning: For this specific one, we do not make use of Facade as this breaks the route creations during the initial loading and phpstan parsing.

Using Features will also ensure that flags are well defined, an exception will be thrown if a flag is not found in the configuration. As usual flags are usually mapped to an optional ENV variable. This will allow us to integrates different code path and functionalities without relying too heavily on databases queries.

### DIFF
--- a/app/Assets/Features.php
+++ b/app/Assets/Features.php
@@ -102,39 +102,23 @@ final class Features
 	 *
 	 * @template T
 	 *
-	 * @param string|array<int,string> $featureNames  to check
-	 * @param \Closure(): T            $callbackTrue  what happens if we features are enabled
-	 * @param \Closure(): T            $callbackFalse what happens if we features are disabled
-	 *
-	 * @return T
-	 */
-	public static function when(string|array $featureNames, \Closure $callbackTrue, \Closure $callbackFalse): mixed
-	{
-		if (is_array($featureNames)) {
-			return self::allAreActive($featureNames) ? $callbackTrue() : $callbackFalse();
-		}
-
-		return self::active($featureNames) ? $callbackTrue() : $callbackFalse();
-	}
-
-	/**
-	 * Determine whether a feature is active.
-	 *
-	 * @template T
-	 *
 	 * @param string|array<int,string> $featureNames to check
-	 * @param T                        $valIfTrue    Value if we features are enabled
-	 * @param T                        $valIfFalse   Value if we features are disabled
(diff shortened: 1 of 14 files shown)


app/Assets/Features.php
```diff
@@ -0,0 +1,156 @@
...
+final class Features
+{
+	/**
+	 * Determine whether a feature is active.
+	 *
+	 * @param string $featureName to check
+	 *
+	 * @return bool is active
+	 */
+	public static function active(string $featureName): bool
```
hereldar, line 19: Is common to add the `is` prefix the methods that return a boolean, but I don't know the code base of this project, so keep the cohesion with the rest of the code.

app/Assets/Features.php
```diff
@@ -0,0 +1,156 @@
...
+	 *
+	 * @template T
+	 *
+	 * @param string|array<int,string> $featureNames to check
+	 * @param T                        $valIfTrue    Value if we features are enabled
+	 * @param T                        $valIfFalse   Value if we features are disabled
+	 *
+	 * @return T
+	 */
+	public static function whenConst(string|array $featureNames, mixed $valIfTrue, mixed $valIfFalse): mixed
```
hereldar: The `const` suffix confused me until I discovered the `when()` method.

You could merge them (detecting that the value is callable).

Another option might be to use a different word, such as "whether".

hereldar (commented):
The changes look good for me. I have added a couple of comments, but they are just suggestions.

app/Assets/Features.php
ildyria: I do agree with you, but in this specific case, I was trying to keep the interface of penant. 
https://laravel.com/docs/10.x/pennant#checking-features
image

app/Assets/Features.php
ildyria: Good point.

ildyria: Fixed: [faee879](https://github.com/LycheeOrg/Lychee/pull/2282/commits/faee87946b9a53277e45f18ed98da788bfdf3684)

hereldar (approved):

d7415 (approved):
Based mostly on @hereldar's review

PR #2291, opened 2024-02-09
If LegacyIdException is thrown provide proper solution

ildyria:
Fixes #2290.

This small handler should resolve some of the exceptions we are having complaints about legacy id.
(though I think we should probably drop the legacy id completely, it is not like we are going to go back at any time.


d7415 (approved):
Looks reasonable.

PR #2298, opened 2024-02-21
Add permission when user is not owner but creating an album.

### LINKED ISSUE #2297: Newly user created album should be opened on create in admin owned album
### Discussed in https://github.com/LycheeOrg/Lychee/discussions/2296

<div type='discussions-op-text'>

<sup>Originally posted by **haivala** February 20, 2024</sup>
### Lychee version

master

### Which PHP version are you using?

PHP 8.2

When I create album as a user in admin owned album I'm redirected to front page.
I think the newly created album should opened instead.

1. create album as admin
2. share album with user
3. make album public by url
4. login as user
5. try to create album to this admin owned album
6. User is redirected to front page 
7. As a user I cannot see this newly created album in the admin owned album.
8. When I login as admin the album has been created but the owner is admin.

Is this expected? Should it be changed?

I have album structure
Album (Event) - Album (photographer) - Pictures
I achieve this now so that I let the users make the Event album where they state their name and as a admin I move the user owned album to the admin owned album, but I think it would be more usable the way I just tried and described as It would give the user all permissions to the albums they own and permissions would not be transferred to admin.

### Diagnostics [REQUIRED]

```
System Information
------------------
Lychee Version (git):                    master (57e3503) -- Up to date (unknown).
DB Version:                              5.1.2

composer install:                        --no-dev
APP_ENV:                                 production
APP_DEBUG:                               false
APP_URL:                                 set
APP_DIR:                                 default
LOG_VIEWER_ENABLED:                      true
LIVEWIRE_ENABLED:                        true

System:                                  Linux
PHP Version:                             8.2.15
PHP User agent:                          Lychee/4 (https://lycheeorg.github.io/)
Timezone:                                Europe/Helsinki
Max uploaded file size:                  500M
Max post size:                           300M
Livewire chunk size:                     12.00 MB
Max execution time:                      0
MySQL Version:                           10.11.7-MariaDB-1:10.11.7+maria~ubu2204

exec() Available:                        yes
Imagick Available:                       1
Imagick Enabled:                         1
Imagick Version:                         1809
GD Version:                              bundled (2.1.0 compatible)
Number of foreign key:                   12 found.
```

### Browser & System [REQUIRED]

FF, Docker

### Please confirm (incomplete submissions will not be addressed)

- [X] I have provided easy and step-by-step instructions to reproduce the bug.
- [X] I understand my bug report will be removed if I haven't met the criteria above.</div>

ildyria:
Edge case discovered by @haivala 

Fixes #2297

### DIFF
--- a/app/Actions/Album/Create.php
+++ b/app/Actions/Album/Create.php
@@ -85,7 +85,7 @@ class Create extends Action
 			$album->access_permissions()->saveMany($this->copyPermission($parentAlbum));
 		}
 
-		$this->fixPermissionIfNotOwner($album);
+		$this->grantFullPermissionsToNewOwner($album);
 	}
 
 	/**
@@ -114,7 +114,7 @@ class Create extends Action
 	 *
 	 * @return void
 	 */
-	private function fixPermissionIfNotOwner(Album $album)
+	private function grantFullPermissionsToNewOwner(Album $album)
 	{
 		if ($album->owner_id === $this->intendedOwnerId) {
 			return;
@@ -125,14 +125,7 @@ class Create extends Action
 			->where(APC::BASE_ALBUM_ID, '=', $album->id)
 			->delete();
 
-		$accessPerm = new AccessPermission([
-			APC::USER_ID => $this->intendedOwnerId,
-			APC::GRANTS_FULL_PHOTO_ACCESS => true,
-			APC::GRANTS_DOWNLOAD => true,
-			APC::GRANTS_UPLOAD => true,
-			APC::GRANTS_EDIT => true,
-			APC::GRANTS_DELETE => true,
-		]);
+		$accessPerm = AccessPermission::withGrantFullPermissionsToUser($this->intendedOwnerId);
 
 		$album->access_permissions()->save($accessPerm);
 	}
--- a/app/Actions/Album/Create.php
+++ b/app/Actions/Album/Create.php
@@ -2,9 +2,12 @@
 
 namespace App\Actions\Album;
 
+use App\Constants\AccessPermissionConstants as APC;
 use App\Enum\DefaultAlbumProtectionType;
+use App\Exceptions\ConfigurationKeyMissingException;
 use App\Exceptions\ModelDBException;
 use App\Exceptions\UnauthenticatedException;
+use App\Exceptions\UnexpectedException;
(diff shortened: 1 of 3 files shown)


app/Actions/Album/Create.php
```diff
@@ -85,4 +105,35 @@ private function copyPermission(?Album $parentAlbum): array
...
+		if ($album->owner_id === $this->intendedOwnerId) {
+			return;
+		}
+
+		$album->access_permissions()
+			->where(APC::USER_ID, '=', $this->intendedOwnerId)
+			->where(APC::BASE_ALBUM_ID, '=', $album->id)
+			->delete();
+
+		$accessPerm = new AccessPermission([
```
codisart: Maybe we could add a static constructor in the AccessPermission class to have with only an id as parameter and which put all grants at true ?

app/Actions/Album/Create.php
```diff
@@ -68,6 +62,32 @@ private function set_parent(Album $album, ?Album $parentAlbum): void
...
+
+		if ($defaultProtectionType === DefaultAlbumProtectionType::PUBLIC) {
+			$album->access_permissions()->saveMany([AccessPermission::ofPublic()]);
+		}
+
+		if ($defaultProtectionType === DefaultAlbumProtectionType::INHERIT && $parentAlbum !== null) {
+			$album->access_permissions()->saveMany($this->copyPermission($parentAlbum));
+		}
+
+		$this->fixPermissionIfNotOwner($album);
```
codisart: Does the fix verb is adequate ? Maybe we could have something like "grantFullPermissionsToNewOwner"

ildyria: Fixed: [d4e2bdf](https://github.com/LycheeOrg/Lychee/pull/2298/commits/d4e2bdf48204ecc6264b7a88cdfdd85cc26e536b)

app/Actions/Album/Create.php
ildyria: Fixed : [d4e2bdf](https://github.com/LycheeOrg/Lychee/pull/2298/commits/d4e2bdf48204ecc6264b7a88cdfdd85cc26e536b)

codisart (approved):

d7415 (approved):

PR #2325, opened 2024-03-20
Pinning dependencies + fix permissions

ildyria:


d7415 (approved):
Pinning to exact hashes seems like more admin, and probably means we should subscribe to their releases, but if you want to, sure.

ildyria:
> Pinning to exact hashes seems like more admin, and probably means we should subscribe to their releases, but if you want to, sure.

It is more to fix one of the OSSF finding.
whether you pin the version of the hash is pretty much the same.

PR #2326, opened 2024-03-21
Fix livewire breaking on non debug with sub folder

ildyria:
Big thanks to @Gendra13 for laying out the ground work.


Gendra13:
I have tested this PR and can confirm that it fixes the issue.
Thank you very much.

d7415 (approved):

PR #2349, opened 2024-04-06
add Cleaning module

ildyria:
Built on top of #2348


app/Livewire/Components/Modules/Maintenance/Cleaning.php
```diff
@@ -0,0 +1,114 @@
...
+
+		$count = 0;
+		foreach (new \DirectoryIterator($this->path) as $fileInfo) {
+			if ($fileInfo->isDot()) {
+				continue;
+			}
+			if (in_array($fileInfo->getFilename(), $this->skip, true)) {
+				continue;
+			}
+			$count = 1;
```
d7415: Could we just return `false` from here?

d7415 (approved):
One query, otherwise looks ok.

app/Livewire/Components/Modules/Maintenance/Cleaning.php
ildyria: true, I should be using a boolean here.

ildyria: see [4dcebc5](https://github.com/LycheeOrg/Lychee/pull/2349/commits/4dcebc5c89d8d4f3d86706561d26509c167d22f7)

PR #2354, opened 2024-04-06
Add module to fix the missing file sizes from size variants

ildyria:
Based on top of #2353


app/Livewire/Components/Modules/Maintenance/MissingFileSizes.php
```diff
@@ -0,0 +1,93 @@
...
+use App\Models\SizeVariant;
+use App\Policies\SettingsPolicy;
+use Illuminate\Support\Facades\Gate;
+use Illuminate\Support\Facades\Log;
+use Illuminate\View\View;
+use Livewire\Component;
+
+/**
+ * We may miss some size variants because of generation problem,
+ * transfer of files, or other.
```
d7415:
```suggestion
 * We may be missing some file sizes because of generation problems,
 * transfer of files, or other.
```

app/Livewire/Components/Modules/Maintenance/MissingFileSizes.php
```diff
@@ -0,0 +1,93 @@
...
+	public function render(): View
+	{
+		Gate::authorize(SettingsPolicy::CAN_UPDATE, Configs::class);
+
+		return view('livewire.modules.maintenance.fill-filesize-sizevariants');
+	}
+
+	/**
+	 * Return the size used by Lychee.
+	 * We now separate this from the initial get() call as this is quite time consuming.
```
d7415:
```suggestion
	 * Calculate missing file sizes
```

lang/cz/maintenance.php
```diff
@@ -25,6 +25,12 @@
 		'button' => 'Generate!',
 		'success' => 'Successfully generated %d %s.',
 	],
+	'fill-filesize-sizevariants' => [
+		'title' => 'File sizes missing',
+		'description' => 'Found %d small variants without file size.',
+		'button' => 'Fetch data!',
+		'success' => 'Successfully computed sizes of %d small variants.',
```
d7415, line 32: Is this only doing `small`?

resources/views/livewire/modules/maintenance/fill-filesize-sizevariants.blade.php
```diff
@@ -0,0 +1,6 @@
+<x-maintenance.card :disabled="$this->number_of_missing_size <= 0">
```
d7415, line 1: I guess the `<` is just in case?

d7415 (approved):

lang/cz/maintenance.php
ildyria: No, it is because size variants are not so clear for users outside of the code. For this reason, I preferred the nomenclature of _small variants_, which is already used in the diagnostics:
https://github.com/LycheeOrg/Lychee/blob/master/app/Actions/Diagnostics/Pipes/Checks/CountSizeVariantsCheck.php#L27

I don't mind changing though.

resources/views/livewire/modules/maintenance/fill-filesize-sizevariants.blade.php
ildyria: I think I reused some of the code from gen-missing-sizevariant which was giving weird negative values.
Originally it was `==` but to be on the safe side, `<=` is stronger. :)

lang/cz/maintenance.php
d7415: Ah, ok. That works.

PR #2363, opened 2024-04-09
[stacked 1] Prepare flag for photo processing by pipes

ildyria:
First step of refactoring photo processing to use pipes.


d7415 (approved):
Ok. Goal is to only have the new pipeline version eventually, right?

ildyria:
> Ok. Goal is to only have the new pipeline version eventually, right?

Yep with a more like canary roll out and ability to switch back easily if things breaks for unknown reasons by setting the flag.

the `PHOTO_PIPES=true` in the `.env.mariadb` ensures that both code paths are tested (with and without the pipelines.)

PR #2367, opened 2024-04-09
[stacked 4] Move old create to Legacy code.

ildyria:
Move old create to Legacy code to avoid breaking processing.


d7415 (approved):
I often think it would be really nice if git or github could detect and indicate when large blocks of code have been moved as-is.

ildyria:
> I often think it would be really nice if git or github could detect and indicate when large blocks of code have been moved as-is.

Isn't it already the case ?
At least in my view it recognises that the files have moved (mostly the namespace changed).

d7415:
> > I often think it would be really nice if git or github could detect and indicate when large blocks of code have been moved as-is.
> 
> Isn't it already the case ? At least in my view it recognises that the files have moved (mostly the namespace changed).

It does for moved files. But the create code that was moved into a new file isn't indicated that way.

PR #2369, opened 2024-04-09
[stacked 6] Support for Duplicate flow using pipes

ildyria:


app/Actions/Photo/Pipes/Shared/HydrateMetadata.php
```diff
@@ -0,0 +1,84 @@
...
+	public function handle(DuplicateDTO $state, \Closure $next): DuplicateDTO
+	{
+		if ($state->photo->title === null) {
+			$state->photo->title = $state->exifInfo->title;
+		}
+		if ($state->photo->description === null) {
+			$state->photo->description = $state->exifInfo->description;
+		}
+		if (count($state->photo->tags) === 0) {
+			$state->photo->tags = $state->exifInfo->tags;
```
d7415, line 34: Not immediately relevant: I wonder if we should merge tags if different. I don't think it's likely to come up often enough to be worth the effort though.

app/Actions/Photo/Pipes/Shared/NotifyAlbums.php
```diff
@@ -0,0 +1,23 @@
...
+use App\Actions\User\Notify;
+use App\Contracts\PhotoCreate\SharedPipe;
+use App\DTO\PhotoCreate\DuplicateDTO;
+
+/**
+ * Notify by email if a picture has been added.
+ */
+class NotifyAlbums implements SharedPipe
+{
+	public function handle(DuplicateDTO $state, \Closure $next): DuplicateDTO
```
d7415, line 14: I guess these can't use a more generic model? (allowing everything to call the same code)

app/Contracts/PhotoCreate/DuplicatePipe.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace App\Contracts\PhotoCreate;
+
+use App\DTO\PhotoCreate\DuplicateDTO;
+
+/**
+ * Basic definition of a Photo creation pipe.
```
d7415:
```suggestion
 * Basic definition of a Duplicate Photo pipe.
```
?

app/Contracts/PhotoCreate/SharedPipe.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace App\Contracts\PhotoCreate;
+
+use App\DTO\PhotoCreate\DuplicateDTO;
+
+/**
+ * Basic definition of a Photo creation pipe.
```
d7415: ?

d7415 (approved):

app/Actions/Photo/Pipes/Shared/NotifyAlbums.php
ildyria: It will be more generic later, I will use Union type in for SharedPipe later, but this is not in the scope of this PR as it only deals with the duplicate flow. :)

app/Contracts/PhotoCreate/SharedPipe.php
ildyria: Good catch. Need to improve the doc here. :D

PR #2372, opened 2024-04-10
[stacked 8] Refactor Save/Notify to use interface DTO.

ildyria:


app/Contracts/PhotoCreate/PhotoDTO.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace App\Contracts\PhotoCreate;
+
+use App\Models\Photo;
+
+interface PhotoDTO
```
d7415, line 7: Should this file be in DTO instead of contracts?

d7415 (approved):

app/Contracts/PhotoCreate/PhotoDTO.php
ildyria: It is an interface, hence a contract.
I am open to another other name.

d7415: Sure. Only because the others were in the DTO folder. If this makes more sense to you I'm happy as-is.

ildyria: Could consider `PhotoDTOInterface` or `HasGetPhoto`. :)

d7415: Could do - I think the current name is nice. Let's just leave it as-is.

PR #2376, opened 2024-04-10
[stacked 12] Use pipes photo partners

ildyria:


d7415 (approved):
This feels very neat.

PR #2396, opened 2024-04-20
Fixes HTTP 500 "Attempt to read property "photo_id" on null" on album page when no photo is found for header

nanawel:
It appears that `\App\Livewire\Components\Pages\Gallery\Album::fetchHeaderUrl()` does not properly handle the situation when no photo of the right ratio/type combination can be found in the content of the album.

And in this case, the result is pretty ugly :confused: 

image

The quick and dirty solution here is just to return `null`, at least until a better way to handle the case is found.

### DIFF
--- a/app/Livewire/Components/Pages/Gallery/Album.php
+++ b/app/Livewire/Components/Pages/Gallery/Album.php
@@ -211,7 +211,7 @@ class Album extends BaseAlbumComponent implements Reloadable
 			->inRandomOrder()
 			->first();
 
-		if (!$photo) {
+		if ($photo === null) {
 			return null;
 		}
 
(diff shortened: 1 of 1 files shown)


ildyria (approved):

ildyria:
Thanks a lot of the investigation.

> The quick and dirty solution here is just to return null, at least until a better way to handle the case is found.

This means that all the pictures in that folder have neither `small`, `small2x` or `medium` versions of size variant in **landscape mode**.

app/Livewire/Components/Pages/Gallery/Album.php
```diff
@@ -211,6 +211,10 @@ private function fetchHeaderUrl(): SizeVariant|null
 			->inRandomOrder()
 			->first();
 
+		if (!$photo) {
```
ildyria:
```suggestion
		if ($photo === null) {
```

nanawel:
> This means that all the pictures in that folder have neither small, small2x or medium versions of size variant **in landscape mode**.

Exactly. Hence the "when no photo of the right ratio/type combination can be found" in my description but it wasn't that precise indeed.

PR #2400, opened 2024-04-20
Fix description empty string creating bug in layout

ildyria:


d7415 (approved):

Gendra13:
Thank you. 
I did test the fix and it works in my special case (no description at all), but there is still an issue when there is a description text present. 
I added a bit more information in the corresponding report: https://github.com/LycheeOrg/Lychee/discussions/2399#discussioncomment-9178818

ildyria:
@Gendra13  should be fixed now. I found the culprit.

Gendra13:
Yes, thank you. That fix work fine.

resources/views/components/gallery/album/details.blade.php
```diff
@@ -38,7 +38,7 @@
     @endif
 </div>
 @if($this->AlbumFormatted->description !== '')
-<div class="w-full px-7 my-4 text-justify text-text-main-200 prose prose-invert prose-sm" {{ $attributes }} >
+<div class="w-full px-7 max-w-full my-4 text-justify text-text-main-200 prose prose-invert prose-sm" {{ $attributes }} >{{-- prose set max-width to 65 chars --}}
```
d7415, line 41: Do we need w-full *and* max-w-full?

d7415 (approved):

resources/views/components/gallery/album/details.blade.php
ildyria: yes.
`w-full` = `width: 100%`
`max-w-full` = `max-width: 100%`

PR #2429, opened 2024-05-10
Use random instead of sorted for thumbs of Smart Albumbs

ildyria:


d7415 (approved):
Looks like the tests are unhappy with the random thumb in recent?

ildyria:
> Looks like the tests are unhappy with the random thumb in recent?

yeah because we are selecting from random now, I need to figure out a way for that or relax the condition. Probably because there are 2 or 3 pictures...

d7415 (approved):

PR #2442, opened 2024-05-23
Small improvements to French translation

anantone:
Small but important, some very visible mistakes for visitors.

One remark: I left "Album Created" empty, because in French it depends on the date format that follows (day, month, year), so there is no way to make it work for all date formats. Whereas leaving it empty just shows the date, which is better than a grammatically incorrect descriptor. 

Hope this is a welcome contribution! Thanks for all your hard work.


ildyria:
Merci, bien que francais, je passe 100% de mon temps en anglais, donc j'ai un peu de mal au retour en Francais, surtout sur certaines traductions. :')

ildyria (approved):

PR #2455, opened 2024-06-08
2 - Remove some queries.

ildyria:
Remove some N+1 queries.
Still a few more left to track...

### DIFF
--- a/app/Livewire/Components/Pages/Gallery/Album.php
+++ b/app/Livewire/Components/Pages/Gallery/Album.php
@@ -14,6 +14,7 @@ use App\Livewire\DTO\AlbumFormatted;
 use App\Livewire\DTO\AlbumRights;
 use App\Livewire\DTO\Layouts;
 use App\Livewire\DTO\PhotoFlags;
+use App\Livewire\DTO\ProtectedCollection;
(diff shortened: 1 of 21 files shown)


ildyria:
Doing `rand()` on `count()` + `skip` is faster than `inRandomOrder()`.

app/Livewire/Components/Pages/Gallery/Album.php
```diff
@@ -203,19 +204,22 @@ private function fetchHeaderUrl(): SizeVariant|null
...
-					->select('photo_id')
-					->whereBelongsTo($this->album->photos)
-					->whereIn('type', [SizeVariantType::MEDIUM, SizeVariantType::SMALL2X, SizeVariantType::SMALL])
-					->inRandomOrder()
-					->first();
+					->whereIn('type', [SizeVariantType::MEDIUM, SizeVariantType::SMALL2X, SizeVariantType::SMALL]);
+		$num = $query_ratio->count() - 1;
+		$photo = $query_ratio->skip(rand(0, $num))->first();
+
+		if ($photo === null) {
```
d7415, line 196: Would this test work with `$num === 0` instead? That might be harder to read though.

d7415 (approved):

PR #2464, opened 2024-06-12
Migrate to Laravel 11

ildyria:
Lots of breaking changes...


app/Providers/AppServiceProvider.php
```diff
@@ -102,6 +102,9 @@ class AppServiceProvider extends ServiceProvider
 	 */
 	public function boot()
 	{
+		// Prohibits: db:wipe, migrate:fresh, migrate:refresh, and migrate:reset
```
d7415, line 105: Do these `migrate:*` commands still exist?

database/migrations/TemporaryModels/OptimizeTables.php
```diff
@@ -46,12 +46,58 @@ public function exec(): void
...
+	{
+		if ($this->driverName === 'sqlite') {
+			return;
+		}
+
+		$fk = collect(Schema::getForeignKeys($table->getTable()))->map(fn ($a) => $a['name'])->all();
+		if (in_array($indexName, $fk, true)) {
+			$table->dropForeign($indexName);
+		}
+	}
```
d7415, line 102: Are these used?

d7415 (approved):

app/Providers/AppServiceProvider.php
ildyria: Yes. It is just a safety instead of having our homebrewed checks

database/migrations/TemporaryModels/OptimizeTables.php
ildyria: Yes in the big migration that Matthias did for refactoring the tables and some other minor migrations. 🙂

app/Providers/AppServiceProvider.php
d7415: Ok :+1:. I just wasn't sure with the deletions of `app/Console/Commands/Laravel/MigrateFresh.php` etc

PR #2475, opened 2024-06-21
Add ability to create user on the fly on Oauth auth step

ildyria:
:warning: Disabled by default.

Provide the ability of Lychee admins to shoot themselves in their foot.
Not my problem.

### DIFF
--- a/app/Actions/User/Create.php
+++ b/app/Actions/User/Create.php
@@ -14,7 +14,7 @@ class Create
 	 * @throws InvalidPropertyException
 	 * @throws ModelDBException
 	 */
-	public function do(string $username, string $password, bool $mayUpload, bool $mayEditOwnSettings): User
+	public function do(string $username, string $password, ?string $email = null, bool $mayUpload = false, bool $mayEditOwnSettings = false): User
 	{
 		if (User::query()->where('username', '=', $username)->count() !== 0) {
 			throw new ConflictingPropertyException('Username already exists');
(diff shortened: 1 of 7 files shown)


d7415 (approved):

mschilli87:
@ildyria: Could you elaborate your security concerns? I did not find any informations in the issues, this PR, the commit messages, or code (comments) on *how* I could 'shoot [myself] in [the] foot' using this feature.

ildyria:
@mschilli87  good question. :)

This PR was a response to the fact that users had to first get an account before being able to link it to their account on google, etc.

If you set `oauth_create_user_on_first_attempt` to `true` users will be created on the fly if they don't exists.
If additionally you set `oauth_grant_new_user_upload_rights` to `true`, **you effectively give everyone with a GitHub account** (if you have enabled Oauth via GitHub) **the ability to upload whatever they want to your Lychee install**...

Need I explain why this would be bad ? :)

mschilli87:
Thank you for the quick response. I get it. That makes sense.

If you don't mind me hijacking this PR for an only semi-related question: Is there any way to add a custom OAuth provider, as in one that I control myself, maybe by hacking the redirect URI of the nextcloud one, or would I have to patch that in myself? By only enabling a single, trusted, provider, the problem you describe would be eliminated as far as I can tell.

ildyria:
Lychee Oauth is built on top of Laravel Socialite.
You should look on https://socialiteproviders.com/about/ See if your Oauth system does not exists already.
If that is the case, it should be a fairly easy integration.

PR #2531, opened 2024-09-08
Fixes RSS Feed's URL

Electro707:
This simple PR fixes the URLs given by an RSS feed, which was changed from `<base>/gallery#` to `<base>/gallery/`

Tested it on my local instance, and works


ildyria (approved):

ildyria:
Thank you. 🙂

PR #2541, opened 2024-09-18
Supplement zh_CN Translation

x1ntt:
I noticed that the new version is missing some Simplified Chinese translations, so I tried to add some.

Most of the supplementary content has been validated in the application, but there are still some that I cannot find under what circumstances they will appear, but they generally do not have ambiguity.
If you can tell me where the prompt below `ERROR_GPX` appears, I will verify it again. If there are any errors, please let me know ❤

In addition, I found the `Album decorations` string, but when I modified its settings, I did not observe any changes, so I did not understand its functionality and did not translate it.


ildyria:
> If you can tell me where the prompt below ERROR_GPX appears, I will verify it again. If there are any errors, please let me know ❤

Probably in the old front-end, if you like paleontology: https://github.com/LycheeOrg/Lychee-front :)

> In addition, I found the `Album decorations` string, but when I modified its settings, I did not observe any changes, so I did not understand its functionality and did not translate it

I have not had time to implement those in the version 5, those are relics of version 4. However I do plan to bring them back in version 6 (soon :tm: ).

ildyria (approved):

PR #2554, opened 2024-10-01
Update HasRandomIDAndLegacyTimeBasedID.php - Album ID Generation

jessienab:
Modification of ID generation based on recommendation by associate who is a PHP developer. This will avoid having a hyphen at the end of URLs which seems to cause issues with certain apps/devices that do URL parsing.

In an exceptionally rare, but not impossible situation, an ID of "--------------" would result in "".

- Title: "Resolves an edge case where hyphens would be generated at end of album IDs"
- See discussion: https://github.com/LycheeOrg/Lychee/discussions/2551#discussioncomment-10791531
- My understanding is that the hyphen is trimmed off the end of the ID if one is present.
----

- Add a test that triggered the bug before fixing it.

I am not sure how to do this...

- Ensure it doesn't break any feature.

I received this as an input from a reliable PHP dev, is all I can say, and I have to assume based on parsing the line that it shouldn't break any feature in this case?

### DIFF
--- a/app/Models/Extensions/HasRandomIDAndLegacyTimeBasedID.php
+++ b/app/Models/Extensions/HasRandomIDAndLegacyTimeBasedID.php
@@ -151,7 +151,7 @@ trait HasRandomIDAndLegacyTimeBasedID
 		// The other characters (a-z, A-Z, 0-9) are legal within an URL.
 		// As the number of bytes is divisible by 3, no trailing `=` occurs.
 		try {
-			$id = strtr(base64_encode(random_bytes(3 * RandomID::ID_LENGTH / 4)), '+/', '-_');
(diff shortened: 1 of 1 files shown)


ildyria:
I can tell you it's going to be breaking, the reason is that we check the length of the string used as if when opening an album.

`rtrim` is removing the character but is not replacing it with something else.

It will not be fully breaking but more of a probabilistic one.

jessienab:
> I can tell you it's going to be breaking, the reason is that we check the length of the string used as if when opening an album.
> 
> `rtrim` is removing the character but is not replacing it with something else.
> 
> It will not be fully breaking but more of a probabilistic one.

In this case I think it would be best to leave everything as is then, as unfortunately I can't provide a better solution, and given the very small error scope of this request, it may not be worth further effort vs better features otherwise.

ildyria (approved):

Issue #2258, opened 2024-01-28
[Enhancement] UX improvement - provide album cover placeholder when the actual cover is being downloaded


pawelmalak:
Every time we open albums view/grid we get this flash of "unstyled content" - default album covers. It takes a while to load all of them, even on local network. With lots of albums, you can see as they get populated one by one.

![lychee](https://github.com/LycheeOrg/Lychee/assets/24780509/ff5675aa-cf99-4d9c-ae6f-1d55c778aa86)

This feature request proposes to provide simplified version of album cover with the initial albums data, so they can get rendered with the album grid. This can be achived by storing cover thumbnail in base64 format or by using something like [blurhash](https://github.com/woltapp/blurhash).

This is how it looks in the [photoview](https://github.com/photoview/photoview) app. They use it for photos in the album but the same approach can be used for the album covers. 

![photoviewblur](https://github.com/LycheeOrg/Lychee/assets/24780509/2a8ca20a-9dc7-4d65-91d8-adb8de2f8e70)

Updating album cover should also update its placeholder.

ildyria:
### Todo:

- [ ] : add blur_hash column to size_variants
- [ ] : add step to generate blur_hash from thumb & small
- [ ] : add console command to generate blur_hash from already existing images
- [ ] : add diagnostic info check to count number of blur_hash missing
- [ ] : insert blur_hash in thumbnail 

### Questions

- [ ] : add JS decoding implementation ??
- [ ] : compute base64 image & store it in DB (replace blur_hash) ??

aSouchereau:
Interested in this...

Looking into it, computing our own base64 images is doable, the trade off being efficiency. In manual testing I was able to reduce an image to around 100B by down-scaling to 16x16 and converting to webp (32x32 at ~135B would be better for higher fidelity and colours). I think this would be acceptable, it would require hundreds of thumbnails to make any significant impact on page load. Compared to blurhash which is only 20-30B and similar quality. However, to address the additional size, we could add a config option to disable the placeholders altogether, for those who need it.

The only downsides to Blurhash I can think of is the decoding implementation v.s. base64 being handled directly by the browser and that it would require separate dependencies for [encoding](https://github.com/kornrunner/php-blurhash) / [decoding](https://github.com/mad-gooze/fast-blurhash).

I'm willing to implement either option, whatever we decide is best.

ildyria:
[...]

PR #2558, opened 2024-10-08
New Feature: Low Quality Image Placeholder

aSouchereau:
Implements backend for #2258 

Adds low quality image placeholder to be displayed while album covers are downloading. 
Improves perceived loading times.

Approach:
- Adds new size variant of type 7 called 'placeholder'
- Adds PlaceholderEncoder class to compress image to usable size, convert to base64 and save to db
- Adds new standalone pipe to process the placeholder
- Adds step in legacy code path to process the placeholder
- Adds command to handle placeholders that exist but have not been encoded yet
- Adds diagnostic info check for determining number of missing and unencoded placeholders
- Modifies SizeVariant getUrlAttribute to use data URI instead of app url for placeholders
- Modifies Thumb to include generated placeholder

**Notes for implementing the frontend**

- requires either a `filter: blur();` or `backdrop-filter: blur();`. I've found that between 8-15px looks best.

### DIFF
--- a/app/Actions/Diagnostics/Errors.php
+++ b/app/Actions/Diagnostics/Errors.php
@@ -15,6 +15,7 @@ use App\Actions\Diagnostics\Pipes\Checks\ImageOptCheck;
 use App\Actions\Diagnostics\Pipes\Checks\IniSettingsCheck;
 use App\Actions\Diagnostics\Pipes\Checks\MigrationCheck;
 use App\Actions\Diagnostics\Pipes\Checks\PHPVersionCheck;
+use App\Actions\Diagnostics\Pipes\Checks\PlaceholderExistsCheck;
 use App\Actions\Diagnostics\Pipes\Checks\SmallMediumExistsCheck;
 use App\Actions\Diagnostics\Pipes\Checks\SupporterCheck;
 use App\Actions\Diagnostics\Pipes\Checks\TimezoneCheck;
@@ -44,6 +45,7 @@ class Errors
 		ForeignKeyListInfo::class,
 		DBIntegrityCheck::class,
 		SmallMediumExistsCheck::class,
+		PlaceholderExistsCheck::class,
 		CountSizeVariantsCheck::class,
 		SupporterCheck::class,
 	];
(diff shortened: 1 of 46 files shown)


aSouchereau:
Apologies, this took a lot longer than I expected. Life gets in the way sometimes, I'm sure you can relate...

ildyria:
Hi, amazing piece of work!

> Life gets in the way sometimes, I'm sure you can relate...

Yep, currently feeling like I have been hit truck. Just got Covid.

None the less, I don't know if you have been following Lychee lately, but we are close to releasing a v6 which will drop all the Livewire crap for a Vue front-end. Can I trouble you later for the implementation of the placeholder for the albums in that case? :)

aSouchereau:
>  Can I trouble you later for the implementation of the placeholder for the albums in that case? :)

Yes of course, would be happy to help. You can ping me here, and I should get to it. 

Get well soon!

ildyria:
@aSouchereau I rebased your branch on top of v6 (future) so that it will be easier to merge, all the Livewire thing is gone. I also updated PhotoResource / SizeVariantsResource to provide the required data.

I think there is still `php artisan typescript:transform` that needs to be run in order to generate the required typescripts definition for the front-end, but in theory all the data are ready to be used in the v6 front-end. :)

It could also be interesting to integrate with the Maintenance page. See:
- `CreateThumbsRequest.php`
- `MaintenanceGenSizevariants.vue`
- `Maintenance.vue`

:) 

For the filter, I guess https://tailwindcss.com/docs/blur will do the job, most likely `blur-md`

aSouchereau:
Sounds good, I agree with adding the maintenance page button. I will have some extra free time in the next week to get started

ildyria:
Crap. I did not expect this to happen. >_<

ildyria:
@aSouchereau  I am sorry. To fix this, I squashed all the commits into one branch and rebased on top of master. https://github.com/LycheeOrg/Lychee/tree/album-cover-placeholder

Feel free to take it over from there.

PR #2572, opened 2024-10-19
Stacked 9: Remove redundant auth check query

ildyria:
Subscribing on LycheeStore was making the request being fired every time the lycheStore changed... definitely not optimal. We only want to reload the data when the user change. So better to just watch the reactive reference to user.


resources/js/menus/LeftMenu.vue
```diff
@@ -72,11 +72,18 @@ const { left_menu_open, clockwork_url, is_se_enabled, is_se_preview_enabled } =
...
-	}
-});
+watch(
+	() => user.value,
+	(newValue, oldValue) => {
+		if (newValue === null) {
+			initData.value = undefined;
+		}
+
+		if (newValue !== null && newValue.id !== oldValue?.id) {
```
d7415: For some reason this feels neater to me. Go with what **you** prefer though as there's no real benefit to either.
```suggestion
		else if (newValue.id !== oldValue?.id) {
```

d7415 (approved):

PR #2581, opened 2024-10-23
Stacked 17: Various fixes.

ildyria:


app/Legacy/V1/Controllers/Administration/SettingsController.php
```diff
@@ -403,6 +404,9 @@ public function getAll(GetSetAllSettingsRequest $request): Collection
 		return Configs::query()
 			->orderBy('cat')
 			->orderBy('id')
+			// Only display settings which are not part of SE
```
d7415, line 407: Do we get those separately? Or should this be behind a gate? Or are we forcing these to be done the "normal" way?

app/Legacy/V1/Controllers/Administration/SettingsController.php
```diff
@@ -419,7 +423,16 @@ public function getAll(GetSetAllSettingsRequest $request): Collection
 	public function saveAll(GetSetAllSettingsRequest $request): void
 	{
 		$lastException = null;
-		foreach ($request->except(['_token', 'function', '/api/Settings::saveAll']) as $key => $value) {
+		// Select all the SE settings.
```
d7415, line 426: Same query I guess

config/log-viewer.php
```diff
@@ -66,7 +66,7 @@ function renv_cond(string $cst): string
 	|
 	*/
 
-	'back_to_system_url' => renv('APP_URL', 'http://localhost') . renv_cond('APP_DIR'),
+	'back_to_system_url' => renv('APP_URL', 'http://localhost') . renv_cond('APP_DIR') . '/gallery',
```
d7415, line 69: :+1:

resources/js/components/modals/UploadPanel.vue
```diff
@@ -3,12 +3,14 @@
 		<template #container="{ closeCallback }">
 			<div v-if="setup">
 				<div v-if="counts.files > 0" class="m-4 flex flex-wrap justify-center">
-					<span class="w-full text-center">Completed: {{ counts.completed }} / {{ counts.files }}</span>
+					<span v-if="counts.completed === counts.files" class="w-full text-center text-muted-color-emphasis font-bold">Completed</span>
```
d7415, line 6: :man_shrugging:

resources/js/components/settings/AllSettings.vue
```diff
@@ -255,6 +257,15 @@ function save() {
 		});
 }
 
+function goto(section: string) {
+	console.log(section);
+	const el = document.getElementById(section.slice(1));
+	console.log(el);
```
d7415: Do we need/want the debug outputs?

resources/js/menus/LeftMenu.vue
```diff
@@ -220,22 +239,29 @@ function loadMenu() {
...
-
 		{
-			label: "lychee.ABOUT_LYCHEE",
-			icon: "info",
-			access: true,
-			command: () => (openLycheeAbout.value = true),
+			label: "Lychee",
+			items: [
+				{
+					label: "About",
```
d7415, line 246: Is `lychee.ABOUT_LYCHEE` no good?

resources/js/views/Sharing.vue
```diff
@@ -1,38 +1,32 @@
...
 			{{ $t("lychee.SHARING") }}
 		</template>
 		<template #end> </template>
 	</Toolbar>
 	<Panel class="border-none p-9 mx-auto max-w-3xl" v-if="perms !== undefined" pt:header:class="hidden">
-		<div class="w-full mb-9 text-center">This page gives an overview and edit the sharing rights associated with albums.</div>
-		<div class="flex flex-col">
-			<div class="flex text-muted-color items-center">
+		<div class="w-full mb-9 text-center text-muted-color-emphasis">
+			This page gives an overview and edit the sharing rights associated with albums.
```
d7415:
```suggestion
			This page gives an overview of and the ability to edit the sharing rights associated with albums.
```

d7415 (approved):

resources/js/components/settings/AllSettings.vue
ildyria: oups. :laughing:

resources/js/menus/LeftMenu.vue
ildyria: About Lychee is returning "About Lychee" and in that specific case it would feel weird to have a sub section "Lychee" and then just inside it "About Lychee" while "About" just get the job done.
But yeah, that is a thing we have to fix later.
Basically redoing the localization of the front-end...

I am really not looking forward to that one. :sweat_smile:

app/Legacy/V1/Controllers/Administration/SettingsController.php
ildyria: When using the Legacy front-end, you would be able to edit the SE settings without going through the middleware support check due to the way the code is implemented in the Legacy side.
As a result, someone would be able to change parameters that they are not supposed to.

In order to solve this, we do two things:
1. we do not display those settings. 
2. As you saw below, we also remove those parameters from the POST request if they were supplied. Effectively making those untouchable.

The idea being that if you want to modify those parameters (for things which are only available on the Vue front-end), you have to go through the new front-end.
The work around otherwise would be to disable v6, edit settings as per preference and re enable v6... Not what we want. :)

app/Legacy/V1/Controllers/Administration/SettingsController.php
ildyria: see above. :)

app/Legacy/V1/Controllers/Administration/SettingsController.php
d7415: Makes sense

PR #2605, opened 2024-10-27
Fix login menu not visible when set to the right.

ildyria:
Fixes #2604


resources/js/components/headers/AlbumsHeader.vue
```diff
@@ -271,7 +270,7 @@ const menu = computed(() =>
 			icon: "pi pi-sign-in",
 			type: "fn",
 			callback: lycheeStore.toggleLogin,
-			if: props.user.id === null && !isLoginLeft,
+			if: props.user.id === null && !isLoginLeft.value,
```
ildyria, line 273: the reason why it was not showing...

d7415 (approved):

PR #2621, opened 2024-10-28
Disable snq if secret is not set

ildyria:
Hopefully this will prevent future dependabots PR to fail and same for contributors.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,14 +3,14 @@ updates:
   - package-ecosystem: github-actions
     directory: /
     schedule:
-      interval: daily
+      interval: weekly
 
   - package-ecosystem: npm
     directory: /
     schedule:
-      interval: daily
+      interval: weekly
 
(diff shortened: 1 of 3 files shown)


d7415 (approved):
Not convinced there's any benefit to making dependabot weekly - they're rarely that frequent, so we'll just learn about problems later?

d7415:
Ok, I see a lot today. But weekly probably means the same number of PRs but concentrated into one day a week.

I mostly see this as "eek, that's a lot of dependencies"

ildyria:
> Ok, I see a lot today. But weekly probably means the same number of PRs but concentrated into one day a week.
> 
> I mostly see this as "eek, that's a lot of dependencies"

I wish some of those PR were compacted in a single one.

d7415:
Definitely. This may be worth a look https://www.hrvey.com/blog/combine-dependabot-prs

Oh, it looks like that schedule doesn't change security PRs, so I retract my comment.

PR #2624, opened 2024-10-29
Fix link to Logs in case of sub folder hosting

ildyria:


d7415 (approved):
`Constants.BASE_URL` includes the leading `/`, right?

d7415:
> `Constants.BASE_URL` includes the leading `/`, right?

Based on other code I'm going to assume yes

ildyria:
> `Constants.BASE_URL` includes the leading `/`, right?

even better, it is fetched directly from here:
https://github.com/LycheeOrg/Lychee/blob/913511a57791eb49634e41f303c8b8963c189aa7/resources/views/components/meta.blade.php#L1

And defined here:
https://github.com/LycheeOrg/Lychee/blob/913511a57791eb49634e41f303c8b8963c189aa7/app/View/Components/Meta.php#L68

PR #2629, opened 2024-10-30
Get automated signed releases

ildyria:
See:
- https://github.com/ossf/scorecard/blob/ea7e27ed41b76ab879c862fa0ca4cc9c61764ee4/docs/checks.md#signed-releases
- https://github.blog/changelog/2024-06-25-artifact-attestations-is-generally-available/

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -159,6 +159,10 @@ jobs:
       - tests_legacy
       - check_js
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+      attestations: write
     env:
       extensions: bcmath, curl, dom, gd, imagick, json, libxml, mbstring, pcntl, pdo, pdo_sqlite, pdo_mysql, pdo_pgsql, pgsql, sqlite3, zip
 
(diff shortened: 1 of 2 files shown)


.github/workflows/CICD.yml
```diff
@@ -178,29 +182,50 @@ jobs:
...
-        with:
-          composer-options: --no-dev
-
       - name: Use Node.js 20
         uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
         with:
             node-version: 20
 
-      - name: Install
-        run: npm ci
```
d7415, line 192: IIRC we only just changed to `ci`. It looks like this block was always redundant (handled by the `make` below) but the `make` uses `npm install`. Do we want this or should we change the Makefile too?

d7415 (approved):

.github/workflows/CICD.yml
ildyria: I think we should change the makefile too.

ildyria: [978559f](https://github.com/LycheeOrg/Lychee/pull/2629/commits/978559f153d4d44a2a668e00b573ae1595bc2de9)

PR #2630, opened 2024-10-30
Fix sub-albums sorting not being respected per album

ildyria:
This thing has been plaguing us for a while.
Finally found the proper fix.

### DIFF
--- a/app/Assets/Helpers.php
+++ b/app/Assets/Helpers.php
@@ -3,6 +3,7 @@
 namespace App\Assets;
 
 use App\Exceptions\Internal\ZeroModuloException;
+use Exception;
 use Illuminate\Support\Facades\File;
 use function Safe\ini_get;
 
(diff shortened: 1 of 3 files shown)


app/Assets/Helpers.php
```diff
@@ -261,4 +262,23 @@ public function censor(string $string, float $percentOfClear = 0.5): string
...
+	/**
+	 * Format exception trace as text.
+	 *
+	 * @param \Exception $e
+	 *
+	 * @return string
+	 *
+	 * @codeCoverageIgnore
+	 */
+	public function exceptionTraceToText(\Exception $e): string
```
ildyria, line 275: Not used in the PR, but quite useful when debugging:
```php
dd($var, Helpers::exceptionTraceToText(new Exception))
```
Will dump the value of $var in a 500 and followed by the trace of where the dump happened.
The front-end is already taking care of making it a nice visual :)

app/Models/Album.php
```diff
@@ -462,4 +462,14 @@ protected function setAlbumSortingAttribute(?AlbumSortingCriterion $sorting): vo
...
 	}
+
+	/**
+	 * Returns the criterion acc. to which **albums** inside the album shall be sorted.
+	 *
+	 * @return AlbumSortingCriterion the attribute acc. to which **albums** inside the album shall be sorted
+	 */
+	public function getEffectiveAlbumSorting(): AlbumSortingCriterion
+	{
+		return $this->getAlbumSortingAttribute() ?? AlbumSortingCriterion::createDefault();
```
ildyria, line 473: `getAlbumSortingAttribute()` is defined on line 448 of the same file.

app/Models/Album.php
```diff
@@ -462,4 +462,14 @@ protected function setAlbumSortingAttribute(?AlbumSortingCriterion $sorting): vo
 		$this->attributes['album_sorting_col'] = $sorting?->column->value;
 		$this->attributes['album_sorting_order'] = $sorting?->order->value;
 	}
+
+	/**
+	 * Returns the criterion acc. to which **albums** inside the album shall be sorted.
+	 *
+	 * @return AlbumSortingCriterion the attribute acc. to which **albums** inside the album shall be sorted
```
d7415:
```suggestion
	 * @return AlbumSortingCriterion
```
Seems redundant

d7415 (approved):

PR #2636, opened 2024-10-31
Vite local dev

ildyria:
This is more of a edge case. This allows people to only contribute on front-end without having to install by leveraging the server proxy option.

This development is defaulted to light mode.
In order to have dark mode, the user will need to trigger it in settings... That's the cost of not running the full server.


resources/js/router/routes.ts
```diff
@@ -123,3 +123,14 @@ export const routes = [
 		component: Users,
 	},
 ];
+
+console.log(import.meta.env);
```
d7415:
```suggestion
```

tsconfig.json
```diff
@@ -26,9 +26,9 @@
     // "moduleDetection": "auto",                        /* Control what method is used to detect module-format JS files. */
 
     /* Modules */
-    "module": "CommonJS",                                /* Specify what module code is generated. */
-    "rootDir": "./resources/js",                                  /* Specify the root folder within your source files. */
-    // "moduleResolution": "node10",                     /* Specify how TypeScript looks up a file from a given module specifier. */
+    "module": "es2020",                                  /* Specify what module code is generated. */
+    "rootDir": "./resources/js",                         /* Specify the root folder within your source files. */
+    "moduleResolution": "bundler",                     /* Specify how TypeScript looks up a file from a given module specifier. */
```
d7415:
```suggestion
    "moduleResolution": "bundler",                       /* Specify how TypeScript looks up a file from a given module specifier. */
```

vite.config.ts
```diff
@@ -67,4 +66,110 @@ export default defineConfig({
...
+//     }
+//   },
+//   css: {
+//     preprocessorOptions: {
+//       scss: {
+//         api: 'modern'
+//       },
+//     }
+//   }
+// });
```
d7415: Do we want any of this?

vite/index.html
d7415, line 1: Are we going to remember to maintain this? Though I guess "It should keep working and anyone using it can tell us when it breaks" is valid for edge cases. Especially developer-focused edge cases.

tsconfig.json
```diff
@@ -12,8 +12,8 @@
     // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
 
     /* Language and Environment */
-    "target": "es6",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
-    "lib": ["ES6", "DOM"],                            /* Specify a set of bundled library declaration files that describe the target runtime environment. */
+    "target": "es2020",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
+    "lib": ["ES6", "DOM"],                               /* Specify a set of bundled library declaration files that describe the target runtime environment. */
```
d7415: Should we add `ES2020` here too? Or even replace `ES6`?

https://www.typescriptlang.org/tsconfig/#high-level-libraries

d7415 (approved):

vite.config.ts
ildyria: Good spot.

vite/index.html
ildyria: > Though I guess "It should keep working and anyone using it can tell us when it breaks" is valid for edge cases. Especially developer-focused edge cases.

Exactly my thought. And I don't think there won't be much to maintain unless we have to completely rewrite the front-end again :laughing:

tsconfig.json
ildyria: Fixed.

vite/index.html
d7415: > unless we have to completely rewrite the front-end again

The gap between complete rewrites has definitely been getting shorter... :thinking: :joy:

PR #2652, opened 2024-11-05
Improve support for free-bsd

ildyria:


vite.config.ts
```diff
@@ -73,33 +73,36 @@ const baseConfig =   {
...
-      }
-      
-    } else { // command === 'build'
-      if (config.plugins === undefined) {
-        throw new Error('plugins list is missing');
-      }
-
-      config.plugins.push(laravelPlugin);
-      return config;
     }
```
d7415, line 103: If the command is `serve` but `VITE_LOCAL_DEV` is not true we continue to the build block, adding the `laravelPlugin`. I'm not convinced that's a problem but it does look like a change. I guess it's a question of what combination of options we support.

d7415 (approved):
One comment, but it doesn't look especially troublesome.

vite.config.ts
ildyria: Actually it is a fix. :laughing:  I found out the hard way...

PR #2686, opened 2024-11-14
Add middleware redirection in vite

ildyria:
support for multiple `/` end-points so that F5 no longer fails when doing local dev with npm only (using `VITE_LOCAL_DEV` and `VITE_HTTP_PROXY_TARGET`)

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,108 +1,142 @@
-import { fileURLToPath, URL } from 'node:url';
-import { defineConfig, loadEnv, UserConfig } from 'vite';
-import laravel from 'laravel-vite-plugin';
-import vue from '@vitejs/plugin-vue';
-import i18n from 'laravel-vue-i18n/vite';
+import { fileURLToPath, URL } from "node:url";
+import { defineConfig, loadEnv, PluginOption, UserConfig } from "vite";
+import laravel from "laravel-vite-plugin";
+import vue from "@vitejs/plugin-vue";
+import i18n from "laravel-vue-i18n/vite";
 
 const laravelPlugin = laravel({
-    input: [
-      'resources/sass/app.scss',
-      'resources/js/app.ts',
-    ],
-    refresh: true,
-  })
+	input: ["resources/sass/app.scss", "resources/js/app.ts"],
+	refresh: true,
+});
 
-const baseConfig =   {
-  base: './',
-  plugins: [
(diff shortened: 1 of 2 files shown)


tinohager (approved):

vite.config.ts
```diff
@@ -1,108 +1,142 @@
...
-                if (id.includes('node_modules')) {
-                    return id.toString().split('node_modules/')[1].split('/')[0].toString();
-                }
-            }
-        }
-    }
+const localDevelopMiddleware: PluginOption = {
+  name: "develop-rewrite-middleware",
+  configureServer(server) {
+    const viteIndexPath = "/vite/index.html";
```
d7415, line 15: I guess if it's local we don't need subpath handling.

vite.config.ts
```diff
@@ -1,108 +1,142 @@
...
+      }
+
+      const requestUrl : string = req.url;
+      const doNotTouch = ["/@", "/api", "/resources", "/node_modules"];
+      if (doNotTouch.some((page) => requestUrl.startsWith(page))) {
+        next();
+        return;
+      }
+
+      console.log(req.url);
```
d7415, line 30: Do we want the debug code?

vite.config.ts
```diff
@@ -1,108 +1,142 @@
...
-    }
+		if (env.VITE_LOCAL_DEV === "true") {
+			if (env.VITE_HTTP_PROXY_TARGET === undefined) {
+				throw new Error("VITE_HTTP_PROXY_TARGET is missing");
+			}
+			config.server.open = "vite/index.html";
+			config.server.proxy = {
+				"/api/": env.VITE_HTTP_PROXY_TARGET,
+			};
+      config.plugins.push(localDevelopMiddleware);
```
d7415:
```suggestion
			config.plugins.push(localDevelopMiddleware);
```

d7415 (approved):

vite.config.ts
ildyria: yes.
in this specific case when you do `npm run dev` it is poping an server and redirecting you to `http://localhost:5173/vite/index.html` which we use to load the app.ts

vite.config.ts
ildyria: better safe on that one, it will only be visible if using local dev without the composer part.

PR #2689, opened 2024-11-14
Minor refactoring

ildyria:
Just type stuff.


d7415 (approved):
```php
function isNotEmpty(link: string | null | undefined): link is string {
```
confuses the hell out of me, but I'm going to assume it makes sense.

PR #2714, opened 2024-11-20
Added Authentik support

nanawel:
This PR adds support for [Authentik](https://goauthentik.io/) as OIDP.
It is heavily inspired from #2386 (Keycloak support).


resources/js/services/oauth-service.ts
```diff
@@ -12,6 +12,8 @@ const OauthService = {
 				return "fa-brands fa-apple";
 			case "amazon":
 				return "fa-brands fa-amazon";
+			case "authentik":
+				return "fa-brands fa-key";
```
ildyria, line 16: Is there any other font awesome icon that could fit too ?
Just wondering. :)

ildyria (approved):

resources/js/services/oauth-service.ts
ildyria: I guess not, helas.

nanawel: There's currently no "Authentik" icon in FA, but I thought about using the OpenID icon instead.
The thing is that it's not exactly equivalent because Authentik does not only support OAuth/OpenID but also SAML, SCIM, LDAP, etc. So a more generic "key" icon seems to be a good compromise.

PR #2717, opened 2024-11-20
Add copy to clipboard button in Diagnostics page.

ildyria:


resources/js/views/Diagnostics.vue
```diff
@@ -28,9 +28,35 @@ import SpaceDiagnostics from "@/components/diagnostics/SpaceDiagnostics.vue";
...
+
+const toast = useToast();
+const errorLoaded = ref<boolean>(false);
+const infoLoaded = ref<boolean>(false);
+const configurationLoaded = ref<boolean>(false);
+
+const canCopy = computed(() => errorLoaded.value && infoLoaded.value && configurationLoaded.value);
+
+function copy() {
+	const errors = document.getElementById("ErrorsData");
```
tinohager: Whoa, that's definitely not the way to do it! 💩 

it is possible to transmit the data to the outside via the emit
https://vuejs.org/guide/typescript/composition-api#typing-component-emits

ildyria: I know, but I didn't feel like it. :laughing: 
And that was simpler to "copy" the content that way.

tinohager: That's still not the way to do it. The other is more standardized 😄

resources/js/views/Diagnostics.vue
```diff
@@ -28,9 +28,35 @@ import SpaceDiagnostics from "@/components/diagnostics/SpaceDiagnostics.vue";
...
+	const info = document.getElementById("InfoData");
+	const Configuration = document.getElementById("ConfigurationData");
+	const errorLines = errors?.innerText.split("\n") ?? [];
+	let errorText = "";
+	for (let i = 0; i < errorLines.length; i++) {
+		errorText += `${errorLines[i].padEnd(7)}: ${errorLines[i + 1]}\n`;
+		i += 1;
```
d7415:
```suggestion
	for (let i = 0; i < errorLines.length; i+=2) {
		errorText += `${errorLines[i].padEnd(7)}: ${errorLines[i + 1]}\n`;
```
?

d7415 (approved):

PR #2768, opened 2024-11-29
:warning:  Add page that allows to fix broken trees

ildyria:
2024-11-29_3456x2160_17:52:19

This page will give you the complete ability to shoot yourself in the foot.


resources/js/views/FixTree.vue
```diff
@@ -0,0 +1,168 @@
...
+		<template #center>
+			{{ $t("maintenance.title") }}
+		</template>
+
+		<template #end> </template>
+	</Toolbar>
+	<div class="text-muted-color">
+		<div class="text-center mt-2 p-2">
+			This page allows you to re-order and fix your albums manually.<br />
+			Before any modifications, we strongly recommend you to read about Nested Set tree structures.<br />
```
d7415, line 17: Is it worth including a link here (e.g. https://en.wikipedia.org/wiki/Nested_set_model )? I can't decide whether it would be useful or if we'd be better off letting users do their own research.

resources/js/views/FixTree.vue
```diff
@@ -0,0 +1,168 @@
...
+function apply() {
+	if (albums.value === undefined) {
+		return;
+	}
+
+	if (!validate()) {
+		toast.add({
+			severity: "error",
+			summary: "Invalid tree!",
+			detail: "We are not applying this as this is guaranteed to be a broken state.",
```
d7415:
```suggestion
			detail: "We are not applying this as it is guaranteed to be a broken state.",
```

resources/js/composables/album/treeOperations.ts
```diff
@@ -0,0 +1,244 @@
...
+
+			const trimmedId = album.id.slice(0, 6);
+			const trimmedParentId = (album.parent_id ?? "root").slice(0, 6);
+			const isDuplicate_lft = hasDuplicateLft(album);
+			const isDuplicate_rgt = hasDuplicateRgt(album);
+
+			let isExpectedParentId = true;
+			// If current lft is greater than the last rgt,
+			// we are no longer a child of the last album. We pop out the pile until we are smaller than _rgt of the pile.
+			while (pile.length > 0 && (album._lft > pile[pile.length - 1].rgt || album._rgt > pile[pile.length - 1].rgt)) {
```
d7415, line 95: Am I right that `album._rgt > pile[pile.length - 1].rgt` *shouldn't* be true unless the `album._lft > pile[pile.length - 1].rgt` is also true? (as` _lft` should be < `_rgt`)

resources/js/composables/album/treeOperations.ts
```diff
@@ -0,0 +1,244 @@
...
+		return originalAlbums.value.filter((a) => a._lft === album._rgt || a._rgt === album._rgt).length > 1;
+	}
+
+	function check() {
+		originalAlbums.value = albums.value?.sort((a, b) => a._lft - b._lft);
+		prepareAlbums();
+		errors.value.forEach((e) => toast.add({ severity: "error", summary: "Error", detail: e, life: 3000 }));
+	}
+
+	// We increment all the nodes left and right by 1.
```
d7415:
```suggestion
	// We increment all the nodes' (>= lft) left and right by 1.
```

d7415 (approved):

resources/js/views/FixTree.vue
ildyria: I did consider it but ultimately this page should not be "found" hence why i wouldn't bother that much.

resources/js/composables/album/treeOperations.ts
ildyria: Yes, but it is more of a "safety" to make sure to unpile if either of the value not right.
The idea being that it avoids make the following parent_is are not all red if the unpiling did not match.

PR #2788, opened 2024-12-07
Fix hidden gallery footer social icons

pmrowla:
Fixes #2501 

The linked issue is reproducible for me in v6.1.2. The icons are visible in the landing page footer, but not in the gallery footer.

I think there's 2 issues with the gallery footer:
https://github.com/LycheeOrg/Lychee/blob/2e873f3a1aea07f41dd1b821f7d550da6679d3c0/resources/js/components/footers/GalleryFooter.vue#L8-L16

The first is that the div has `style="display: none"`. This hides the entire div even when `footer_show_social_media` is enabled. If I override the style to display the div in chrome devtools, I can see the div, but get the square boxes as shown in the original report.

The second issue is that the font with the icons is tied to the `socials` class, but the gallery footer links only set `socialicons`. (This is what causes the square boxes)

https://github.com/LycheeOrg/Lychee/blob/2e873f3a1aea07f41dd1b821f7d550da6679d3c0/resources/sass/fonts.css#L46-L48

You can see that the correctly working landing page footer sets both `socials` and `socialicons` classes:
https://github.com/LycheeOrg/Lychee/blob/2e873f3a1aea07f41dd1b821f7d550da6679d3c0/resources/js/components/footers/LandingFooter.vue#L3-L11

### DIFF
--- a/resources/js/components/footers/GalleryFooter.vue
+++ b/resources/js/components/footers/GalleryFooter.vue
@@ -5,11 +5,11 @@
 			The height of the footer is always the natural height
 			of its child elements
 			-->
-		<div id="home_socials" class="w-full" style="display: none" v-if="footerData.footer_show_social_media">
+		<div id="home_socials" class="w-full" v-if="footerData.footer_show_social_media">
 			<a
 				v-if="footerData.sm_facebook_url !== ''"
 				:href="footerData.sm_facebook_url"
-				class="socialicons"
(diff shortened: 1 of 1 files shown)


ildyria:
Thank you for your investigation and PR. I'll meege as soon as this passes the CI checks.

ildyria (approved):

PR #2849, opened 2024-12-27
Fixes upload fails for non-existent partner video upload to S3

kiancross:
Previously, the S3 upload flow attempted to upload a partner video regardless of its existence, resulting in an error when the video was missing. This error prevented the job responsible for uploading the original variant to S3 from completing successfully.

This commit adds a check to ensure the partner video exists before attempting to upload, resolving the issue and allowing the job to complete as expected.

Fixes one of the bugs described here: https://github.com/LycheeOrg/Lychee/discussions/2521.

### DIFF
--- a/app/Jobs/UploadSizeVariantToS3Job.php
+++ b/app/Jobs/UploadSizeVariantToS3Job.php
@@ -86,6 +86,10 @@ class UploadSizeVariantToS3Job implements ShouldQueue
 
 		$photo = Photo::query()->where('id', '=', $this->variant->photo_id)->first();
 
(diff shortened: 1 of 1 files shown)


kiancross:
I have no idea what the test error means.

ildyria:
Hi @kiancross  

thanks for the PR. I'll try to investigate hat is going wrong.
I do agree it is a weird one which makes no sense.

ildyria (approved):

ildyria:
_Tests pass on my machine..._

PR #2850, opened 2024-12-27
Add scroll to top when pressing the 'i' or clicking on details

ildyria:
QoL improvement when being in an album and wanting to edit.


ildyria:
@tinohager

d7415 (approved):

PR #2857, opened 2024-12-28
The eye does nothing if there are no hidden albums

ildyria:
… to avoid confusing users


ildyria:
On request of @tinohager

resources/js/components/gallery/AlbumHero.vue
```diff
@@ -59,6 +59,24 @@
 				>
 					<i class="pi pi-chart-scatter" />
 				</a>
+				<template v-if="isTouchDevice() && user?.id !== null">
+					<a
+						v-if="props.hasHidden && lycheeStore.are_nsfw_visible"
+						class="flex-shrink-0 px-3 cursor-pointer text-muted-color inline-block transform duration-300 hover:scale-150 hover:text-color"
+						:title="'show hidden'"
```
d7415: I _think_ this way is best, with the "toggle" effect showing what will happen, but we could change the tooltip here to "hide hidden".. :man_shrugging:

d7415 (approved):

PR #2858, opened 2024-12-28
Fix webauthn not showing up

ildyria:


ildyria:
Quite surprising no one noticed that one yet...

d7415 (approved):

PR #2870, opened 2025-01-02
Fix missing include in `hook-redirection.blade.php`

kiancross:
In commit b560ff5, the `layouts.app` base template was added to `hook-redirection.blade.php`. However, in commit b156ec4, the base template was removed, leaving a broken include reference.

This causes commands like `php artisan view:cache` to fail due to the missing template.

This commit removes the invalid include, restoring the file to its state prior to b560ff5 and resolving the issue.

### DIFF
--- a/resources/views/hook-redirection.blade.php
+++ b/resources/views/hook-redirection.blade.php
@@ -1,4 +1,3 @@
-<x-layouts.app>
 <div 
 	id="redirectData"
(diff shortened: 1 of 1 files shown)


ildyria:
Good spot, that's a left over from the Livewire part...

ildyria (approved):

PR #2879, opened 2025-01-04
Fixes "Videos not loading when going consecutively"

sancsin:
Fixes #2878 by monitoring the current photo type and if it is video triggers the video loading.

This fix will allow users to navigate through consecutive videos without needing to refresh page manually to load the current video. Hence, this fix should enhance user experience when navigating through an album.

### DIFF
--- a/resources/js/composables/photo/basePhoto.ts
+++ b/resources/js/composables/photo/basePhoto.ts
@@ -1,6 +1,6 @@
 import { computed, Ref, ref } from "vue";
 
-export function usePhotoBaseFunction(photoId: Ref<string>) {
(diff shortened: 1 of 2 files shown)


ildyria:
Hi @sancsin , Thank you for your fix, I updated to avoid the double `watch()`, could you please check if this is working as expected for you ? :)

Once you give me green light, I'll merge (I'm pretty sure tests will pass).

sancsin:
> Hi @sancsin , Thank you for your fix, I updated to avoid the double `watch()`, could you please check if this is working as expected for you ? :)
> 
> Once you give me green light, I'll merge (I'm pretty sure tests will pass).

Thanks for optimizing the code @ildyria. Looks good to me and works fine. Ok from my side to go ahead and merge.

ildyria (approved):

PR #2889, opened 2025-01-10
Backend part of duplicate finder

ildyria:


app/Actions/Photo/DuplicateFinder.php
```diff
@@ -0,0 +1,117 @@
...
+	 * Quickly count the number of duplicates candidates.
+	 *
+	 * @param bool $with_album_constraint    Requires the duplicates to be in the same album
+	 * @param bool $with_checksum_constraint Requires the duplicates to have the same checksum
+	 * @param bool $with_title_constraint    Requires the duplicates to have the same title
+	 *
+	 * @return int
+	 */
+	public function checkCount(
+		bool $with_album_constraint,
```
ppshobi: naming is hard but i'd call these `$mustBeWithinSameAlbum`, `$mustBeSameChecksum`, `$mustBeSameTitle`

ildyria: Fixed [e04205d](https://github.com/LycheeOrg/Lychee/pull/2889/commits/e04205d7307acfd5b6ee0a73107bf7d3b25ae497).

tinohager:
@ildyria To speed up the release cycle, maybe we can postpone this PR to 6.4 and make all the previous optimizations available again in one release?

ildyria:
That's also a possibility indeed. But I first need to fix the building of the nightly image. It is broken on ARM due to the Tailwindcss v4 migration. 🤬

This one: https://github.com/tailwindlabs/tailwindcss/issues/15806

app/Http/Controllers/Admin/Maintenance/DuplicateFinder.php
```diff
@@ -0,0 +1,59 @@
...
+{
+	/**
+	 * Get the number of duplicates.
+	 *
+	 * @return DuplicateCount
+	 */
+	public function check(MaintenanceRequest $request, PhotoDuplicateFinder $duplicateFinder): DuplicateCount
+	{
+		$pure_duplicates = $duplicateFinder->checkCount(must_be_within_same_album: false, must_have_same_checksum: true, must_have_same_title: false);
+		$title_duplicates = $duplicateFinder->checkCount(must_be_within_same_album: true, must_have_same_checksum: false, must_have_same_title: true);
```
sancsin, line 33: Couldn't title duplicates be present in other albums as well?

tests/Feature_v2/Maintenance/DuplicateFinderTest.php
```diff
@@ -0,0 +1,98 @@
...
+		]);
+		$this->assertOk($response);
+
+		$response = $this->getJsonWithData('Maintenance::searchDuplicates', [
+			'with_album_constraint' => true,
+			'with_checksum_constraint' => false,
+			'with_title_constraint' => true,
+		]);
+		$this->assertUnprocessable($response);
+	}
```
sancsin, line 101: Do we have the possibility in our test framework to create a dummy database to create duplicates and test the actual duplicate finding functionality of DuplicateFinder action class?

sancsin (approved):
Much awaited feature. Really nice that you could find time to implement this feature!

app/Http/Controllers/Admin/Maintenance/DuplicateFinder.php
ildyria: Yes, but the duplicate we are looking for here are checksum based.

The problem with considering _only_ title duplicates (so no checksum) is that if you shoot a lot and do not rename your pictures, you will have multiple _r5_1234.jpg for a lot of images.

tests/Feature_v2/Maintenance/DuplicateFinderTest.php
ildyria: There is always a dummy database, but we can generate temporary photos data.
We already do that here:
https://github.com/LycheeOrg/Lychee/blob/master/tests/Feature_v2/Base/BaseV2Test.php#L79

app/Http/Controllers/Admin/Maintenance/DuplicateFinder.php
ildyria: It is easier to limit this search to only title duplicates within the same album.
I just tested on my DB with 50 000 pictures, it takes wayyyy too long to load the data.

If you do not rename your pictures, in my case all my pictures are of the shape `_r5_1234.jpg` as a result I am bound to have duplicates accross a LOT of pictures.

PR #2891, opened 2025-01-10
Add license + copyright to all files

ildyria:
This will give us Gold badge :)


d7415 (approved):

d7415:
I wonder whether the end date on the copyright is needed. That's a big patch every year. But other than that sure.

ildyria:
Not specifically needed but it helps possible to see that the project is alive too. 🙂

PR #2892, opened 2025-01-10
Fix diagnostics not being complete on version 6

ildyria:
Some info were incomplete due to formatting...
Solved with some refactoring to later allow discarding messages.

### DIFF
--- a/app/Actions/Diagnostics/Pipes/Checks/ForeignKeyListInfo.php
+++ b/app/Actions/Diagnostics/Pipes/Checks/ForeignKeyListInfo.php
@@ -3,6 +3,8 @@
 namespace App\Actions\Diagnostics\Pipes\Checks;
 
 use App\Contracts\DiagnosticPipe;
+use App\DTO\DiagnosticData;
+use App\Enum\MessageType;
 use Illuminate\Support\Facades\DB;
 
 /**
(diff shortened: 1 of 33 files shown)


app/Actions/Diagnostics/Pipes/Checks/ForeignKeyListInfo.php
```diff
@@ -40,12 +42,16 @@ private function sqlite(array &$data): void
 		$fks = DB::select("SELECT m.name , p.* FROM sqlite_master m JOIN pragma_foreign_key_list(m.name) p ON m.name != p.\"table\" WHERE m.type = 'table' ORDER BY m.name;");
 
 		foreach ($fks as $fk) {
-			$data[] = sprintf('Foreign key: %-30s → %-20s : %s', $fk->name . '.' . $fk->from, $fk->table . '.' . $fk->to, strval($fk->on_update));
+			$data[] = DiagnosticData::from(
```
d7415: Why are these using `from` instead of `info`? Looks like the result would be the same?

d7415 (approved):

app/Actions/Diagnostics/Pipes/Checks/ForeignKeyListInfo.php
ildyria: Good spot !

PR #2899, opened 2025-01-12
Fix docker status always returning custom.

ildyria:
Bug detected on latest nightly build.
Tested the fix locally.


d7415 (approved):
:joy:

PR #2922, opened 2025-01-17
Add configuration integrity check

ildyria:
Just checking that we are not missing values.


s-payyeri (approved):
tests to make sure that the `ConfigIntegrity::SE_FIELDS` is same as the keys present in the database. Looks good

d7415 (approved):

PR #2929, opened 2025-01-18
Fix links from LycheeOrg.github.io to LycheeOrg.dev

ildyria:
Thanks @jphuguet for the spot!


tinohager (approved):

d7415 (approved):
We probably don't need translations for a URL, but we hopefully won't need to change it again!

PR #2934, opened 2025-01-19
Php 32bit - Zip-stream with version 2.1 instead of 3.1

### LINKED ISSUE #2933: 32 bit no longer supported
Hello,

Reading through the documentation, it mentions 
> While Lychee works on 32bit systems, we strongly recommend the use of a 64bit OS.
>> [Installation.md](https://github.com/LycheeOrg/LycheeOrg.github.io/blob/master/docs/installation.md) Line 45

However, while attempting to load this project onto my machine, it failed at the composer step since it now depends on zipstream-php version 3.x.x. As of their update to 3.x.x, they now made it a requirement that it runs on 64bit machines ([docs](https://github.com/maennchen/ZipStream-PHP/blob/main/README.md#general)), and so Lychee by extension has to obey that. 

I'd recommend removing that line and instead writing in it's place that this needs to be run on 64bit machines. 

Now time for me to either try downgrading Lychee or finding a different software. Thank you for your great project.

ildyria:
Fixes #2933

We leverage 2 traits one for each implementation of ZipStream.
The code is moved to a BaseArchive class and the installation decides which class to use.

### DIFF
--- a/app/Actions/Album/Archive32.php
+++ b/app/Actions/Album/Archive32.php
@@ -1,0 +1,16 @@
+<?php
+
+/**
+ * SPDX-License-Identifier: MIT
+ * Copyright (c) 2017-2018 Tobias Reich
+ * Copyright (c) 2018-2025 LycheeOrg.
+ */
+
+namespace App\Actions\Album;
+
+use App\Services\Archives\Zip21Trait;
+
+final class Archive32 extends BaseArchive
+{
+	use Zip21Trait;
+}
--- a/app/Actions/Album/Archive64.php
+++ b/app/Actions/Album/Archive64.php
@@ -1,0 +1,16 @@
+<?php
+
+/**
+ * SPDX-License-Identifier: MIT
+ * Copyright (c) 2017-2018 Tobias Reich
+ * Copyright (c) 2018-2025 LycheeOrg.
+ */
+
+namespace App\Actions\Album;
+
+use App\Services\Archives\Zip31Trait;
+
+final class Archive64 extends BaseArchive
+{
+	use Zip31Trait;
+}
--- a/app/Actions/Album/BaseArchive.php
+++ b/app/Actions/Album/BaseArchive.php
@@ -12,6 +12,9 @@ use App\Contracts\Models\AbstractAlbum;
 use App\Exceptions\ConfigurationKeyMissingException;
 use App\Exceptions\Handler;
 use App\Exceptions\Internal\FrameworkException;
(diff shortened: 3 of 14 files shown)


ildyria:
@RustyPotato could you try this branch ?

ildyria:
@RustyPotato you may need to remove the composer.lock before doing your install, but in theory it should now support your 32 bit install.

tinohager:
Couldn't the zip creation also be outsourced to the client?
https://github.com/gildas-lormeau/zip.js

the latest 2 version would be 2.4.0
https://github.com/maennchen/ZipStream-PHP/releases/tag/2.4.0

ildyria:
> Couldn't the zip creation also be outsourced to the client? [gildas-lormeau/zip.js](https://github.com/gildas-lormeau/zip.js?rgh-link-date=2025-01-20T06%3A57%3A34Z)

The problem is that the response is in streaming mode, In version 3 of Lychee, we were creating the zip file serverside before sending it. This has changed by the use of ZipStream, but I am not knowledgeable enough of how streaming files works with http, not how it would be possible for the client to know, that is one file, that is the next file.
The advantage of streaming the zip file directly is that the client no longer need to keep the page open and can just have it  run in a background tab.

ildyria:
@RustyPotato please confirm otherwise I won't be able to merge this. :)

RustyPotato:
My apologies for the long delay, this took longer than expected to get back to. I have gotten it to work, and it's running successfully. ([Link](https://lychee.kylefawcett.dev/gallery/D_IVgom0_XAvOr93NmVmrYCw)). Now just for me to set all the configs properly.

As we were messing with the zip library, had to double check what that'd look like. It gives off an empty zip file. So if I want that fixed, I'll probably need to do so myself eventually, but it's not an essential feature for me.

Having said that, it seems the demo site is also having difficulties with it too. ([Link](https://lychee-demo.fly.dev/gallery/eTNMmoiQzPbqZAbQMDxqKZKS))

ildyria:
> As we were messing with the zip library, had to double check what that'd look like. It gives off an empty zip file. So if I want that fixed, I'll probably need to do so myself eventually, but it's not an essential feature for me.

I checked on 64bit install it is working as expected. I will try to see with the 32bit version...

I can confirm the empty zip behaviour on the 32 bit.

ildyria:
@RustyPotato  this should fix the empty zip issue you were observing.

d7415 (approved):
Untested

PR #2938, opened 2025-01-20
Add modal for creating sharing

ildyria:
This replaces the sharing form directly in the page.
Yes, it adds a click but it makes it easier for the user to understand the process.

### DIFF
--- a/lang/cz/sharing.php
+++ b/lang/cz/sharing.php
@@ -19,6 +19,7 @@ return [
 	'username' => 'Username',
 	'no_data' => 'Sharing list is empty.',
 	'share' => 'Share',
+	'add_new_access_permission' => 'Add a new access permission',
 	'permission_deleted' => 'Permission deleted!',
 	'permission_created' => 'Permission created!',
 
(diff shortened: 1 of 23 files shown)


tinohager:
The user experience could be improved.

- Missing loading, the screen is white.
- I would not have the feeling that pressing the “Share” button opens a dialog.

image

ildyria:
> Missing loading, the screen is white.

What do you mean ?

tinohager:
image

ildyria:
> image

there is clearly something going wrong there.

tinohager:
> > image
> 
> there is clearly something going wrong there.

The correct dialog appears a second later but no loading process is displayed.

tinohager:
image

In my view, there would now also be enough space to make the dialog easier to understand.
The user selection can only allow one user. It is just an idea to add the possibility to authorize several users at the same time.
image

ildyria:
> image
> 
> In my view, there would now also be enough space to make the dialog easier to understand. The user selection can only allow one user. It is just an idea to add the possibility to authorize several users at the same time. image

That I will not do because I am planning to implement user groups and **permission propagation**.
Too much work for not much gain on that one.

> > > image
> > 
> > 
> > there is clearly something going wrong there.
> 
> The correct dialog appears a second later but no loading process is displayed.

Fixed with latest commit.

ildyria:
@LycheeOrg/reviewers  ready for reviews. :)

d7415 (approved):

PR #2950, opened 2025-01-23
Sharing propagation backend.

### LINKED ISSUE #292: [Enhancement] Permit sharing of subalbums at creation
Say I've shared album X with user A. Now, when I create a subalbum in X, let's call it Y. However, it will not be shared with A by default, it has to be manually shared with user A. This goes for all subalbums under X..

Now, it would be a massive help if every new subalbum under X would automatically be shared with everyone who can access X (Y inherits access rights of X, if you will).

Does this seem like a reasonable request?

ildyria:
Adds a single end-point (PUT) on Sharing to support propagation.
a boolean is required to define whether the overwritten or update strategy is chosen.

Contains #2938 

Fixes #292

### DIFF
--- a/app/Actions/Sharing/Propagate.php
+++ b/app/Actions/Sharing/Propagate.php
@@ -15,10 +15,11 @@ use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\App;
 use Illuminate\Support\Facades\DB;
 
-class Propagate
+final class Propagate
 {
 	/**
 	 * Update all descendants with the current album access permissions.
+	 * This is run in a DB transaction for safety.
 	 *
 	 * @param Album $album
 	 *
@@ -28,9 +29,24 @@ class Propagate
 	{
 		if (!App::runningUnitTests()) {
 			// @codeCoverageIgnoreStart
-			DB::beginTransaction();
+			DB::transaction(fn () => $this->applyUpdate($album));
+
+			return;
 			// @codeCoverageIgnoreEnd
 		}
+
+		$this->applyUpdate($album);
+	}
+
+	/**
+	 * Apply the current album access permissions to all descendants.
+	 *
+	 * @param Album $album
+	 *
+	 * @return void
+	 */
+	private function applyUpdate(Album $album): void
+	{
 		// for each descendant, create a new permission if it does not exist.
 		// or update the existing permission.
 		/** @var Collection<int,string> $descendants */
@@ -55,12 +71,6 @@ class Propagate
 				$perm->save();
 			});
 		});
-
-		if (!App::runningUnitTests()) {
-			// @codeCoverageIgnoreStart
-			DB::commit();
-			// @codeCoverageIgnoreEnd
-		}
 	}
 
 	/**
@@ -74,10 +84,24 @@ class Propagate
 	{
 		if (!App::runningUnitTests()) {
 			// @codeCoverageIgnoreStart
-			DB::beginTransaction();
+			DB::transaction(fn () => $this->applyOverwrite($album));
+
+			return;
 			// @codeCoverageIgnoreEnd
 		}
 
+		$this->applyOverwrite($album);
+	}
+
+	/**
+	 * Apply the overwrite of all descendants with the current album access permissions.
+	 *
+	 * @param Album $album
+	 *
+	 * @return void
+	 */
+	private function applyOverwrite(Album $album): void
+	{
 		// override permission for all descendants albums.
 		// Faster done by:
 		// 1. clearing all the permissions.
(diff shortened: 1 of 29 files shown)


tests/Feature_v2/Album/SharingTest.php
```diff
@@ -74,4 +77,127 @@ public function testUserGet(): void
...
+		$this->assertOk($response);
+		self::assertEquals(2, AccessPermission::where(APC::BASE_ALBUM_ID, '=', $this->album1->id)->count());
+
+		// Update sub album permission.
+		$response = $this->actingAs($this->userMayUpload1)->putJson('Sharing', [
+			'album_id' => $this->album1->id,
+			'shall_override' => true,
+		]);
+		$this->assertNoContent($response);
+		// self::assertEquals(2, AccessPermission::where(APC::BASE_ALBUM_ID, '=', $this->subAlbum1->id)->count());
```
JasonMillward: Commented code. Is this required or can it be cleaned up and removed?

ildyria: Good spot! 
Fixed [9d66a20](https://github.com/LycheeOrg/Lychee/pull/2950/commits/9d66a20640c328d1c336d9115ca1787703ea1f48)

That test is no longer necessary because of the stricter condition asserted later.

app/Actions/Sharing/Propagate.php
```diff
@@ -0,0 +1,130 @@
...
+				]);
+				$perm->save();
+			});
+		});
+
+		if (!App::runningUnitTests()) {
+			// @codeCoverageIgnoreStart
+			DB::commit();
+			// @codeCoverageIgnoreEnd
+		}
```
sancsin: Should we allow rollback in-case the transaction fails?

```suggestion
try {
        // for each descendant, create a new permission if it does not exist.
        // or update the existing permission.
        /** @var Collection<int,string> $descendants */
        $descendants = $album->descendants()->select('id')->pluck('id');
        $permissions = $album->access_permissions()->whereNotNull('user_id')->get();

        // This is super inefficient.
        // It would be better to do it in a single query...
        // But how?
        $descendants->each(function (string $descendant, int|string $idx) use ($permissions) {
            $permissions->each(function (AccessPermission $permission) use ($descendant) {
                $perm = AccessPermission::updateOrCreate([
                    APC::BASE_ALBUM_ID => $descendant,
                    APC::USER_ID => $permission->user_id,
                ], [
                    APC::GRANTS_FULL_PHOTO_ACCESS => $permission->grants_full_photo_access,
                    APC::GRANTS_DOWNLOAD => $permission->grants_download,
                    APC::GRANTS_UPLOAD => $permission->grants_upload,
                    APC::GRANTS_EDIT => $permission->grants_edit,
                    APC::GRANTS_DELETE => $permission->grants_delete,
                ]);
                $perm->save();
            });
        });

        if (!App::runningUnitTests()) {
            // @codeCoverageIgnoreStart
            DB::commit();
            // @codeCoverageIgnoreEnd
        }
    } catch (\Exception $e) {
        if (!App::runningUnitTests()) {
            // @codeCoverageIgnoreStart
            DB::rollBack();
            // @codeCoverageIgnoreEnd
        }
        // Add some kind of logging here or throw an exception for front-end to catch and inform the user?
```

app/Actions/Sharing/Propagate.php
```diff
@@ -0,0 +1,130 @@
...
+						APC::GRANTS_DOWNLOAD => $permission->grants_download,
+						APC::GRANTS_UPLOAD => $permission->grants_upload,
+						APC::GRANTS_EDIT => $permission->grants_edit,
+						APC::GRANTS_DELETE => $permission->grants_delete,
+					]
+				)->all()
+			)
+		);
+
+		DB::table(APC::ACCESS_PERMISSIONS)->insert($new_perm);
```
sancsin, line 146: Wrap this in try-catch block maybe and roll-back if something fails. Just like the proposal in `Propagate.php`?

JasonMillward (approved):

sancsin (approved):

app/Actions/Sharing/Propagate.php
ildyria: good idea!

sancsin:
Awesome feature! Thanks for creating this feature @ildyria.

app/Actions/Sharing/Propagate.php
ildyria: >         // Add some kind of logging here?

not needed, this is taken care of in the error handler, :)

app/Actions/Sharing/Propagate.php
ildyria: Fixed [6220a05](https://github.com/LycheeOrg/Lychee/pull/2950/commits/6220a059df2578bf80ca7e16d8f7cf463c2c2b1b).

ildyria: Good read: https://stackoverflow.com/a/40772361

PR #2952, opened 2025-01-24
Update csp config with base.

ildyria:
Copied from:

https://raw.githubusercontent.com/bepsvpt/secure-headers/refs/heads/main/config/secure-headers.php

### DIFF
--- a/config/secure-headers.php
+++ b/config/secure-headers.php
@@ -3,112 +3,181 @@
 use function Safe\parse_url;
 
 return [
-	/*
-	 * Server
+	/**
+	 * Server.
 	 *
-	 * Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server
+	 * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server
 	 *
 	 * Note: when server is empty string, it will not add to response header
 	 */
-
 	'server' => '',
 
-	/*
-	 * X-Content-Type-Options
+	/**
+	 * X-Content-Type-Options.
 	 *
-	 * Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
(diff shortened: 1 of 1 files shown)


config/secure-headers.php
```diff
@@ -3,319 +3,515 @@
 use function Safe\parse_url;
 
 return [
-	/*
-	 * Server
+	/**
+	 * Server.
```
d7415, line 7: I'm tempted to PR the source just to remove these full stops :facepalm:

config/secure-headers.php
```diff
@@ -3,319 +3,515 @@
...
+		// The name of reporting API, not the endpoint URL.
+		'report-to' => '',
+		'max-age' => 86400,
+		'include-subdomains' => false,
+		'success-fraction' => 0.0,
+		'failure-fraction' => 1.0,
+	],
 
+	/**
+	 * Expect-CT.
```
d7415, line 177: Probably better to keep with the copy, but this is deprecated.

config/secure-headers.php
```diff
@@ -3,319 +3,515 @@
...
+		'enable' => true,
 
-	'hpkp' => [
-		'hashes' => [
-			// 'sha256-hash-value',
+		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Permissions-Policy/accelerometer
+		'accelerometer' => [
+			'none' => false,
+			'*' => false,
+			'self' => true,
```
d7415, line 203: I won't add for all of these, but I don't think we use many of these, so should probably change to `none`.

d7415 (approved):
Not entirely convinced (see comments), but not bad :man_shrugging:

config/secure-headers.php
ildyria: the one i am wondering about is the auth key one.

config/secure-headers.php
ildyria: yeah :/

PR #2955, opened 2025-01-25
Add a feature flag to disable the content-type verification

ildyria:
The middleware was preventing requests to go through when testing from the `/docs/api` page.
This allows to do the requests.


RonnieTaz (commented):
This checks out well, I see no issue in either execution or documentation from the scramble document specification.

d7415 (approved):
*grumble*

PR #2960, opened 2025-01-26
Add license checker

ildyria:
We do not want to have any GPL thing in our distributed software.

Thanks to @dominikb for this nice piece of code. :)


.github/workflows/CICD_no_legacy.yml
```diff
@@ -131,6 +131,33 @@ jobs:
...
+        with:
+          php-version: 8.3
+          coverage: none
+
+      - name: Install Composer dependencies
+        uses: ramsey/composer-install@57532f8be5bda426838819c5ee9afb8af389d51a # 3.0.0
+
+      # We ban GPL-2.0-or-later, we don't want that viral thing in Lychee: We want to keep our MIT license.
+      - name: Run License Checker
+        run: vendor/bin/composer-licence-checker check -b GPL-2.0-or-later --no-dev
```
d7415: Is this right, or do you just want "GPL" as in the examples? That looks like a specific licence (different to 2 or 3)

d7415 (approved):

.github/workflows/CICD_no_legacy.yml
ildyria: Good spot, that is indeed a problem.
It seems we need to make a refuse list. I hoped that GPL-2.0-or-later was also taking all the LGPL, Allftredo etc but that does not seem to be the case. That is annoying.

PR #2962, opened 2025-01-27
Fix cors setting on Vite 6.0.9

ildyria:
@sancsin  Can you check if that fixes it for you ? 
You will probably need to define your `APP_URL` in your `.env`


sancsin (approved):
Seems to be working fine now. Thanks for this fix! :)

PR #2978, opened 2025-02-01
Fixes Displayed Times in Jobs page

RustyPotato:
Change from using substring to instead having JS parse the date.

Previously entries would've appeared as such:
firefox_ubKiz7fc2z
Take note of the last one, which is at 4:9, not how anyone would write 04:09 (or 4:09 AM even). 

This was due to us parsing the string into numbers, and forgetting to place leading 0s as appropriate. 

This has been fixed now, as seen in this preview:
firefox_m2sZoQKf95
It does so by letting the Javascript DateTime object do all the heavy lifting and work. 

Furthermore, this change has a "feature" addition to it as well, as now datetimes are shown in the local datetime of the user instead of GMT, which will help keep it more readable. Furthermore respecting the users' locale, so Europeans will get DD MM YYYY and Americans MM DD YYYY, etc.
chrome_BYZnTxUHas
(Note the changing datetime formats based off Locale, as well as changing hours when GMT is converted to "local" timezone).

### DIFF
--- a/resources/js/views/Jobs.vue
+++ b/resources/js/views/Jobs.vue
@@ -13,7 +13,7 @@
 	<Panel class="max-w-5xl mx-auto border-none">
 		<div v-if="jobs.length === 0" class="text-center">{{ $t("jobs.no_data") }}</div>
 		<div class="flex text-xs sm:text-base flex-wrap sm:flex-nowrap" v-for="job in jobs">
-			<span class="hidden sm:inline-block sm:w-1/4 text-muted-color">{{ prettyDate(job.created_at) }}</span>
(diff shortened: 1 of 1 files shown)


RustyPotato:
I've also found similar datetime manipulation done [here](https://github.com/LycheeOrg/Lychee/blob/a6d7c128a071706daa9feb04131988800f544ba1/resources/js/components/forms/profile/SetSecondFactorLine.vue#L40), but was unable to get my environment working such that I could test any changes there, so forewent attempting to change what I cannot (personally) test.

ildyria (approved):

PR #2981, opened 2025-02-01
Make sure that all v2 API routes are covered by the caching.

ildyria:


sancsin (approved):
That's a nice test to make sure that all GET routes are cached! 🙂

PR #2983, opened 2025-02-02
Add more statistics: punch card a la GitHub (SE).

ildyria:
Add more statistics: a punch card.
image

- Allows to filter on upload/taken date.
- Allows to look through the previous years.

The low, medium, high thresholds are configurable too (for the shading). Defaults are 10, 50, 100.

### DIFF
--- a/resources/js/components/statistics/PunchCard.vue
+++ b/resources/js/components/statistics/PunchCard.vue
@@ -37,11 +37,10 @@ const props = defineProps<{
 	low: number;
 	medium: number;
 	high: number;
-	data: DayCount[];
+	data: App.Http.Resources.Statistics.DayCount[];
 	year: number | undefined;
 }>();
 
-type DayCount = { date: string; count: number };
 type DayData = { date: Date; count: number };
 type MonthCol = { month: string; colspan: number };
 
@@ -51,7 +50,7 @@ const weeks = ref<DayData[][]>([]);
 const months = ref<MonthCol[]>([]);
 const days = ref<number[]>([0, 1, 2, 3, 4, 5, 6]);
 
-function transformData(data: DayCount[], year: number | undefined) {
+function transformData(data: App.Http.Resources.Statistics.DayCount[], year: number | undefined) {
 	weeks.value = [];
 	let startDate: Date;
 
--- a/resources/js/components/statistics/PunchCard.vue
+++ b/resources/js/components/statistics/PunchCard.vue
@@ -3,14 +3,19 @@
 		<thead>
 			<tr>
 				<td></td>
-				<td v-for="(m, i) in months" :key="`mo${i}${m.month}${m.colspan}`" :colspan="m.colspan" class="text-xs">{{ m.month }}</td>
+				<template v-for="(m, i) in months" :key="`mo${i}${m.month}${m.colspan}`">
+					<td :colspan="m.colspan" class="text-xs">{{ m.colspan > 1 ? m.month : "" }}</td>
+				</template>
 			</tr>
 		</thead>
 		<tbody>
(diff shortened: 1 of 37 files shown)


resources/js/components/statistics/PunchCard.vue
```diff
@@ -0,0 +1,138 @@
+<template>
+	<table class="border-separate border-spacing-0.5">
+		<thead>
+			<tr>
+				<td></td>
+				<td v-for="(m, i) in months" :key="`mo${i}${m.month}${m.colspan}`" :colspan="m.colspan" class="text-xs">{{ m.month }}</td>
+			</tr>
```
sancsin: Shall we have week numbers? Maybe it is easier to understand then what each box means? Like below example? I can add this code, if you agree to it.

image

resources/js/components/statistics/PunchCard.vue
```diff
@@ -0,0 +1,138 @@
...
+<script setup lang="ts">
+import { trans } from "laravel-vue-i18n";
+import { sprintf } from "sprintf-js";
+import { onMounted, ref } from "vue";
+
+const props = defineProps<{
+	low: number;
+	medium: number;
+	high: number;
+	data: DayCount[];
```
sancsin: Why don't we use `App.Http.Resources.Statistics.DayCount[]` type here?

sancsin (changes requested):
Looking really nice. It was just the Activity vue that was a bit difficult to grasp at first glance. I  had a suggestion for how it might look a little more intuitive in one of the comments.

resources/js/components/statistics/PunchCard.vue
ildyria: We could, but that would increase the "weight" of the table quite a bit.
Also if you mouse-over the tile you already have which date it corresponds to.

Or would increasing the day size be better ? (full instead of just 3 letters)

resources/js/components/statistics/PunchCard.vue
ildyria: good spot.

ildyria: Fixed : [335ae6d](https://github.com/LycheeOrg/Lychee/pull/2983/commits/335ae6d97a8fbff968eaad339840cbf5f6964a09)

resources/js/components/statistics/PunchCard.vue
ildyria: The other problem is that the week number do not start at 1 as we are listing the last 365 days.

sancsin: > We could, but that would increase the "weight" of the table quite a bit. Also if you mouse-over the tile you already have which date it corresponds to.
> 
> Or would increasing the day size be better ? (full instead of just 3 letters)

Day size is fine. It's only difficult to understand at first glance that each element in the table corresponds to a week. By showing the week number, for me atleast it became instantly more clear. Mouse-over, I was already aware of, that is quite nice and does make it more clear for the user. But still the user needs to take an action to get a better understanding. At first glance, the table still feels a bit confusing.

In the end it's a tradeoff. More clarity vs more weight to the table. Since you have done the development of this nice feature, I would like you to take the final decision. Or maybe go for voting within our team, if you don't want to take the decision yourself :)

sancsin (approved):
Great job with this new addition to the statistics page.

resources/js/components/statistics/PunchCard.vue
sancsin: > The other problem is that the week number do not start at 1 as we are listing the last 365 days.

I think that shouldn't matter that much. If the week numbers start at 1 or something else. As long the week numbers are showing correctly for the corresponding months.

ildyria: image
@sancsin  a middle ground ? :)

PR #2994, opened 2025-02-06
Fixes #2993 and #2992

sancsin:


ildyria (approved):
I see, I used to do CTRL + click for the first and shift click for the second. Hence the behaviour.

PR #3012, opened 2025-02-17
Add config option for the number of operations done

ildyria:
This should resolve the "I can only process 100 at the same time complaint".


d7415 (approved):
And change the default from 100/500 to 5000 :joy:

PR #3013, opened 2025-02-17
Full permission check is now disabled by default.

ildyria:
As it slows down quite a bit the diagnostic page.


d7415 (approved):
This looks fine, but I wonder if it would be better behind a button rather than a config so that it can be run as a one-off.

That would also be more effort though.

PR #3014, opened 2025-02-17
[SE] Add more privacy options

ildyria:
- allow to disable the right info panel in photo view + overlay.
- allow to hide the photo title for guest users.


database/migrations/2025_02_17_080713_more_privacy_options.php
```diff
@@ -0,0 +1,37 @@
...
+				'level' => 1,
+			],
+			[
+				'key' => 'file_name_hidden',
+				'value' => '0',
+				'cat' => 'Mod Privacy',
+				'type_range' => self::BOOL,
+				'description' => 'Do not show the photo title to anonymous users.',
+				'details' => 'Logged in user will still have access to the title.',
+				'is_secret' => true,
```
d7415: I don't see a problem with it, but why secret?

d7415 (approved):

database/migrations/2025_02_17_080713_more_privacy_options.php
ildyria: oversight, there is no need for that to be secret.

ildyria: fixed : [f3bb7da](https://github.com/LycheeOrg/Lychee/pull/3014/commits/f3bb7da1c06078c9756e10784cf48a299db7408f)

PR #3027, opened 2025-02-19
Add Diagnostics cache checks

ildyria:
- Add diagnostic cache checks
- Disable password albums when cache is enabled (do not force it but make it not possible to set a new password)
- Add cache into the easy settings.

### DIFF
--- a/resources/js/components/settings/EasySettings.vue
+++ b/resources/js/components/settings/EasySettings.vue
@@ -10,10 +10,17 @@
 				/>
 				<SelectLang v-if="lang !== undefined" :label="$t('settings.system.language')" :config="lang" />
 				<div class="flex flex-wrap justify-between">
-					<label for="pp_dialog_nsfw_visible">{{ $t("settings.system.nsfw_album_visibility") }}</label>
+					<label for="pp_dialog_nsfw_visible" class="text-muted-color-emphasis">{{ $t("settings.system.nsfw_album_visibility") }}</label>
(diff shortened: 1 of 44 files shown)


resources/js/components/settings/EasySettings.vue
```diff
@@ -10,10 +10,17 @@
...
+					<label for="pp_dialog_nsfw_visible" class="text-muted-color-emphasis">{{ $t("settings.system.nsfw_album_visibility") }}</label>
 					<ToggleSwitch id="pp_dialog_nsfw_visible" v-model="nsfwVisible" class="text-sm" @update:model-value="updateNSFW" />
 					<p class="my-1.5 text-muted-color w-full" v-html="$t('settings.system.nsfw_album_explanation')"></p>
 				</div>
+				<BoolField
+					v-if="cache_enabled !== undefined"
+					:label="$t('settings.system.cache_enabled')"
+					:config="cache_enabled"
+					:details="$t('settings.system.cache_enabled_details')"
+					@filled="save"
```
sancsin, line 22: We really want to show cache enable settings both in Basic and Advanced Settings?

sancsin (approved):

resources/js/components/settings/EasySettings.vue
ildyria: It is in Advanced by default (because we list all settings).

But I would say so for basic, it makes it more visible as people may think it is enabled by default in version 6.3.0 (which is not the case).

PR #3028, opened 2025-02-19
Offboarding Sonar - done with it breaking randomly...

ildyria:
Maybe it will come back later when I am in a better mood...


.github/workflows/php_no_legacy_tests.yml
```diff
@@ -26,7 +23,7 @@ jobs:
   tests:
     permissions:
       contents: read  # for actions/checkout to fetch code
-      pull-requests: read  # for SonarSource/sonarqube-scan-action to determine which PR to decorate
+      pull-requests: read
```
d7415, line 26: Do we still need this permission here (for codecov maybe?)?

.github/workflows/php_tests.yml
```diff
@@ -26,7 +23,7 @@ jobs:
   tests:
     permissions:
       contents: read  # for actions/checkout to fetch code
-      pull-requests: read  # for SonarSource/sonarcloud-github-action to determine which PR to decorate
+      pull-requests: read
```
d7415, line 26: Ditto

d7415 (approved):

.github/workflows/php_no_legacy_tests.yml
ildyria: yeah codecov is the reason I left it.

.github/workflows/php_tests.yml
ildyria: ditto. :)

PR #3062, opened 2025-03-07
Special characters cannot be used for the database password

### LINKED ISSUE #1248: at web install page,mysql password cannot include special charactor like !@#$%^&
at web install page,mysql password cannot include special charactor like !@#$%^&

ildyria:
Fixes #1248

Simply put: do not use special characters. If the password is long enough it should not matter.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,9 +17,6 @@ APP_FORCE_HTTPS=false
 # Do note that this disable CSP!!
 DEBUGBAR_ENABLED=false
 
-# enable or disable the v6 layout.
-VUEJS_ENABLED=true
-
 # enable or disable log viewer. By default it is enabled.
 LOG_VIEWER_ENABLED=true
 
@@ -55,9 +52,10 @@ LEGACY_V4_REDIRECT=false
 DB_OLD_LYCHEE_PREFIX=
 
 # DB_CONNECTION can be sqlite, mysql or pgsql. For sqlite the other entries are
-# not required, but an existing sqlite3 database may be specified if desired. In
-# this case, please use an absolute path. DB_DATABASE may be omitted but should
+# not required, but an existing sqlite3 database may be specified if desired.
+# In this case, please use an absolute path. DB_DATABASE may be omitted but should
 # *not* be left blank.
+# Note that DB_PASSWORD must not include special charactor like !@#$%^&
 DB_CONNECTION=sqlite
 DB_HOST=
 DB_PORT=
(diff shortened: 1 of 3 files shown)


.env.example
```diff
@@ -17,9 +17,6 @@ APP_FORCE_HTTPS=false
 # Do note that this disable CSP!!
 DEBUGBAR_ENABLED=false
 
-# enable or disable the v6 layout.
```
ildyria, line 20: hide this option, we do not want to advertise that anymore. Let's make the vue front-end strong default.

resources/views/vueapp.blade.php
```diff
@@ -9,6 +9,25 @@
     <x-meta />
     @vite(['resources/js/app.ts','resources/sass/app.css'])
 </head>
+{{-- Dirty work around but hey, backward compatibility... --}}
```
ildyria, line 12: Work around that was present in v5 but absent in v6.
Fixed.

d7415:
They could probably just quote or escape them.

ildyria:
I'll double check :)

ildyria:
@d7415  indeed. Tested locally.
Fixed: [1d1b5da](https://github.com/LycheeOrg/Lychee/pull/3062/commits/1d1b5daf93f151aa9ca0f2b7f6d92133e08ec164) & [3812f99](https://github.com/LycheeOrg/Lychee/pull/3062/commits/3812f9939649beb89f69e2d5556d7ea2d150c07c)

Note that this is slightly more tricky on the docker side, I'll make a PR for that there.

resources/views/install/env.blade.php
```diff
@@ -7,6 +7,8 @@
 @section('content')
 <ul class="list">
 	<li class="list__item list__item--env"><span><i class="fa fa-warning" aria-hidden="true"></i> <strong>Lychee does not create the database.</strong></span></li>
+	<li class="list__item list__item--env"><span><i class="fa fa-warning" aria-hidden="true"></i> Database password can include special characters like <code class="folder">!@#$%^&</code> but then the string must be enclodes in quotes.
+		For example: <code class="folder">DB_PASSWORD="lychee!@#$%^&"</code></span></li>
```
d7415, line 11: From the (limited) documentation I've seen, single quotes are a bit stronger. Though I suppose they're also more likely to be included in a password, so on balance, yeah, maybe double is better for the example.

d7415 (approved):

PR #3081, opened 2025-03-15
Refactoring Settings page.

ildyria:
Complete refactoring of the setting page.

- Settings are now properly ordered.
- One "tab" per Module.
- Difference between General, Modules and Advanced.
- Add an expert toggle to hide the more complex settings by default
- Add a all settings toggle to revert back to single page.

https://github.com/user-attachments/assets/c2bdcb6e-962e-45a7-9461-48741f8aa729


app/Http/Controllers/Admin/SettingsController.php
```diff
@@ -90,9 +99,12 @@ public function setJS(SetJSSettingRequest $request): void
 	{
 		$js = $request->getJs();
 		if (Storage::disk('dist')->put('custom.js', $js) === false) {
```
ppshobi: can we keep the disk names in a cosntant

app/Http/Resources/GalleryConfigs/SettingsConfig.php
```diff
@@ -0,0 +1,28 @@
...
+namespace App\Http\Resources\GalleryConfigs;
+
+use App\Models\Configs;
+use Spatie\LaravelData\Data;
+use Spatie\TypeScriptTransformer\Attributes\TypeScript;
+
+#[TypeScript()]
+class SettingsConfig extends Data
+{
+	public bool $default_old_settings;
```
ppshobi, line 18: I'm not sure what the parent class `Data` do but, why not private properties with getters & setters?

database/migrations/2025_03_18_173343_create_config_categories_table.php
```diff
@@ -0,0 +1,128 @@
...
+		$this->downExpert();
+		$this->downOrdering();
+		$this->downHideForDockerInstall();
+		$this->downConfigCategories();
+	}
+
+	private function upConfigCategories(): void
+	{
+		Schema::create('config_categories', function (Blueprint $table) {
+			$table->string('cat');
```
ppshobi: what is the `cat` column ? is it id? then either `id` or `category_id`?  also laravel has `$table->id()` or `$table->increments()` or even a uuid option

app/Http/Controllers/Admin/SettingsController.php
```diff
@@ -111,9 +123,22 @@ public function setCSS(SetCSSSettingRequest $request): void
...
 	}
-}
\ No newline at end of file
+
+	/**
+	 * Return the necessary information to configure the settings page.
+	 *
+	 * @return SettingsConfig
+	 */
+	public function getConfig(GetAllConfigsRequest $request): SettingsConfig
```
ppshobi, line 140: I think we don't need the `$request` parameter? beacuse we are returning new Settings config all the time?

app/Http/Resources/GalleryConfigs/SettingsConfig.php
ildyria: https://spatie.be/docs/laravel-data/v4/introduction the Data class transforms any public properties into JSON attributes at serialisation. What we could consider is making them read only (but didn't feel like experimenting :) )

app/Http/Controllers/Admin/SettingsController.php
ildyria: The `$request` parameter ensure that the authorization layer in GetAllConfigs request is applied.

database/migrations/2025_03_18_173343_create_config_categories_table.php
ildyria: **Before:**

Settings are contained in a single table:
- `id` - duh
- `key` - key of the config
- `value` - value of the config
- `cat` - String - category of the setting
- `type_range` - a string containing an array by `explode(..., "|")`  
- `is_secret` - whether the value is a secret and should be not be displayed in diagnostic page
- `description` - Description (or the pretty value of the key)
- `details` - Additional info with regard to the config.
- `level` - Level of support require to modify that setting 0 = global, 1 = supporter

**This PR adds**:

- `not_on_docker` - Defines that this setting is not used/displayed in docker installations
- `is_expert` - Defines that this setting is only visible in expert view
- `order` which allows us to re-order the display of settings

It also adds a `config_categories` table which allows us to re-order the settings based on the category (rather than to have to play with the order in a single table.
In that table

- `cat` directly references the `config.cat` (or the opposite actually, but that is the relation between those two tables).
- `name` is the friendlier version of `cat`
- `description` allows us to add a description on top of the Settings
- `order` allows us to re-order the categories. The following logic is done in front-end:
order < 10 = Basic, order < 100 = Modules, order >= 100 are Advanced.

I don't mind renaming the `cat` column into `id` or `category_id`. The thing is that for legacy reasons, while we still maintain the version 4 front-end we need to keep `cat` as a string. Once we drop the Legacy v4, then we have nothing holding us back to migrate to another id system for categories (the config table would need to be updated though).

ppshobi: thanks for the clarification. so AFAIK `cat` is a primary key string. and changing it to an int Id is not straightforward

ppshobi: maybe still makes sense to add an `id` column thats an integer, and we keep the cat column? at some point in the future it will be easier to centralize the categories?

ildyria: sure :)

app/Http/Controllers/Admin/SettingsController.php
ildyria: Fixed [b2ebf48](https://github.com/LycheeOrg/Lychee/pull/3081/commits/b2ebf4810a96d58314d687062707211f502cc489)

database/migrations/2025_03_18_173343_create_config_categories_table.php
ildyria: Fixed [b2ebf48](https://github.com/LycheeOrg/Lychee/pull/3081/commits/b2ebf4810a96d58314d687062707211f502cc489)

tinohager:
Here is my feedback as discussed

- I find the top 3 switches disturbing I don't know if this is planned to be removed in the future “Expert Mode, Old Settings style, ...”
- Some Setting Text is blue?
- After change something, the box "Some settings changed." is visible. Add a second button in the box to reset all my changes. Maybe show me all the changes.

| Group | Area | Setting | Remark | 
|----------|:-------------|:------|:------|
| Genral | - | Move Dropbox to Module section | 
| System |  Basics |  Use dark mode for Lychee. | Remove dot at the end "." | 
| System |  Basics |  Website Owner | Move to position 1  | 
| System |  Basics |  Website title | Move to position 2  | 
| System |  Basics |  Language used by Lychee | Move to position 3  | 
| System |  Basics |  Allowed extra formats, will not be processed | Move into other area, is not basics  | 
| System |  Basics |  Frequency of Lychee update checks | Move into other area, is not basics  | 
| System |  Landing page |  - |  Add some information what is the Langing page  | 
| System |  Footer |  - |  Move the full menu into a new Config Group "General Settings"  | 
| System |  Footer |  - |  Url of twitter profile, twitter is now X  | 
| General Settings |  Layout |  - |  Use dark mode and Position of the login button  | 
| Modules |  Symbolic Link |  - |  Is this required?  |

ildyria:
> I find the top 3 switches disturbing I don't know if this is planned to be removed in the future “Expert Mode, Old Settings style, ...”

I had two options: 
1. use toggle switches
2. use icons (a bit like the layout changes)

As there are no easy icons that are very explicit, toggle switches clarifies what each does.

About "old settings style", you don't realize it, but it is very useful for devs to quickly figure out what key the setting was corresponding to.  
About "expert" it is pretty obvious this one is going to stay.  
About "all settings" I am still unsure at the moment, note that it only appears when you toggle expert mode.  

> Some Setting Text is blue?

Those are Lychee SE settings.

> After change something, the box "Some settings changed." is visible. Add a second button in the box to reset all my changes. Maybe show me all the changes.

Reset is easy~ ish. Show all the change is quite some troubles.

> Move Dropbox to Module section	

It is only a single setting, it does not make sense to have it as a module.

> Frequency of Lychee update checks -> Move into other area, is not basics

Considered moving that into Advanced.

> Symbolic Link	 -> Is this required?

Sadly yes until I figure out how to use Laravel storage private link

ildyria:
> I find the top 3 switches disturbing I don't know if this is planned to be removed in the future “Expert Mode, Old Settings style, ...”

Fixed: [7299044](https://github.com/LycheeOrg/Lychee/pull/3081/commits/72990446f35f80ee6e17e41b730fc8e3a0d503bc)

> Reset is easy~ ish. Show all the change is quite some troubles.

_Left as an exercise to the reader._ Would rather see this functionality merged first as this is already quite a big chunk of code...

> Considered moving that into Advanced.

Fixed (and other settings moved): [7299044](https://github.com/LycheeOrg/Lychee/pull/3081/commits/72990446f35f80ee6e17e41b730fc8e3a0d503bc)

tinohager:
> Symbolic Link -> Is this required?
> Sadly yes until I figure out how to use Laravel storage private link

If it is needed, does it make sense if I can switch it off?

ildyria:
> If it is needed, does it make sense if I can switch it off?

It is needed as a protection mechanism to avoid people guessing the url of the original image from the thumb. It is an opt-in option that was requested in the past (since version 3 actually).

tinohager:
> > If it is needed, does it make sense if I can switch it off?
> 
> It is needed as a protection mechanism to avoid people guessing the url of the original image from the thumb. It is an opt-in option that was requested in the past (since version 3 actually).

I like it! Does it make sense to switch it off?

ildyria:
> I like it! Does it make sense to switch it off?

It is a speed trade off. Personally I don't use it.

resources/js/components/settings/AllSettings.vue
```diff
@@ -288,4 +97,8 @@ onUpdated(function () {
 	const adminMenuItem = spy.getCurrentMenuItem(admin);
 	spy.setActive(adminMenuItem, admin);
 });
+
+// watch(() => [props.configs, props.hash], ([newConfigs, newHash], _) => {
+// 	load(newConfigs as App.Http.Resources.Models.ConfigCategoryResource[]);
+// });
```
sancsin: Why do we want to keep this commented out code?

resources/js/components/settings/EasySettings.vue
sancsin, line 1: Wouldn't it be more intuituve with filename as "System.vue" since the settings category header is "System"?

sancsin (changes requested):
Looks like a great improvement for the settings panel.

Should we btw have a way to go back to the view once user clicks on "All settings"? Right now, the user needs to navigate back to the side-menu and click on "Settings" again to go back to the category view.

resources/js/components/settings/AllSettings.vue
ildyria: Fixed: [13d4bac](https://github.com/LycheeOrg/Lychee/pull/3081/commits/13d4bac2c8ce8e8070e353a2bad411ab8a395e82)

resources/js/components/settings/EasySettings.vue
ildyria: Fixed [13d4bac](https://github.com/LycheeOrg/Lychee/pull/3081/commits/13d4bac2c8ce8e8070e353a2bad411ab8a395e82)
I went for General as it is not just the System tab.

ildyria:
@sancsin  Updated :)

ildyria:
> Should we btw have a way to go back to the view once user clicks on "All settings"? Right now, the user needs to navigate back to the side-menu and click on "Settings" again to go back to the category view.

Fixed with the addition of a small double back arrow button (with tooltip) :)

resources/js/components/settings/ConfirmSave.vue
```diff
@@ -49,6 +59,7 @@
 <script setup lang="ts">
 import SettingsService from "@/services/settings-service";
 import { useLycheeStateStore } from "@/stores/LycheeState";
+import { trans } from "laravel-vue-i18n";
```
sancsin: Is this import used somwhere?

sancsin (changes requested):

resources/js/components/settings/ConfirmSave.vue
ildyria: nope: [b6eb73b](https://github.com/LycheeOrg/Lychee/pull/3081/commits/b6eb73b0f6d2b0e038a28e944f41a6a725e4763e)

sancsin (approved):

PR #3088, opened 2025-03-16
install rector & update phpstan -> set level to 1 and fix errors

ppshobi:
this looks like a huge PR, but for phpstan to run without errors, I had to remove a lot of dead `ignore-phpstan-errors`

+ a rector run which added `void` return types to some closures. Should not affect any features, but lets check

follow up PR's will have further increments of phpstan level's (so might be the case that some of these removed phpstan-ignore comments will come back).

Also see: https://github.com/LycheeOrg/phpstan-lychee/pull/24


phpstan.neon
```diff
@@ -1,18 +1,12 @@
 includes:
 	- vendor/larastan/larastan/extension.neon
 	- vendor/lychee-org/phpstan-lychee/phpstan.neon
-
 parameters:
+	level: 1
 	treatPhpDocTypesAsCertain: false
 	paths:
 		- app
-		- scripts
```
ppshobi, line 9: I removed all these folders from phpstan check, as they are not logic related files , also `tests` folder as the tests usually containts hacks that phpstan have hard time detecting

app/Relations/HasManySizeVariants.php
```diff
@@ -39,16 +39,7 @@ public function __construct(Photo $owningPhoto)
 	/**
 	 * Get the results of the relationship.
 	 *
-	 * @internal The parent class allows to return `mixed`, hence it is
```
ildyria, line 42: I would be happy if this comment was kept as it explains why this is happenning, because I do expect this error to come back with phpstan. If not we can always remove it later. :)

phpstan/stubs/image.stub
```diff
@@ -115,17 +115,6 @@ function imagegif(\GdImage $image, $to = null): void {}
...
 
-/**
- * @param \GdImage $image An image, returned by one of the image creation functions, such as imagecreatetruecolor.
- * @param float $angle Rotation angle, in degrees. The rotation angle is interpreted as the number of degrees to rotate the image anticlockwise.
- * @param int $bgd_color Specifies the color of the uncovered zone after the rotation
- * @param int $dummy This parameter is unused.
- * @return \GdImage Returns an image for the rotated image.
- * @throws ImageException
- */
-function imagerotate(\GdImage $image, float $angle, int $bgd_color, int $dummy = 0): \GdImage {}
```
ildyria, line 126: Instead of deleting that one, just comment it please, same resonnning as above.

phpstan.neon
ildyria: I disagree.

lang: phpstan checks that the arrays used in there do not contains duplicates keys.
database/migrations: phpstan ensures that the types of arrays when creating configs are properly respected. Also that some part of the logic which relates to the app are also still present.
tests: same as database/migration.
configs: same principle as lang.

phpstan/stubs/image.stub
ppshobi: this one was deleted, because its duplicate function definition, if you can expand to see the full file, then you will see it.

phpstan.neon
ppshobi: I am not so familiar with the codebase, so not sure what is included in there (those in config/lang etc...), but tbh except the `tests/` I agree if there is logic in there, it makes sense to cover them with phpstan (While ideally there should not be any logic in there) 

from my POV for `tests/` they should not directly contain logic related to the app. The tests are to make sure the app works fine, but if we are covering  tests folders too, then it means we are testing the tests 🤔. The tests are in a different namespace, whatever in the `Tests` namespace should never appear in the app. 

the test frameworks usually do a lot of magic to setup the test environment. this involves dynamic calls (e.g., mocks, reflection, or @dataProvider methods), which throws a lot of false positives and phpstan struggles to understand this. then we have to add a lot of these ignore errors in the test code as well as the phpstan config. 

While I disagree,  if you say it needs to be in there, I can put them back and we can open a discussion later.

ppshobi: I generated a baseline with already existing errors from database/migrations folder. so we are only ignoring exisiting errors, and in the new code we should not be making new mistakes

ildyria: Let's leave tests out for now, and see how it goes later. :)

phpstan.neon
```diff
@@ -1,138 +1,25 @@
...
 	excludePaths:
-		- database/migrations/2021_12_04_181200_refactor_models.php
 		- app/Services/Archives # This is creating errors depending if version 2.1 or 3.1 is installed...
 	stubFiles:
 		# these can be removed after https://github.com/thecodingmachine/safe/issues/283 has been merged
 		- phpstan/stubs/Wireable.stub
 		- phpstan/stubs/image.stub
 		- phpstan/stubs/imageexception.stub
 	ignoreErrors:
-		# False positive php8.2 + phpstan on interface @property
```
ildyria, line 23: I strongly suggest that you keep those but just comment them. They might come handy later when you increase the levels.
Also that is very likely going to be breaking the script/delete_legacy.sh

phpstan/stubs/image.stub
ildyria: Good spot, my mistake :+1:

ildyria (approved):
So far LGTM.

ildyria:
Good news, It passes the tests: https://github.com/LycheeOrg/Lychee/actions/runs/13886807201

phpstan.neon
ppshobi: I mean, we can always come back, git should help us then

ildyria: Sure.

PR #3090, opened 2025-03-17
fix: rector default run

ppshobi:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ test:
 		echo ""; \
 	fi
 
-formatting:
+formatting: rector
 	@rm .php_cs.cache 2> /dev/null || true
 	@if [ -x "vendor/bin/php-cs-fixer" ]; then \
 		PHP_CS_FIXER_IGNORE_ENV=1 ./vendor/bin/php-cs-fixer fix -v --config=.php-cs-fixer.php; \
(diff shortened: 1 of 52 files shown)


ildyria:
Are the void in the arrow function required by PhpStan ? Because those feel a bit superfluous to me.

ildyria:
Having a look at Rector rules, those are the ones that I would disable.
- AddArrowFunctionReturnTypeRector
- AddClosureNeverReturnTypeRector
- AddClosureVoidReturnTypeWhereNoReturnRector
- AddClosureNeverReturnTypeRector

Reasoning? We mostly use closures in `->where()` queries, which adds unnecessary noise.

ppshobi:
the higher we go with phpstan levels it needs us to add return types to all/any functions, so I think its already nice that rector doing it for us. also its a double clarity that the closure does not return anything.

without right return types phpstan can't function properly

for example `function a() {}` vs `function a(): void {}` tells phpstan 2 different things, one is that it *might* return soemthing vs it returns nothing,

if at some point a closure changes its return values type, (lets say from `void` -> `null`) phpstan won't be able to detect bug if its not typehinted  properly.

ildyria:
But we got to phpstan 6 without complaints on those return types. :)
Though I see what you mean. But then it would be nice if that rector step is added to the makefile to the `formatting` step :)

s-payyeri:
> But then it would be nice if that rector step is added to the makefile to the formatting step :)

totally doable

ppshobi:
fixed

ildyria (approved):
LGTM.

ildyria:
Waiting for CICD codecov report (not that I am worried).

PR #3091, opened 2025-03-17
feat: basic docker-compose setup

ppshobi:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,3 +155,11 @@ gen_typescript_types:
 
 class-leak:
 	vendor/bin/class-leak check app database/migrations config --skip-type Illuminate\\View\\Component
+
+docker-build-base-image:
+	docker build -t lychee-base:latest ./docker/base
+
+docker-build-dev-image:
+	docker build ./docker/dev -t lychee-dev:latest
+
+docker-build-dev: docker-build-base-image docker-build-dev-image
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -3,7 +3,9 @@ FROM lychee-base:latest AS base
 # Install development dependencies
 
 RUN \
-	apt-get install -y git composer nodejs
+    curl -fsSL https://deb.nodesource.com/setup_23.x -o nodesource_setup.sh \
+    && bash nodesource_setup.sh \
+    && apt-get install -y git composer nodejs
 
 RUN \
     rm -rf /var/lib/apt/lists/* \
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -1,0 +1,10 @@
+FROM lychee-base:latest AS base
+
+# Install development dependencies
+
+RUN \
+	apt-get install -y git composer nodejs
+
+RUN \
+    rm -rf /var/lib/apt/lists/* \
+	&& apt-get clean -qy
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,77 @@
+
+services:
+    lychee_cache:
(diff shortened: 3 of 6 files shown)


Makefile
```diff
@@ -154,4 +154,12 @@ gen_typescript_types:
 	php artisan typescript:transform
 
 class-leak:
-	vendor/bin/class-leak check app database/migrations config --skip-type Illuminate\\View\\Component
\ No newline at end of file
+	vendor/bin/class-leak check app database/migrations config --skip-type Illuminate\\View\\Component
+
+docker-build-base-image:
+	docker build -t lychee-base:latest ./docker/base
```
ppshobi, line 160: during each merge to master, a docker image has to be pushed to docker hub. but TBD later. 


right now if I do `make docker-build-dev` and then a `docker-compose up` + changes in the .env file with database value. then a debug version of the Lychee should start up

docker/dev/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM lychee-base:latest AS base
+
+# Install development dependencies
+
+RUN \
+	apt-get install -y git composer nodejs
+
+RUN \
```
ildyria, line 10: Is that necessary ? I am not even sure that reduces the size of the image due to layering.

docker/dev/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM lychee-base:latest AS base
+
+# Install development dependencies
+
+RUN \
+	apt-get install -y git composer nodejs
```
ildyria: Unless I am wrong, you may end up with an old version of nodejs if you use bookworm as base.
https://github.com/nodesource/distributions#debian-and-ubuntu-based-distributions < what I use locally to get an updated version.

https://github.com/LycheeOrg/Lychee/blob/dockerfile/docker-context/Dockerfile see some work I had done a few weeks ago in that direction (line 25)

docker/dev/Dockerfile
ppshobi: there was no intention of reducing the image size, just so that only dev image contains composer, git, nodejs etc..

docker/dev/Dockerfile
ppshobi: great, I took it from the Lychee-Docker repo. but updated with your suggestion, using the node23 from nodesource

d7415:
This feels a bit messy - are we going to maintain this in both places now?

docker/dev/Dockerfile
ildyria: The way it works on the Lychee-Docker repo is that we use a node image inside the docker file to compile the necessary files before pulling them to the base image, that way we don't have to deal with the nodejs layer.

ppshobi: I understand what you mean, but this image is only for development, lets say you want to run some nodejs commands during the development inside the container. then we need the nodejs cli. For building production images ofc we dont need them, thats Why I split docker files into two. (base & dev)

ildyria: I know, I was just explaining why we didn't need the node source. while for the dev image, it is indeed necessary. :)

ppshobi:
@d7415 

I did not wanted to make changes to the Docker repo because lots of hardcoded stuff are in there and did not wanted to break them for users. 

I only wanted it to be a developement image, At some point we can combine them (maybe the base image), I had a hard time dealing with all the symlinks in the Docker repo.

If you have a solution for a dev image also happy to use it.

ildyria (approved):

ildyria:
LGTM. No strong opinion, that helps others to join on Lychee dev, I am all for it :)

PR #3098, opened 2025-03-17
feat: custom variable casing rector to convert snake_case vars to  camelCase

ppshobi:


ppshobi:
see https://github.com/LycheeOrg/Lychee/pull/3100

PR #3130, opened 2025-03-22
snakecasify app/DTO

ildyria:


tests/LoadedSubscriber.php
```diff
@@ -25,7 +25,15 @@ public function notify(Loaded $event): void
 		$this->createApplication();
 		$this->migrateApplication();
 
-		if (config('features.vuejs') === true) {
+		// Detect local run base on the phpunit.xml file
+		if (str_contains($event->testSuite()->name(), 'phpunit.xml') &&
+			str_contains($event->testSuite()->tests()->asArray()[0]->file(), 'Feature_v1') &&
+			config('features.vuejs')) {
+			dd('Wrong configuration, tests are not compatible with VueJS routes');
```
ppshobi:
```suggestion
			throw new RuntimeException ('Wrong configuration, tests are not compatible with VueJS routes');
```

ildyria:
@ppshobi After further test, we really need the `exit(1)`. I tried with the others solution (like `throw` and they were not behaving the way I want them to: still start the tests... )

ppshobi:
I'd rather do an `die('some message')`, becuase the `dd` is a developement utility that prints out lots of unneccessary stuff, maybe even a security loophole

ppshobi (approved):

PR #3134, opened 2025-03-22
🚧  phpstan: phpstan level 2

ppshobi:


app/Models/Builders/AccessPermissionBuilder.php
```diff
@@ -9,11 +9,14 @@
 namespace App\Models\Builders;
 
 use App\Eloquent\FixedQueryBuilder;
+use App\Models\AccessPermission;
 
 /**
  * Specialized query builder for {@link \App\Models\AccessPermission}.
  *
- * @extends FixedQueryBuilder<\App\Models\AccessPermission>
+ * @template TModelClass of AccessPermission
```
ildyria, line 17: really ? damn phpstan 2 is annoying :(

app/SmartAlbums/BaseSmartAlbum.php
```diff
@@ -73,6 +74,8 @@ protected function __construct(SmartAlbumType $id, \Closure $smart_condition)
 			throw new FrameworkException('Laravel\'s service container', $e);
 		}
 		// @codeCoverageIgnoreEnd
+
+		parent::__construct();
```
ildyria, line 78: I am not sure about that one.

app/SmartAlbums/BaseSmartAlbum.php
```diff
@@ -39,7 +40,7 @@
  * Photos belong to these albums due to certain properties like being
  * starred, being recently added, etc.
  */
-abstract class BaseSmartAlbum implements AbstractAlbum
+abstract class BaseSmartAlbum extends Model implements AbstractAlbum
```
ildyria, line 43: The problem with that one is that it implies that there is a table property... etc.

app/SmartAlbums/Utils/MimicModel.php
```diff
@@ -25,11 +25,9 @@ abstract public function toArray(): array;
 	 *
 	 * @param string $key
 	 *
-	 * @return mixed
-	 *
 	 * @throws InternalLycheeException
 	 */
-	public function __get(string $key)
+	public function __get(mixed $key)
```
ildyria, line 30: Then you need to throw a `LycheeInvalidArgumentException()` if `$key` is not of type string.

app/SmartAlbums/BaseSmartAlbum.php
ppshobi: I note this one, was trying to make phpstan happy, I will come back to this

ildyria: https://github.com/LycheeOrg/Lychee/blob/master/app/Models/BaseAlbumImpl.php#L33 here is some reading :)

app/Models/Builders/AccessPermissionBuilder.php
ppshobi: I'm not even sure what does it mean, but just grunting though the errors for now

ppshobi:
see https://github.com/LycheeOrg/Lychee/pull/3134

PR #3145, opened 2025-03-23
phpstan: level 2

ppshobi:


app/SmartAlbums/BaseSmartAlbum.php
```diff
@@ -39,7 +40,7 @@
  * Photos belong to these albums due to certain properties like being
  * starred, being recently added, etc.
  */
-abstract class BaseSmartAlbum implements AbstractAlbum
+abstract class BaseSmartAlbum extends Model implements AbstractAlbum
```
ppshobi: but isn't this kind of expected? (I mean `extends Model`) 

because the extra logic in  the constructor of (`BaseSmartAlbum`) is doing some database stuff and it kinda expect this class to be the Model no?

also the `app/Models/Extensions/BaseAlbum.php` already extends `Model`, 

the structure is  supposed to be 

```
             AbstractAlbum
                 |
|---------------|------------------|
BaseAlbum                       BaseSmartAlbum 
```

no? or am I doing something else?

ppshobi: refering to the other discussion https://github.com/LycheeOrg/Lychee/pull/3134/files#r2009076078

app/SmartAlbums/Utils/MimicModel.php
```diff
@@ -29,6 +29,9 @@ abstract public function toArray();
 	 */
 	public function __get(mixed $key)
 	{
+		if (is_string($key)) {
```
ildyria: Don't you want... the negative? :laughing:

ppshobi: so fast 😆 , just fixed it in #2707fde

app/Actions/Album/Delete.php
```diff
@@ -92,7 +92,6 @@ public function do(array $album_ids): FileDeleter
 			// find all photos in those and their descendants
 			// Only load necessary attributes for tree; in particular avoid
 			// loading expensive `min_taken_at` and `max_taken_at`.
-			/** @var Collection<int,Album> $albums */
```
ildyria, line 95: might be necessary later, we shall see. :)

app/Actions/Album/Merge.php
```diff
@@ -21,9 +21,6 @@ class Merge
 	 * Merges the content of the given source albums (photos and sub-albums)
 	 * into the target.
 	 *
-	 * @param Album                 $targetAlbum
-	 * @param Collection<int,Album> $albums
```
ildyria, line 25: I am pretty sure that one will be needed later.

app/Actions/Album/Move.php
```diff
@@ -18,9 +18,6 @@ class Move
 	/**
 	 * Moves the given albums into the target.
 	 *
-	 * @param Album|null            $targetAlbum
-	 * @param Collection<int,Album> $albums
```
ildyria, line 22: Same.

app/Actions/Album/SetSmartProtectionPolicy.php
```diff
@@ -17,7 +17,7 @@ class SetSmartProtectionPolicy
 {
 	/**
 	 * @param BaseSmartAlbum $album
-	 * @param bool           $isPublic
```
ildyria, line 20: I was expecting those to go :)

app/Actions/Photo/Pipes/Standalone/CreateOriginalSizeVariant.php
```diff
@@ -12,11 +12,14 @@
...
 use App\DTO\PhotoCreate\StandaloneDTO;
 use App\Enum\SizeVariantType;
+use App\Image\StreamStat;
 
 class CreateOriginalSizeVariant implements StandalonePipe
 {
 	public function handle(StandaloneDTO $state, \Closure $next): StandaloneDTO
 	{
+		/** @var StreamStat $stat */
+		$stat = $state->stream_stat;
```
ildyria, line 22: I guess because it is a "pointer" to the object, the stat object is automatically updated, but you can read it on line 35.

app/Actions/Photo/DuplicateFinder.php
```diff
@@ -52,7 +52,6 @@ public function search(
 		bool $must_have_same_checksum,
 		bool $must_have_same_title,
 	): Collection {
-		/** @var Collection<int,object{album_id:string,album_title:string,photo_id:string,photo_title:string,checksum:string,short_path:string|null,storage_disk:string|null}> */
```
ildyria, line 55: is this causing problems ? :o

app/Actions/Sharing/ListShare.php
```diff
@@ -38,7 +38,7 @@ public function do(?User $participant, ?User $owner, ?BaseAlbum $base_album): Sh
 		try {
 			// Active shares, optionally filtered by album ID, participant ID
 			// and or owner ID
-			/** @var Collection<int,object{id:int,user_id:int,album_id:string,username:string,title:string}> $shared */
+			/** @var Collection<int, AccessPermission> $shared */
```
ildyria: No, you do not get the full access permission properties, that is incorrect. See the Select query.

app/Actions/Album/SetSmartProtectionPolicy.php
ppshobi: we can run it through cs fixer to remove those no value phpdoc stuff

app/Actions/Sharing/ListShare.php
```diff
@@ -107,7 +104,7 @@ private function breadcrumbPath(object $album): string
 	{
 		$title = [$album->title];
 		$parent = $album->parent;
-		while ($parent) {
+		while (!is_null($parent)) {
```
ildyria:
```suggestion
		while ($parent !== null) {
```

app/Casts/DateTimeWithTimezoneCast.php
```diff
@@ -34,7 +36,7 @@ class DateTimeWithTimezoneCast implements CastsAttributes
 	 * $key . '_orig_tz' and which stores the original timezone of the
 	 * (key, value)-pair at hand.
 	 *
-	 * @param Model               $model      the associated model class
+	 * @param Photo               $model      the associated model class // it should be only Model instance, but phpstan does not see the trait UTCBasedTimes applied to it
```
ildyria: No the solution for that is to create an interface with the methods defined in the trait and use an intersection type Model&ThatInterface.

app/Contracts/Models/AbstractAlbum.php
```diff
@@ -34,6 +34,8 @@
  * @property Collection<int,Photo>            $photos
  * @property Thumb|null                       $thumb
  * @property Collection<int,AccessPermission> $access_permissions
+ *
+ * @phpstan-require-extends Model
```
ildyria: That should not be needed.

app/Http/Controllers/Admin/UserManagementController.php
```diff
@@ -32,17 +32,13 @@ class UserManagementController extends Controller
...
-	 * @param Spaces                    $spaces
-	 *
 	 * @return Collection<array-key, UserManagementResource>
 	 */
 	public function list(ManagmentListUsersRequest $request, Spaces $spaces): Collection
 	{
 		/** @var Collection<int,User> $users */
 		$users = User::select(['id', 'username', 'may_administrate', 'may_upload', 'may_edit_own_settings', 'quota_kb', 'description', 'note'])->orderBy('id', 'asc')->get();
 		$spaces_per_user = $spaces->getFullSpacePerUser();
-		/** @var Collection<int,array{0:User,1:array{id:int,username:string,size:int}}> $zipped */
```
ildyria, line 45: I would keep that one if possible, maybe remove the @var ? 
```suggestion
		/** Collection<int,array{0:User,1:array{id:int,username:string,size:int}}> $zipped */
```

app/Http/Controllers/StatisticsController.php
```diff
@@ -81,7 +72,6 @@ public function getSpacePerAlbum(SpacePerAlbumRequest $request, Spaces $spaces):
 			album_id: $album_id,
 			owner_id: $owner_id);
 
-		/** @var Collection<int,array{0:array{id:string,left:int,right:int,size:int},1:array{id:string,username:string,title:string,is_nsfw:bool,left:int,right:int,num_photos:int,num_descendants:int}}> $zipped */
```
ildyria, line 84: I would keep that one...

app/Http/Controllers/StatisticsController.php
```diff
@@ -108,7 +95,6 @@ public function getTotalSpacePerAlbum(SpacePerAlbumRequest $request, Spaces $spa
...
```
ildyria, line 111: same.

app/Jobs/ProcessImageJob.php
```diff
@@ -72,6 +71,7 @@ public function __construct(
 			$album = $abstract_album;
 		}
 
+		/** @var Album|TagAlbum $album */
```
ildyria: That is incorrect, you can also upload in some of the smart albums.
Notably `unsorted` and `starred` 

This is also directly invalidated by line 70.

app/Legacy/V1/Requests/Photo/DuplicatePhotosRequest.php
```diff
@@ -38,7 +38,6 @@ public function rules(): array
 	 */
 	protected function processValidatedValues(array $values, array $files): void
 	{
-		/** @var array<int,string> $photo_ids */
```
ildyria, line 41: I am pretty sure you want to keep that one.

app/Legacy/V1/Requests/Photo/MovePhotosRequest.php
```diff
@@ -38,12 +38,11 @@ public function rules(): array
 	 */
 	protected function processValidatedValues(array $values, array $files): void
 	{
-		/** @var array<int,string> $photo_ids */
```
ildyria, line 41: idem

app/Models/Extensions/SizeVariants.php
```diff
@@ -196,17 +187,11 @@ public function getPlaceholder(): ?SizeVariant
 	 * Creates a new instance of {@link \App\Models\SizeVariant} for the
 	 * associated photo and persists it to DB.
 	 *
-	 * @param SizeVariantType $sizeVariantType the type of the desired size variant;
-	 * @param string          $shortPath       the short path of the media file this
```
ildyria, line 200: can we keep more of those description?

app/Models/Extensions/Thumb.php
```diff
@@ -44,11 +45,11 @@ protected function __construct(string $id, string $type, string $thumb_url, ?str
 	/**
 	 * Restricts the given relation for size variants such that only the
 	 * necessary variants for a thumbnail are selected.
+	 * /**
```
ildyria:
```suggestion
	 *
```

app/Models/Extensions/Thumb.php
```diff
@@ -44,11 +45,11 @@ protected function __construct(string $id, string $type, string $thumb_url, ?str
...
+	 * @param HasMany<Photo, Model> $relation
 	 *
-	 * @param HasMany<Photo,$this> $relation
-	 *
-	 * @return HasMany<Photo,$this>
-	 */
+	 * @return HasMany<Photo, Model>
+	 * */
```
ildyria:
```suggestion
	 */
```

app/Models/Extensions/Thumb.php
```diff
@@ -44,11 +45,11 @@ protected function __construct(string $id, string $type, string $thumb_url, ?str
 	/**
 	 * Restricts the given relation for size variants such that only the
 	 * necessary variants for a thumbnail are selected.
+	 * /**
+	 * @param HasMany<Photo, Model> $relation
```
ildyria: I don't think this is correct.

app/Models/Photo.php
```diff
@@ -30,6 +30,7 @@
 use App\Models\Extensions\ToArrayThrowsNotImplemented;
 use App\Models\Extensions\UTCBasedTimes;
 use App\Relations\HasManySizeVariants;
+use Barryvdh\LaravelIdeHelper\Eloquent;
```
ildyria: I am pretty sure that is not the one you want.

app/Policies/AlbumPolicy.php
```diff
@@ -268,7 +268,6 @@ public function canDelete(User $user, ?AbstractAlbum $abstract_album = null): bo
 			return true;
 		}
 
-		/** @var Album $abstractAlbum */
```
ildyria, line 271: You want to keep that one. Otherwise phpstan is going to complain about parent_id later.

ildyria:
```suggestion
		/** @var Album $abstract_album */
```

app/SmartAlbums/Utils/MimicModel.php
```diff
@@ -14,7 +14,7 @@
 
 trait MimicModel
 {
-	abstract public function toArray(): array;
```
ildyria, line 17: Why is the return type dropped ?

app/View/Components/Meta.php
```diff
@@ -38,7 +38,7 @@ class Meta extends Component
 	public string $user_js_url;
 
 	private bool $access = true;
-	private ?AbstractAlbum $album = null;
+	private ?Album $album = null;
```
ildyria: Are you sure about that change?

app/Actions/Photo/Pipes/Standalone/CreateOriginalSizeVariant.php
ppshobi: The problem is that the `$state->stream_stat` is typehinted with the Interface `StreamStats`, which of course not wokring with the @property phpdoc. hence I typehint it against the actual implementation

app/Actions/Photo/DuplicateFinder.php
ppshobi: yes, the error says 

> PHPDoc tag @var with type Illuminate\Support\Collection<int, object{album_id: string, album_title: string, photo_id: string, photo_title: string, checksum: string, short_path:  
string|null, storage_disk: string|null}> is not subtype of type Illuminate\Support\Collection<int, stdClass>. 

because the custom select in the `query()` builder makes the collection to be returned as `Collection<int, stdClass>` of course phpstan donot see the shape of that. now I have 2 options 

1. to typehint `Collection<int, stdClass>`
      1.2 now we don't have to explicitly tell phpstan that because its automatically inferered
2. we can have a DTO and create that object right after the `->get()` and typehint as `Collection<int, PhotoDiskDTO>` or something like this. now that is a whole lot of refcatoring
I'd rather make phpstan happy for now and take another hit at it with higher levels of phpstan

ildyria: That is the stdClass again... Great...

app/Casts/DateTimeWithTimezoneCast.php
ppshobi: fixed fdbe63e131988ffa987bcae568646e754e008da3

app/Http/Controllers/StatisticsController.php
ppshobi: fixed 662215efe0c7ff907220ef8c2f373ac078c1c9fb

app/Http/Controllers/StatisticsController.php
ppshobi: fixed : 662215efe0c7ff907220ef8c2f373ac078c1c9fb

app/Jobs/ProcessImageJob.php
ppshobi: removed typehint to solve it for now

app/Models/Photo.php
ppshobi: Good catch

ppshobi: fixed

app/SmartAlbums/Utils/MimicModel.php
ppshobi: because the MimicModel when used in `abstract class BaseSmartAlbum extends Model` class it overrides the toArray methid and it needs to match the parent defintiion

app/View/Components/Meta.php
ppshobi: reverted, not sure why i changed it

app/SmartAlbums/Utils/MimicModel.php
ildyria: Ohhh right.

app/Models/Extensions/Thumb.php
```diff
@@ -45,9 +46,9 @@ protected function __construct(string $id, string $type, string $thumb_url, ?str
 	 * Restricts the given relation for size variants such that only the
 	 * necessary variants for a thumbnail are selected.
 	 *
-	 * @param HasMany<Photo,$this> $relation
+	 * @param HasMany $relation
 	 *
-	 * @return HasMany<Photo,$this>
```
ildyria, line 50: lvl 6 will complain about that one.

ildyria (approved):
LGTM.

PR #3166, opened 2025-03-28
test: AddAlbumRequestTest

ppshobi:
Just some unit tests


app/Http/Requests/Album/MergeAlbumsRequest.php
```diff
@@ -29,6 +30,13 @@ class MergeAlbumsRequest extends BaseApiRequest implements HasAlbum, HasAlbums
 	use HasAlbumsTrait;
 	use AuthorizeCanEditAlbumAlbumsTrait;
 
+	public function __construct(array $query = [], array $request = [], array $attributes = [], array $cookies = [], array $files = [], array $server = [], $content = null)
+	{
+		parent::__construct($query, $request, $attributes, $cookies, $files, $server, $content);
+
+		$this->albums = new Collection(); // initialize with empty collection to avoid error MergeAlbumsRequest::$albums must not be accessed before initialization
```
ppshobi, line 37: I'm not sure if i am doing this correctly, I was getting an error 
> MergeAlbumsRequest::$albums must not be accessed before initialization

when i comment this line and run MerglbRequestTest

ildyria:
It took me a bit of time to understand why you wanted those.
image
Looking at the coverage I understand better.

However this is not the best way to do it.
It is way more interesting to directly make "queries" in Feature_v2.
The reason being that you are able to also at the same time cover the missing functions here:
https://app.codecov.io/gh/LycheeOrg/Lychee/blob/master/app%2FHttp%2FControllers%2FGallery%2FAlbumController.php
And the associated actions (which are currently only green because we still test with Legacy endpoints).

tests/Unit/Http/Requests/Album/AddAlbumRequestTest.php
```diff
@@ -0,0 +1,57 @@
...
+use LycheeVerify\Contract\VerifyInterface;
+use Tests\AbstractTestCase;
+
+class AddAlbumRequestTest extends AbstractTestCase
+{
+	protected function setUp(): void
+	{
+		parent::setUp();
+		$mockVerify = $this->createMock(VerifyInterface::class);
+		App::instance(VerifyInterface::class, $mockVerify); // VerifyInterface is talking to DB & that is not needed for Request classes
```
ildyria, line 30: > that is not needed for Request classes

What do you mean ? :)

ildyria: @ppshobi  ? :)

ppshobi: was with some work, what I mean here is that these tests are unit tests. (which should only test that particualr class). Request classes like `AddalbumRequest` those are  small and only need to be tested on the functionalities that they actually contain. but becauase of the class hierarchy, those classes have the following parent
```php
abstract class BaseApiRequest extends FormRequest
{
	protected readonly VerifyInterface $verify;

	/**
	 * @throws FrameworkException
	 */
	public function __construct(
	....
	) {
	....
			$this->verify = resolve(Verify::class); // now this resolve() makes it hard to test because this object can't be passed from outside
    }
...
```

while the `resolve(Verify::class)` is looking in the container and trying to get it. its getting the class `Verify implements VerifyInterface` from LycheeVerify package

in that class the constructor talks to the database to fetch some config values
```php
public function __construct(
		....
       ....
	) {
		$this->config_email = $config_email ?? DB::table('configs')->where('key', 'email')->first()?->value ?? ''; // @phpstan-ignore-line
		$this->license_key = $license_key ?? DB::table('configs')->where('key', 'license_key')->first()?->value ?? ''; // @phpstan-ignore-line
		$this->validateSignature = new ValidateSignature($public_key);
		$this->validateHash = new ValidateHash($hash);
	}
```

makes complete sense for an integration test, but for a unit test connecting to database is not needed, Hence I mock the `Verify` class and repalce it in the container for the unit test. 

so I can test only the `AddAlbumRequest` class. 

these tests not only helps for increasing the coverage rather makes sure [something like this does not happen](https://github.com/LycheeOrg/Lychee/pull/3167/files#r2027509753) when we do a whole project level refactoring.

ildyria (approved):

PR #3168, opened 2025-03-28
Pin dependencies for docker

ildyria:
Not the prettiest code...

As we are not publishing the Lychee-base yet we cannot pin it...
As a result, the only way to have pinned dependency on the base is to literally copy paste the code.
ARG allows to set up the base.

Could be refactored later.

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -1,12 +1,74 @@
-FROM lychee-base:latest AS base
+# Later when we use lychee-base, We can just remove this big block of Dockerfile.
+# FROM lychee-base:latest AS base
+
+### REMOVE ME LATER
+### FROM HERE
+FROM debian:bookworm-slim@sha256:1209d8fd77def86ceb6663deef7956481cc6c14a25e1e64daec12c0ceffcc19d AS base
+
+ARG DEV_IMAGE 
+
+# Set version label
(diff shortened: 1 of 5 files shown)


d7415 (approved):
That's painful... But 🤷‍♂️ 
May want to get @ppshobi's view before merging though

docker/dev/nginx.conf
ppshobi, line 1: can we not reuse the nginx conf file in the base folder?

docker/dev/start.sh
ppshobi, line 1: same with start., file

docker/dev/nginx.conf
ppshobi: they just needs to be mounted to contianer,

docker/dev/Dockerfile
```diff
@@ -1,12 +1,74 @@
-FROM lychee-base:latest AS base
+# Later when we use lychee-base, We can just remove this big block of Dockerfile.
```
ppshobi, line 1: I'm not sure why we are copying the whole docker file. because woth the make command it will build both lychee base and lychee dev. and tag them accordingly

d7415: True. I'm not entirely sure what we need to pin a version of an internally defined Docker image.

PR #3170, opened 2025-03-28
Add change log link to left menu

ildyria:


app/Metadata/Json/ExternalRequestFunctions.php
```diff
@@ -0,0 +1,128 @@
...
+				(int) now()->diffInMonths($age) => (int) -now()->diffInWeeks($age) . ' weeks',
+				(int) now()->diffInYears($age) => (int) -now()->diffInMonths($age) . ' months',
+				default => now()->diffInYears($age) . ' years',
+				// @codeCoverageIgnoreEnd
+			};
+
+			return $text . ' ago';
+			// @codeCoverageIgnoreStart
+		} catch (\Throwable) {
+			return 'unknown';
```
ppshobi, line 66: if we return unknown , that means that we have to handle this unknow case somehwere else

like ` if (get_age_text() == unknown) { //do shizzle}` either throw an exception or return null

app/Metadata/Json/ExternalRequestFunctions.php
```diff
@@ -0,0 +1,128 @@
...
+use function Safe\file_get_contents;
+use function Safe\ini_get;
+
+class ExternalRequestFunctions implements ExternalRequest
+{
+	protected ?string $data = null;
+
+	public function __construct(
+		private string $url,
+		private int $ttl,
```
ppshobi:
```suggestion
		private int $ttlDays,
```

ppshobi (approved):

app/Metadata/Json/ExternalRequestFunctions.php
ildyria: Nah, this is directly returned to the front-end.
Ideally this function should actually be localized...

app/Metadata/Json/ExternalRequestFunctions.php
ildyria: [d4faf3a](https://github.com/LycheeOrg/Lychee/pull/3170/commits/d4faf3ac96534b4acd9976015361670463a7319e)

PR #3186, opened 2025-04-01
Clarify the photo view modes

ildyria:
Proper strings instead of numbers (numbers were used in Livewire front-end).

### DIFF
--- a/resources/js/composables/photo/basePhoto.ts
+++ b/resources/js/composables/photo/basePhoto.ts
@@ -59,25 +59,27 @@ export function usePhotoBaseFunction(
 
 	const imageViewMode = computed(() => {
 		if (photo.value?.precomputed.is_video) {
-			return 0;
-		}
-		if (photo.value?.precomputed.is_raw) {
-			if (photo.value?.size_variants.medium !== null) {
-				return 2;
-			}
-			return 1;
+			return "video";
 		}
 
-		if (!photo.value?.precomputed.is_livephoto) {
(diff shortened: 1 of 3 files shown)


tinohager:
I think it would make sense to create an enum for the imageViewMode

ildyria:
> I think it would make sense to create an enum for the imageViewMode

Fixed : [cd395fd](https://github.com/LycheeOrg/Lychee/pull/3186/commits/cd395fdb94615e8d865fe66bca2d508cf1b035eb)

resources/js/composables/photo/basePhoto.ts
```diff
@@ -57,27 +67,32 @@ export function usePhotoBaseFunction(
...
 			}
-			return 1;
+			if (photo.value?.size_variants.original?.url?.endsWith(".pdf")) {
+				return ImageViewMode.Pdf;
+			}
+			return ImageViewMode.Raw;
 		}
 
-		if (!photo.value?.precomputed.is_livephoto) {
+		if (photo.value?.precomputed.is_livephoto === true) {
```
tinohager, line 85: Is there a reason why you explicitly wrote === true here?

s-payyeri (approved):

ppshobi (approved):

resources/js/composables/photo/basePhoto.ts
ildyria: I flipped the conditions and wanted to be exact in the behaviour. I prefer clarity even thought it might be too much in this case.

PR #3190, opened 2025-04-01
Fix 500 when opening directly a smart album

ildyria:


sancsin (approved):

sancsin (approved):
Checked in local environment. The bug seems to be fixed! :)

PR #3207, opened 2025-04-05
Improve left menu

ildyria:
If we want to add more functionalities to Lychee we need to make more use of the Left menu, we cannot add more and more buttons to the header as it does not makes it explicit what functionalities does what. As a result, this PR creates the following changes on front-end:

1. make left menu always available.
2. move the login action to the left menu (I will admit that this one sucks, fortunately there is the `L` shortcut).
3. move the map button on Albums page to the left menu.
4. move the frame button on Albums page to the left menu.
5. move the map button on Album page near the download etc. buttons.
6. move the frame button on Album page near the download etc. buttons.

on backend: this PR creates a new ModulesRightsResource which is embeded in the GlobalRightsResources. That way we are able to extend with more configuration options. This makes it easier to track what is enabled or not. 

<table>
<tr><td>Logged in</td><td>Logged out</td></tr>
<tr>
<td>
	
</td>
<td>
	
</td>
</tr>
</table>

### DIFF
--- a/app/Http/Resources/GalleryConfigs/RootConfig.php
+++ b/app/Http/Resources/GalleryConfigs/RootConfig.php
@@ -8,17 +8,11 @@
 
 namespace App\Http\Resources\GalleryConfigs;
 
-use App\Contracts\Models\AbstractAlbum;
 use App\Enum\AspectRatioCSSType;
 use App\Enum\AspectRatioType;
 use App\Enum\TimelineAlbumGranularity;
-use App\Factories\AlbumFactory;
 use App\Models\Configs;
-use App\Models\Photo;
-use App\Policies\AlbumPolicy;
 use Illuminate\Support\Facades\Auth;
-use Illuminate\Support\Facades\Gate;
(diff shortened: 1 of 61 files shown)


tinohager:
I would like to reiterate my suggestion that in my opinion it makes sense to place the **Login button** at the top right as this is a much more widespread standard. 

| GitHub | Microsoft | Google | Facebook |
| -------- | ------- | ------- | ------- |
| GitHub  | Microsoft   | Google| Facebook |

ildyria:
> I would like to reiterate my suggestion that in my opinion it makes sense to place the **Login button** at the top right as this is a much more widespread standard.
> 
> GitHub	Microsoft	Google	Facebook
> GitHub	Microsoft	Google	Facebook

It is planned on a longer run. :)

But I believe that small refactoring are easier to review.

RonnieTaz (approved):
This looks good to me, but as @tinohager commented, we should look at changing the login button position for better UX. Since it is in the roadmap already, we can just go forth with this.

PR #3214, opened 2025-04-06
Add conditionals in a Pipe instead of unsets

ildyria:


JasonMillward (approved):
Much cleaner than my attempt.

PR #3217, opened 2025-04-07
Fix smart albums not having thumbs

ildyria:
Fixes #3204

### DIFF
--- a/app/SmartAlbums/BaseSmartAlbum.php
+++ b/app/SmartAlbums/BaseSmartAlbum.php
@@ -91,7 +91,11 @@ abstract class BaseSmartAlbum implements AbstractAlbum
 	 */
 	public function get_access_permissions(): Collection
 	{
-		return $this->access_permissions ?? resolve(Collection::class);
+		if ($this->public_permissions === null) {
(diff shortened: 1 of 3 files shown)


s-payyeri (approved):

ppshobi (approved):

app/SmartAlbums/BaseSmartAlbum.php
```diff
@@ -77,6 +75,37 @@ protected function __construct(SmartAlbumType $id, \Closure $smart_condition)
...
+		return $this->title;
+	}
+
+	/**
+	 * @return Collection<int,AccessPermission>
+	 */
+	public function get_access_permissions(): Collection
+	{
+		if ($this->public_permissions === null) {
+			return resolve(Collection::class);
```
ppshobi, line 95:
```suggestion
			return collect([]);
```

ildyria: The problem with that one is that it resolves into an `\Illuminate\Support\Collection` instead of `Illuminate\Database\Eloquent\Collection`.
Then it crashes. :(

PR #3218, opened 2025-04-07
Do not fetch data from internet when running tests

ildyria:
Tired of the 416 error code...

### DIFF
--- a/app/Metadata/Json/ExternalRequestFunctions.php
+++ b/app/Metadata/Json/ExternalRequestFunctions.php
@@ -105,7 +105,7 @@ class ExternalRequestFunctions implements ExternalRequest
 	private function fetchFromServer(): string
 	{
 		if (app()->runningUnitTests()) {
-			throw new RequestFailedException('file_get_contents() failed');
+			throw new RequestFailedException('file_get_contents() failed: "testing" environment detected.');
 		}
 		// @codeCoverageIgnoreStart
 		// we cannot code cov this part. APP_ENV is `testing` in testing mode.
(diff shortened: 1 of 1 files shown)


app/Metadata/Json/ExternalRequestFunctions.php
```diff
@@ -104,6 +104,12 @@ public function get_data(bool $use_cache = false): string|null
 	 */
 	private function fetchFromServer(): string
 	{
+		if (app()->runningUnitTests()) {
+			throw new RequestFailedException('file_get_contents() failed');
```
sancsin: Should the exception message maybe also clarify that the request is failing since unit tests are being run? Something like, "file_get_contents() failed due to unit tests being run"

sancsin (commented):
Which test is giving 512 error. I couldn't locate.

app/Metadata/Json/ExternalRequestFunctions.php
ildyria: good point.

ildyria: @sancsin  fixed [3bd0385](https://github.com/LycheeOrg/Lychee/pull/3218/commits/3bd0385f391e99ee96c75cfc678483dde4228854)

ildyria:
> Which test is giving 512 error. I couldn't locate.

416.

When fetching data from github (for checking the updates available) we are running quite a few tests in parallel: `[8.3, 8.4] x [pgsql, sqlite, mysql] x [legacy, non legacy]`

That combined with running multiple pipeline in parallel, makes us hit the rate limit of Github hence the 416. It happens rarely, but still annoying to see build breaking for 416 reason.

The test in question is "UpdateTest" in `Feature_v2/Maintenance`.
```php
	public function testAdmin(): void
	{
		$response = $this->actingAs($this->admin)->getJsonWithData('Maintenance::update');
		$this->assertOk($response);

		$response = $this->actingAs($this->admin)->postJson('Maintenance::update');
		$this->assertCreated($response);
	}
```

sancsin (approved):

PR #3228, opened 2025-04-09
SE: add download/view/share/favourite metrics on photos/albums. 

ildyria:
This PR contains 3 changes:

1. The refactoring of the side bar for a more compact/efficient look:
image

2. The addition of statistics on photo page (see above) and album page (see below)
image

3. The addition of an activity side panel.
image

Those come with a set of associated settings:
image

Note that the user tracking is privacy preserving by keeping only the hash of the following data: user agent & ip.

```php
public function visitorId(): string
{
	if ($this->visitor_id === null) {
		// Hash the ip of the the request with the user agent to create a unique visitor id
		// This is not a secure way to create a unique id, but it is good enough for our purpose.
		// For privacy reasons we don't want to store the ip or user agent in the database.

		/** @var Request $request */
		$request = request();

		// xxh64 is a fast hash function that is not cryptographically secure.
		// The cryptographic properties of xxh64 is actually not important to us.
		// We do not need to have a strong non-collision guarantee. (Yes, that is counter intuitive, lol)
		// Collisions are actually a "good thing" as it gives plausible deniability.
		$this->visitor_id = hash('xxh64', $request->ip() . $request->userAgent());
	}

	return $this->visitor_id;
}
```


app/Contracts/Events/HasTable.php
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * SPDX-License-Identifier: MIT
+ * Copyright (c) 2017-2018 Tobias Reich
+ * Copyright (c) 2018-2025 LycheeOrg.
+ */
+
+namespace App\Contracts\Events;
+
+interface HasTable
+{
```
ppshobi: why not an abstract class that has both `public function metricAction(): MetricsAction;` & `public function table(): string;` and called `MetricEvent` for example? si there going to be an a MetricAction without storign it anywhere?

app/Listeners/MetricsListener.php
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Just logging of Cache events.
+ */
+class MetricsListener
+{
+	/**
+	 * Handle the event.
+	 */
+	public function handle(AlbumDownload|AlbumShared|AlbumVisit|PhotoDownload|PhotoFavourite|PhotoShared|PhotoVisit $event): void
```
ppshobi: my above comment will save modifying this parameter type every time we add a new event ;)

ildyria: WTRC: [a4a4e18](https://github.com/LycheeOrg/Lychee/pull/3228/commits/a4a4e18a59583d98569932f5bb11a7eea3137965)

app/Contracts/Events/HasTable.php
ildyria: WRTC: [a4a4e18](https://github.com/LycheeOrg/Lychee/pull/3228/commits/a4a4e18a59583d98569932f5bb11a7eea3137965)

database/migrations/2025_04_07_211144_add_stats_columns.php
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * Run the migrations.
+	 */
+	public function up(): void
+	{
+		Schema::create('statistics', function (Blueprint $table) {
+			$table->id();
+			$table->char('album_id', self::RANDOM_ID_LENGTH)->index()->nullable(true);
+			$table->char('photo_id', self::RANDOM_ID_LENGTH)->index()->nullable(true);
```
ppshobi: is there a case when both columns are null? beacuse then what is this metrics for?

ppshobi: maybe some unique index also makes sense here

ppshobi: I think the correct way of doing this in the relational world is to add a pivot table
```
statstics_type
=========
id, type
1,  album
2, photo

stats table
=====
id, statistic_type_id, statisticable_id, ..., view_count
  1, 1 (beacuse album), 10-album_uuid-01
```

and then map the correct foreign keys, and use the morph type, or hasManyThrough in laravel (maybe can do without it)

beacuse otherwise what happens if a new type is coming in, lets say video

ildyria: > is there a case when both columns are null? beacuse then what is this metrics for?

not supposed to happen. :laughing: 

> maybe some unique index also makes sense here

True.

> I think the correct way of doing this in the relational world is to add a pivot table

You mean like this?
```
photo <-> pivot <-> stats
album <-> pivot <-> stats
```

I did consider it, but it felt to me that it was quite an overhead for not much gain (yet), also I believe it can easily be refactored later into using a pivot table if necessary. YAGNI principle, maybe YAGNIY.
Especially when you see how easy it is to transfer data from a table to another:
```php
DB::statement('INSERT INTO statistics (photo_id) SELECT id FROM photos');
DB::statement('INSERT INTO statistics (album_id) SELECT id FROM base_albums');
```

There would be a third approach which would be using a table with (id, created_at, action, userhash), and (pivot stat_id, id_able, album/photo), But then that would mean that we need to do counts etc at every query, which is going to slowdown things a lot. :( 

> because otherwise what happens if a new type is coming in, lets say video

Video is already supported, it is over loading photos...
But I see what you mean, if we create e.g. a posts feed, well 2 options:

Option 1: add a post_id column. :laughing: 
Option 2: move the ids to a pivot table.

But once again, in both cases, this is assuming that the current 3/4 choices of observable statistics are matching that new item which we want to track.

ildyria:
This PR has been split for readability.
See:
https://github.com/LycheeOrg/Lychee/pull/3244
https://github.com/LycheeOrg/Lychee/pull/3245
https://github.com/LycheeOrg/Lychee/pull/3246
https://github.com/LycheeOrg/Lychee/pull/3247
https://github.com/LycheeOrg/Lychee/pull/3248

PR #3229, opened 2025-04-11
more http tests

ppshobi:


ildyria (approved):
Not a big fan of the constructors additions. But well... 🤷

ppshobi:
I can fix the constructor one, I found a way for it

ildyria:
> I can fix the constructor one, I found a way for it

Do you want to do it in this PR or should we merge this one and you do it on top ?
Open to both options.

tests/Unit/Http/Requests/Album/MoveAlbumsRequestTest.php
```diff
@@ -23,11 +23,22 @@ class MoveAlbumsRequestTest extends BaseRequestTest
...
+		$request->merge([
+			RequestAttribute::PARENT_ID_ATTRIBUTE => null,
+			RequestAttribute::ALBUM_IDS_ATTRIBUTE => [$album->id],
+			RequestAttribute::ALBUM_ID_ATTRIBUTE => $album->id,
+			RequestAttribute::ALBUM_IDS_ATTRIBUTE . '.*' => $album->id,
+		]);
+
+		$request->validateResolved(); // hydrate the request Class with the data before authorizing . Fighting the framework a bit
 
 		$this->assertTrue($request->authorize());
```
ildyria, line 58: No longer necessary, because this is called directly in `validateResolved()`

If `$request->authorize()` returns false, an exception is raised in `validateResolved()`.

ppshobi: weird class inheritance problem, 

if you don't call this `$request->validateResolved()` then the `processValidatedValues()` method will not get called. then 

`$this->album`
`$this->albums`

will not get initialized 😬 

then we get this error where it says `$this->albums` can;t be used before initializing it. 


but its a good thing, the logic inside `processValidatedValues()` will also gets covered with this setup, but all in all this setup looks a bit weird to me., (overriding the `processValidatedValues()` method) . even though right now the tests cover them logic

ildyria: >  (overriding the `processValidatedValues()` method) 

Understandable, but that is because Laravel does not resolve the values then Authorize, it first authorize then validates. Unfortunately, this also means that you would be double querying to validate whether a user can access an album: first by fetching the Id to validate ownership etc, then again to fetch the data...
Hence why we first fetch, then validate.

PR #3240, opened 2025-04-15
Refactor tests to not drop middleware

ildyria:

### DIFF
--- a/tests/Traits/RequireSE.php
+++ b/tests/Traits/RequireSE.php
@@ -25,7 +25,7 @@ use LycheeVerify\Verify;
 /**
  * Some of our tests require that the user is a subscriber.
  */
-trait RequireSe
+trait RequireSE
 {
 	protected function requireSe(Status $status = Status::SUPPORTER_EDITION): Verify
 	{
--- a/tests/Traits/RequireSe.php
+++ b/tests/Traits/RequireSe.php
@@ -1,0 +1,62 @@
+<?php
+
+/**
+ * SPDX-License-Identifier: MIT
+ * Copyright (c) 2017-2018 Tobias Reich
+ * Copyright (c) 2018-2025 LycheeOrg.
+ */
+
+/**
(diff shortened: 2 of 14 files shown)


tests/Traits/RequireSe.php
```diff
@@ -0,0 +1,62 @@
...
+namespace Tests\Traits;
+
+use Illuminate\Support\Facades\App;
+use LycheeVerify\Contract\Status;
+use LycheeVerify\Verify;
+
+/**
+ * Some of our tests require that the user is a subscriber.
+ */
+trait RequireSe
```
s-payyeri:
```suggestion
trait RequireSE
```

ildyria: Sure :)

s-payyeri (approved):

ppshobi (approved):

tests/Traits/RequireSe.php
ildyria: WRTC [907965e](https://github.com/LycheeOrg/Lychee/pull/3240/commits/907965e411232e15ac6882321b1b90addd47250a)

tests/Traits/RequireSE.php
```diff
@@ -0,0 +1,62 @@
...
+				return $this->status;
+			}
+
+			public function validate(): bool
+			{
+				return app()->runningUnitTests();
+			}
+		};
+
+		App::instance(Verify::class, $fake_verify);
```
ppshobi, line 50: Lets set this in the setup of tests (in the BaseRequestTest) because whats happening in the traits its hard to see. unless you particularly look for it

ildyria: But then it makes it less easy to toggle it: You need to get the `$fake` and set it as instance.
While with the current implementation, you can just do `$this->requireSe()` and you are done. :)

The init settings are also set with sane defaults: 
```php
$this->mock_verify = $this->requireSe(Status::FREE_EDITION);
```

PR #3260, opened 2025-04-21
Add end points for LiveMetrics

ildyria:
This pull request introduces a new "Live Metrics" feature, allowing for semi-real-time tracking of user interactions with photos and albums. The changes include the addition of a new database table, backend logic for managing metrics, and APIs for fetching and cleaning up metrics. Below is a breakdown of the most important changes:

### Core Feature Implementation

* **`app/Models/LiveMetrics.php`**: Added a new `LiveMetrics` model to represent the live metrics data. This includes relationships to photos and albums and custom query builder support.
* **`database/migrations/2025_04_07_211142_create_live_metrics_table.php`**: Added a migration to create the `live_metrics` table with fields for `visitor_id`, `action`, `album_id`, `photo_id`, and timestamps.

### API and Business Logic

* **`app/Actions/Metrics/CleanupMetrics.php`**: Introduced a new action to clean up old metrics based on a configurable retention period.
* **`app/Actions/Metrics/GetMetrics.php`**: Added logic to fetch live metrics, including filtering by user permissions and excluding specific types of noisy data.
* **`app/Http/Controllers/MetricsController.php`**: Updated the controller to include a new `get` endpoint for fetching live metrics, with cleanup logic executed before retrieval.

### Authorization and Policies

* **`app/Policies/MetricsPolicy.php`**: Added a new policy to control access to live metrics based on user roles and configuration settings.
* **`app/Http/Requests/Metrics/MetricsRequest.php`**: Created a request class to authorize access to live metrics using the new policy.

### Resource and Caching

* **`app/Http/Resources/Models/LiveMetricsResource.php`**: Added a resource class to format live metrics data for API responses, including photo and album details.
* **`app/Metadata/Cache/RouteCacheManager.php`**: Excluded the new live metrics API route from caching.

### DIFF
--- a/app/Http/Controllers/MetricsController.php
+++ b/app/Http/Controllers/MetricsController.php
@@ -8,11 +8,17 @@
 
 namespace App\Http\Controllers;
 
+use App\Actions\Metrics\CleanupMetrics;
+use App\Actions\Metrics\GetMetrics;
 use App\Events\Metrics\PhotoFavourite;
 use App\Events\Metrics\PhotoVisit;
+use App\Exceptions\UnauthorizedException;
+use App\Http\Requests\Metrics\MetricsRequest;
 use App\Http\Requests\Metrics\PhotoMetricsRequest;
+use App\Http\Resources\Models\LiveMetricsResource;
 use App\Models\Configs;
 use Illuminate\Routing\Controller;
+use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Auth;
 
 /**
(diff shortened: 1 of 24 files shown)


app/Http/Controllers/MetricsController.php
```diff
@@ -21,6 +27,19 @@
...
 class MetricsController extends Controller
 {
+	public function get(MetricsRequest $request, GetMetrics $get_metrics, CleanupMetrics $cleanup_metrics): Collection
+	{
+		if (Configs::getValueAsBool('live_metrics_enabled') === false) {
+			throw new UnauthorizedException('Live metrics are not enabled.');
+		}
+
+		// First clean up.
+		$cleanup_metrics->do();
```
JasonMillward, line 37: Is there a risk of this delete query negatively impacting user experience, and would it be worth moving the cleanup to a scheduled command or job so it runs in the background?

ildyria: The clean up is just a DELETE on the db with anything older than XXX.
That endpoint is also only queried on demand if the user has access rather than for everyone.
So I am not sure whether it would impact negatively or not.

I do agree that having a scheduled command/job could also do the trick (or do a `defer()` call).

JasonMillward (approved):

PR #3296, opened 2025-04-28
Propose revising the footer text as it could be misunderstood 

Novapixel1010:
I propose revising the footer text so that end-users—especially clients—clearly understand their photos are stored on our infrastructure, not on Lychee’s servers. The current phrase, “Hosted with Lychee,” could be misconstrued as implying that Lychee itself provides the hosting, which might expose Lychee to unnecessary legal risk. In addition, family members and other non-technical users may feel uneasy if they believe their images are being held on an unknown third-party platform. Updating the wording will avoid confusion and reinforce that the photos remain securely within our own environment.


ildyria:
Good idea! However, this needs to be changed in all the other languages too (a quick search and replace will do it easily).
and here:
- https://github.com/LycheeOrg/Lychee/blob/master/resources/js/components/footers/GalleryFooter.vue#L63
- https://github.com/LycheeOrg/Lychee/blob/master/resources/views/includes/footer.blade.php#L36

see here for all ocurences: https://github.com/search?q=repo%3ALycheeOrg%2FLychee+hosted_with_lychee&type=code&p=1

d7415:
I'm not convinced it's an issue, but if we do change, the underscores need to go.

Novapixel1010:
> I'm not convinced it's an issue, but if we do change, the underscores need to go.

have something against underscores 🤣

Novapixel1010 (commented):
ops underscores

ildyria (approved):

PR #3303, opened 2025-04-30
Version 6.6.0

ildyria:
Release minor dropping Legacy support.


database/migrations/2025_04_30_211901_bump_version060600.php
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up(): void
+	{
+		DB::table('configs')->where('key', 'version')->update(['value' => '060600']);
+		Artisan::call('cache:clear');
```
tinohager, line 33: The database migration script may not be the right place to the `Artisan::call('cache:clear');` command.

d7415 (approved):

database/migrations/2025_04_30_211901_bump_version060600.php
ildyria: The thing is we want to guarantee that this is run every version bump as the backend responses are likely to have changed. This literally made one of our release broken. See https://lycheeorg.dev/docs/releases.html#v651

tinohager: This is absolutely understandable for me I just wanted to bring in that technically it has nothing to do with the database migration and it might make sense to run a separate migration script that performs cleanups in the cache filesystem etc.


```php
<?php

define('PRODUCT_VERSION', '2.3.0');

if (version_compare(PRODUCT_VERSION, '3.0.0', '>=')) {
    // For version 3.0.0 or higher
    include 'features_v3.php';
} elseif (version_compare(PRODUCT_VERSION, '2.0.0', '>=')) {
    // For version 2.x
    include 'features_v2.php';
} else {
    // For older versions
    include 'features_legacy.php';
}
```

PR #3311, opened 2025-05-04
Forgot to add 'self' to csp object policies. 

ildyria:
Fixes pdf integration:
image

@mitpjones this should be the reason why it was not working...
Expected display.
image

This pull request updates the Content Security Policy (CSP) configuration in `config/secure-headers.php` to allow embedded PDF files and improve support for media sources such as live photos and S3-hosted content.

### Content Security Policy updates:
* Modified the `object-src` directive to allow `self` for embedded PDF files and added support for `blob:` URLs to enable "live" photos.
* Dynamically included the S3 URL and additional media sources from the `SECURITY_HEADER_CSP_MEDIA_SRC` environment variable in the `object-src` directive.

### DIFF
--- a/config/secure-headers.php
+++ b/config/secure-headers.php
@@ -629,7 +629,18 @@ return [
 
 		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/object-src
 		'object-src' => [
-			'none' => 'self', // Needed to display embeded pdf files.
+			'self' => true, // needed to display embeded pdf files
+			'allow' => array_merge(
+				[
+					'blob:', // required for "live" photos
+				],
+				// Add the S3 URL to the list of allowed media sources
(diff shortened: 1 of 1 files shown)


config/secure-headers.php
```diff
@@ -629,7 +629,18 @@
 
 		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/object-src
 		'object-src' => [
-			'none' => true,
+			'self' => true, // needed to display embeded pdf files
+			'allow' => array_merge(
+				[
+					'blob:', // required for "live" photos
+				],
```
d7415, line 636: Is this one needed for objects?

d7415 (approved):

config/secure-headers.php
ildyria: Probably not, but I wanted to uniformize with media.

PR #3322, opened 2025-05-07
Remove copyright notice from lang to avoid confusing weblate

ildyria:


.php-cs-fixer.php
```diff
@@ -4,7 +4,7 @@
 	[
 		__DIR__ . '/app/',
 		__DIR__ . '/database/',
-		__DIR__ . '/lang/',
+		// __DIR__ . '/lang/', // We no longer format lang files as they are from weblate and weblate does not support our formatting.
```
JasonMillward: Is this worth leaving as a comment or should it be removed entirely?

ildyria: I would like to keep it because at a later stage I would be happy to have this formatted with a secondary workflow: 
if files is not formatted, automatically create a PR to format it so that formatting can be enforced in such case (or figure out how to fix the formatting on weblate side :/ )

ildyria: I am also not closed to removing the lang from formatting completely.

ildyria: Fixed: [60488cb](https://github.com/LycheeOrg/Lychee/pull/3322/commits/60488cb8a0bd89dfa0e49c42e257da657975948c)

JasonMillward (approved):

.php-cs-fixer.php
```diff
@@ -1,10 +1,11 @@
 <?php
 
+// Note: We no longer format lang files as they are from weblate and weblate does not support our formatting.
+// Todo: later create automation that applies formatting on PR (and create a pull request on top if not formatted properly)
```
d7415, line 4: :+1:

d7415 (approved):

PR #3348, opened 2025-05-15
Fix secure link missing route

ildyria:


ildyria:
@rmspeers here is your fix.

d7415 (approved):

PR #3375, opened 2025-05-27
Drop legacy id

ildyria:
Cleaning  up dead code.


JasonMillward (approved):

haivala:
I have links in my other service to galleries. Is there a way to update the current https://lychee/#16230578300845 links to the new format?

ildyria:
> I have links in my other service to galleries. Is there a way to update the current [lychee#16230578300845](https://lychee/#16230578300845) links to the new format?

Unfortunately not without manual work while staying temporarily on version before 6.6.0.
On 6.6.0 we dropped all the legacy API and in that one was contained also the legacy id redirection.

Note that the older link of style `lychee#<24-alpha-num-chars>` are still converted by the front-end into `lychee/gallery/<24-alpha-num-chars>`: https://github.com/LycheeOrg/Lychee/blob/master/resources/views/vueapp.blade.php#L14

However older links like `lychee#<0-9 digits style>` no longer works already since version 6.6.0. It is frustrating and we really try to keep braking changes at minimum, but unfortunately on that one, we need to bite the bullet. :|
Sorry.

PR #3389, opened 2025-05-30
Compact left menu

ildyria:
|  Before | After |
| -------- | ------- |
| image|image |

### DIFF
--- a/resources/js/menus/LeftMenu.vue
+++ b/resources/js/menus/LeftMenu.vue
@@ -15,7 +15,7 @@
 				</router-link>
 			</div>
 		</template>
-		<Menu :model="items" v-if="initData" class="!border-none">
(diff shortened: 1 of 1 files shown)


RonnieTaz (commented):
This looks well with me, but in future, could we reimplement the relaxed menu and make as an option of sorts so user can choose if they want compact or relaxed menu

RonnieTaz (approved):
Looks good to me

ildyria:
> This looks well with me, but in future, could we reimplement the relaxed menu and make as an option of sorts so user can choose if they want compact or relaxed menu

It is actually just changing the padding, not much implementation needed. the thing is that the "Lychee" part is taking quite some  space and I am planning to move it else where.
And the rest is the Admin part which "normal users" registered in Lychee do not see :)

PR #3392, opened 2025-05-31
Add hint to discussion template that they should update Lychee

ildyria:


.github/DISCUSSION_TEMPLATE/bugs.yml
```diff
@@ -11,7 +11,18 @@ body:
     attributes:
       label: Lychee version
       description: Which version of Lychee are you using? Please provide the full version, e.g. v3.2.6.
```
d7415:
```suggestion
      description: Which version of Lychee are you using? Please provide the full version, e.g. v6.6.5.
```
Just to avoid reference to Lychee v3...

d7415 (approved):

PR #3409, opened 2025-06-03
Add color extraction from pictures.

ildyria:


app/Models/Palette.php
```diff
@@ -0,0 +1,59 @@
...
+		'photo_id',
+		'colour_1',
+		'colour_2',
+		'colour_3',
+		'colour_4',
+		'colour_5',
+	];
+
+	/**
+	 * @return array{colour_1:string,colour_2:string,colour_3:string,colour_4:string,colour_5:string}
```
sancsin: What is the background behind 5 colors for the color palette?

sancsin (approved):

app/Models/Palette.php
ildyria: @sancsin what do you mean ? :)

ildyria: This is a stacked PR so I added all the functionalities on top of each other. :)

ildyria: image

PR #3420, opened 2025-06-07
Add Persian translations

ildyria:
GPT generated. Will ask an Iranian friend to double check them.


lang/fa/changelogs.php
```diff
@@ -3,9 +3,9 @@
 return [
 	/*
 	|--------------------------------------------------------------------------
-	| Changelogs page
```
d7415, line 6: I don't think we normally translate the section headers? I don't mind, but for consistency. Also the Aspect Ratio one here isn't translated.

lang/fa/toasts.php
```diff
@@ -3,9 +3,9 @@
 return [
 	/*
 	|--------------------------------------------------------------------------
-	| Toasts
```
d7415, line 6: :thinking:

d7415 (approved):
Subject to someone who knows the language looking, lgtm.

lang/fa/changelogs.php
ildyria: yeah GPT got a bit overzealous. 😆

PR #3423, opened 2025-06-08
Persian localization reviewed

foadalavi:
I reviewed the Persian (fa) translation that was generated using Chat-GPT. 
Due to the right-to-left (rtl) nature of Farsi localization, some corrections may be needed after merging it into the master branch and checking the final result.


ildyria (approved):
Thanks a lot @foadalavi !

PR #3441, opened 2025-06-12
Fix 500 on search

ildyria:
Consequence of the album requirement on PhotoResource... 

### Backend Changes:

* **Enhanced `ResultsResource` to support album context:**
  - Updated the `fromData` method in `ResultsResource` to accept an optional `AbstractAlbum` parameter. This allows the resource to include album-specific context when constructing photo resources.
  - Modified the `SearchController` to pass the album parameter to `ResultsResource` when constructing the response.

* **Dependency inclusion:**
  - Added the `AbstractAlbum` contract import in `ResultsResource` to support the new parameter.

### Frontend Changes:

* **Improved global search handling in `MetricsService`:**
  - Updated the logic to treat "all" as a valid album ID for global search mode in two methods (`photo` and `favourite`). This ensures consistent behavior when no specific album is targeted. [[1]](diffhunk://#diff-e081103bbf5b4a618c2510746556b2a0d3f16d3eb95f48d1066d9f8847590b9eL16-R16) [[2]](diffhunk://#diff-e081103bbf5b4a618c2510746556b2a0d3f16d3eb95f48d1066d9f8847590b9eL25-R25)

### DIFF
--- a/resources/js/services/metrics-service.ts
+++ b/resources/js/services/metrics-service.ts
@@ -1,5 +1,6 @@
 import axios, { type AxiosResponse } from "axios";
 import Constants from "./constants";
+import { ALL } from "@/config/constants";
 
 const MetricsService = {
 	get(): Promise<AxiosResponse<App.Http.Resources.Models.LiveMetricsResource[]>> {
@@ -13,7 +14,7 @@ const MetricsService = {
 			return Promise.resolve(null);
 		}
 		// This is the case if we are in global search mode.
-		if (album_id === undefined || album_id === "all") {
+		if (album_id === undefined || album_id === ALL) {
 			return Promise.resolve(null);
 		}
 
@@ -22,7 +23,7 @@ const MetricsService = {
 
 	favourite(photo_id: string, album_id: string | undefined): Promise<AxiosResponse<null> | null> {
 		// This is the case if we are in global search mode.
-		if (album_id === undefined || album_id === "all") {
+		if (album_id === undefined || album_id === ALL) {
 			return Promise.resolve(null);
 		}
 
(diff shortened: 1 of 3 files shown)


JasonMillward (approved):

resources/js/services/metrics-service.ts
```diff
@@ -13,7 +13,7 @@ const MetricsService = {
 			return Promise.resolve(null);
 		}
 		// This is the case if we are in global search mode.
-		if (album_id === undefined) {
+		if (album_id === undefined || album_id === "all") {
```
tinohager: not use `"all"`

```js
import { ALL } from "@/config/constants";

if (album_id === undefined || album_id === ALL) {
```

resources/js/services/metrics-service.ts
```diff
@@ -22,7 +22,7 @@ const MetricsService = {
 
 	favourite(photo_id: string, album_id: string | undefined): Promise<AxiosResponse<null> | null> {
 		// This is the case if we are in global search mode.
-		if (album_id === undefined) {
+		if (album_id === undefined || album_id === "all") {
```
tinohager: not use "all"

```js
import { ALL } from "@/config/constants";

if (album_id === undefined || album_id === ALL) {
```

ildyria: Fixed [60af847](https://github.com/LycheeOrg/Lychee/pull/3441/commits/60af847b84b8a947e6e2563a787f35241b7601ad).
Thank you :)

resources/js/services/metrics-service.ts
ildyria: Fixed [60af847](https://github.com/LycheeOrg/Lychee/pull/3441/commits/60af847b84b8a947e6e2563a787f35241b7601ad).
Thank you :)

tinohager (approved):

PR #3455, opened 2025-06-14
Check Imagick version 7 in diagnostics.

ildyria:


app/Actions/Diagnostics/Pipes/Checks/ImagickPdfCheck.php
```diff
@@ -22,8 +22,13 @@
  */
 class ImagickPdfCheck implements DiagnosticPipe
 {
+	public const IMAGICK_POLICY_LOCATIONS = [
+		'/etc/ImageMagick-6/policy.xml',
+		'/etc/ImageMagick-7/policy.xml',
```
d7415: Is it worth swapping these so that it checks 7 first? Not sure how likely/possible it is to install both, but I wouldn't be surprised if old, unused config files were left behind.

d7415 (approved):

app/Actions/Diagnostics/Pipes/Checks/ImagickPdfCheck.php
ildyria: hummm good point.

ildyria: Fixed: [fc50d97](https://github.com/LycheeOrg/Lychee/pull/3455/commits/fc50d97f62792a5c48045404ba6058a7f741dd56)

PR #3458, opened 2025-06-15
feat: Add quick setup to run pgsql locally with tests

ildyria:


d7415 (approved):
Untested, but looks good and hopefully you have!

ildyria:
> Untested, but looks good and hopefully you have!

I'm using it pretty much as soon as I dev when using no data. :D
Sqlite for my unit test.
MySQL for when I want to test with large amount of data (copy of my prod DB).

PR #3469, opened 2025-06-17
Fix path traversal attack + bump version.

ildyria:


app/Http/Requests/SecurePath/SecurePathRequest.php
```diff
@@ -0,0 +1,30 @@
...
+class SecurePathRequest extends AbstractEmptyRequest
+{
+	/**
+	 * {@inheritDoc}
+	 */
+	public function authorize(): bool
+	{
+		// We make sure that at least secure_image_link is enabled or temporary_image_link is enabled.
+		// This ensures that the path provided by the potential attacker is encrypted with the key of the server or
+		// That it the url is signed.
```
d7415:
```suggestion
		// that the url is signed.
```

app/Http/Requests/SecurePath/SecurePathRequest.php
```diff
@@ -0,0 +1,30 @@
...
+
+		// These two options options should guarantee that the path only under control of the owner of the server (hopefully).
```
d7415:
```suggestion
		// This should guarantee that only paths that should be shared are accessible.
```
Maybe? I don't think I like my version any better...

app/Exceptions/SecurePaths/PathTraversalException.php
```diff
@@ -0,0 +1,26 @@
...
+ * PathTraversalException.
+ *
+ * This exception is thrown when a path traversal attack is detected.
+ * We throw an error 418 because we use this with fail-to-ban for the honeypot.
+ */
+class PathTraversalException extends BaseLycheeException
+{
+	public function __construct(string $msg, ?\Throwable $previous = null)
+	{
+		parent::__construct(Response::HTTP_I_AM_A_TEAPOT, $msg, $previous);
```
d7415, line 24: I don't like misusing 418. Is 403 not sufficient?

d7415 (approved):

app/Exceptions/SecurePaths/PathTraversalException.php
ildyria: 418 is used by the Honey pot functionality of Lychee.

PR #3475, opened 2025-06-21
Parse all folders before starting actual import.

ildyria:


ildyria:
Coverage will be provided at a later PR.

PR #3517, opened 2025-07-02
Remove annoying check preventing migration on prod database

ildyria:
It is so rarely used that most of the time it creates problems for our users.
Special case of software getting in the way of its admin. Let's ditch that.


d7415 (approved):
Sure

PR #3533, opened 2025-07-13
While processing in CLI, do not crash on large photos

ildyria:
Avoid the following (which could be resolved by checking if the image size is small enough for Imagick.
See more here:
https://stackoverflow.com/questions/77578848/imagemagick-throws-width-or-height-exceeds-when-converting-svg-to-png-with-ver

Let's just not fail for now. We later can consider a fix or a warning in the settings.
```
▶ docker exec -it 394cec php artisan lychee:extract_colour_palette       
Extracting Color Palette for IMG_1363-Pano [0VmiYZJs-xGaFypAGiddDcOe].

In ExtractColourPalette.php line 68:
                                                     
  Unknown Lychee exception (this is probably a bug)  
                                                     

In ImageFactoryForColourExtraction.php line 119:
                        
  Failed to load image  
                        

In ImageFactoryForColourExtraction.php line 107:
                                                                                                                    
  width or height exceeds limit `/tmp/magick-xfA9Pq9r7doI2aNcxSwY9sYPDRh85scS' @ error/cache.c/OpenPixelCache/3909  
```

### DIFF
--- a/app/Console/Commands/ImageProcessing/ExtractColourPalette.php
+++ b/app/Console/Commands/ImageProcessing/ExtractColourPalette.php
@@ -59,8 +59,14 @@ class ExtractColourPalette extends Command
 			}
 
 			foreach ($photos as $photo) {
-				$this->line(sprintf('Extracting Color Palette for %s [%s].', $photo->title, $photo->id));
-				ExtractColoursJob::dispatchSync($photo);
+				try {
+					$this->line(sprintf('Extracting Color Palette for %s [%s].', $photo->title, $photo->id));
+					ExtractColoursJob::dispatchSync($photo);
+					// @codeCoverageIgnoreStart
+				} catch (\Throwable $e) {
+					$this->error(sprintf('Error extracting color palette for %s [%s]: %s', $photo->title, $photo->id, $e->getMessage()));
+				}
+				// @codeCoverageIgnoreEnd
 			}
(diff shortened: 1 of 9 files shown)


app/Console/Commands/ImageProcessing/ExtractColourPalette.php
```diff
@@ -59,8 +60,14 @@ public function handle(): int
 			}
 
 			foreach ($photos as $photo) {
-				$this->line(sprintf('Extracting Color Palette for %s [%s].', $photo->title, $photo->id));
-				ExtractColoursJob::dispatchSync($photo);
+				try {
+					$this->line(sprintf('Extracting Color Palette for %s [%s].', $photo->title, $photo->id));
+					ExtractColoursJob::dispatchSync($photo);
```
JasonMillward, line 65: Do we want to check `Configs::getValueAsBool('use_job_queues')` here to dispatch jobs to queues?

ildyria: I am considering dropping the `use_job_queues` config actually. :)

ildyria: and have it determined directly by the `.env` config

JasonMillward (approved):

PR #3545, opened 2025-07-19
Implement pinned albums

cdzombak:
This PR is an initial implementation of "pinned albums" (see: https://github.com/LycheeOrg/Lychee/discussions/3326 ). It allows users to pin certain albums above the gallery's timeline view.

Notes:
- Tag albums cannot be pinned, since this would make the frontend work harder. That's an area for potential improvement.
- Albums within other albums cannot be pinned. I think that would make the album UI feel cluttered.
- Should we call this section of the UI "Pinned Albums" or "Featured Albums"?

## GIF Screenshot

pinning


ildyria:
Really good job! 

- Tag albums cannot be pinned, since this would make the frontend work harder. That's an area for potential improvement.
  > Nah, totally good right now.
- Albums within other albums cannot be pinned. I think that would make the album UI feel cluttered.
  > Sure. No strong opinion.
- Should we call this section of the UI "Pinned Albums" or "Featured Albums"?
  > No preference, as long as it is localized ;)

Makefile
```diff
@@ -105,7 +105,7 @@ rector:
 	 vendor/bin/rector process
 
 phpstan:
-	vendor/bin/phpstan analyze
+	vendor/bin/phpstan analyze --memory-limit 512M
```
ildyria: needed ?

app/Http/Resources/Models/ThumbAlbumResource.php
```diff
@@ -46,6 +47,7 @@ class ThumbAlbumResource extends Data
 	private ?string $max_taken_at = null;
 	public ?string $formatted_min_max = null;
 	public ?string $owner = null;
+	public ?string $parent_id = null;
```
ildyria: why is this necessary ? :)

Makefile
cdzombak: On my machine the PHP memory limit defaulted to 128MB, and PHP ran out of memory when running PHPStan. Perhaps other devs' machines have a higher limit set in `php.ini`?

app/Http/Resources/Models/ThumbAlbumResource.php
cdzombak: I need `parent_id` on `ThumbAlbumResource` so that I can hide the Pin/Unpin item of the context menu on sub-albums: https://github.com/LycheeOrg/Lychee/pull/3545/files#diff-c0b153753f9c8b289358c9921a85890017db8ed4f4e42c5ea6212448ab5e97a4R278

I did this because, though the backend technically supports setting `is_pinned` on sub-albums, the frontend doesn't have support for pinned sub-albums right now.

cdzombak:
> > Should we call this section of the UI "Pinned Albums" or "Featured Albums"?
>
> No preference, as long as it is localized ;)

https://github.com/LycheeOrg/Lychee/pull/3545/commits/4a89bd05ab17b75c4b0c19edda3d534dc196fdff changes it to "Featured Albums," since that's my personal preference :)

Makefile
cdzombak: Screenshot 2025-07-24 at 1 32 59 PM

ildyria: image

ildyria: Then I would just put that one change in it's own PR (separation of concerns ;) ).
I have no issue in limiting the memory consumption. :)

app/Http/Resources/Models/ThumbAlbumResource.php
ildyria: Then it is better to directly use the composable:
```ts
export function usePhotoRoute(router: Router) {
	function getParentId(): string | undefined {
		return router.currentRoute.value.params.albumId as string | undefined;
	}
```

That way if `getParentId()` is not undefined, then you know that you are in an album.

```ts
import { usePhotoRoute } from "@/composables/photo/photoRoute";
const { getParentId } = usePhotoRoute(router);
```

Makefile
cdzombak: ✅ 
- removed in https://github.com/LycheeOrg/Lychee/pull/3545/commits/6743b81be347b27a80b146beca99b7f0e2bf6078 
- see #3561

app/Http/Resources/Models/ThumbAlbumResource.php
cdzombak: Hmm. Here's where we hit the limits of my Vue knowledge.

How do I get `router` from within `useContextMenu`? https://github.com/cdzombak/Lychee/blob/351b3b80857e9e0adb702ee3956348e06b75e815/resources/js/composables/contextMenus/contextMenu.ts#L67

cdzombak: ✅ Done in https://github.com/LycheeOrg/Lychee/pull/3545/commits/9332647e3fefc4a2e0cbe7e593ced1da2f7c4650

resources/js/composables/album/albumsRefresher.ts
```diff
@@ -34,14 +38,23 @@ export function useAlbumsRefresher(auth: AuthStore, lycheeStore: LycheeStateStor
 		const getAlbums = AlbumService.getAll()
 			.then((data) => {
 				smartAlbums.value = (data.data.smart_albums as App.Http.Resources.Models.ThumbAlbumResource[]) ?? [];
-				albums.value = data.data.albums as App.Http.Resources.Models.ThumbAlbumResource[];
 				smartAlbums.value = smartAlbums.value.concat(data.data.tag_albums as App.Http.Resources.Models.ThumbAlbumResource[]);
 				sharedAlbums.value = spliter(
 					(data.data.shared_albums as App.Http.Resources.Models.ThumbAlbumResource[]) ?? [],
 					(d) => d.owner as string, // mapper
 					(d) => d.owner as string, // formatter
 					albums.value.length,
```
ildyria, line 45: That is a problem. The length here is read before the assignment of `albums.value` which was happening line 37.

cdzombak: 🤦🏻 

✅ Fixed in d9176bb832d5e303f591603dc6edfb9f6b84af3a

ildyria (approved):
LGTM.

PR #3550, opened 2025-07-21
Improve scrolling UX when exiting photo lightbox

cdzombak:
Prior to this PR, the album page unconditionally scrolled when exiting the photo lightbox. This scrolling was jarring — the page appeared to jump instantly — and often scrolled the page for no reason, since the photo might have been visible already.

This PR makes this scrolling logic smarter. Now:

- scrolling only happens if the photo that was viewed isn't fully visible already
- scrolling happens smoothly
- the minimum amount of scrolling necessary to make the photo visible is used


cdzombak:
GIF demonstrating new scroll behavior

ildyria (approved):

PR #3577, opened 2025-08-03
Tags: Add autosuggestion to the tag selection

ildyria:
image

### DIFF
--- a/resources/js/components/drawers/PhotoEdit.vue
+++ b/resources/js/components/drawers/PhotoEdit.vue
@@ -10,16 +10,7 @@
 					<Textarea id="description" class="w-full h-48" v-model="description" :rows="5" :cols="30" />
 
 					<label for="tags" class="font-bold h-11 mt-4 md:mt-0 self-center">{{ $t("gallery.photo.edit.set_tags") }}</label>
-					<AutoComplete
-						id="tags"
-						v-model="tags"
-						:typeahead="false"
-						multiple
-						class="border-b hover:border-b-0"
-						:placeholder="$t('gallery.photo.edit.no_tags')"
-						pt:inputmultiple:class="w-full border-t-0 border-l-0 border-r-0 border-b hover:border-b-primary-400 focus:border-b-primary-400"
-					/>
-
+					<TagsInput id="tags" v-model="tags" :add="true" :placeholder="$t('gallery.photo.edit.no_tags')" />
 					<label for="uploadDate" class="font-bold mt-4 md:mt-0 self-center">{{ $t("gallery.photo.edit.set_created_at") }}</label>
 					<DatePicker
 						id="uploadDate"
@@ -112,7 +103,6 @@ import { licenseOptions, SelectOption, SelectBuilders, timeZoneOptions } from "@
 import Select from "primevue/select";
 import Textarea from "@/components/forms/basic/Textarea.vue";
 import DatePicker from "primevue/datepicker";
-import AutoComplete from "primevue/autocomplete";
 import PhotoService from "@/services/photo-service";
 import Button from "primevue/button";
 import InputGroup from "primevue/inputgroup";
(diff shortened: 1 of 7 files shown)


resources/js/components/drawers/PhotoEdit.vue
```diff
@@ -122,6 +112,9 @@ import Checkbox from "primevue/checkbox";
 import { sprintf } from "sprintf-js";
 import { useRouter } from "vue-router";
 import { usePhotoRoute } from "@/composables/photo/photoRoute";
+import TagsInput from "../forms/basic/TagsInput.vue";
```
JasonMillward, line 115: `@/components/forms/basic/TagsInput.vue`

JasonMillward (approved):

resources/js/services/tags-service.ts
```diff
@@ -0,0 +1,17 @@
+import axios, { type AxiosResponse } from "axios";
+import Constants from "./constants";
+import { AxiosCacheInstance } from "axios-cache-interceptor";
+
+const TagsService = {
+	clearCache(): void {
+		const axiosWithCache = axios as unknown as AxiosCacheInstance;
```
cdzombak, line 7: This cast seems like a code smell.

resources/js/components/forms/photo/PhotoTagDialog.vue
```diff
@@ -97,6 +90,7 @@ function execute() {
 			life: 3000,
 		});
 		AlbumService.clearCache(props.parentId);
+		TagsService.clearCache();
```
cdzombak, line 93: This is a general question as I get oriented in the codebase, not a problem with this PR:

How well do we expect the Lychee UI to scale as more albums and tags are added? I have noticed a number of spots where we effectively clear the cache then re-fetch _all_ albums/tags/etc, and I am wondering how well I should expect my Lychee instance's UI to scale.

Or, put another way: how many albums are in the largest known Lychee instances?

resources/js/components/forms/basic/TagsInput.vue
```diff
@@ -0,0 +1,93 @@
...
+// List of all tags fetched from the server.
+// This contains the full list of tags available in the system and the total number of photos under it.
+const tags = ref<App.Http.Resources.Tags.TagResource[]>([]);
+
+// This is the filtered list of tags, it needs to be string, to match the modelValue type.
+const filteredTags = ref<string[]>([]);
+
+// In order to display the number of photos, we need to search into the tags list.
+// This is not optimal, maybe we can consider using a map later?
+function getNum(tag: string) {
```
cdzombak, line 48:
```suggestion
function getNum(tag: string): number {
```

cdzombak (approved):
With the caveat that I am _not_ an experienced Vue user, this PR looks reasonable to me. My only concern would be efficiency in a Lychee instance with a large number of tags, but you've already called that out as an area for future improvement.

resources/js/components/forms/photo/PhotoTagDialog.vue
ildyria: That's my own instance. :)
image

ildyria: > I have noticed a number of spots where we effectively clear the cache then re-fetch all albums/tags/etc, and I am wondering how well I should expect my Lychee instance's UI to scale.

Cache management is always a difficult thing. :| The implementation in Lychee is not perfect, but it aims to help.

resources/js/services/tags-service.ts
ildyria: Unfortunately, that is the only way I found to do it. :(
I am open to other solution. The thing is that the caching within axios is a plugin/decorator, as a result when loading axios (with the decorator), it is not recognized by typescript.

Option one, cast via `unknown`, option 2, cast without `unknown` and with a `@ts-ignore` ... pick your poison :(

resources/js/components/forms/basic/TagsInput.vue
ildyria: Will fix in the major re-linting of the code. 👍

ildyria:
See #3616

PR #3582, opened 2025-08-03
feature: allow tag renaming backend.

ildyria:
Figured it might be useful. :)

### DIFF
--- a/app/Http/Controllers/TagController.php
+++ b/app/Http/Controllers/TagController.php
@@ -8,9 +8,12 @@
 
 namespace App\Http\Controllers;
 
+use App\Exceptions\TagAlreadyExistException;
 use App\Http\Requests\Tags\DeleteTagRequest;
+use App\Http\Requests\Tags\EditTagRequest;
 use App\Http\Requests\Tags\ListTagRequest;
 use App\Http\Resources\Tags\TagResource;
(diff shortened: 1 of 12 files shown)


app/Http/Controllers/TagController.php
```diff
@@ -38,6 +41,20 @@ public function list(ListTagRequest $request): Collection
 		));
 	}
 
+	public function edit(EditTagRequest $request): void
+	{
+		$tag = $request->tag();
+
+		if (Tag::where('name', $request->name())->exists()) {
+			throw new TagAlreadyExistException();
+		}
```
JasonMillward, line 50: Could this be left to the validator? https://laravel.com/docs/10.x/validation#rule-unique

JasonMillward:
Happy with everything else

app/Http/Controllers/TagController.php
ildyria: good spot ! :)

ildyria: @JasonMillward this is refactored heavily in #3599 and thus no longer applicable. :)

ildyria: image
image

ildyria:
See #3616

PR #3599, opened 2025-08-07
fix access rights so it does not leak

ildyria:
This refactors the logic to avoid leaking tags between users.

While the logic is slightly more complex, I do hope the comments helps understanding the reasoning and the code. :)
Tests have been expanded to really lock down the expected behaviour.


JasonMillward (approved):

ildyria:
See #3616

ildyria:
See #3616

Issue #3542, opened 2025-07-16
Untagged Album


ildyria:
### Discussed in https://github.com/LycheeOrg/Lychee/discussions/3541

<div type='discussions-op-text'>

<sup>Originally posted by **MissCatTrap** July 16, 2025</sup>
### Enhancement

When you get a massive library of images, can tag them in batches it's very easy for a few to slip through the net. A private, untagged album, in similar vain to a Unsorted x Tag Album crossover would help solve this issue

<img width="436" height="228" alt="Screenshot 2025-07-16 at 08 41 26" src="https://github.com/user-attachments/assets/aff26ad4-93ed-494a-8e65-2f6efe60457d" />
</div>

PeterMMM:
Hi @ildyria ,
May I work on this issue, this one look fine for me.

ildyria:
Go for it ! :)
Feel free to join discord or ping me if you have any questions. :)

ildyria:
@PeterMMM This one is probably not tricky, but do be aware that you may need to implement pagination later as the query will explode on installs with a lot of pictures...

For this, I recommend you check this:
https://laravel.com/docs/12.x/pagination

And https://github.com/LycheeOrg/Lychee/blob/master/app/Http/Controllers/Gallery/SearchController.php which also implements it. 👍

PeterMMM:
Okay, sure @ildyria , Thanks for information. Well noted for pagination.
I will join discord too.

ildyria:
Done :)

PR #3718, opened 2025-09-24
Untagged Album: Add private untagged album

PeterMMM:
### Untagged Album
**Description**
This album functions as a crossover between the "Unsorted" and "Tag" albums, automatically collecting any images that have not been assigned a tag.

Reference #3542

## Summary by CodeRabbit

* **New Features**
  * Added a toggleable "Untagged" Smart Album and album pagination (server-side paginated photo lists).

* **UI**
  * Visible paginators in album view, new "Untagged" badge, and header image selection now supports paginated photo sources.

* **Settings**
  * New enable_untagged toggle and photos-per-page configuration.

* **Documentation**
  * Smart Albums docs and localization entries updated for “Untagged”.

* **Tests**
  * Tests updated to cover the new album type and related validations.


ildyria:
Rather than creating a Model for those, I think it is better to create a SmartAlbum, you can take example on 
https://github.com/LycheeOrg/Lychee/blob/master/app/SmartAlbums/UnsortedAlbum.php
https://github.com/LycheeOrg/Lychee/blob/master/app/SmartAlbums/StarredAlbum.php

Which both extends :
https://github.com/LycheeOrg/Lychee/blob/master/app/SmartAlbums/BaseSmartAlbum.php

That should already simplify some of the work.

Model expects to have a table etc. which would not make sense as there is only one "Untagged album".

app/Models/UnTaggedAlbum.php
```diff
@@ -0,0 +1,163 @@
+<?php
```
ildyria: I assume this one should go as you have the other ones in SmartAlbums. :)

app/SmartAlbums/UntaggedAlbum.php
```diff
@@ -0,0 +1,57 @@
...
+	/**
+	 * @throws ConfigurationKeyMissingException
+	 * @throws FrameworkException
+	 */
+	public function __construct()
+	{
+		parent::__construct(
+			id: SmartAlbumType::UNTAGGED,
+			smart_condition: function (Builder $query): void {
+				$query->doesntHave('tags');
```
ildyria: 👍

app/Relations/HasManyUnTaggedPhotos.php
```diff
@@ -0,0 +1,71 @@
+<?php
```
ildyria: Becomes unnecessary since you have the UntaggedAlbum (smart)

ildyria:
Thank you for your PR and first draft. 👍 
I like the direction you are taking, I see that you tried to approach and I believe the Smart album one is better.

app/Http/Resources/Editable/EditableBaseAlbumResource.php
```diff
@@ -43,7 +43,7 @@ class EditableBaseAlbumResource extends Data
 	public bool $is_model_album;
 	public bool $is_pinned;
```
ildyria, line 44: Untagged albums do not have the EditableBaseAlbumResource. That one is used to specify the "description" and other configurations...

And Untagged album is effectively a smart album this one won't be populated.

ildyria:
FYI: I am working on a significant refactoring of the front-end to avoid the prop drilling bad practice.
I moved quite a bit of the logic and removed quite a few props (which are now contained in pinia states).
https://github.com/LycheeOrg/Lychee/pull/3721

In think the best/cleanest way to integrate your PR would be by first integrate pagination on all albums,
and then the Unsorted smart albums should be really easy to add as the main pagination logic will be there.
What do you think?

PeterMMM:
> FYI: I am working on a significant refactoring of the front-end to avoid the prop drilling bad practice. I moved quite a bit of the logic and removed quite a few props (which are now contained in pinia states). #3721
> 
> In think the best/cleanest way to integrate your PR would be by first integrate pagination on all albums, and then the Unsorted smart albums should be really easy to add as the main pagination logic will be there. What do you think?

@ildyria Well noted for information.
For pagination, I think that's good idea. Keeping pagination logic to common, so we can use pagination to every other albums.

ildyria:
@PeterMMM  Code rabbit reviews are pretty good, don't you think? :)

resources/js/views/gallery-panels/Album.vue
```diff
@@ -24,6 +24,11 @@
 		@toggle-edit="toggleEdit"
 		@open-search="openSearch"
 		@go-back="goBack"
+		:total="albumStore.total"
+		v-model:rows="albumStore.per_page"
+		v-model:first="albumStore.from"
+		@update:rows="(val) => { albumStore.per_page = val; refresh(); }"
+		@update:first="(val) => { albumStore.from = val; refresh(); }"
```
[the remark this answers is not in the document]
ildyria: I am not sure about that one, we will need to double check in testing.

PeterMMM:
@ildyria Yes, they are quite valid. 🤔 I will check and try to fix them.

ildyria:
From here it should be also fairly easy to extend the pagination to all albums (but that's for another PR)

app/Models/Extensions/SortingDecorator.php
```diff
@@ -186,4 +187,62 @@ public function get(array $columns = ['*']): Collection
...
+			$options = in_array($column_sorting_type, self::POSTPONE_COLUMNS, true) ? SORT_NATURAL | SORT_FLAG_CASE : SORT_REGULAR;
+			$items = $items->sortBy(
+				$column_sorting_name,
+				$options,
+				$this->order_by[$i]['direction'] === OrderSortingType::DESC->value
+			)->values();
+		}
+
+		return new LengthAwarePaginator($items, $result->total(), $result->perPage(), $result->currentPage(), $result->getOptions());
+	}
```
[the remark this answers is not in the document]
ildyria: Your solution is not a viable solution. Memory use is a problem, hence why we need pagination. It is unfortunate but we will have to bite the bullet and place a disclaimer later in the config.

resources/js/stores/AlbumState.ts
```diff
@@ -32,6 +44,9 @@ export const useAlbumStore = defineStore("album-store", {
 			this.config = undefined;
 			this.isLoading = false;
 		},
+		updateCurrentPage(page: number) {
```
ildyria: dead code.

resources/js/stores/AlbumState.ts
```diff
@@ -18,12 +18,24 @@ export const useAlbumStore = defineStore("album-store", {
...
+		last_page: 0,
+		per_page: 0,
+		total: 0,
 	}),
 	actions: {
 		refresh(): Promise<void> {
 			this.reset();
 			return this.load();
 		},
+		resetPages() {
```
ildyria: dead code.

ildyria (approved):

PR #3746, opened 2025-10-09
Handle negative (= unlimited) `memory_limit` during computation of upload chunk size

schuetzm:
This fixes the error "The total chunks must be an integer." as described in [this discussion](https://github.com/LycheeOrg/Lychee/discussions/2961).

It is caused by `ini_get('memory_limit')` returning `-1`, which will then be divided by `10` and converted to an integer. The final chunk size is the minimum of this value and the other inputs, thus `0`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of upload size limits when no custom chunk size is set, deriving a safe value from server memory settings with safeguards against invalid values.
  * Prevents stalls or failures during large uploads and improves stability across varied server configurations.
  * Retains a safe 1 MB fallback on configuration/info errors to ensure uploads continue to work reliably.

### DIFF
--- a/app/Http/Resources/GalleryConfigs/UploadConfig.php
+++ b/app/Http/Resources/GalleryConfigs/UploadConfig.php
@@ -32,10 +32,14 @@ class UploadConfig extends Data
 		$size = Configs::getValueAsInt('upload_chunk_size');
 		if ($size === 0) {
 			try {
+				$memory_size = Helpers::convertSize(ini_get('memory_limit')) / 10;
+				if ($memory_size < 0) {
+					$memory_size = INF;
+				}
 				$size = (int) min(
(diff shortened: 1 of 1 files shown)


ildyria:
@schuetzm  First and foremost, thank you for debugging this one!

What do you think of the suggestion for CodeRabbit ?
Do you want to integrate it or should I ? :)

schuetzm:
The suggested changes are wrong. [`post_max_size`](https://www.php.net/manual/en/ini.core.php#ini.post-max-size) uses `0` as "no limit" value (see the "Changelog" section in the link). For [`upload_max_filesize`](https://www.php.net/manual/en/ini.core.php#ini.upload-max-filesize), the documentation doesn't even mention that this is an option.

ildyria (approved):
Lgtm

PR #3946, opened 2026-01-06
feat: ability to upload  and work with HEIF images by convert them to JPEG 

nbalkandzhiyski:
This new feature adds a additional pre-processing step in **app/Actions/Photo/Create.php** ** $pre_pipes** to detect HEIF file types and convert them to jpeg in order to support working with these file types.

Also during our work we found that the **photos.title** is **varchar(100)** but our test images from **IPhone 15+** had longer titles, so we had to add a migration to change the title to **text** column in order to support longer titles.
Ability to add more convertible file types in **app/Actions/Photo/Convert/ImageTypeFactory.php**

Added fallback image conversion via **Go** https://github.com/MaestroError/php-heic-to-jpg#php-heic-to-jpg because ImageMagick fails to convert some HEIF images, mostly from IPhone 15+ devices 

Key points: 
• Pre-processing happens early in the upload/import lifecycle, before AssertSupportedMedia.
• Conversion uses existing ImageMagick support when available (no new hard dependency).
• Converted file transparently replaces the original HEIC/HEIF file for downstream processing.
• Original file remains untouched if conversion fails.
• Fully compatible with S3 / remote storage setups, as conversion happens while the file is still local/temporary.
• Designed as an extensible pre-processing stage, not a HEIC-only hack.

## Summary by CodeRabbit

* **New Features**
  * Added support for HEIF/HEIC uploads with automatic conversion to JPEG during photo import.
  * Upload pipeline now normalizes unsupported image formats before validation.

* **Database**
  * Expanded photo title length and updated related indexes to support longer titles.

* **Tests**
  * Added unit and integration tests covering HEIF/HEIC upload and conversion flows.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


ildyria:
Hi,

Thank you for your pull request. I really agree that we need to have a pre-processing step that can be further extended.

However, I have a few issues that needs to be resolved:
- the vulnerabilities in the golang binaries see the findings from trivy. Those are simply not acceptable.
- https://github.com/MaestroError/php-heic-to-jpg/pull/37 < this update does not inspire me trust in the maintaining dev.
This has been dragging for more than 1.5 year, we are now on go 1.25, this should be kept up to date.

Looking more into details: 
- https://github.com/MaestroError/php-heic-to-jpg/tree/maestro/bin < the library ships binaries. That is a unacceptable risk (cf xz util backdoor). The binaries should be compiled from the CICD and put at release file (not for you to do but for MaestroError) 
- https://github.com/MaestroError/php-heic-to-jpg/blob/maestro/go.mod shows a dependency on https://github.com/adrium/goheif which does not contains a license file.
- https://github.com/strukturag/libheif is **GPL v3**.

So converting via Imagick, no issue, totally approved.
Converting with this tool is going to be a big nope. 

With regard to Imagick failing:
https://github.com/ImageMagick/ImageMagick/discussions/7426#discussioncomment-11599911
It seems that dependencies needs to be updated mostly.

nbalkandzhiyski:
@ildyria 
Hi, 
Thank you for your response. 
I suspected we'll have issues with the fallback package.

I'll explore the possible solutions for libhief. 
Currently the issue seems indeed to be related with libheif and not ImageMagick itself. 

libheif version greater than 1.15 is not able to get installed on Debian versions <= 12 or Ubuntu  <= 25.04 out of the box and as from your comment the possible solution will be to update libheif to >= 1.18.2

I'll update the PR once I have a resolution regarding this and update the docs if needed 

Regards,

ildyria:
> libheif version greater than 1.15 is not able to get installed on Debian versions <= 12 or Ubuntu <= 25.04 out of the box and as from your comment the possible solution will be to update libheif to >= 1.18.2

Yeah, but Debian versions <= 12 are no longer considered. 😃 
We consider only Debian stable which is Trixie - aka version 13! 🥳 

https://tracker.debian.org/pkg/libheif
stable: [1.19.8-1](https://packages.debian.org/source/stable/libheif)

Ubuntu is not in the scope of Lychee as our base images are on alpine or debian stable.

What I would do instead is to see if we can have a check in diagnostics for the version of libheif, and display a warning if it is too old. That way we do not have to deal with the fallback package. 👍

ildyria:
@nbalkandzhiyski  I worked on your PR and simplified it. The code is 99% here, however I am not able to push to your repo on that specific branch. :/

I you can either unlock it or merge the branch `feature/...` from Lychee into your PR. then I will be able to merge it.
(I still need to have a look at the foreign constraint issue)

ildyria (approved):

PR #3985, opened 2026-01-17
Speed-up space computation, at the cost no data from Unsorted

ildyria:
## Summary by CodeRabbit

* **Refactor**
  * Space statistics now exclude photos not assigned to any album; storage usage totals are built from per-album and per-size-type aggregates for clearer reporting.

* **Performance**
  * Statistics calculation reworked for more efficient aggregation, improving reporting speed and consistency.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


sancsin (approved):
Alright from my side. But why we don't want to include file size from Unsorted?

ildyria:
@sancsin Welcome back!

I am doing some testing on a very heavy database with more than 1 million Size Variant records and I am trying to speed up that query.

As we are now using a proper photo-album relationship, it requires more joins which makes the sum over all the size variant explode. I am experimenting on that side, see what can be done for improvements.

Ideally we would want to have that data I agree. But I don't think people puts that many pictures in the unsorted folder (assumption). As a result the reported size should not differ much.

ildyria:
image

Query in question in a database on SSD with 3,285,540 size_variants records.

PR #4170, opened 2026-03-10
fix(csp): add 'self' to connect-src so API calls to own origin are allowed

cdzombak:
Commit fa1da391 ("Fix Connect src csp #4161") corrected the `connect-src` config format from a flat array to a structured array, which caused the secure-headers library to actually emit the directive.

However, it omitted `'self' => true`, so all XHR/fetch requests to the site's own origin (e.g. `/api/v2/*`) were blocked by the Content Security Policy.


cdzombak:
This change was required after pulling the latest `master` for my website and is currently running at https://pictures.dzombak.com.

ildyria (approved):

PR #4205, opened 2026-03-20
Fix ssrf lookback edge case

ildyria:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced security for URL imports: stricter host/IP validation now blocks private, reserved and localhost loopback addresses.

* **Chores**
  * Bumped application version to 7.5.1.
  * Updated production base container image digest.
  * Added a migration to update the stored version and attempt a safe cache clear.
  * Added a new vulnerability ignore entry for CVE-2026-33186.


app/Rules/PhotoUrlRule.php
```diff
@@ -95,7 +95,7 @@ public function validate(string $attribute, mixed $value, \Closure $fail): void
 
 		if (
 			$this->config_manager->getValueAsBool('import_via_url_forbidden_localhost') &&
-			$host === 'localhost'
+			in_array(strtolower($host), ['localhost', '127.0.0.1', '::1'], true)
```
d7415, line 98: Probably not worth it, but should we block the rest of 127.0.0.0/8 here?

ildyria: it should be filtered with the above set in theory. No ?

ildyria: FILTER_FLAG_NO_RES_RANGE is a PHP filter flag used with filter_var() to validate IP addresses by excluding reserved ranges as defined in RFC 6890. 

IPv4 Reserved Ranges:
0.0.0.0/8 (default route)
127.0.0.0/8 (loopback)
169.254.0.0/16 (link-local)
240.0.0.0/4 (reserved for future use) 

IPv6 Reserved Ranges:
::1/128 (loopback)
::/128 (unspecified)
::ffff:0:0/96 (IPv4-mapped addresses)
fe80::/10 (link-local unicast)

d7415: Looks like this one wouldn't block e.g. 127.0.0.2 here, which is loopback, but I guess if people are that concerned they can turn on the local_ip  one which will cover it in `FILTER_FLAG_NO_RES_RANGE` :+1:

d7415 (approved):

PR #4214, opened 2026-03-22
Fix dns-resolving to localip

ildyria:
## Summary by CodeRabbit

* **Security Improvements**
  * Photo URL validation now resolves hostnames and blocks targets that resolve to private/reserved IPs or localhost, improving protection against DNS rebinding and updating failure messages accordingly.

* **Version Update**
  * Application version bumped to 7.5.2; migration updates stored version and attempts to clear application cache with a safe fallback on failure.

* **Tests**
  * Unit tests updated and new cases added to mock DNS resolutions and verify rejection of private/reserved and localhost targets.


app/Rules/PhotoUrlRule.php
```diff
@@ -83,23 +91,107 @@ public function validate(string $attribute, mixed $value, \Closure $fail): void
...
+				foreach ($aaaa_records as $record) {
+					$ips[] = $record['ipv6'];
+				}
+			}
+		} catch (\ErrorException) {
+			// DNS resolution failed — return empty array.
+			// The hostname checks (e.g. literal "localhost") still apply.
+		}
+
+		return $ips;
```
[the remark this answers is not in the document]
ildyria: No, if there is no resolution, then it does not matter anymore.

tests/Unit/Rules/PhotoUrlRuleTest.php
```diff
@@ -149,17 +169,51 @@ public function testForbiddenPrivateIp(): void
...
+		self::assertEquals('photo_url must not resolve to localhost.', $this->failMessage);
+	}
+
+	/**
+	 * Test that hostnames resolving to private IPs are blocked (DNS rebinding protection).
+	 */
+	public function testForbiddenPrivateIpViaHostname(): void
+	{
+		$this->rule = $this->makeRule(fn (string $hostname, int $type = DNS_A) => match ($type) {
+			DNS_A => [['ip' => '127.0.0.1']],
```
d7415: Should this use 192.168.x.x or 10.x.x.x for this test?

d7415 (approved):
Just the one query

tests/Unit/Rules/PhotoUrlRuleTest.php
ildyria: haaa yeah you are right would be good to also test those. :)

Issue #4260, opened 2026-04-04
Deep link to photo fails silently when photo is not on first page


Blogem:
## Bug Description

When navigating directly to a photo URL (`/gallery/{albumID}/{photoID}`), the photo viewer shows a blank page (no photo, no error, no spinner) if the target photo is not on the first page of the album's paginated photos. Navigating to the same photo by browsing the album and clicking on it always works.

## Steps to Reproduce

1. Have an album with more photos than `photos_per_page` (default 100)
2. Find a photo that falls on page 2+ of the album's sort order
3. Open its direct URL: `/gallery/{albumID}/{photoID}`
4. **Expected:** Photo displays in the viewer
5. **Actual:** Blank page — no photo, no error, no spinner

Photos on page 1 of the same album work fine as direct links.

## Root Cause

In `resources/js/stores/AlbumState.ts`, the `load()` method always loads only page 1:

```js
const r = [this.loadPhotos(1, false)];
```

Then in `resources/js/stores/PhotoState.ts`, the `load()` method searches for the target photo in the loaded photos array:

```js
this.photo = l.photos.find(t => t.id === this.photoId);
```

If the photo is on page 2+, `find()` returns `undefined`, `isLoaded` stays `false`, and the `PhotoPanel` component never renders.

When navigating from within the album, infinite scroll has already loaded more pages via `loadMorePhotos()`, so the photo is in the array by the time you click it.

## Additional Factor: POSTPONE_COLUMNS sorting

This is especially impactful when the album's sort column is `title` or `description` (in `SortingDecorator::POSTPONE_COLUMNS`). For these columns, `SortingDecorator::paginate()` applies no SQL `ORDER BY` — it paginates first, then sorts in PHP. This means the page boundaries are based on arbitrary database order rather than the visible sort order, making it unpredictable which photos will work as deep links.

## Suggested Fix

In `PhotoState.ts`, when `find()` returns `undefined`, fall back to fetching the photo individually via the existing `PhotoService.get(photoId)` API endpoint:

```typescript
async load() {
    if (this.photoId === undefined) {
        this.photo = undefined;
        return;
    }
    if (this.photo?.id === this.photoId) return;

    const photosState = usePhotosStore();
    this.photo = photosState.photos.find((p) => p.id === this.photoId);

    if (this.photo === undefined) {
        // Photo not in loaded pages — fetch individually
        try {
            const response = await PhotoService.get(this.photoId);
            this.photo = response.data;
[...]

PR #4272, opened 2026-04-08
Fix deep link failure for non-initial page photos (#4260)

diusazzad:
This PR resolves an issue where direct links (deep links) to photos would fail if the target photo was not part of the initial paginated load of the album. This happened because the frontend relied solely on the local photosStore cache and lacked a mechanism to fetch individual photo details from the backend when missing.

## Summary by CodeRabbit

* **New Features**
  * Added a new API endpoint to fetch individual photos.
  * Improved photo loading behavior with dynamic API fetching when photos are not cached locally.

* **Chores**
  * Enhanced system tool detection during database configuration to better identify platform-specific executable paths.

### DIFF
--- a/app/Http/Controllers/Gallery/PhotoController.php
+++ b/app/Http/Controllers/Gallery/PhotoController.php
@@ -24,6 +24,7 @@ use App\Http\Requests\Photo\DeletePhotosRequest;
 use App\Http\Requests\Photo\EditPhotoRequest;
 use App\Http\Requests\Photo\FromUrlRequest;
 use App\Http\Requests\Photo\GetPhotoAlbumsRequest;
+use App\Http\Requests\Photo\GetPhotoRequest;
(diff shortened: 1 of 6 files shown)


ildyria:
Hi, thank you for the tentative, but this is not the way to solve this issue. Yes you get the photo, however you do not have the next-previous behaviour. A better approach would be to query the page the photo is in and load that page (before loading the previous pages of that album).

Once you have the data of the album in, you can fetch from the cache. Yes it is more complex, but it aligns with the current behavior of the front-end rather than making it an exception for a single photo.

ildyria:
You open #4273 which is pretty much this with a patch to make the left right work. This is NOT the solution we want. We want something that is properly structured. I gave you directions in which to focus on, you could have for example added a query string which reference the current photo page in the album. This could be updated when scrolling down for example, but no you ignored and persisted.

Closing.

PR #4273, opened 2026-04-09
fix: Populate photo neighbors and album context when deep linking

### LINKED ISSUE #4260: Deep link to photo fails silently when photo is not on first page

diusazzad:
This Pull Request addresses the issue where deep-linking directly to a photo (e.g., gallery/<photo_id>) results in a lack of navigation context (Next/Previous photo) and a null album ID on the frontend.

Previously, the backend's PhotoController@get method hardcoded album_id as null and did not provide information about surrounding photos. This forced the frontend to manually manage state, which failed for non-initial page photos and direct entry points.

Changes:
Backend Calculation of Neighbors: Introduced a new action app/Actions/Photo/GetNeighbors.php that calculates the previous and next photo IDs within a given album's context, respecting its specific sorting settings (column and order).
PhotoResource Update: Modified the PhotoResource constructor to accept and return next_photo_id and previous_photo_id.
Controller Logic: Updated PhotoController@get to identify a suitable album for the photo (if it belongs to any) and use the GetNeighbors action to populate the navigation context.
This ensures that direct links have fully functional navigation buttons from the moment they are loaded.

Related Issues
Fixes #4260 (indirectly, as part of the discussion in #4272).

## Release Notes

* **New Features**
  * Introduced API endpoint to fetch individual photos, including information about neighboring photos for seamless album navigation
  * Improved photo loading mechanism with enhanced asynchronous synchronization when switching between photos in albums

* **Bug Fixes**
  * Resolved exiftool detection compatibility issues on Windows systems

### DIFF
--- a/app/Actions/Photo/GetNeighbors.php
+++ b/app/Actions/Photo/GetNeighbors.php
@@ -1,0 +1,83 @@
+<?php
+
+/**
+ * SPDX-License-Identifier: MIT
+ * Copyright (c) 2017-2018 Tobias Reich
+ * Copyright (c) 2018-2026 LycheeOrg.
+ */
+
+namespace App\Actions\Photo;
+
+use App\Contracts\Models\AbstractAlbum;
+use App\Models\Photo;
+use Illuminate\Database\Eloquent\Builder;
+
+class GetNeighbors
+{
+	/**
+	 * Returns the ID of the next and previous photo in the sorted album.
+	 *
+	 * @param Photo         $photo
+	 * @param AbstractAlbum $album
+	 *
+	 * @return array{previous_photo_id: ?string, next_photo_id: ?string}
+	 */
+	public function do(Photo $photo, AbstractAlbum $album): array
+	{
+		$sorting = $album->getEffectivePhotoSorting();
+		$column = $sorting->column->toColumn();
+		$order = $sorting->order->toOrder();
+
+		$prev = $this->getNeighbor($photo, $album, $column, $order, true);
+		$next = $this->getNeighbor($photo, $album, $column, $order, false);
+
+		return [
+			'previous_photo_id' => $prev?->id,
+			'next_photo_id' => $next?->id,
+		];
+	}
+
+	/**
+	 * @param Photo         $photo
+	 * @param AbstractAlbum $album
+	 * @param string        $column
+	 * @param string        $order
+	 * @param bool          $is_previous
+	 *
+	 * @return Photo|null
+	 */
+	private function getNeighbor(Photo $photo, AbstractAlbum $album, string $column, string $order, bool $is_previous): ?Photo
+	{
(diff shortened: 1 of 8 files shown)


ildyria:
I already replied on #4272 that this was not the way to do it. You completely ignored my advises and implementation directions. Closing.

PR #4278, opened 2026-04-10
Fixes the wrong aspect_ratio (16by9) in translations

mhepp63:
There is an inconsistency between lang's and code - the `aspect_ratio.1byx9` is missing language field, while there is `16by9` sitting in translations, which is not used elsewhere.

I rather fixed translations then code.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal aspect ratio configuration identifiers across language files and IDE configuration.


ildyria:
Hi, thank you for your PR.

I think it is better to keep `16by9` instead of `1byx9` and figure out where the `1byx9` is used and replace it by `16by9`.
:)

(and if this require a database migration, so be it)

mhepp63:
OK, no problem. It's only one sed ;) . And it seem like in the DB is stored "string" representation (`16/9`).

mhepp63:
... I will create a new request.

PR #4334, opened 2026-05-04
fix: use file path instead of stream for PDF thumbnail generation

mitpjones:
## Summary

 - **Before:** Uploading a large (> 50 MB) multi-page PDF produced no thumbnail — the upload appeared to succeed but the thumb was silently missing                                                            
  - **After:** PDF thumbnails are generated correctly regardless of file size (up to 100 MB limit imposed by docker/nginx.conf and PHP settings post_max_size=100M / upload_max_filesize=100M)                                                                                                                              
  - **Root cause:** Ghostscript (Imagick's PDF delegate) cannot seek backwards in an anonymous PHP stream, causing silent "Page drawing error" failures. Fixed by passing PDFs to Imagick via a real filesystem path instead of a stream; remote/non-local files are copied to a named temp file first (cleaned up in a `finally` block)                                                                                                      
  - Non-PDF uploads are unaffected    

  ## Root cause

  When `readImageFile()` is called with an anonymous stream, Ghostscript receives no filename or format hint and cannot seek backward. Modern PDFs commonly require non-linear reads to resolve cross-reference tables and compressed object streams, so any such PDF fails silently.

  ## Test plan

  - [ ] Manual regression test: upload a multi-page PDF > 50 MB **without** this fix and confirm no thumbnail is generated; apply the fix and confirm a thumbnail is generated for the same file
  - [x] Added `testPdfUploadCreatesThumbnail()` to `PhotosAddHandlerImagickTest` — uploads a PDF and asserts a 200×200 thumb is produced
  - [x] Existing image-handler tests continue to pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed PDF thumbnail generation so first-page thumbnails reliably render for complex/large PDFs and temporary files are always cleaned up on failure.

* **Tests**
  * Added a test verifying PDF uploads produce a 200x200 thumbnail variant.

* **Documentation**
  * Added feature spec, plan, and task docs describing the PDF thumbnail fix, local vs remote PDF behavior, and acceptance criteria.

* **Chores**
  * CI now installs Ghostscript in test runners to support PDF rendering.

### DIFF
--- a/.github/workflows/php_tests.yml
+++ b/.github/workflows/php_tests.yml
@@ -79,7 +79,7 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get --fix-broken install
-          sudo apt-get -y install ffmpeg libimage-exiftool-perl
+          sudo apt-get -y install ffmpeg libimage-exiftool-perl ghostscript
 
       - name: setup Databases
         run: |
--- a/app/Image/Handlers/ImagickHandler.php
+++ b/app/Image/Handlers/ImagickHandler.php
@@ -20,6 +20,8 @@ use App\Image\Files\NativeLocalFile;
 use App\Repositories\ConfigManager;
 use App\Services\Image\FileExtensionService;
 use Imagick;
+use function Safe\fopen;
+use function Safe\stream_copy_to_stream;
 
 class ImagickHandler extends BaseImageHandler
 {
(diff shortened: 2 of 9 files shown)


ildyria:
Hi @mitpjones  thank you for your PR.

Can I just ask you to extract the content of the if ($is_pdf) into a private function and similar for the normal path in order to reduce the complexity of the function. 

Other than that LGTM. :)

mitpjones:
Hi @ildyria,

getting a bit late for me here now. I'll check tomorrow if you want any further changes.

Thanks,
Tim

ildyria (approved):

PR #4453, opened 2026-06-22
v7.6.2 : Revert validate-env.sh 

ildyria:


ajohnson30:
Ok, good news/bad news? (maybe just for me)

Locally http://x.x.x.x:8000 is working again, just fine.  I browsed, uploaded just fine like before.

I was able to get an external https configuration working - it appeared to need both the docker-compose.minimal.yaml and .env file entries (I suppose if I figured out how to embed the .env stuff in the yaml file they could be in there)

relavent sections in yaml file:

  env_file:
    - path: /mypath/mydir/lychee/.env
x-common-env: &common-env
  APP_URL: "${APP_URL:-https://www.myurl.com}"
  APP_FORCE_HTTPS: "${APP_FORCE_HTTPS:-true}"

.env file:
APP_URL=https://www.myurl.com
TRUSTED_PROXIES=*
ASSET_URL=https://www.myurl.com/photos
APP_DIR=/photos
APP_FORCE_HTTPS=true

Now, I still have one problem:  when I'm external, some of my big picture files wont upload.  A few years back I got a 50megapixel camera and I like some of the big pictures :p.  Storage isnt an issue so I wasn't really worried about it, and the site is just for me and a few friends so I'm not concerned about bandwidth.  They upload just fine on the ip version.  I think I just missed a setting/toggle somewhere maybe?

We had a discussion a while back about running two instances on two different ports, one internal (ip) on external (https) which I never got around to, that may be an option too.

ildyria:
> Now, I still have one problem: when I'm external, some of my big picture files wont upload. A few years back I got a 50megapixel camera and I like some of the big pictures :p. Storage isnt an issue so I wasn't really worried about it, and the site is just for me and a few friends so I'm not concerned about bandwidth. They upload just fine on the ip version. I think I just missed a setting/toggle somewhere maybe?

Sizes is no longer a problem, Lychee supports chuncked uploads.

E.g. this 224MP photo 
image

> some of my big picture files wont upload

Have a look at your reverse proxy settings see if there is a limit there?

Also in Settings => Image Processing => Toggle expert => "Size of chunks when uploading in bytes: 0 is auto"
You can set the size of your chunks.

ajohnson30:
Oops, you're right
By default, NGINX has a file upload size limit of 1 MB per request. When a user tries to upload a file larger than 1 MB, NGINX blocks the request and triggers a 413 Request Entity Too Large error. 

Thats exactly what I got.

I'll need to add something like this to the config file
client_max_body_size 100M;

Thanks!

PR #4500, opened 2026-07-07
RSS Feed includes one <item> per photo

cdzombak:
Prior to this PR, the RSS feed included one item for every album each photo belonged to -- that is, if a photo belonged to 3 albums, it appeared 3 times in the RSS feed.

This change causes each photo to appear exactly once in the RSS feed, with multiple `<category>` tags if the photo belongs in multiple albums.

Care is taken to avoid pitfalls:
- The feed includes up to `$rss_max` photos (not `(photo, album)` pairs).
- We include _all_ albums a photo belongs to in the feed, as categories.
- The link for each photo is generated deterministically, by linking to the photo in the newest album (by created_at) it belongs to.
- The GUID for each RSS <item> is a deterministically-generated `tag:` URI, ensuring no GUID churn as photos' album memberships change.
- The `applySearchabilityFilter` logic/arguments are unchanged in this PR; visibility/NSFW filtering is unchanged.
- The only test that changes is the test for multi-album behavior added in https://github.com/LycheeOrg/Lychee/pull/4499 ; other tests covering RSS remain unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * RSS feeds now show each photo only once when it belongs to multiple albums.
  * RSS item GUIDs are stable and remain photo-scoped even as album memberships change.
  * Feed links consistently point to the newest relevant album while still listing all associated album names as categories.
* **Refactor**
  * Streamlined RSS item generation to improve identifier reliability and album/category selection.
* **Tests**
  * Updated RSS tests to assert single-item behavior, correct category listing, and GUID stability across membership changes.

### DIFF
--- a/app/Actions/RSS/Generate.php
+++ b/app/Actions/RSS/Generate.php
@@ -22,14 +22,16 @@ use Carbon\Exceptions\InvalidFormatException;
 use Carbon\Exceptions\UnitException;
 use GrahamCampbell\Markdown\Facades\Markdown;
 use Illuminate\Contracts\Container\BindingResolutionException;
+use Illuminate\Database\Query\Builder as BaseBuilder;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\DB;
+use function Safe\parse_url;
 use Spatie\Feed\FeedItem;
 
 /**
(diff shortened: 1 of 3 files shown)


cdzombak:
This has been deployed to pictures.dzombak.com; you can see the resulting RSS feed at https://pictures.dzombak.com/feed .

PR #4501, opened 2026-07-08
Allow customizing RSS feed title & description

cdzombak:
This PR allows customizing the RSS feed's title and description, via Settings -> RSS. Previously the title was hardcoded as "Latest pictures" and description was hardcoded as "Latest added pictures." regardless of Lychee's language setting.

**Disclosure:** this approach was suggested by Claude Code, as I don't have the specific Laravel expertise needed to resolve this on my own. I have tested and validated the changes myself, but I welcome any feedback on better ways to handle this feature.

## Summary by CodeRabbit

* **New Features**
  * Added configurable RSS feed title and description fields in the settings area.
  * RSS feeds can now use custom channel metadata while still falling back to defaults when left blank.

* **Bug Fixes**
  * RSS feed pages now consistently apply the saved feed title and description at request time.

* **Documentation**
  * Updated settings labels and help text across multiple languages to describe the new RSS options.


ildyria:
It is a possibility. The other approach would be to use the `.env`.

In the `config/feed.php`
replace
```php
			'title' => 'Latest pictures',
			'description' => 'Latest added pictures.',
```
with:
```php
			'title' => env('FEED_TITLE', 'Latest pictures'),
			'description' => env('FEED_DESCRIPTION', 'Latest added pictures.'),
```
And then the user can just set it in their `.env`.
I am not 100% sure which approach is the best.

ildyria (approved):

PR #4575, opened 2026-07-28
Add memory profiler to help @FredPraca

ildyria:


ildyria:
Dropped, this is not compatible with ZTS (Zend Thread Safety) which is the build that is used by frankenphp.
We are not recompiling a frankenphp image without ZTS for this.

PR #4592, opened 2026-08-06
Fix slipery zip extraction

ildyria:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved uploaded filename validation to reject unsafe names, path traversal, directory paths, and invalid values.
  * Strengthened ZIP extraction safeguards to keep archive contents within the permitted extraction location.
  * Improved handling of invalid extraction destinations, preventing unsafe or misconfigured archive extraction.
  * Added protection against archive entries resolving outside the designated extraction directory.

* **Tests**
  * Added coverage for valid and invalid filenames, including traversal attempts, directory paths, empty values, and non-string input.


Alpastx (approved):
the updated PR looks good. The pre-create `realpath` issue is addressed, and `FilenameRule` + base-dir containment in `SafeZipExtractor` close the reported path.

PR #4596, opened 2026-08-07
Upgrade typescript types generations

ildyria:
## Summary by CodeRabbit

- **New Features**
  - Added consistent person details, including IDs and names, across editable album and album header data.
  - Added automatic TypeScript definitions for supported application data, improving consistency between backend responses and frontend usage.

- **Improvements**
  - Standardized person information formats across album-related responses.
  - Improved type generation configuration and output formatting for more reliable frontend integration.
  - Clarified internal data labels for improved maintainability without changing user-facing behavior.


RonnieTaz (commented):
Code looks okay. A small question about whether the `space_usage` parameter on `app/Http/Resources/Models/UserManagementResource.php` should be turned into a Value Object like what you did with `app/Http/Resources/Models/Utils/PersonNameResource.php`

ildyria:
Hi @RonnieTaz ! 
That's a valid point. I will add a small DTO for this one too. :)

ildyria:
Actually, not really in that specific case as we extract the space value from the array.
I checked the code upstream, this is the result of:
```php
$users = User::select(['id', 'username', 'may_administrate', 'may_upload', 'may_edit_own_settings', 'quota_kb', 'description', 'note', 'upload_trust_level'])->orderBy('id', 'asc')->get();
$spaces_per_user = $spaces->getFullSpacePerUser();
$zipped = $users->zip($spaces_per_user);

return $zipped->map(fn ($item) => new UserManagementResource($item[0], $item[1], $request->verify()->is_supporter()));
```
and further up:
```php
	public function getFullSpacePerUser(?int $owner_id = null): Collection
	{
		return DB::table('users')
			->when($owner_id !== null, fn ($query) => $query->where('users.id', '=', $owner_id))
			->leftJoinSub(
				query: DB::table('base_albums')->select(['base_albums.id', 'base_albums.owner_id']),
				as: 'base_albums',
				first: 'base_albums.owner_id',
				operator: '=',
				second: 'users.id'
			)
			->leftJoin('album_size_statistics', 'album_size_statistics.album_id', '=', 'base_albums.id')
			->select(
				'users.id',
				'username',
				DB::raw('SUM(COALESCE(album_size_statistics.size_thumb, 0) +
					COALESCE(album_size_statistics.size_thumb2x, 0) +
					COALESCE(album_size_statistics.size_small, 0) +
					COALESCE(album_size_statistics.size_small2x, 0) +
					COALESCE(album_size_statistics.size_medium, 0) +
					COALESCE(album_size_statistics.size_medium2x, 0) +
					COALESCE(album_size_statistics.size_original, 0) +
					COALESCE(album_size_statistics.size_raw, 0)) as size')
			)
			->groupBy('users.id', 'username')
			->orderBy('users.id', 'asc')
			->get()
			->map(fn ($item) => [
				'id' => intval($item->id),
				'username' => strval($item->username),
				'size' => intval($item->size),
			]);
	}
```
So aside from changing those few lines, this is different.

The reason I went for a DTO in the case of `app/Http/Resources/Models/Utils/PersonNameResource.php` is because we do return both, but in the case of the space, we only return the space. So this has no impact on the typescript generated (the reason of the code change was a bug fix of the migration from spatie typescript generator version 2 to 3).

PR #4598, opened 2026-08-08
Adding OpenTelemetry and provide traces

FredPraca:
This PR provides OpenTelemetry functionalities to Lychee.
It allows to send logs, metrics and traces to any OpenTelemetry collector.

It has been tested with Loki, Tempo and Mimir for logs, traces and metrics respectively.

The best way to test it is to use Grafana Alloy as a collector and then dispatch to other tools such as the ones mentioned before.

It requires to install the opentelemetry extension as described [here](https://opentelemetry.io/docs/languages/php/getting-started/).

The following environment variables may be set:
```
OTEL_PHP_AUTOLOAD_ENABLED=true
OTEL_TRACES_EXPORTER=otlp
OTEL_METRICS_EXPORTER=none
OTEL_LOGS_EXPORTER=console
OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318/
OTEL_SERVICE_NAME=lychee
OTEL_PHP_DISABLED_INSTRUMENTATIONS=pdo
```

Please note that exporters may be set to console or otlp. When using otlp, you may provide the _OTEL_EXPORTER_OTLP_ENDPOINT_ variable to point to an OTLP collector. Here _http://localhost:4318/_ corresponds to a local setup for Tempo.

Moreover, _OTEL_PHP_DISABLED_INSTRUMENTATIONS_ allows to disable some instrumentation. Here, I disabled _pdo_ to avoid having all SQL operations in my traces. Currently, only _laravel_, _guzzle_ and _pdo_ are configured.


ildyria:
@FredPraca  I am not able to push changes to enable the CICD missing extension. Would you mind ticking the box or modify the php-test and php-dist yaml files :)

FredPraca:
> @FredPraca I am not able to push changes to enable the CICD missing extension. Would you mind ticking the box or modify the php-test and php-dist yaml files :)

The _Allow edits by maintainers_ is checked so I don't know why you're unable to change it.
I didn't have time to dig into the tests and dist things. I'll take a look at it.

The main problem is that _opentelemetry_ extensions is not a standard one. The process is described [here](https://opentelemetry.io/docs/languages/php/getting-started/#add-zero-code-instrumentation) but I don't know if I can do something like this Github actions.
Moreover, I don't know how to test without committing and pushing to GH.

FredPraca:
I'm also trying to include a package providing metrics about PHP runtime, especially RAM but I face a problem with initialization order.

app/Actions/Photo/Pipes/Init/LoadFileMetadata.php
```diff
@@ -25,27 +26,31 @@ class LoadFileMetadata implements InitPipe
 	 */
 	public function handle(InitDTO $state, \Closure $next): InitDTO
 	{
-		if ($state->exif_info !== null) {
-			// Metadata already loaded
-			return $next($state);
-		}
+		$trace = app(TraceService::class);
+
+		return $trace->traceMethod('photo.load_file_metadata', function () use ($state, $next) {
```
ildyria: I think here it might be cleaner to rename 
`handle(InitDTO $state, \Closure $next): InitDTO`
into
`_handle(InitDTO $state, \Closure $next): InitDTO`

and then have
```php
public function handle(InitDTO $state, \Closure $next): InitDTO
{
		$trace = app(TraceService::class);
		return $trace->traceMethod('photo.load_file_metadata',
			fn () => $this->_handle($state, $next)); 
}
```

ildyria: or indeed, rename into `execute()` instead of `_handle` :)

app/Actions/Photo/Pipes/Standalone/AbstractStandalonePipe.php
```diff
@@ -0,0 +1,35 @@
...
+	 *
+	 * @return StandaloneDTO
+	 */
+	public function handle(StandaloneDTO $state, \Closure $next): StandaloneDTO
+	{
+		$trace = app(TraceService::class);
+
+		return $trace->traceMethod($this->getSpanName(), function () use ($state, $next) {
+			return $this->execute($state, $next);
+		});
```
ildyria:
```suggestion
		return $trace->traceMethod($this->getSpanName(), fn () => $this->execute($state, $next));
```
You can use the arrow functions instead to avoid the `use (...)`

app/Actions/Photo/Pipes/Standalone/AutoRenamer.php
```diff
@@ -20,18 +20,22 @@
...
-class AutoRenamer implements StandalonePipe
+class AutoRenamer extends AbstractStandalonePipe
 {
-	public function handle(StandaloneDTO $state, \Closure $next): StandaloneDTO
+	protected function execute(StandaloneDTO $state, \Closure $next): StandaloneDTO
 	{
 		// Skip if the caller explicitly provided a title at upload time (FR-041-06).
 		// User-supplied titles take precedence and must not be overwritten by renamer rules.
 		if ($state->title !== null) {
+			Log::info('Photo has a title, we don\'t rename it, skipping');
```
ildyria: I don't think this is necessary :)

app/Actions/Photo/Pipes/Standalone/AutoRenamer.php
```diff
@@ -20,18 +20,22 @@
...
+
 			return $next($state);
 		}
 
 		// Skip if not enabled.
 		if (!$state->shall_rename_photo_title) {
+			Log::info('renaming not necessary, skipping');
```
ildyria: idem :)

app/Actions/Photo/Pipes/Standalone/SetOriginalChecksum.php
```diff
@@ -8,14 +8,16 @@
...
 use App\Image\StreamStat;
+use Illuminate\Support\Facades\Log;
 
-class SetOriginalChecksum implements StandalonePipe
+class SetOriginalChecksum extends AbstractStandalonePipe
 {
-	public function handle(StandaloneDTO $state, \Closure $next): StandaloneDTO
+	protected function execute(StandaloneDTO $state, \Closure $next): StandaloneDTO
 	{
+		Log::info('Computing original checksum');
```
ildyria: Instead of `info` maybe we want to keep those at `debug` level?

app/Http/Controllers/Gallery/PhotoController.php
```diff
@@ -65,45 +66,59 @@
...
+	}
+
 	/**
 	 * Upload a picture.
 	 */
 	public function upload(UploadPhotoRequest $request): UploadMetaResource
 	{
-		$meta = $request->meta();
-		$file = new UploadedFile($request->uploaded_file_chunk());
+		// Reset max memory usage before measuring
```
ildyria, line 79: Should this be in a middleware instead?

ildyria: See here some ideas I had:
https://github.com/LycheeOrg/Lychee/pull/4577

app/Http/Controllers/Gallery/PhotoController.php
```diff
@@ -115,34 +130,49 @@ private function process(
...
-			$final->getOriginalExtension(),
-			$meta->file_name
-		);
-		$processable_file->write($final->read());
-
-		$final->close();
-		$final->delete();
-		$processable_file->close();
-		// End of work-around
+		return $this->trace->traceMethod('photo.process', function () use ($verify,
```
ildyria, line 133: It is important to note that here you are measuring either of two cases:
if the processing is done sync, you will have the full processing memory, but if it is dispatch to a queue you won't see the computation anymore.

app/Actions/Photo/Pipes/Standalone/AbstractStandalonePipe.php
FredPraca: Ahaha, I  am not really a PHP dev as you may have seen :)

app/Http/Controllers/Gallery/PhotoController.php
FredPraca: Yes, I figured it out and maybe we will use traces to provide time consumption instead of memory consumption.
That's the reason why I would like to import the php metrics module and experiment around it.

app/Actions/Photo/Pipes/Standalone/AutoRenamer.php
FredPraca: Maybe at debug level  or even trace ?

app/Actions/Photo/Pipes/Standalone/AutoRenamer.php
FredPraca: Same as before

app/Actions/Photo/Pipes/Standalone/SetOriginalChecksum.php
FredPraca: As it is standard flow, yes, debug is enough

app/Actions/Photo/Pipes/Standalone/AutoRenamer.php
FredPraca: Ahaha, there is no trace :)
Notice maybe ?

app/Http/Controllers/Gallery/PhotoController.php
FredPraca: I'll read it as I'm not aware of all Laravel possibilities and especially with the middleware and so on stuff.

Besides this, I was not super happy with this way of tracing memory usage.
So we might maybe leave the trace without performing these memory features and keep tracing telemetry for what it is, following calls durations.

Your way of profiling might be a better way that we may merge with telemetry afterwards. 

What do you think ?

app/Actions/Photo/Pipes/Init/LoadFileMetadata.php
FredPraca: In fact, I started with it for telemetry tests and switched to Standalone pipes which were more promising in terms of trace and memory investigations.
If the part on Standalone pipes was considered well done by the main maintainer of Lychee, I was considering using the same pattern as for standalone pipes :)

We could also generalize this "pattern" to other types of pipes.
The drawback is that it will generate a lot of traces :smile:

ildyria: > We could also generalize this "pattern" to other types of pipes.
> The drawback is that it will generate a lot of traces 😄

You can solve this by creating a `.env` variable like: `OPENTELEMETRY_ENABLED`
and in `config/features.php` fetch it and set it to false by default (see the other feature flags).

And then in your general pipe use `config('features.telemetry') === true` to apply whether to do the log trace or not.

app/Actions/Photo/Pipes/Standalone/AutoRenamer.php
ildyria: Those are the levels of log in Laravel:
```
Log::emergency($message);
Log::alert($message);
Log::critical($message);
Log::error($message);
Log::warning($message);
Log::notice($message);
Log::info($message);
Log::debug($message);
```
I am fine with debug :)

app/Actions/Photo/Pipes/Init/LoadFileMetadata.php
FredPraca: In fact, that's already the goal of these variables
```
OTEL_TRACES_EXPORTER
OTEL_METRICS_EXPORTER
OTEL_LOGS_EXPORTER
```
If you set them to _none_, then nothing is sent, globally.
This way, you can choose which feature you enable.

But if you choose to set _OTEL_TRACES_EXPORTER_ to _otlp_ or _console_, then you'll have all traces.

ildyria:
@FredPraca  I'm considering creating a docker image with and without the instrumentation.
What do you think about it ?
There are also a few bugs I found which needs fixing (including an infinite recursion loop :'D )

PR #4665, opened 2026-08-25
Show inline error for wrong album password instead of opening login

matthewbolding:
## Summary
When an album is password-protected and the wrong password is entered, the unlock dialog currently closes and the site-wide login form opens instead, with no indication the password itself was wrong.

This makes the unlock dialog stay open and show an inline error on a 403 response, so the user can just retry the password. Other failure types keep the existing fallback behavior. Applied to both the v7 and v8 Unlock dialogs, following the same pattern already used in LoginForm.vue.

## Test plan
- [x] `npm run format` — no changes
- [x] `npm run check` (vue-tsc) — passes
- [x] `npm run build` — succeeds
- [x] Manually verified against a running v7.7.5 instance: entering a wrong album password now shows the inline error and keeps the dialog open; correct password still unlocks normally

## Summary by CodeRabbit

* **Bug Fixes**
  * Incorrect passwords entered when unlocking a protected album now display a clear inline error message.
  * The unlock dialog remains open so the password can be corrected and resubmitted.
  * The error clears automatically when the password is edited.
  * Other unlock failures continue to follow the existing error-handling behavior.

* **Localization**
  * Added incorrect-password messages across supported languages.

### DIFF
--- a/lang/en/dialogs.php
+++ b/lang/en/dialogs.php
@@ -226,6 +226,7 @@ return [
         'password_required' => 'This album is protected by a password. Enter the password below to view the photos of this album:',
         'password' => 'Password',
         'unlock' => 'Unlock',
+        'invalid_password' => 'Incorrect password for this album. Please try again.',
     ],
     'photo_tags' => [
         'question' => 'Enter your tags for this photo.',
--- a/resources/js/v8/components/forms/album/Unlock.vue
+++ b/resources/js/v8/components/forms/album/Unlock.vue
@@ -4,6 +4,7 @@
 			<p class="mb-5">{{ $t("dialogs.unlock.password_required") }}</p>
 			<UFormField :label="$t('dialogs.unlock.password')">
 				<InputPassword id="albumPassword" v-model="password" @keydown.enter="unlock" />
+				<UAlert v-if="invalidPassword" color="error" variant="soft" class="mt-2" :description="$t('dialogs.unlock.invalid_password')" />
 			</UFormField>
 		</template>
 		<template #footer>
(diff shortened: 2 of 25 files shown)


resources/js/v8/components/forms/album/Unlock.vue
```diff
@@ -4,6 +4,7 @@
 			<p class="mb-5">{{ $t("dialogs.unlock.password_required") }}</p>
 			<UFormField :label="$t('dialogs.unlock.password')">
 				<InputPassword id="albumPassword" v-model="password" @keydown.enter="unlock" />
+				<UAlert v-if="invalidPassword" color="error" variant="soft" class="mt-2" :description="$t('dialogs.unlock.invalid_password')" />
```
matthewbolding, line 7: Since the element I'm working inside is a UModal, I believe UAlert would be the most appropriate child element to insert if the user provides an incorrect password.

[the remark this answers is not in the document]
ildyria: @coderabbitai the v8 folder does not use Primevue but NuxtUI.

ildyria: good point. :)

lang/en/dialogs.php
```diff
@@ -226,6 +226,7 @@
         'password_required' => 'This album is protected by a password. Enter the password below to view the photos of this album:',
         'password' => 'Password',
         'unlock' => 'Unlock',
+        'invalid_password' => 'Incorrect password for this album. Please try again.',
```
ildyria, line 229: you will need to add this to all the other languages (no need to translate it, just put it in plain English)

ildyria: Otherwise it will fail the tests. you can run 
`php artisan test --filter=Lang` to check that.

ildyria (commented):
LGTM, just need to fix the translations keys :)

matthewbolding:
> LGTM, just need to fix the translations keys :)

Remaining langs updated to have the `invalid_password` field. @ildyria

ildyria (approved):

matthewbolding:
Disregard request to re-review; misclick. @ildyria

PR #4704, opened 2026-09-01
Show album cover art while password-protected

matthewbolding:
Adds a "Show cover when locked" toggle to the Visibility tab (v7 and
v8) for password-protected albums. When enabled, the album's cover
thumbnail is shown to anonymous visitors on gallery listings even
before they enter the password — the photos themselves stay hidden
until unlock. This replaces the need for a custom-JS workaround some
users (myself included) had resorted to.

- Off by default for every album, existing and new — no behavior
  change unless an admin explicitly opts an album in.
- Scoped to gallery listings (root albums, search, tags) via
  ThumbAlbumResource. The embed endpoint is intentionally out of
  scope for this PR: EmbededRequest::authorize() already rejects any
  password-protected album outright before the resource layer is
  ever reached, and the embed DTO doesn't carry a cover/thumbnail
  concept in the first place. Wiring cover-while-locked into that
  flow is a separate design decision, not a small addition.
- Not exposed in the bulk-edit admin action; per-album only for now.
- Fixed a latent bug this change surfaced: AlbumThumb.vue and
  AlbumListItem.vue (v7 + v8) inferred "is this album locked" from
  `thumb === null`, which was only ever true by coincidence — the
  old ThumbAlbumResource had exactly one code path that nulled out
  `thumb`, and it was the same path that ran when the album was
  locked, so the two facts always lined up. This PR breaks that
  coincidence on purpose: a locked album with the new toggle on now
  has `is_password_required = true` and a non-null `thumb` at the
  same time. Run that state through the old badge logic and it shows
  the green "unlocked" badge on an album that still requires a
  password to view any photos — a real, user-visible bug that simply
  had no way to occur before now. Fixed by adding an explicit
  `is_locked` field on ThumbAlbumResource and pointing all four
  components at it instead of inferring from `thumb`.

Note for anyone driving the API directly: `grants_cover_access` is
now a required field on `Album::updateProtectionPolicy`, matching
the existing validation on `grants_full_photo_access`.

Tests added: `AlbumsTest::testLockedAlbumHidesThumbByDefault` and
`::testLockedAlbumWithGrantsCoverAccessShowsThumb`, plus updates to
existing protection-policy tests for the new required field.

## Summary by CodeRabbit

* **New Features**
  * Added album cover selection and removal through album properties and context menus.
  * Added settings to control cover visibility for locked albums, including manually selected covers.
  * Added localized labels and descriptions for cover actions and locked-album settings.

* **Bug Fixes**
  * Album lock indicators now accurately reflect locked or unlocked status, even when thumbnails are hidden.
  * Cover selections remain synchronized across album views, search, and editing forms.
  * Prevented cover selections from being unintentionally cleared when updating other album details.
  * Improved album listings and thumbnails after album organization changes.


ildyria:
Hi @matthewbolding  I am currently in the middle of a huge refactoring of the way albums are actually being displayed in the front-end.

Do you mind we wait for it to be merged as it will have a LOT of changes with regard to your proposition?

PS: I really like the idea BTW, it is just that it arrives at the worse moment and I don't want you to waste your time on something which is going to be removed soon.

See here for what is coming: https://github.com/LycheeOrg/Lychee/pull/4705
If you join discord, feel free to ping me so we can see how to synchronize. :)

And yes, I am planning the same change for the photo display...

matthewbolding:
> Hi @matthewbolding I am currently in the middle of a huge refactoring of the way albums are actually being displayed in the front-end.
> 
> Do you mind we wait for it to be merged as it will have a LOT of changes with regard to your proposition?
> 
> PS: I really like the idea BTW, it is just that it arrives at the worse moment and I don't want you to waste your time on something which is going to be removed soon.
> 
> See here for what is coming: #4705 If you join discord, feel free to ping me so we can see how to synchronize. :)
> 
> And yes, I am planning the same change for the photo display...

Hi @ildyria. Thanks for the reply. Yes, I have no issue waiting on other changes to the codebase to be applied before this feature gets incorporated. I'll likely reach out to you on Discord late on. Glad you like the idea!

Another idea I had was to permit fully custom album covers, i.e., the system shall accept a new image file to be used for the sole purpose of an album cover.

ildyria:
> Another idea I had was to permit fully custom album covers, i.e., the system shall accept a new image file to be used for the sole purpose of an album cover.

Isn't cover_id already enough for that ?

matthewbolding:
> > Another idea I had was to permit fully custom album covers, i.e., the system shall accept a new image file to be used for the sole purpose of an album cover.
> 
> Isn't cover_id already enough for that ?

Not necessarily. To my understanding, cover_id can only be a photo that already exists in the album (or sub-album, I suppose), but the cover_id cannot be, for instance, some standalone photo that's not meant to be part of the album.

In my own deployments of Lychee, I've inserted custom JS to force the covers of certain top-level password-protected albums to be 1) visible to anonymous users and 2) be a custom image that's not part of the gallery.

app/Http/Resources/Models/ThumbAlbumResource.php
```diff
@@ -37,6 +37,8 @@ class ThumbAlbumResource extends Data
 	public bool $is_public;
 	public bool $is_link_required;
 	public bool $is_password_required;
+	public bool $grants_cover_access;
```
ildyria: by making it public, you are exposing this to the front-end. Do we want to expose the `grant_cover_access` ?
It does not seem that you are using it in the front-end.

ildyria:
I had a look, pretty damn, good!
Though I am wondering, should this be a per album setting or a global setting ? 

On one end, a global setting would make it easier and avoid the Access rights propagation.
On the other end, a per album setting allows for more configurable setup.

In that specific case, given the complexity added by the flexibility I would suggest to go for the global config instead.
This will also limit the refactoring needed with the Struct-of-Array migration.

Other questions are:

- Should we consider that all albums with a cover_id set and protected by password should display the thumb?
- Still on that vibe, should we make that a separate setting actually?  
  two separate settings:
  - show cover of locked albums => on = always display the cover on locked albums, off = do not display the cover
  - show selected cover on locked albums => on = display the cover even if the above is set to off, off = do not display the cover.

app/Http/Resources/Models/ThumbAlbumResource.php
matthewbolding: Ah, good catch. Since that is the case, exposing `grant_cover_access` really isn't necessary. I'll make the appropriate changes.

matthewbolding:
>  Though I am wondering, should this be a per album setting or a global setting ?

I think that having the option global would likely be the best good idea, at least initially, although I could certainly see a user make a compelling case for the setting to be album-by-album.

> Should we consider that all albums with a cover_id set and protected by password should display the thumb?

Yes, that's what I was thinking; if the option is enabled for the album cover to show while locked, then display it. I'm not particularly sure if differentiating between a specifically set cover_id and the auto-selected thumbnail is necessary, but that's just me. Did I understand your question properly?

And another thought crossed my mind: replacing the lock icon with the cover will give the user no indication that a particular album is locked. Perhaps overlaying a transparent, small lock icon that's positioned, say, in the bottom right might be prudent.

ildyria:
I think then the following is the best approach:

Two separate settings:
- show cover of locked albums => on = always display the cover on locked albums, off = do not display the cover
- show selected cover on locked albums => on = display the cover even if the above is set to off, off = do not display the cover.

The reason why the selected cover is interesting is because it allows the granularity you have from the per-album setting and it gives control on which thumb is selected, while if you just have the per album setting, you are exposed to a bit of randomness (the thumb selected depends of the ordering inside the album).

With regard to locked, you can probably just put a span 50% with `pi pi-lock` and on v8 similar approach but with `lucide:lock`. With the change of license of PrimeVue from v4 => v5, I no longer want this anywhere close to Lychee, so I migrated to NuxtUI, hence also the change of icons for which we use mostly `lucid` from https://iconify.design/

ildyria:
The two migrations which add and drop the `grants_cover_access` column should be obviously removed. ;)
And you still have some tests to fix which are returning 422 due to the missing argument.

app/Http/Requests/Album/UpdateAlbumRequest.php
```diff
@@ -107,6 +117,13 @@ public function authorize(): bool
 					->where(PA::ALBUM_ID, $this->album->id)
 					->where(PA::PHOTO_ID, $this->photo->id)
 					->count() > 0)
+			) &&
+			(
+				$this->cover_photo === null ||
+				(DB::table(PA::PHOTO_ALBUM)
```
ildyria: I think this is not correct.

The reason is because you you can right click on an album to set its cover as the cover of the album (transitivity by the parent).

So instead you should make a query which considers all descendants using `_lft` and `_rgt`.
Something like:
```php
DB::table(PA::PHOTO_ALBUM)
	->where(PA::PHOTO_ID, $this->cover_photo->id)
	->joins('albums', 'PA::ALBUM_ID', '=', 'albums.id')
	->where('_lft', '>=', $this->album->_lft) // album containing the photo needs to have lft >= than current album
	->where('_rgt', '<=', $this->album->_rgt) // album containing the photo needs to have rgt <= than current album 
	->exists();
```

ildyria (commented):
Just one fix on the validation that the photo belongs to albums (and descendants)

app/Http/Requests/Album/UpdateAlbumRequest.php
matthewbolding: I hadn't considered that as a possible route to set cover album. Good catch! Nevertheless, fix pushed below.

ildyria (approved):

Issue #4712, opened 2026-09-02
Crypto/stablecoin payments via Omnipay (Payzum) — how would you like async settlement handled?


payzum-hq:
Hi! Lychee's shop already speaks Omnipay, and I'd like to contribute support for **Payzum**, a non-custodial crypto/stablecoin gateway (USDC, USDT and more) — the [Omnipay driver](https://github.com/payzum-dev/omnipay-payzum) exists and the config/enum part is a tiny diff following the Mollie pattern.

The one design question is settlement: crypto confirmation is asynchronous, and right now `handlePaymentReturn` marks an order FAILED whenever `completePurchase()` is not successful on the buyer's return — which for a crypto payment is the normal case (the invoice is still confirming on-chain when the buyer comes back). Before opening a PR I'd love your preference between:

1. an **inbound notification endpoint** (the gateway signs each notification with HMAC-SHA-512; the driver verifies it over the raw bytes), following the custom-gateway pattern your `app/Actions/Shop/Gateway/README.md` describes, or
2. keeping it **poll-based**: leave the order in PROCESSING on return and re-check the invoice status (a scheduled job or on-view refresh).

Happy to implement whichever fits Lychee's architecture best — tests included.

Disclosure: I work on Payzum and would maintain the integration long-term.

ildyria:
> an inbound notification endpoint (the gateway signs each notification with HMAC-SHA-512; the driver verifies it over the raw bytes), following the custom-gateway pattern your app/Actions/Shop/Gateway/README.md describes, or

I think this is a better approach.
polling is very unreliable because it requires scheduled jobs to be setup etc.

Also a HMAC is not a signature. It is an authentication code. ;)

payzum-hq:
Perfect — inbound endpoint it is, thanks for the quick guidance. I'll open a PR following the custom-gateway pattern described in `app/Actions/Shop/Gateway/README.md`: the Payzum config/enum part (Mollie-style), a notification route that verifies the HMAC-SHA-512 signature over the raw bytes and completes the order idempotently, and `handlePaymentReturn` leaving the order in PROCESSING until the verified notification lands. Tests included.

PR #4713, opened 2026-09-03
Add Payzum as a crypto/stablecoin payment provider (async settlement via signed notification)

payzum-hq:
Implements the plan agreed in #4712 (option 1, the inbound notification endpoint @ildyria preferred). Adds **Payzum**, a non-custodial crypto/stablecoin gateway (USDC, USDT and more, multi-chain — funds settle directly to the photographer's own wallet), as an Omnipay provider.

Since crypto confirmation is asynchronous, the interesting part is settlement — following the pattern discussed in the issue:

- **New notification endpoint** `POST /api/v2/Shop/Checkout/Notify/Payzum/{order_id}` (`shop.checkout.notify`; keyed by order id, because completing an order replaces its transaction id with the provider reference and the gateway retries to the original URL): the [Omnipay driver](https://github.com/payzum-dev/omnipay-payzum) verifies the HMAC-SHA-512 signature over the raw request bytes (with a replay window) before any payload field is readable. After signature, order, amount and currency checks, the order is completed idempotently — a redelivered notification can never fulfil twice. `OrderCompleted` is dispatched here (guarded by `wasChanged('status')`) because the buyer may never come back to the browser.
- **`handlePaymentReturn` no longer fails confirming payments**: for Payzum the return handler polls the invoice once (`fetchTransaction`) and either completes, fails (expired/cancelled), or **leaves the order in PROCESSING** — never FAILED just because the buyer returned before the chain confirmed. `finalize` then redirects to the checkout status page, where a new `processing` state in `CancelledFailed.vue` (v7+v8, i18n key added to all locales with the English fallback) tells the buyer the payment is being confirmed and the page can be closed safely.
- **Config/enum**: `PAYZUM` case in `OmnipayProviderType` (requires `apiKey` + `webhookSecret`), `config/omnipay.php` block, `.env.example` entries. `composer.json`/lock gain [`payzum/omnipay-payzum`](https://packagist.org/packages/payzum/omnipay-payzum) (MIT) and its only dependency, the official `payzum/payzum-php` SDK (zero runtime deps); nothing else in the lock changes.
- `FinalizeRequest` additionally authorizes `COMPLETED` Payzum orders, because the notification can legitimately complete the order before the buyer's browser returns.

Tests: `CheckoutNotifyControllerTest` covers the happy path, forged signatures, pending statuses, expiry, amount mismatches, redelivery idempotency, unknown orders, and the finalize-keeps-processing behaviour. The whole `tests/Webshop` suite passes (295 tests), as do `phpstan` and `php-cs-fixer`.

Disclosure: I work on Payzum and will maintain the integration long-term.

## Summary by CodeRabbit

- **New Features**
  - Added Payzum as a payment provider.
  - Added asynchronous payment confirmation through secure payment notifications.
  - Orders can remain pending while payment confirmation is in progress and complete automatically afterward.
  - Added localized payment-processing status messages across supported storefront languages.

- **Bug Fixes**
  - Prevented duplicate order completion during asynchronous payment updates.
  - Added validation for payment signatures, order details, amounts, and currencies.
  - Improved browser-return handling for payments completed or still being confirmed.

### DIFF
--- a/app/Actions/Shop/CheckoutService.php
+++ b/app/Actions/Shop/CheckoutService.php
@@ -17,6 +17,7 @@ use App\Exceptions\Internal\LycheeLogicException;
 use App\Factories\OmnipayFactory;
 use App\Models\Order;
 use App\Services\MoneyService;
+use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Facades\Session;
 use Omnipay\Common\Exception\InvalidCreditCardException;
@@ -174,9 +175,49 @@ class CheckoutService
 	public function completePayment(Order $order, ResponseInterface $response): Order
 	{
 		$transaction_id = $response->getTransactionReference();
-		$order->markAsPaid($transaction_id);
+		$this->settle($order, $transaction_id);
 
 		return $order;
+	}
+
+	/**
+	 * Mark an order as paid, at most once.
+	 *
+	 * The browser return and an inbound payment notification can arrive at the
+	 * same moment, and both would otherwise observe PROCESSING and settle the
+	 * order — dispatching OrderCompleted (and thus fulfilling) twice. The row
+	 * is locked and re-read inside a transaction, so exactly one caller
+	 * performs the transition; the loser sees the fresh state and reports that
+	 * it changed nothing.
+	 *
+	 * @param Order  $order          The order to settle
+	 * @param string $transaction_id The reference to store on the order
+	 *
+	 * @return bool Whether THIS call completed the order
+	 */
+	private function settle(Order $order, string $transaction_id): bool
+	{
+		return DB::transaction(function () use ($order, $transaction_id): bool {
+			$fresh = Order::query()->whereKey($order->getKey())->lockForUpdate()->first();
+
+			if ($fresh === null) {
+				return false;
+			}
+
+			if (in_array($fresh->status, [PaymentStatusType::COMPLETED, PaymentStatusType::CLOSED], true)) {
+				// Someone else settled it first; adopt their state without
+				// claiming the transition.
+				$order->refresh();
+
+				return false;
+			}
+
+			// Saved through the caller's instance, while this transaction holds
+			// the row lock, so wasChanged('status') is true for the winner only.
+			$order->markAsPaid($transaction_id);
+
+			return true;
+		});
 	}
 
 	/**
@@ -253,7 +294,11 @@ class CheckoutService
 			$response = $gateway->fetchTransaction(['transactionReference' => $transaction_reference])->send();
 
 			if ($response->isSuccessful()) {
-				return $this->completePayment($order, $response);
+				// Same reasoning as in handlePaymentNotification(): the order's
+				// own transaction id has to stay the stable lookup key.
+				$this->settle($order, $order->transaction_id);
+
+				return $order;
 			}
 
 			if ($response instanceof PayzumFetchTransactionResponse && ($response->isExpired() || $response->isCancelled())) {
@@ -337,12 +382,18 @@ class CheckoutService
 			abort(400, 'Amount mismatch');
 		}
 
-		$transaction_reference = $notification->getTransactionReference();
-		if ($transaction_reference === null) {
+		if ($notification->getTransactionReference() === null) {
 			abort(400, 'Missing payment reference');
 		}
 
-		$order->markAsPaid($transaction_reference);
+		// Settled with the order's own transaction id rather than the gateway
+		// reference: that id is the lookup key of both the return and the
+		// notification URLs, and replacing it would make every later request
+		// for this order — including the buyer's browser return after an
+		// early notification — fail to resolve. The Payzum invoice stays
+		// reachable by it, since their API reads an invoice by payment id or
+		// by order id.
+		$this->settle($order, $order->transaction_id);
 
 		return $order;
 	}
--- a/app/Actions/Shop/CheckoutService.php
+++ b/app/Actions/Shop/CheckoutService.php
@@ -20,11 +20,16 @@ use App\Services\MoneyService;
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Facades\Session;
 use Omnipay\Common\Exception\InvalidCreditCardException;
+use Omnipay\Common\Exception\InvalidRequestException;
 use Omnipay\Common\GatewayInterface;
+use Omnipay\Common\Message\NotificationInterface;
 use Omnipay\Common\Message\RedirectResponseInterface;
 use Omnipay\Common\Message\ResponseInterface;
 use Omnipay\Dummy\Message\Response as DummyResponse;
 use Omnipay\Mollie\Message\Response\FetchTransactionResponse;
+use Omnipay\Payzum\Gateway as PayzumGateway;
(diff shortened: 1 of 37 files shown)


payzum-hq:
Thanks — all three were worth fixing, and two of them were real defects. Pushed in 5a8e46d.

**Atomic settlement.** Settlement now runs inside a transaction that locks and re-reads the order (`CheckoutService::settle()`), so the browser return and an inbound notification racing each other can no longer both transition it. Exactly one caller performs the transition and reports it; the loser adopts the fresh state and reports that it changed nothing, so `OrderCompleted` is dispatched once. `completePayment()` routes through it too, which means the other providers get the same guarantee.

**Stable lookup key.** You were right, and this was the more serious one: completing the order replaced `transaction_id` with the gateway reference, so once a notification arrived before the buyer's browser, their return URL — built with the original id — no longer resolved and they landed on a "not found" after paying successfully. Payments are now settled with the order's own transaction id, which stays the lookup key of both the return and the notification URLs for the order's whole life. Nothing is lost by this: Payzum's API reads an invoice by payment id *or* by order id, so the invoice remains reachable from the stored value. Added the regression test you asked for (notification first, then the browser return), plus one asserting the return does not re-fulfil.

**Redelivery test.** Added `testItFulfilsExactlyOnceAcrossRedeliveries`, which enables `webshop_auto_fulfill_enabled`, fakes `OrderCompleted` and asserts a single dispatch across two deliveries.

Translation keys renamed to `payment_processing` / `payment_processing_message` in all 23 locales and both Vue consumers.

`tests/Webshop` passes (298 tests), as do `phpstan` and `php-cs-fixer`.

ildyria:
@payzum-hq it will take me a bit of time to review this PR, but it is on my radar.

app/Actions/Shop/CheckoutService.php
```diff
@@ -204,6 +260,181 @@ public function handlePaymentReturn(Order $order, OmnipayProviderType $provider)
...
+	 * @param Order         $order    The order being processed
+	 * @param PayzumGateway $gateway  The initialized gateway
+	 * @param array         $metadata Session metadata stored at purchase time
+	 *
+	 * @return Order The refreshed order
+	 */
+	private function handleAsyncPaymentReturn(Order $order, PayzumGateway $gateway, array $metadata): Order
+	{
+		if ($order->status !== PaymentStatusType::PROCESSING) {
+			// Already settled, e.g. the notification landed before the redirect.
```
ildyria, line 282: Are you 100% it is already settled ? 

In normal flow we just throw:
`throw new LycheeLogicException('Order with invalid status.');`

app/Actions/Shop/CheckoutService.php
```diff
@@ -204,6 +260,181 @@ public function handlePaymentReturn(Order $order, OmnipayProviderType $provider)
...
+	 * readable — a forged, stale, or malformed delivery aborts with 400
+	 * without touching the order. Deliveries are retried by the gateway,
+	 * so a redelivered notification for an already-completed order is a
+	 * no-op.
+	 *
+	 * @param Order $order The order the notification URL points at
+	 *
+	 * @return Order The updated order
+	 */
+	public function handlePaymentNotification(Order $order): Order
```
ildyria, line 334: This is adding to much code which is Payzum specific into the CheckoutService.
This approach is not clean and is more of a hack and will prevent Lychee from being maintainable in the future.

The flow of Mollie and Paypal are relatively similar.
Hower in this specific case this is WIDELY different due to the asynchronization.

I would recommend to put those functions in a different class and use injection for it so that the logic of Payzum is centralized.

app/Http/Controllers/Shop/CheckoutController.php
```diff
@@ -173,13 +178,45 @@ public function finalize(FinalizeRequest $request, string $provider, string $tra
 			);
 		}
 
+		if (!$success && $order->status === PaymentStatusType::PROCESSING) {
```
ildyria, line 181: You need to add a gate on the provider here.

app/Http/Requests/Checkout/FinalizeRequest.php
```diff
@@ -38,7 +38,13 @@ class FinalizeRequest extends BaseApiRequest implements HasBasket
 	 */
 	public function authorize(): bool
 	{
-		return $this->order?->status === PaymentStatusType::PROCESSING && $this->order?->provider === $this->provider_type && $this->provider_type !== null;
+		// Asynchronous providers can complete the order from the payment
+		// notification before the buyer's browser returns; that return is
+		// still legitimate and must be able to show the completed order.
+		$is_valid_status = $this->order?->status === PaymentStatusType::PROCESSING ||
+			($this->order?->status === PaymentStatusType::COMPLETED && $this->provider_type === OmnipayProviderType::PAYZUM);
```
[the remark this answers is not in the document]
ildyria: ☝️

app/Http/Requests/Checkout/NotifyRequest.php
```diff
@@ -0,0 +1,89 @@
...
+ * CheckoutService::handlePaymentNotification().
+ *
+ * @property string $order_id
+ *
+ * @method merge(array $values)
+ * @method route(string $key)
+ */
+class NotifyRequest extends BaseApiRequest implements HasBasket
+{
+	public const ORDER_ID_ATTRIBUTE = 'order_id';
```
ildyria, line 34: There is Request constant for this ? 
If not, add it to the contracts instead of here.

app/Http/Requests/Checkout/NotifyRequest.php
```diff
@@ -0,0 +1,89 @@
...
+	 *
+	 * CANCELLED is allowed on purpose: a buyer can abandon the browser flow
+	 * and still pay from the hosted checkout page that is already open — the
+	 * money moved, so the notification must be able to complete the order.
+	 *
+	 * @return bool
+	 */
+	public function authorize(): bool
+	{
+		return $this->order?->provider === OmnipayProviderType::PAYZUM &&
```
ildyria, line 49: This is very much Payzum oriented, I would be more in favour of having the asynchronous flow available  _globally_ with a provider such as Payzum as one of the option.

That would allow more extensibility in the future.
Look at strategy design pattern: https://refactoring.guru/design-patterns/strategy

ildyria (changes requested):
The PR is in good shape, however there are parts which need refactoring otherwise the overall quality of Lychee will make it an unmaintainable dragon. And I would like to keep that one tamed.

app/Http/Requests/Checkout/FinalizeRequest.php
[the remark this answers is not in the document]
ildyria: @coderabbitai  I did not verify this, I highlighed this needs fixing.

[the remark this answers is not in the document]
payzum-hq: @ildyria Sorry for the bot noise — to clarify from our side: this was fixed in 5a8e46df (pushed Sep 3, the day the finding was raised). `handlePaymentNotification()` now settles with the order's own `transaction_id` instead of overwriting it with the provider reference, so the buyer's browser return still resolves after an early notification. `testTheBrowserReturnStillResolvesAfterAnEarlyNotification()` covers the notification-first race. Happy to adjust if you'd prefer a different approach (e.g. using the order ID in the return route instead).

